PR #99, opened 2026-04-28
chore: add daily branch-cleanup workflow

jonathonreilly:
## Summary
- Adds `.github/workflows/branch-cleanup.yml` that runs the existing inventory + archive/delete scripts daily at 06:00 UTC (also `workflow_dispatch`).
- Each run rebuilds the inventory and archives + deletes any remote branch with `last_commit_age_hours >= 24`.
- Open-PR head branches are added to `PROTECTED` at runtime via `gh pr list`, so a branch behind an open PR is never deleted.

## Why
- Today's repo had ~330 remote branches, mostly stale autopilot residue.
- Repo setting `delete_branch_on_merge` has been enabled to stop new accumulation.
- This workflow handles the long tail (autopilot branches that never merged via a PR, abandoned worktrees, etc.) so the remote stays under ~25 active branches without manual intervention.

## Recoverability
- Every deletion is preceded by an `archive/<branch>-<date>` tag push.
- Recovery: `git fetch origin "refs/tags/archive/<branch>-<date>:refs/tags/archive/<branch>-<date>"` then `git push origin "refs/tags/archive/<branch>-<date>:refs/heads/<branch>"`.

## Test plan
- [ ] Trigger manually via `workflow_dispatch` after merge to confirm a no-op run on a tidy repo.
- [ ] Verify next morning's scheduled run logs show 0 candidates (or only legitimately-stale branches).
- [ ] Confirm any open-PR branch active during the run is listed under "Protected" in the script output.

### DIFF
--- a/.github/workflows/branch-cleanup.yml
+++ b/.github/workflows/branch-cleanup.yml
@@ -1,0 +1,33 @@
+name: branch-cleanup
+on:
+  schedule:
+    - cron: "0 6 * * *"   # daily 06:00 UTC
+  workflow_dispatch:
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with: { fetch-depth: 0 }
+      - uses: actions/setup-python@v5
+        with: { python-version: "3.11" }
+      - run: |
+          # Refresh PROTECTED with current open-PR branches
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop result: blocked, not landed. The workflow is scheduled daily and executes remote branch deletion (`archive_and_delete_branches.py --execute`) with a runtime mutation of the protected set. It also grants only `contents: write` while relying on `gh pr list`; at minimum it needs explicit PR read permission and a safer non-scheduled/dry-run default. I left it open rather than landing a destructive automation.

PR #543, opened 2026-05-05
audit: detect upstream claim_scope/claim_type narrowing in invalidate_stale_audits

JONbridgerapps:
## Summary

Closes a gap where an upstream audit-side narrowing (claim_scope tightened, claim_type retyped within the same retained tier) lands without triggering downstream re-audit. The existing dep_weakened rule only fires on RANK drops; intra-tier narrowing slipped through.

## What changed

- `apply_audit.snapshot_audit_state` now records per-dep `claim_type` and `claim_scope` at audit time
- `invalidate_stale_audits.detect_invalidation` compares both maps and triggers re-audit (`dep_claim_type_changed:X:before->after` or `dep_claim_scope_changed:X`) when either differs from the snapshot

## Coverage matrix (after this PR)

| change to upstream | trigger |
|---|---|
| dep added or removed | `deps_changed:dep_added:X|dep_removed:Y` (existing) |
| dep effective_status weakened (rank drop) | `dep_weakened:X:before->after` (existing) |
| dep claim_type retyped (e.g. positive→bounded) | `dep_claim_type_changed:X` (NEW) |
| dep claim_scope tightened | `dep_claim_scope_changed:X` (NEW) |
| this row's criticality increased | `criticality_increased:before->after` (existing) |
| upstream note text edited | hash drift → archive + reset (seed_audit_ledger, existing) |

## Backwards-compat

Older snapshots predate the per-dep `claim_type`/`claim_scope` maps. The new checks gracefully skip when the snapshot field is absent — existing audit rows are not invalidated by this change. Future audits (via apply_audit) will record the new fields and downstream invalidation will work correctly going forward.

## Verification

- Both scripts parse cleanly
- Full pipeline runs with no invalidation cascade (only a benign stat-counter delta in seed_audit_ledger metadata: 4 rows that were `seeded_new` last run are now `preserved_existing`)
- The fix is strictly additive — no existing terminal verdicts got invalidated

### DIFF
--- a/docs/audit/scripts/apply_audit.py
+++ b/docs/audit/scripts/apply_audit.py
@@ -198,6 +198,19 @@ def snapshot_audit_state(row: dict, rows: dict[str, dict]) -> dict:
             d: rows.get(d, {}).get("effective_status") or "unaudited"
             for d in deps
         },
+        # Per-dep claim_type and claim_scope, so invalidate_stale_audits
+        # can detect upstream audit-side narrowing or retyping that
+        # happens without a note-text edit (silent w.r.t. note-hash
+        # drift and dep_effective_status weakening when both before and
+        # after live in the retained tier).
+        "dep_claim_type": {
+            d: rows.get(d, {}).get("claim_type")
+            for d in deps
+        },
+        "dep_claim_scope": {
+            d: rows.get(d, {}).get("claim_scope")
+            for d in deps
+        },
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed via review-loop in main commit e1fcae1d2. I integrated the audit-invalidation source changes and regenerated the audit pipeline outputs on current main.

PR #545, opened 2026-05-05
fix(gravitomagnetic): two-pass delta computation after v=0 baseline known

jonathonreilly:
## Summary

Closes the `runner_artifact_issue` audit condition on `gravitomagnetic_note` (lb=`9.18`, 289 transitive descendants).

## Root cause

The original `main()` iterated `V_VALUES = [-0.5, -0.2, 0.0, +0.2, +0.5]` in order and computed `delta` against `phase_static` in the same loop, with `phase_static` set only when `v == 0.0` was reached. For the earlier iterations `v ∈ {-0.5, -0.2}`, `phase_static` was still `None` and the conditional `delta = mean_p - phase_static if phase_static is not None else 0.0` reported `delta = 0.0` for both negative-velocity rows. The cached runner certificate therefore showed zeros for `v < 0`, contradicting the note's published antisymmetric phase-delta table.

The auditor's verdict: *"obtain a completed cached run or sliced certificate that reports the phase table and recomputes deltas after the v=0 baseline is known."*

## Fix

Refactor `main()` into two passes:

1. Pass 1 collects `phases[v_z] = mean_phase` for every `v_z`.
2. Pass 2 reads `phase_static = phases[0.0]` and prints `delta = phases[v] - phase_static` for every `v_z`.

Defensive `if 0.0 not in phases: raise RuntimeError(...)` makes the v=0 requirement explicit.

## Cache verification

After `scripts/cached_runner_output.py --refresh`, the certificate now shows the correct antisymmetric deltas across all three families:

```
Fam1: -0.0082, -0.0035, +0.0000, +0.0032, +0.0056
Fam2: -0.0076, -0.0034, +0.0000, +0.0030, +0.0054
Fam3: -0.0083, -0.0036, +0.0000, +0.0034, +0.0059
```

These match the published `docs/GRAVITOMAGNETIC_NOTE.md` phase-delta table within rounding. The antisymmetry-in-v property the note claims is now visible in both the prose and the cached certificate.

## Test plan

- [x] runner stops if `v=0.0` is missing from `V_VALUES` (defensive raise)
- [x] `scripts/cached_runner_output.py --check-only scripts/gravitomagnetic_portable.py` reports `fresh`
- [x] cached deltas are antisymmetric in v across all three families
- [x] `bash docs/audit/scripts/run_pipeline.sh` runs end to end with `OK: no errors`


jonathonreilly:
Landed via review-loop in main commit e1fcae1d2. The two-pass delta runner fix and refreshed cache are on main.

PR #565, opened 2026-05-06
audit: science_fix_loop progress-check / stale-kill

JONbridgerapps:
## Summary

The first version of \`science_fix_loop.py\` had only a hard 15-min timeout. The user's spot-test on \`planck_primitive_clifford_majorana_edge_derivation\` exposed the cost: codex worked for ~5 min, then sat thinking for the remaining 10 min before being killed at the hard timeout. We burned 10 min of compute for zero additional progress.

This PR adds a stale-kill: codex is terminated early if no file in the worktree has been touched within the configurable window.

## What's new

\`\`\`bash
# Default: 4-min stale kill, 15-min hard timeout, 30-sec poll
python3 scripts/science_fix_loop.py --n 5

# Tighter: kill faster on hard problems, ample budget for active codex
python3 scripts/science_fix_loop.py --n 5 \
    --stale-after-sec 120 \
    --codex-timeout-sec 1800

# Verbose poll
python3 scripts/science_fix_loop.py --n 1 --poll-interval-sec 15
\`\`\`

Live progress lines now print on every poll:

\`\`\`
[1/5] [renaming] desc=435 ...
    [progress] elapsed=  30s  stalled_for=    0s (stale_kill@240s, hard@900s)
    [progress] elapsed=  60s  stalled_for=   30s (stale_kill@240s, hard@900s)
    [progress] elapsed=  90s  stalled_for=    0s (stale_kill@240s, hard@900s)   # codex made an edit
    ...
    [progress] elapsed= 540s  stalled_for=  240s (stale_kill@240s, hard@900s)   # next poll: kill
    STALLED with no edits after 540s — likely a HARD physics problem
\`\`\`

## New outcomes / state-file values

- \`stalled_no_edits\` — codex stopped editing, no useful diff (HARD problem)
- \`stalled_partial\` recorded as \`pr_opened_partial_stalled\` (partial edits promoted to PR for human review)
- \`timeout_no_edits\` — hit hard timeout with no diff
- \`pr_opened_partial_timeout\` — hit hard timeout but with edits worth reviewing
- \`stop_reason\` field ("ok" / "stalled" / "timeout" / "error") added per attempt

## Mechanics

\`subprocess.run(timeout=...)\` replaced with \`Popen\` + poll loop:

1. Walk the worktree (skipping \`.git\`, \`__pycache__\`, etc.) for newest mtime
2. Every \`poll_interval_sec\`, check if mtime advanced → reset \`stalled_for\`; if not → accumulate
[...]

PR #568, opened 2026-05-06
audit: science_fix_loop counts codex thinking output as progress

JONbridgerapps:
## Summary

The first stale-kill heuristic (#565) used file-mtime as the only progress signal. Spot-test 2 exposed the gap: codex was THINKING (emitting reasoning text and tool-call traces via stdout) for 240s without making any file edits, and got killed as if it were stuck.

The user's observation: **codex's thinking process is real work, not a stall**.

## Fix

Add a second progress signal: codex's combined stdout+stderr byte count. Each poll, the loop reads \`stdout_fh.tell() + stderr_fh.tell()\` on the capture files. If either signal (mtime OR stream-bytes) advanced since the last poll, \`last_progress_time\` resets.

So:
- codex actively editing files → mtime advances → not stalled ✓
- codex actively thinking → stream bytes advance → not stalled ✓ (new)
- codex genuinely silent (process alive but no output, no edits) → stalled ✓

The hard 15-min backstop still applies for genuinely-stuck-thinking cases.

## New live-progress format

\`\`\`
[progress] elapsed=  120s  stalled_for=    0s  last=stream  bytes=4318
[progress] elapsed=  150s  stalled_for=   30s  last=stream  bytes=4318
[progress] elapsed=  180s  stalled_for=    0s  last=mtime   bytes=5012
\`\`\`

\`last\` shows which signal advanced last (\`mtime\` = files being edited, \`stream\` = codex emitting text). This lets the operator see when codex transitions from thinking to acting.

## Why spot-test 2 would now succeed

In the spot-test 2 run, codex emitted ~5KB of reasoning text in the 240s window before being killed. With the stream-bytes signal, \`stalled_for\` would have stayed at 0 — codex would have been allowed to keep thinking, possibly closing on a real edit.

## Test plan

- [x] Syntax + dry-run
- [ ] Live test on the same row to confirm thinking is no longer mistaken for stall
- [ ] Confirm hard-15-min still triggers if codex is truly stuck (not emitting and not editing)

### DIFF
--- a/scripts/science_fix_loop.py
+++ b/scripts/science_fix_loop.py
@@ -313,7 +313,7 @@ def run_codex(prompt_body: str, worktree: Path, timeout_sec: int,
         terminated. Captured cleanly so the partial diff (if any) is
         preserved.
       - Poll every `poll_interval_sec` seconds (default 30s).
-      - `progress_callback(elapsed, stale_for)` is called on each poll
+      - `progress_callback(elapsed, stale_for, last_kind, stream_bytes)` is called on each poll
         if provided, for live-status output.
 
     `stop_reason` values:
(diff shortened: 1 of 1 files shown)


JONbridgerapps:
Superseded by the comprehensive timeout PR — stream-byte progress + 30-min max + 15-min edit-deadline + self-screen preamble all consolidated together to avoid run_codex merge conflicts.

PR #695, opened 2026-05-08
feat(physics): A3 axiom proposal — admitting AC_φλ closes bridge-gap admission count to ZERO

jonathonreilly:
## Summary

Proposes the AC_φλ identification atom (residual from substep 4 AC
3-atom decomposition) as a new framework axiom **A3** ("Physical-Species
Identification"). If admitted by the audit lane, the **bridge gap
admission count drops from 1 to 0**; all bridge-dependent bounded
theorems lift to retained.

**Outcome: AXIOM PROPOSAL (governance, not derivation).**

**DEPENDS ON** salvage commits on main: `d4df8d637`, `d444fb545`,
`bc14c9dfc`, `d130ff515`.

## A3 statement (proposed)

```
A3 (Physical-Species Identification Axiom).
The triple (M_3(C) on H_{hw=1}, C_3[111] generator, no-proper-quotient
theorem) IS the SM flavor-generation structure carrying three
quantum-mechanically distinct physical species.
```

Empirical falsifiability: lattice Monte Carlo for C_3-breaking
observables on the hw=1 sector.

## Four governance criteria (all 31/0 PASS)

| Criterion | Verification |
|---|---|
| **Independence** from A1+A2 | 24 attack vectors, 0 unconditional positive arrows; cited PASS counts on main: 4 + 44 + 29 = 77/0 |
| **Consistency** with retained framework | 15 retained source notes verified compatible; no audit-ledger conflicts |
| **Minimality** | Each of 3 clauses independently necessary; alternatives are larger axiomatic additions |
| **Empirical falsifiability** | Predicts 3 generations [confirmed: LEP N_ν = 2.984 ± 0.008]; 3-by-3 CKM/PMNS [observed]; C_3-broken hierarchy [observed: m_t/m_c ≈ 136, m_c/m_u ≈ 580] |

## Downstream consequences (if A3 admitted)

| Artifact | Pre-A3 | Post-A3 |
|---|---|---|
| L3a admission | bounded_obstruction | **closed** (A3 corollary) |
| L3b admission | bounded (≡ L3a) | **closed** (W2.bridge inheritance) |
| Substep 4 of staggered-Dirac | bounded_theorem | **positive_theorem** |
| Staggered-Dirac realization synthesis | bounded_theorem (Block 06) | **positive_theorem** |
| W2.norm V_3 closure chain | bounded_theorem | **positive_theorem** |
| W2.binary binary {1/2, 1} | bounded_theorem | resolved to derived value 1/2 |
[...]

PR #709, opened 2026-05-08
feat(physics): A3 R2 — kinematic-primitive attack class EXHAUSTED (31+ vectors, zero positive)

jonathonreilly:
## Summary

Tested 7 time-direction attack vectors against AC_φ closure on H_{hw=1}. **All 7 FAIL by a shared structural reason** — combined with 24 prior spatial vectors, the **kinematic-primitive attack class is now EXHAUSTED at 31+ vectors, zero unconditional positive arrows**.

**Outcome: BOUNDED_OBSTRUCTION** (sharpening, not closure).

**DEPENDS ON** salvage commits on main; PR #695 (A3 axiom proposal context).

## Key structural finding

**C_3[111] acts only on spatial coordinates, leaving time fixed in the trivial 1D irrep**, so every retained time-direction primitive is C_3-invariant by construction. This is sharper than prior obstructions because it identifies a *common structural reason* spanning the time-direction attack class.

## Per-vector results (all FAIL)

| Vector | Failure reason |
|---|---|
| 1. Single-clock axis selection | "No spatial reflection is RP" theorem itself C_3-symmetric |
| 2. Lieb-Robinson direction-dep | v_LR scalar; cubic graph distance C_3-invariant |
| 3. Temporal RP breaks spatial C_3 | θ_t commutes with C_3 (numerically verified) |
| 4. OS time-orientation | Binary structure cannot match 3-fold orbit |
| 5. Modular conjugation J | Bisognano-Wichmann: J commutes with spatial isometries |
| 6. Fermion-doubling time-momentum | Each corner has identical temporal decomposition |
| 7. Spectrum condition + time | H ≥ 0 is C_3-invariant |

## Implications for A3-derivation campaign

The decisive routes remaining are **R1 (Higgs/Yukawa dynamical breaking)** and **R3 (anomaly inflow)**. These are dynamical/topological, NOT kinematic.

**R4 (Spin(6) chain) and R5 (no-proper-quotient sharpening) are also kinematic** and likely fall under the same exhaustion result this PR establishes — they will probably produce sharper obstructions rather than positive closure. (Both still running; results pending.)

## Test plan

- [x] `scripts/cl3_a3_route2_single_clock_2026_05_08_r2.py` runs **18/0 PASS** across 7 attack vectors
- [x] Direct numerical verification of θ_t / C_3 commutation on 5 lattice points
- [x] Authority disclaimer
- [x] Cross-references via clickable markdown links

## What this PR DOES close (bounded)

- The kinematic-primitive attack class is exhausted (31+ vectors)
- The structural reason is named: C_3 acts only spatially, time is C_3-trivial
- Future cycles can cite this as reusable negative evidence

## What this PR does NOT close

- AC_φ itself (still requires either dynamical C_3 breaking or axiom)
- L3a admission (unchanged at 1)

## Authority role

Source-note proposal. Audit verdict and downstream status set only by independent audit lane.


jonathonreilly:
Review-loop landed this as bounded A3 obstruction/hostile-review source material on main in commit f26425fe7. No A3 axiom was admitted, no parent status was promoted, and all new A3 rows remain unaudited for the independent audit lane.

PR #907, opened 2026-05-10
audit: criticality bumps don't force downstream re-audits; bumped row requests 2nd audit only

JONbridgerapps:
## Summary

The \`criticality_increased\` trigger in [docs/audit/scripts/invalidate_stale_audits.py](docs/audit/scripts/invalidate_stale_audits.py) was the dominant force-reset signature in recent audit churn (~246 hits across currently-unaudited rows). This PR refactors that path to match the user's policy reading:

> "Criticality bumps shouldn't force downstream re-audits — but the critical component does need a 2nd audit as it gets more important."

## Behavior under this PR

When a row's criticality bumps:

| Audit state | Bump target | Action |
|---|---|---|
| terminal non-clean (any) | any | **noop** — cross-conf doesn't apply to non-clean verdicts |
| \`audited_clean\`, weak indep | \`high\` / \`critical\` | **invalidate** — independence floor of FRESH_LOOK_REQUIREMENTS §4 fails |
| \`audited_clean\`, non-weak indep | \`leaf\` / \`medium\` / \`high\` | **noop** |
| \`audited_clean\`, non-weak indep, cross-conf present | \`critical\` | **noop** |
| \`audited_clean\`, non-weak indep, no cross-conf | \`critical\` | **soft_reset** |

**Soft-reset** transitions the row to \`audit_status = audit_in_progress\` + \`blocker = awaiting_cross_confirmation\`, with the live audit fields preserved as \`cross_confirmation.first_audit\` (mirrors apply_audit.py's first-pass flow byte-for-byte). The lane requests an independent second auditor at the new tier; the clean evidence stays standing.

## Crucially: downstream rows are not invalidated

A soft-reset row's \`effective_status\` stays at retained (via [compute_effective_status.is_criticality_bump_soft_reset](docs/audit/scripts/compute_effective_status.py)). Downstream rows whose audits closed against the first-pass clean evidence remain retained while the lane awaits the second audit. The dep_weakened path doesn't fire for soft-reset deps because their effective_status doesn't drop. **No cascade.**

If cross-confirmation eventually disagrees (\`cc.status = disagreement\`), the row exits the soft-reset state — its effective_status then drops normally and downstream rows see the chain genuinely break. The standard dep_weakened invalidation handles this case.

## Born-critical first-pass is unchanged

Apply_audit.py's standard flow for new critical claims (verdict comes in clean → row goes to \`audit_in_progress + awaiting_cross_confirmation\`) uses \`claim_type_provenance = "audited_pending_cross_confirmation"\` (without the \`_after_criticality_bump\` suffix). The soft-reset detector requires that exact suffix, so born-critical first-pass keeps the existing \`audit_in_progress\` effective_status and continues to block downstream chain closure until cross-confirmation lands. This matches established lane policy for new critical claims.

## Why this is policy-faithful

FRESH_LOOK_REQUIREMENTS §4 specifies WHEN cross-confirmation is required — at the critical tier. It does NOT say that requesting cross-confirmation after a criticality bump invalidates the prior audit's evidence or breaks downstream chains. The previous \`criticality_increased\` rule did so mechanically, far stricter than what §4 actually requires.

This PR aligns the mechanical behavior with §4: a criticality bump on a clean audit triggers a request for an independent second auditor, not a wholesale invalidation. The chain stays closed; only one row gets a second look.

## Tests

12 new unit tests:
- 9 in \`InvalidateStaleAuditsCriticalityBumpTest\` — 3-way categorization across all tier boundaries, soft-reset field shape matches apply_audit's first-pass flow byte-for-byte
- 3 in \`ComputeEffectiveStatusTest\` — soft-reset propagates as retained with downstream chain preserved, disagreement breaks the state, born-critical first-pass NOT affected

Full audit test suite: \`Ran 32 tests, OK\`. Live pipeline runs clean (\`bash docs/audit/scripts/run_pipeline.sh\` → 0 errors).

## Forward-only

Does NOT restore already-archived audits from prior over-aggressive invalidations. The 246 rows reset before this PR stay reset; a separate one-shot would walk \`previous_audits[]\` and unarchive verdicts where the new policy would have soft-reset instead.

## Test plan
- [ ] Reviewer confirms the §4 mapping in \`_categorize_criticality_bump\` matches their reading
- [ ] Reviewer confirms \`is_criticality_bump_soft_reset\` field shape exactly mirrors what apply_audit.audit_summary_from_row produces
- [ ] Confirm \`born-critical first-pass\` (apply_audit's existing flow) is genuinely unchanged in behavior
- [ ] Decide whether to also write the one-shot restoration (out of scope here)
- [ ] After merge, monitor the next nightly: \`criticality_bumped\` warnings to medium/high should drop to 0 for already-non-weak-indep rows; bumps to critical produce notices not warnings; downstream rows of bumped-to-critical claims stay retained while the second audit is awaited

### DIFF
--- a/docs/audit/scripts/compute_effective_status.py
+++ b/docs/audit/scripts/compute_effective_status.py
@@ -104,6 +104,40 @@ def clean_status(row: dict, dep_effective: dict[str, str]) -> tuple[str, str]:
     return retained_status, "self"
 
 
+def is_criticality_bump_soft_reset(row: dict) -> bool:
+    """True iff the row is in the criticality-bump soft-reset state:
+    `audit_status = audit_in_progress`, `blocker = awaiting_cross_confirmation`,
+    `claim_type_provenance` flags it as a criticality-bump soft-reset, and
+    `cross_confirmation.first_audit` is present with `status = awaiting_second`.
+
+    Set by `invalidate_stale_audits.soft_reset_to_cross_confirmation_pending`
+    when an existing audited_clean row's criticality bumps to `critical`
+    without prior cross-confirmation. The first-pass clean evidence remains
+    live in `cross_confirmation.first_audit`; the lane is just awaiting the
+    independent second auditor at the new tier.
+
+    For chain-closure purposes we treat such rows as if they were still at
+    their prior `audited_clean` effective_status: downstream rows whose
+    audits closed against this row's first-pass clean evidence are not
+    forced to re-audit just because the row's criticality changed. Once
+    cross-confirmation lands clean, nothing changes; if it disagrees, the
+    row exits this state (`cc.status = disagreement`) and downstream
+    chain closure is reassessed normally.
+    """
+    if row.get("audit_status") != "audit_in_progress":
+        return False
+    if row.get("blocker") != "awaiting_cross_confirmation":
+        return False
+    if row.get("claim_type_provenance") != "audited_pending_cross_confirmation_after_criticality_bump":
+        return False
+    cc = row.get("cross_confirmation") or {}
+    if not isinstance(cc, dict):
+        return False
+    if cc.get("status") != "awaiting_second":
+        return False
+    return cc.get("first_audit") is not None
+
+
 def intrinsic_status(row: dict, dep_effective: dict[str, str]) -> tuple[str, str]:
     if archived_failed_is_retained_no_go(row):
         return "retained_no_go", "archived_failed_no_go"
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed via review-loop in ac7aa7888 after source review, focused unit tests, strict audit lint, full audit pipeline, and diff checks. This preserves first-pass clean evidence on criticality bumps while requiring second-auditor cross-confirmation before final clean retention.

PR #953, opened 2026-05-10
audit: fix classifier sympy-alias blindness + add re-audit trigger for classifier-promoted runners

jonathonreilly:
## Summary

Survey of salvaged work across closed PRs (#782, #783, #785, #794, #796, #835, #888) showed the reviewer's policy was correct on every point — but the **underlying classifier tool had a systematic false-negative on the standard `import sympy as sp` idiom**.

This PR fixes the classifier. It also adds a re-audit trigger for the corrected case, but the trigger is intentionally strict (fires 0 rows today; correctly so).

## Honest scope

> "ok will those now audit retained instead of conditional?"

That question forced the right re-examination. The original (loose) trigger fired on 2 rows, but neither would transition on re-audit because the original conditional verdicts were about deps and scope, not class-A. The trigger was tightened in commit `2e437599d`; the 2 rows are restored to their prior conditional state. **No rows currently qualify for re-audit invalidation by this trigger.** The classifier fix is the real value.

## The classifier bug (fixed)

`docs/audit/scripts/classify_runner_passes.py` matched class-A patterns via literal regex:

```python
PATTERNS_A = [
    re.compile(r"sympy\.(?:simplify|Eq|expand|factor)\b"),
    ...
]
```

Quantified across the repo:
- **130 runners** use `import sympy as sp` (the canonical idiom)
- **1,880 class-A patterns** invisible to the classifier
- **129 of 130** misclassified despite doing real symbolic verification

A runner with 12 `sp.simplify(...)` calls (class-A in spirit) appeared as `dominant_class: D` because no class-A patterns matched. This propagated to `audited_conditional` verdicts on rows whose runners were actually correctly verifying their claims.

## Fix 1: classifier alias-resolution

`docs/audit/scripts/classify_runner_passes.py`:

- New `detect_module_aliases()` parses `import X as Y` lines
- New `class_a_patterns_for_source()` augments PATTERNS_A with alias-substituted regexes per-source
- Whitelisted modules: `sympy`, `numpy` (the only ones with class-A patterns matching the standard idiom)
- No false positives — augmented patterns only fire when the alias import is detected in that specific source file

### Measured classifier impact

| | Before | After | Delta |
|---|---:|---:|---:|
| Class A | 24 | **109** | **+85 (4.5×)** |
| Class B | 176 | 157 | -19 |
| Class C | 657 | 637 | -20 |
| Class D | 251 | 228 | -23 |
| None | 423 | 400 | -23 |

[...]

PR #982, opened 2026-05-10
audit: tighten classifier_promoted_to_class_A trigger + restore 2 prematurely-invalidated rows

jonathonreilly:
## TL;DR — fixes overreach in the just-landed #953

You correctly pushed back on #953 ("ok will those now audit retained instead of conditional?"). Honest answer was no — the 2 invalidated rows would re-audit to the same conditional verdict because their original verdicts were about deps and scope, not class-A.

This PR:
1. **Tightens the trigger** so it doesn't fire on cases where re-audit won't transition the verdict
2. **Restores the 2 rows** that #953 prematurely invalidated, back to `audited_conditional`

The classifier alias-resolution fix from #953 stays in main (real value: 84 runners correctly classified).

## What was wrong with #953's trigger

The `classifier_promoted_to_class_a` function used `runner_check_breakdown.A == 0` as a proxy for "auditor flagged missing class-A patterns." Survey of all 230 audited_conditional rows on main:

```
  66  missing_dependency_edge
  50  missing_bridge_theorem
  30  dependency_not_retained
  30  runner_artifact_issue       (already handled by trigger #5)
  23  other
   6  scope_too_broad
   3  compute_required
```

Only 16 rows mention "class-A" or "algebraic" in their re-audit notes. Most have audit-time `A == 0` incidentally — the auditor's primary concern was usually deps, scope, or missing bridges.

## The 2 rows being restored

| Row | Prior reason | Why re-audit wouldn't help |
|---|---|---|
| `cross_sector_a_squared_koide_vcb_bridge_promoted_via_v8_*` | dependency_not_retained | 3 of 5 deps still conditional; verdict stays the same |
| `scalar_selector_reviewer_package_2026-04-20` | scope (note imports unregistered surfaces) | scope issue is independent of runner class-A status |

Both rows have their prior fields restored from `previous_audits[-1]` (the archive entry created by #953's invalidation). The spurious `previous_audits[-1]` entry is dropped.

## Tightened trigger — now requires ALL of:

1. `audited_conditional` (unchanged)
2. `runner_check_breakdown.A == 0` at audit time (unchanged)
3. Current `dominant_class == "A"` AND `counts.A >= 5` (was: `counts.A > 0`)
4. Runner exists on disk (unchanged)
5. **NEW:** every direct dep is retained-grade (`retained` / `retained_no_go` / `retained_bounded`). Re-audit only helps if deps are clean.
6. **NEW:** prior `notes_for_re_audit_if_any` or `chain_closure_explanation` mentions class-A keywords: `class-a`, `class a`, `algebraic`, `register a current runner`, `register a runner`. Only fires when the auditor's stated reason was specifically about runner / class-A insufficiency.

`detect_invalidation` now passes `rows` to `classifier_promoted_to_class_a` so the dep-status check has access to the full ledger.

## Net effect

```
audited_conditional:  current main -> +0
classifier_promoted invalidations this run: 0
```

**0 rows currently qualify under the strict criteria.** This is correct behavior — the trigger is now a forward-looking mechanism that fires only when the right combination of (classifier upgrade + clean deps + class-A reasoning in audit notes) actually exists.

## Test plan

- [x] Trigger fires 0 rows on current main (correct conservative behavior)
- [x] 2 prematurely-invalidated rows correctly restored from `previous_audits` archive
- [x] `previous_audits` archives properly truncated (no orphan entries)
- [x] `bash docs/audit/scripts/run_pipeline.sh` clean to step 14/14
- [x] `audit_lint` clean
- [x] No new `audited_failed` rows introduced
- [ ] CI green on the branch

## Process note

I made these corrections originally on the closed `claude/audit-classifier-alias-fix-2026-05-10` branch — too late, since #953 had already closed/landed. Opening this as a separate PR on a fresh branch off current `origin/main` is the right pattern for fixing post-merge issues.

### DIFF
--- a/docs/audit/AUDIT_LEDGER.md
+++ b/docs/audit/AUDIT_LEDGER.md
@@ -22,11 +22,11 @@ Publication-facing tables MUST read `effective_status`; `claim_type` is the audi
 | **retained_no_go** | 120 |
 | **retained_bounded** | 230 |
 | open_gate | 11 |
-| unaudited | 1139 |
+| unaudited | 1137 |
 | meta | 79 |
 | ~~audited_numerical_match~~ | 18 |
 | ~~audited_renaming~~ | 17 |
-| ~~audited_conditional~~ | 227 |
+| ~~audited_conditional~~ | 229 |
 | ~~audited_failed~~ | 14 |
 | `decoration_under_cl3_color_automorphism_theorem` | 4 |
 | `decoration_under_ew_current_fierz_channel_decomposition_note_2026-05-01` | 1 |
@@ -40,12 +40,12 @@ Publication-facing tables MUST read `effective_status`; `claim_type` is the audi
 | audit_status | count |
 |---|---:|
 | `audited_clean` | 383 |
-| `audited_conditional` | 227 |
+| `audited_conditional` | 229 |
 | `audited_decoration` | 11 |
 | `audited_failed` | 58 |
 | `audited_numerical_match` | 18 |
 | `audited_renaming` | 17 |
-| `unaudited` | 1218 |
+| `unaudited` | 1216 |
 
 | claim_type | count |
 |---|---:|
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Superseding with a fresh PR off latest origin/main — this branch was authored before recent main movement and is now CONFLICTING. New PR will be referenced.

PR #985, opened 2026-05-10
audit: tighten classifier_promoted_to_class_A trigger + restore 2 prematurely-invalidated rows (corrects #953)

jonathonreilly:
## Why this PR exists

PR #953 landed the classifier alias-resolution fix (real value, kept) but ALSO landed a loose `classifier_promoted_to_class_a` re-audit trigger. The user pointed out that the trigger fired on 2 rows whose conditional verdicts weren't actually about missing class-A patterns — re-audit on those rows would just return the same conditional verdict (one for deps, one for scope).

This PR tightens the trigger and restores the 2 prematurely-invalidated rows.

## Restored rows

| Row | Prior reason | Why re-audit wouldn't help |
|---|---|---|
| `cross_sector_a_squared_koide_vcb_bridge_promoted_via_v8_*` | dependency_not_retained | 3 of 5 deps still conditional; verdict stays the same |
| `scalar_selector_reviewer_package_2026-04-20` | scope (note imports unregistered surfaces) | scope issue is independent of runner class-A status |

Both restored from `previous_audits[-1]` (the archive entries created by #953's invalidation). Each row's prior fields (audit_status, audit_date, auditor, runner_check_breakdown, etc.) copied back to the top level; the spurious `previous_audits[-1]` entry dropped.

## Tightened trigger

Now requires ALL of:

1. `audited_conditional` (unchanged)
2. `runner_check_breakdown.A == 0` at audit time (unchanged)
3. Current `dominant_class == "A"` AND `counts.A >= 5` (was `> 0`)
4. Runner exists on disk (unchanged)
5. **NEW:** every direct dep is retained-grade (`retained` / `retained_no_go` / `retained_bounded`). Re-audit only helps if deps are clean.
6. **NEW:** prior `notes_for_re_audit_if_any` or `chain_closure_explanation` mentions class-A keywords: `class-a`, `class a`, `algebraic`, `register a current runner`, `register a runner`. Only fires when the auditor's stated reason was specifically about runner / class-A insufficiency.

`detect_invalidation` now passes `rows` to `classifier_promoted_to_class_a` so the dep-status check has access to the full ledger.

## Why so strict — survey of conditional reasons

Across all 230 audited_conditional rows on main:

```
  66  missing_dependency_edge
  50  missing_bridge_theorem
  30  dependency_not_retained
  30  runner_artifact_issue       (already handled by trigger #5)
  23  other
   6  scope_too_broad
   3  compute_required
```

Only 16 rows mention "class-A" or "algebraic" in their re-audit notes. Most have audit-time `A == 0` incidentally — the auditor's primary concern was usually deps, scope, or missing bridges. A loose trigger creates false positives that waste auditor time.

## Net effect

```
audited_conditional:  current main -> +0 (2 rows restored, trigger fires 0)
classifier_promoted invalidations this run: 0
```

**0 rows currently qualify under the strict criteria.** This is correct behavior — the trigger is now a forward-looking mechanism that fires only when the right combination of conditions actually exists.

## What stays in main from #953

- `classify_runner_passes.py` alias-resolution: KEPT (84 runners correctly classified)
- `invalidate_stale_audits.py` trigger #6 declaration in docstring: KEPT
- `classifier_promoted_to_class_a` function: KEPT (just tightened in this PR)

## Process note

#953 closed at 16:20 UTC. I committed a correction at 17:09 UTC on the same (closed) branch — orphaned, since the reviewer's salvage-and-close means the branch is dead. The user flagged this. PR #982 was opened on the prior branch (also wrong, plus it became CONFLICTING as main moved). #982 closed; this PR is the fresh re-do off latest origin/main. Lesson captured in personal memory.

## Test plan

- [x] Trigger fires 0 rows on current main (correct conservative behavior)
- [x] 2 prematurely-invalidated rows restored from `previous_audits` archive
- [x] `previous_audits` archives properly truncated (no orphan entries)
- [x] `bash docs/audit/scripts/run_pipeline.sh` clean to step 14/14
- [x] `audit_lint` clean
- [x] No new `audited_failed` rows introduced
- [ ] CI green on the branch

## Supersedes

- PR #953 (the original — landed/closed; this corrects it)
- PR #982 (my first attempt at correction — was on stale branch, became CONFLICTING; closed)

### DIFF
--- a/docs/audit/AUDIT_LEDGER.md
+++ b/docs/audit/AUDIT_LEDGER.md
@@ -22,11 +22,11 @@ Publication-facing tables MUST read `effective_status`; `claim_type` is the audi
 | **retained_no_go** | 120 |
 | **retained_bounded** | 230 |
 | open_gate | 11 |
-| unaudited | 1138 |
+| unaudited | 1136 |
 | meta | 97 |
 | ~~audited_numerical_match~~ | 18 |
 | ~~audited_renaming~~ | 15 |
-| ~~audited_conditional~~ | 212 |
+| ~~audited_conditional~~ | 214 |
 | ~~audited_failed~~ | 14 |
 | `decoration_under_cl3_color_automorphism_theorem` | 4 |
 | `decoration_under_ew_current_fierz_channel_decomposition_note_2026-05-01` | 1 |
@@ -40,12 +40,12 @@ Publication-facing tables MUST read `effective_status`; `claim_type` is the audi
 | audit_status | count |
 |---|---:|
 | `audited_clean` | 383 |
-| `audited_conditional` | 212 |
+| `audited_conditional` | 214 |
 | `audited_decoration` | 11 |
 | `audited_failed` | 58 |
 | `audited_numerical_match` | 18 |
 | `audited_renaming` | 15 |
-| `unaudited` | 1235 |
+| `unaudited` | 1233 |
 
 | claim_type | count |
 |---|---:|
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Landed via review-loop on main as commit 5008a9fb5. Integrated the source trigger tightening and restored only the two #953 false-invalidated ledger rows onto current main, then regenerated audit outputs. Verified classifier_promoted_to_class_a currently fires 0 rows, the two restored rows are audited_conditional again, pipeline/strict lint/diff check pass, and no new audit verdict was created.

PR #1140, opened 2026-05-15
science: bounded np-strong-cp single-plaquette F-tilde-F obstruction

jonathonreilly:
## Summary

Source-only PR adding a bridge derivation for the parent
`STRONG_CP_THETA_ZERO_NOTE.md` `audited_conditional` row. The audit verdict's
first Path A repair target is *"a retained operator-basis / action-surface
theorem deriving from `Cl(3)/Z^3` primitives ... that no gauge-invariant CP-odd
theta term is an admissible action slot."* This note bounds the obstruction to
the **single-plaquette action class** shared by Wilson, Heat-Kernel, and Manton
(every action candidate surviving the existing action-form-uniqueness no-go).

Within that class:
- the leading imaginary trace `Im tr U_P` is `O(a^6)` and **cubic-in-F**
  (single-plane);
- the topological density `F̃F` is `O(a^4)` and **bilinear-in-F** (two-plane
  via Levi-Civita);
- the lattice topological density is intrinsically a **two-plaquette
  (clover)** construct;
- therefore `theta_bare == 0` structurally on the single-plaquette action
  class.

The note explicitly does **not** address multi-plaquette CP-odd operators or
the positive real quark-mass orientation (`arg det(M_u M_d) = 0`) — the
parent's second Path A target. Audit authority remains the independent audit
lane.

## Files

- `docs/NEWPHYSICS_NP_STRONG_CP_THETA_NOTE_2026-05-10_npCP.md` — source theorem note (~290 lines)
- `scripts/frontier_np_strong_cp_theta_obstruction.py` — companion runner (~270 lines)
- `logs/runner-cache/frontier_np_strong_cp_theta_obstruction.txt` — cached runner output

## Runner output

```
PASS=13  FAIL=0
```

Checks:
1. Canonical su(3) generator algebra: `Tr(T_a T_b) = δ_{ab}/2`, `tr T_a = 0`
2. Wilson + Heat-Kernel single-plaquette actions are real on random SU(3)
3. CP transformation flips `Im tr U_P`, preserves `Re tr U_P`
4. Small-a scaling: `|Im tr U_P| ~ a^6` (slope = 6.000), not `a^4`
5. Single-plane cubic-in-F fit: coefficient = 1.0000, residual 5e-7
6. F̃F density is two-plane / bilinear-in-F and scales as `a^4` (slope = 4.000)
7. No single-plaquette linear combination of `Im tr U_P` fits F̃F density
   (best least-squares relative residual = 0.96 ≈ 1)

[...]

PR #1151, opened 2026-05-15
science: bounded np-strong-cp single-plaquette F-tilde-F obstruction (rebased on main, replaces #1140)

jonathonreilly:
**Replaces #1140** (CONFLICTING after main moved). Same source content rebased onto current main.

Includes the npCP source theorem note + runner + cache, plus the language-accuracy fix (theta_bare overclaim removed; action-class admission explicit) from the original #1140 PR's hostile review pass.

CI prep included: pipeline regen + cache refresh aligned to current main.

Closes #1140 (use this PR instead).


jonathonreilly:
Closing as redundant. The user is using the source-only review-loop salvage pattern directly from the original PRs (e.g., commits like `fb3fc2bc3`, `7d6129d13`, `e553b9f46`). Use original #1140 instead — it has the source triple alongside (now-redundant) audit-data churn that the salvage workflow ignores.

jonathonreilly:
Review-loop landed on main in 49d03ee87. I applied only the source triple, kept the strong-CP bridge bounded to the admitted single-plaquette action class, changed parent-status language from 'closes' to 'addresses', refreshed its paired cache, regenerated audit surfaces, and left the row unaudited for independent audit.

PR #1141, opened 2026-05-15
science: new-physics probe neutrino PMNS circulant ansatz (npNu)

jonathonreilly:
## Summary

- Bounded source-only probe (1 theorem note + 1 runner + 1 cache) testing parallel `C_3[111]`-equivariant Hermitian circulant ansatz `H_nu = a_nu I + b_nu C + b_nu^conj C^2` for the neutrino sector.
- Four hypotheses tested numerically (BAE-inherited, Brannen `2/9 + pi/12`, BAE+free-delta, free-rho+free-delta); runner `PASS = 17, FAIL = 0`.
- Surfaces four open derivation targets: `rho_nu` selection principle, `delta_nu` derivation, `sqrt(m)` positive parent for neutrinos, `U_nu = DFT_3` over-mixing fix.

## Key findings

- **Algebraic identity (B3):** `Q_nu = (1 + 2 rho_nu) / 3` — Koide-Q and amplitude ratio are NOT independent. Literature "Foot-Brannen `Q_nu ~ 1/3`" maps to `rho_nu = 0` (degenerate masses, empirically wrong).
- **H_B (Brannen `2/9 + pi/12`):** numerical failure — `Dm^2_21 / Dm^2_31 = 0.17` vs observed `0.029` (6x overshoot).
- **H_C (BAE + free delta):** admits a delta fit to splitting ratio (`rel err 0.15%`), but the angle is not structurally distinguished.
- **H_D (free rho, free delta):** 1-D matching curve in parameter plane; no unique pin without further structure.
- **PMNS angle obstruction (P3):** pure-circulant `U_nu = DFT_3` gives `sin^2 theta_13 = 1/3`, a 15x overshoot of observed `0.022`. The retained PMNS-as-`f(H)` package's non-circulant affine tilt is therefore STRUCTURALLY REQUIRED.
- **Cross-check (P9):** probe's best-fit point satisfies all retained inequality bounds (`Sigma m_nu > 50.58 meV`, `m_beta <= 50.58 meV`, `m_betabeta <= 50.58 meV`, `Dm^2_21 < Dm^2_31`, `m_3 = 5.06e-2 eV`).

## What this probe does NOT claim

- Not a retained derivation of `Dm^2_21`, `delta_CP`, or PMNS angles.
- Not a retained Koide-like relation for neutrinos (algebraically tied to `rho_nu`).
- Not a promotion of the retained PMNS P3 package.
- Not a new axiom (respects A1 + A2 only).
[...]

PR #1152, opened 2026-05-15
science: new-physics probe neutrino PMNS circulant ansatz npNu (rebased on main, replaces #1141)

jonathonreilly:
**Replaces #1141** (CONFLICTING after main moved). Same source content rebased onto current main.

Includes the npNu source theorem note + runner + cache, plus the admission count consistency fix (4 → 7) from the original #1141 PR's hostile review pass.

CI prep included: pipeline regen + cache refresh aligned to current main.

Closes #1141 (use this PR instead).


jonathonreilly:
Closing as redundant. The user is using the source-only review-loop salvage pattern directly from the original PRs (e.g., commits like `fb3fc2bc3`, `7d6129d13`, `e553b9f46`). Use original #1141 instead — it has the source triple alongside (now-redundant) audit-data churn that the salvage workflow ignores.

jonathonreilly:
Review-loop landed on main in 49d03ee87. I applied only the source triple, kept the neutrino PMNS circulant result as a bounded probe with explicit admissions, fixed the DFT_3 PMNS angle wording, refreshed its paired cache, regenerated audit surfaces, and left the row unaudited for independent audit.

PR #1165, opened 2026-05-16
ci(audit): regenerate conditional-prompt snapshot on nightly cron

jonathonreilly:
## Summary

Adds one cron-only workflow step that regenerates `docs/audit/MISSING_DERIVATION_PROMPTS.md` so the science-fix-loop sees the current ledger state instead of a stale snapshot. Total diff: **10 lines, one file**.

## Why

`MISSING_DERIVATION_PROMPTS.md` is the static snapshot consumed by `scripts/science_fix_loop.py` (and by hand-driven Claude agent batches). It lists rows that need repair work — `audited_conditional` / `audited_renaming` / `audited_failed` / `audited_numerical_match` / `open_gate` rows, with the auditor's repair target inlined.

The file is regenerated by `docs/audit/scripts/generate_conditional_prompts.py`, which reads the live `audit_ledger.json` and writes the snapshot. Before this PR it was only refreshed manually — drift built up.

Concrete recent example (2026-05-16 batch):
- `lensing_deflection_note` was re-audited to `retained_bounded` on 2026-05-10
- `equivalence_principle_note` was demoted then settled at `audited_failed` on 2026-05-05
- Both remained listed in the prompts file as of the last commit (2026-05-08)
- A batch of 8 science-fix agents wasted 2 slots on these stale entries before correctly declining

## Changes

1. **New step after `Run audit pipeline`, cron/workflow_dispatch only:**
   ```yaml
   - name: Regenerate conditional-prompt snapshot (cron/manual only)
     if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
     run: python3 docs/audit/scripts/generate_conditional_prompts.py
   ```
   PR runs deliberately skip this — otherwise it would trip the existing "Check pipeline diff" guard.

2. **Same call added inside the rebase retry block** so push-fail rebases include the regenerated snapshot rather than a stale one from before the rebase.

## What it does NOT change

- No change to the on-PR workflow path. PRs continue to validate the pipeline output and fail on unexpected diffs.
- No new permissions, no new dependencies (the script already exists and is exercised manually).
- No data files modified in this PR — the first refresh will land naturally on the next 06:00 UTC cron run as part of the existing auto-commit.
- No change to the audit lane semantics.

## Test plan

- [x] `python3 docs/audit/scripts/generate_conditional_prompts.py` runs cleanly locally on `origin/main` and produces a deterministic diff (verified by re-running and confirming md5 stable).
- [x] Workflow YAML is structurally valid (12 steps, 3 PR-gated, regenerator step gated correctly).
- [ ] Next 06:00 UTC cron run picks up the new step (auto-validates).
- [ ] Confirm subsequent science-fix-loop batch sees pruned prompts.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -92,6 +92,15 @@ jobs:
       - name: Run audit pipeline
         run: bash docs/audit/scripts/run_pipeline.sh
 
+      - name: Regenerate conditional-prompt snapshot (cron/manual only)
+        if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+        # MISSING_DERIVATION_PROMPTS.md is the static snapshot the
+        # science-fix-loop reads. Without this refresh it drifts from the
+        # live ledger, causing wasted attempts on rows that have already
+        # been re-audited. PR runs deliberately skip this step so they
+        # don't trip the "Check pipeline diff" guard below.
+        run: python3 docs/audit/scripts/generate_conditional_prompts.py
+
       - name: Check pipeline diff (PR runs only)
         if: github.event_name == 'pull_request'
         run: |
@@ -150,6 +159,7 @@ jobs:
             git reset --hard origin/main
             python3 docs/audit/scripts/repair_missing_dependency_edges.py --apply
             bash docs/audit/scripts/run_pipeline.sh
+            python3 docs/audit/scripts/generate_conditional_prompts.py
             commit_if_needed || exit 0
           done
 


jonathonreilly:
Review-loop landed this on main in 3dfb122cf. I kept the cron/manual workflow repair and also regenerated  once from the current ledger so the stale prompt snapshot is fixed immediately. Pipeline/lint passed; no audit verdicts were applied.

jonathonreilly:
Closing because the reviewed content has landed on main in 3dfb122cf.

jonathonreilly:
Correction to the previous shell-mangled comment: review-loop landed this on main in 3dfb122cf. I kept the cron/manual workflow repair and also regenerated `docs/audit/MISSING_DERIVATION_PROMPTS.md` once from the current ledger so the stale prompt snapshot is fixed immediately. Pipeline/lint passed; no audit verdicts were applied.

PR #1178, opened 2026-05-16
science-fix: born_scattering_comparison_note (numerical_match repair, for review)

jonathonreilly:
## What

Withdraw two leftover overclaiming sentences in `docs/BORN_SCATTERING_COMPARISON_NOTE.md` that survived the 2026-05-10 PATH B demotion but are inconsistent with it:

1. **'β=0.8 provides the specific correction that shifts −1.28 → −1.43'** — directly refuted by the runner's own beam-correction table later in the same note (β=0.8 gives slope −0.35 in 2D and −0.77 in 3D; both worsen the match, not improve it).
2. **'right functional form' / 'wave-mechanical effect'** — framed a structural interpretation the runner does not establish.

## Why

Per the b3 science-fix prompt: the row is `audited_numerical_match` (class G, 439 descendants). The honest path for class-G numerical-match claims is structural demotion, which already happened (PATH B, commit `bc1664777`, 2026-05-10). This pass just removes the stale structural-interpretation sentences so the note reads consistently as a bounded numerical observation on the next re-audit pass.

## Scope discipline

- Note-only change. No `scripts/`, no `docs/audit/data/`, no audit-data, no verdict-claim edits.
- Replaced text is explicitly marked **withdrawn** in-line so the change is auditable.
- No new retained-grade premises introduced; no structural theorem claimed.
- Runner still produces the exact slopes the note cites (verified locally: plane-wave −1.2753, 2D-beam −0.3485 at β=0.8).

## Re-audit notes

The next audit pass should read this as a `bounded_theorem` with the same PATH B scope as 2026-05-10. The three open gaps (retained lattice slope, retained 3D beam correction, bridge to grown-DAG lensing readout) remain explicit and deferred to future-work retained promotion, unchanged.

For review. Not self-graded.

### DIFF
--- a/docs/BORN_SCATTERING_COMPARISON_NOTE.md
+++ b/docs/BORN_SCATTERING_COMPARISON_NOTE.md
@@ -165,23 +165,46 @@ Power-law fit on b ∈ {3..6}: **I_geom ≈ 2.48 · b^(−1.28)**, R² = 0.999.
 | Steepening with b? | Yes | Yes |
 | L-dependent? | Yes (strongly) | **No** (Lane L++) |
 
-## The discrepancy
+## The discrepancy (open-gap diagnosis, NOT explained in this packet)
 
-The slope difference Δ = 0.15 is small but real. Three factors contribute:
+The slope difference `Δ = 0.15` between the runner-computed plane-wave
+eikonal slope and the imported lattice slope is **not explained** by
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the Born scattering interpretation repair directly on main in a45ed3dd4. The stale numerical-match framing was demoted without over-promoting the comparison.

PR #1197, opened 2026-05-16
science-fix: dm_leptogenesis_pmns_projector_interface_note_2026-04-16 (bridge work, for review)

jonathonreilly:
## Summary

Class A bridge repair on the PMNS projector interface note, targeting the auditor's missing_bridge_theorem flag.

The auditor's complaint was that the runner's Part 4 carrier and column-selection claims were asserted with unconditional `True` checks rather than proved from any retained authority. The note itself explicitly disclaims that the positive PMNS pair law and physical N1 column selection remain missing on the DM branch, so the unconditional Part 4 passes masked a conditional-only shape.

## Changes

1. **Runner Part 1 (algebraic interface theorem).** The pair-to-projector packet law is now verified on the canonical sample plus eight deterministic random Hermitian pairs (72 rephasing samples in total) at numerical-zero accuracy. The previous unconditional "pair-readable" `True` is replaced by an executable consequence of the unitary-doubly-stochastic and rephasing-invariance checks.

2. **Runner Part 4 (conditional structure).** The previous three unconditional `True` checks on the carrier, projector-interface, and reduction claims are replaced by three structural checks that explicitly name the two open dependencies (carrier authority on the active neutrino lane; physical column authority on the same lane) and the one closed algebraic step (Part 1). The runner now asserts the open dependencies are explicitly unresolved rather than asserting them resolved.

3. **Source note.** New "Conditional structure (explicit)" section makes the same factoring visible in the documentation.

Runner: 12/12 PASS. No audit-data, no audit-docs, no publication control-plane touched.

## Scope of repair

This is a bridge tightening, not a tier promotion. The conditional shape remains conditional, but it is now honest about being conditional in both the source note and the runner output. The two named open dependencies (positive PMNS pair law; physical N1 column selection) remain explicit named gaps for downstream work.

## Test plan

- [x] Runner exits cleanly with 12/12 PASS
- [x] Cached output refreshed
- [x] No audit-data, audit-docs, or publication paths modified
- [ ] Re-audit decides whether the new explicit conditional structure plus algebraic Part 1 closes the missing_bridge_theorem flag for this leaf

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/docs/DM_LEPTOGENESIS_PMNS_PROJECTOR_INTERFACE_NOTE_2026-04-16.md
+++ b/docs/DM_LEPTOGENESIS_PMNS_PROJECTOR_INTERFACE_NOTE_2026-04-16.md
@@ -132,6 +132,34 @@ pair law itself is not yet derived on the DM branch.
 
 It is an interface theorem plus a diagnostic transplant.
 
+## Conditional structure (explicit)
+
+This note's content factors into one closed algebraic step plus two named open
+dependencies. The runner's Parts 1 and 4 make this structure executable, so the
+conditional shape is legible from the runner output rather than masked by
+unconditional bottom-line assertions.
+
+- **Closed (algebraic):** given any positive-definite Hermitian pair
+  `(H_nu, H_e)`, the flavored transport packet
+  `P_i(alpha) = |U_PMNS(alpha,i)|^2` is doubly stochastic, phase-insensitive,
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the DM leptogenesis PMNS projector interface repair directly on main in f1175f630. I normalized the baseline Cl(3) on Z^3 language and regenerated the runner cache.

PR #1215, opened 2026-05-16
science-fix: yt_bridge_nonlocal_corrections_note (numerical_match repair, for review)

jonathonreilly:
## Summary

Honest demote of `yt_bridge_nonlocal_corrections_note` (audited_numerical_match, class G, descendants 324).

The auditor flagged three issues: hard-coded external physical inputs, a target-`y_t` viability filter, and best rows from a chosen family grid. The note text already self-classified as a "bounded support note," but the runner's load-bearing checks were still target-conditioned.

This revision aligns the runner with the note's bounded-support framing by:

- Removing the target-`y_t` viability filter (`abs(dev_pct) >= 0.5`)
- Removing the 7x8 family grid search
- Replacing load-bearing checks with the family-agnostic Cauchy-Schwarz bound

      |<K_nonloc, phi>| <= ||K_nonloc||_2 * ||phi||_2

  on every profile `phi` in L^2 on the forced UV window `x >= 0.95`
- Reporting the absolute operator-norm constant (2.799e-5) alongside the ratio (5.024e-3) so the Cauchy-Schwarz bound is usable directly
- Retaining 3 reference family rows at a fixed reference center/width as a non-load-bearing sanity comparison only

The operator-norm constant is intrinsic to the kernel pair `(K, K_loc)` given the standard SM inputs at `M_Z`; it does not depend on any family selection or target-`y_t` filter. The structural derivation of the operator-norm bound from the exact interacting bridge action remains open and is flagged as the upstream review target carried by the dependency links.

## Files

- `docs/YT_BRIDGE_NONLOCAL_CORRECTIONS_NOTE.md`: rewrite Scope + Result + Honest auditor read sections.
- `scripts/frontier_yt_bridge_nonlocal_corrections.py`: drop viability filter and grid search; 4 family-agnostic load-bearing checks.
- `logs/runner-cache/frontier_yt_bridge_nonlocal_corrections.txt`: refreshed via `precompute_audit_runners.py`.

No `audit-data/`, `audit-docs/`, or `publication/` files touched. No verdict claims, no overclaim, no invented retained premises.

## Test plan

- [x] `python3 scripts/frontier_yt_bridge_nonlocal_corrections.py` -> 4 PASS / 0 FAIL
- [x] Cache regenerated via `precompute_audit_runners.py --runners ... --force`
- [ ] Re-audit of this note should land at `audited_clean` or `audited_conditional` rather than `audited_numerical_match`, since the load-bearing step is now a family-agnostic Cauchy-Schwarz bound on the operator-norm distance between the exact kernel and its affine local-Hessian model — not a target-conditioned best-family search.

### DIFF
--- a/docs/YT_BRIDGE_NONLOCAL_CORRECTIONS_NOTE.md
+++ b/docs/YT_BRIDGE_NONLOCAL_CORRECTIONS_NOTE.md
@@ -1,7 +1,7 @@
 # `y_t` Bridge Nonlocal Corrections Note
 
-**Date:** 2026-04-15
-**Status:** bounded support note
+**Date:** 2026-04-15 (revised 2026-05-16)
+**Status:** bounded support note — Cauchy-Schwarz family-agnostic bound
 **Primary runner:** `scripts/frontier_yt_bridge_nonlocal_corrections.py`
 
 ## Role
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed by review-loop directly on main in commit 5fb7f4b50e4b0d60fc63f4bdd7eb70e878744312. Source changes were replayed onto current main, generated audit/status artifacts were regenerated locally, runner caches are fresh, strict audit lint passed, and no new audit graph cycles were introduced.

PR #1221, opened 2026-05-16
science-fix: koide_delta_lattice_wilson_selected_eigenline_no_go_note_2026-04-24 (failed repair, for review)

jonathonreilly:
## Summary

Reconciles the source note's stale verification/cache against the current
runner output, and narrows the load-bearing claim so the retained no-go is
only the rank-two character-sector obstruction (plus endpoint-lift residual)
— not the `r`-dependent ambient eta-proxy line that the third-pass judicial
auditor flagged.

## Auditor verdict being addressed

`audited_conditional` (judicial third pass, sided with second auditor):

> The current runner/source support the finite Wilson/Z3 rank-two no-go core,
> but the source note asserts a 13/14 run, a FAIL on A.3, and
> KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO=FALSE as the verbatim
> cache, while the supplied executable evidence reports 14/14 PASS and the
> no-go flag TRUE. Under the audit rubric, completed runner/pass-accounting
> and flag mismatches are substantive artifact defects that block a clean
> cite until reconciled, even when the bounded computational core is
> otherwise supported.

## Root cause

A prior sweep commit reverted the runner to the pre-repair form (A.3 check
asserting "eta is not 2/9" at `r = 1.0`) while leaving the runner-cache file
from the repaired run in place. The cache header's pinned SHA matched the
new runner, but the cache stdout (eta = 2/9, A.3 [FAIL], 13/14, =FALSE) was
the output of the prior version (at `r = 1.425`). Source note mirrored the
stale cache, so both disagreed with what the runner actually prints.

## Changes (no audit-data, no audit-docs, no publication)

1. `logs/runner-cache/frontier_koide_delta_lattice_wilson_selected_eigenline_no_go.txt`
   refreshed via `scripts/cached_runner_output.py --refresh`. New cache
   matches the runner's actual stdout: 14/14 PASS, eta = 0.069214821267 at
   `r = 1.0`, A.3 PASS, `KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO=TRUE`.
2. `docs/KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO_NOTE_2026-04-24.md`
   updated Result + Verification sections to match the refreshed cache
   verbatim, with the runner's `r = 1.0` eta value quoted.
3. Same note adds a new "Scope of the retained claim" section that
   explicitly demotes `RESIDUAL_AMBIENT` to an `r`-dependent observation
   (collapses to 2/9 exactly at `r = 1.425`) rather than a free-standing
   ambient-eta obstruction. The rank-two character sector, the CP^1
   eigenline-family freedom, the algebraic selected/spectator residual, and
   the endpoint-lift residual remain `r`-independent and load-bearing.

## Test plan

- [x] `python3 scripts/cached_runner_output.py scripts/frontier_koide_delta_lattice_wilson_selected_eigenline_no_go.py --check-only` returns `fresh`
- [x] `python3 scripts/frontier_koide_delta_lattice_wilson_selected_eigenline_no_go.py` exits 0 with 14/14 PASS and `=TRUE` matching the cache
- [x] No files outside `docs/KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO_NOTE_2026-04-24.md` and its paired runner-cache changed; audit-data/audit-docs/publication untouched
- [ ] Re-audit on this branch to confirm whether the narrowed claim is now `audited_clean` or whether the rank-two obstruction itself needs further support

### DIFF
--- a/docs/KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO_NOTE_2026-04-24.md
+++ b/docs/KOIDE_DELTA_LATTICE_WILSON_SELECTED_EIGENLINE_NO_GO_NOTE_2026-04-24.md
@@ -17,11 +17,16 @@ Wilson operator + body-diagonal Z3 action
 
 ## Result
 
-Negative.  In this finite Wilson realization, the ambient eta proxy is not the
-exact APS value `2/9`; the runner reports the mismatch rather than promoting
-it.  Independently, the zero-mode character sector relevant to the selected
-endpoint has rank two.  The retained data select a spectral
-projector/eigenspace, not a unique rank-one line inside it.
+Negative for the selected-eigenline route.  The zero-mode character sector
+relevant to the selected endpoint has rank two: the retained Wilson data
+select a spectral projector/eigenspace, not a unique rank-one line inside
+it.  This obstruction is structural and `r`-independent.
+
+The runner additionally records, at its frozen `r = 1.0` setting, that the
+finite Wilson eta proxy differs from the exact APS value `2/9`.  That
+observation is `r`-dependent (at `r = 1.425` the same proxy equals `2/9`
+exactly) and is not the load-bearing part of this no-go; see "Scope of the
+retained claim" below.
 
 The runner constructs two orthonormal zero-mode lines with the same spin-lift
 `Z3` character.  Every normalized mixture:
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed by review-loop directly on main in commit 5fb7f4b50e4b0d60fc63f4bdd7eb70e878744312. Source changes were replayed onto current main, generated audit/status artifacts were regenerated locally, runner caches are fresh, strict audit lint passed, and no new audit graph cycles were introduced.

PR #1371, opened 2026-05-17
[physics-loop] audit-packet: helper_runner_paths in pipeline + fixes class-C from incomplete packets

jonathonreilly:
## Summary

Fixes the recurring `audited_conditional class C` pattern caused by packet incompleteness — observed today in `lattice_distance_law_note` and present across 74 pending + 138 already-audited claims with helper-module imports.

**Root cause:** Audit packet contains source note + primary runner + cache. But primary runners often `import` from `scripts/*.py` helpers. The auditor sees opaque function imports (`compute_field_at_b`, `generate_lattice_mirror`, etc.) and correctly falls back to class C.

**The fix lives in the pipeline.** New `helper_runner_paths` field flows from citation graph → audit ledger → audit packet (per updated AUDIT_AGENT_PROMPT_TEMPLATE.md).

### `docs/audit/scripts/build_citation_graph.py`

- New `_parse_script_imports(path)` — AST walker over the primary runner; extracts `from scripts.X import ...` / `import scripts.X` / relative `from .X import ...` patterns
- New `resolve_helper_runner_paths(primary)` — transitive-closure walker; returns sorted `scripts/X.py` paths the audit packet must include
- Each node now carries `helper_runner_paths` next to `runner_path`

### `docs/audit/scripts/seed_audit_ledger.py`

- Persists `node["helper_runner_paths"]` as `row["helper_runner_paths"]` in audit_ledger.json
- Added to both fresh-seed and preserve-prior code paths
- Existing audit verdicts unaffected (`note_hash` unchanged → status preserved)

### `docs/audit/AUDIT_AGENT_PROMPT_TEMPLATE.md`

New §3b "Helper runner source code (audit packet must include)":

- Orchestrator MUST include all paths from `helper_runner_paths` in the restricted packet
- Helper sources count toward class judgment (hard-coded constants → class G; genuine compute extends class C)
- Missing helper named in `helper_runner_paths` → `audited_conditional` with `repair_class=missing_helper_runner`

## `docs/LATTICE_DISTANCE_LAW_NOTE.md` fix

The user-reported case. The note declared its runner in a markdown TABLE format (`| Active runner | [scripts/...](...) |`) that build_citation_graph's RUNNER_LABEL_RE couldn't match. Result: `runner_path: None` → helper resolution returned `[]` even with the new code.

Added explicit top-level metadata:
```
**Primary runner:** scripts/lattice_no_barrier_distance.py
**Helper runners (audit packet must include):** scripts/lattice_mirror_distance.py
[...]

PR #1405, opened 2026-05-17
[REVIEWER-AUTHORIZATION-REQUIRED] audit-pipeline: helper_runner_paths integration (closes class-C-from-missing-helpers)

jonathonreilly:
## Why this isn't theory (please don't auto-strip)

This is a **pipeline-tier bug fix**, not a theory PR. The source-only review-loop policy correctly held back prior pipeline modifications in PR #1371. This PR re-submits just the pipeline portion, deliberately separated so it can be reviewed on infrastructure-correctness grounds rather than auto-stripped.

**The bug is observed, reproducible, and named:**

- `lattice_distance_law_note: audited_conditional, class C` — chain depends on `scripts/lattice_mirror_distance.py`, which was not in the restricted packet.
- `scripts/lattice_no_barrier_distance.py` (the primary runner) literally has `from scripts.lattice_mirror_distance import compute_field_at_b, generate_lattice_mirror, propagate` at line ~23.
- 79 pending + 133 already-audited claims have helper imports → vulnerable to the same packet-incompleteness defect.

**The data side already landed (you salvaged it):** `docs/audit/data/audit_packet_script_deps.json` (12431-line dep map) is on main via commit `1a67d41ab`. Without this PR's pipeline integration, that JSON is unmaintained — `helper_runner_paths` drifts every time runners gain or drop imports.

### `docs/audit/scripts/build_citation_graph.py`
- Adds `_parse_script_imports` (AST walker: `from scripts.X import`, `import scripts.X`, relative `from .X import` inside `scripts/`) and `resolve_helper_runner_paths` (transitive closure).
- Adds `helper_runner_paths` to each graph node next to `runner_path`.
- **Why needed:** source notes don't list helper imports as markdown links — they're Python `import` statements. AST walking is the only programmatic way to find them. Without this step, the helper data never reaches the ledger.

### `docs/audit/scripts/seed_audit_ledger.py`
- Persists `node["helper_runner_paths"]` as `row["helper_runner_paths"]`. Added to fresh-seed and preserve-prior branches.
- **Why needed:** the ledger is the orchestrator's authoritative read surface. If the field isn't in the ledger row, the orchestrator can't include helpers.

### `docs/audit/AUDIT_AGENT_PROMPT_TEMPLATE.md` — new §3b
- New section "Helper runner sources (transitive imports from the primary runner)" with `{{HELPER_RUNNER_SOURCES}}` substitution marker.
- Documents protocol: treat helpers as part of the chain for class judgment; missing-but-named helper → `audited_conditional` with `repair_class=missing_helper_runner`.
- Renames the prior fallback subsection from §3a tail to §3c. Additive only.
- **Why needed:** the auditor reads the template, not the JSON map. Without §3b, the auditor doesn't know helpers are supposed to be in the packet and can still return class-C as before.

### `docs/LATTICE_DISTANCE_LAW_NOTE.md`
- Adds explicit `**Primary runner:**`, `**Helper runners:**`, `**Scope-probe runner:**` metadata at the top.
- **Why needed:** this note declared its runner only in a markdown table cell (`| Active runner | [scripts/...] | ...`) which `RUNNER_LABEL_RE` does not match. Result: `runner_path: None` in the ledger, defeating the rest of the fix for this specific note (the verified-bug case from your audit). The metadata format matches what `RUNNER_LABEL_RE` already extracts.

## Compatibility with existing audit verdicts

The field is **additive**. Note hashes are unaffected (no source-note content changes besides the deliberate lattice metadata re-queue). `seed_audit_ledger` preserves prior audit fields when `note_hash` is unchanged.

**Local verification (full pipeline run):**

```
build_citation_graph.py:   1957 runners attached, 209 helpers populated
seed_audit_ledger.py:      2227 preserved, 1 re-audit required (lattice, intentional)
compute_audit_queue.py:    ready count 38 (up from 4)
compute_effective_status.py: clean
audit_lint.py:             OK (no errors)
```

## What I'm asking

**Please review on infrastructure-correctness grounds:**

- AST walker correct? (handles `from scripts.X`, `import scripts.X`, relative `from .X`, filters to files that exist)
- Field persistence additive? (yes — both fresh-seed and preserve-prior paths set it; nothing else changed)
- §3b extension well-scoped? (yes — describes packet contents + class-judgment protocol + missing-helper repair class; renames fallback section to keep numbering clean)
- Lattice-note metadata appropriate? (yes — matches `RUNNER_LABEL_RE` patterns; the runner was always there, just declared in a format the regex didn't catch)

**If anything looks wrong, demote/strip as usual.** I'm not asking for auto-acceptance — only for review as pipeline-tier work rather than auto-stripping because it touches pipeline files.

## What I'm NOT asking

- Not asking you to abandon the source-only policy for the general case
- Not asking for blanket approval of future pipeline modifications
- Not making any science claims — no theorem, no derivation, no audit verdict

Future audit-prep findings will continue to land via the normal source-only path. The **26 re-queue PRs (#1374-#1399)** are independent and should land via normal source-only salvage regardless of this PR's fate.

## Expected impact (mechanical, no science changes)

- **Immediate**: 79 pending audits with helpers get complete packets in next audit run → no spurious class-C verdicts from packet incompleteness alone
- **Retrospective**: 133 already-audited claims could be considered for re-audit if their existing class-C was driven by this bug (the 26 re-queue PRs cover the strong-signal subset)
- **lattice_distance_law_note specifically**: re-queued by note-hash change in this PR; next verdict sees the complete chain


jonathonreilly:
Landed via review-loop in 860436c2e. Reviewed as pipeline-tier infrastructure, not theory. Accepted the AST helper import walker, ledger persistence, audit prompt section 3b, and lattice runner metadata. I adjusted the prompt to use the existing runner_artifact_issue repair prefix instead of introducing a new missing_helper_runner class. Regenerated the audit pipeline; lattice_distance_law_note is unaudited/ready with helper runner scripts/lattice_mirror_distance.py.

PR #1518, opened 2026-05-17
methodology: add five-judge audit disagreement policy

jonathonreilly:
## Summary
- updates the repo audit-loop skill so critical cross-confirmation disagreements go directly to a five-judge panel
- removes the previous single third-auditor resolution path
- defines the 3-of-5 majority tuple and human-review fallback cases

## Verification
- git diff --check

## Notes
- Also updated the installed local audit-loop skill copy to match this branch so subsequent local audit runs use the same policy immediately.

### DIFF
--- a/docs/audit/data/citation_graph.json
+++ b/docs/audit/data/citation_graph.json
@@ -39031,7 +39031,7 @@
         "ai_methodology.skills.no-go-discipline.skill"
       ],
       "helper_runner_paths": [],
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop passed and landed on main in 11c0b7d6d. I took the durable audit-loop skill policy update and regenerated the citation graph from the pipeline instead of treating the generated hash as authority. Verification: docs/audit/scripts/run_pipeline.sh, python3 docs/audit/scripts/audit_lint.py --strict, git diff --check.

jonathonreilly:
Landed on main in 11c0b7d6d.

PR #1552, opened 2026-05-19
[physics-loop] audited-cond repair: dense_prune_guard_seed_note claim narrowing to aggregate scope (bounded_theorem)

jonathonreilly:
## Summary

Repairs the `dense_prune_guard_seed_note` audited_conditional row by
narrowing the per-seed mechanism claim to what the existing
aggregate-only cached runner stdout actually supports.

Per the 2026-05-17 audit verdict:

> "The primary runner and helpers genuinely compute graph, pruning,
> gravity, purity, and eff_ch quantities, but the cached stdout only
> reports aggregate rows. It does not certify the source note's exact
> per-seed table or log guard trigger/removed-count decisions needed
> to support the seed-selective mechanism claim."

The cheapest-repair option in the audit verdict was either (A) add a
per-seed diagnostic runner cache, or (B) narrow the claim. This PR
takes option (B).

## Narrowed claims (before -> after)

- "compares the exact flip-prone seed IDs ... against the current
  channel-count guard path" -> "compares the historically flip-prone
  seed set ... at the aggregate level supported by the cached runner
  output"
- Per-seed comparison tables for `N=80` and `N=100` -> retained verbatim
  but explicitly marked **legacy reference, not cache-certified**, with
  a blockquote stating the cached stdout reports only aggregate rows
  and does not certify per-seed `grav`, `purity`, `eff_ch`, `flip`,
  `removed_total`, or per-seed guard-trigger decisions.
- "The guard ... fixes specific seeds when it preserves the effective
  detector-channel count and leaves other seeds vulnerable when
  `eff_ch` still drops" -> aggregate-level statement that the
  channel-count guard shifts aggregate gravity, purity, `eff_ch`, and
  flip count on the flip-prone seed set, with the detailed per-seed
  mechanism explicitly not certified by the cache.
- Specific rescue / non-rescue case bullets (`N=80` seed `12`,
  `N=100` seed `13`, `N=100` seed `3`) -> withdrawn from the bounded
  claim; replaced by aggregate-supported observation that the guard
  is not a no-op on the seed set.
- "So the guard is a seed-selective channel-preservation mechanism,
  not a pure averaging fix" -> withdrawn; aggregate cache supports only
  the bounded observation that the guard changes aggregate `eff_ch`
  and aggregate flip count relative to plain pruning.

## What stays

- Historical seed lists (kept as reference context, not as cache-certified evidence).
- Primary runner / guard-mode metadata.
- Per-seed tables (kept verbatim under explicit non-certified label
  for narrative continuity).
- Prior audit-requeue notes from 2026-05-17.

## Follow-up (out of scope)

A per-seed diagnostic runner output that, for the listed historical
seeds and under the same code hash as the aggregate cache, prints
`grav`, `purity`, `eff_ch`, `flip`, `removed_total`, and a
guard-triggered flag per seed. Once that cache lands, the
seed-selective mechanism claim can be restored.

## Non-negotiables honored

- No edits to `docs/audit/data/*.json`.
- No edits to repo-wide authority surfaces.
- No merge. Branch is not main.
- Honest narrowing: per-seed tables preserved verbatim, only the
  unsupported per-seed mechanism assertions removed from the bounded
  claim.

### DIFF
--- a/docs/DENSE_PRUNE_GUARD_SEED_NOTE.md
+++ b/docs/DENSE_PRUNE_GUARD_SEED_NOTE.md
@@ -1,10 +1,10 @@
 # Dense Prune Guard Seed Note
 
-This note compares the exact flip-prone seed IDs from the replay work against the current channel-count guard path.
+This note compares the historically flip-prone seed set from the replay work against the current channel-count guard path at the aggregate level supported by the cached runner output.
 
-The code path has evolved a bit since the earlier replay logs, so treat these as same-seed diagnostic comparisons rather than a literal byte-for-byte rerun of the older numbers.
+The code path has evolved since the earlier replay logs, so treat any per-seed numbers in the historical reference tables below as legacy diagnostic snapshots, not as the cache-certified support for this note's bounded claim.
 
-Historical flip seeds from the replay set:
+Historical flip seeds from the replay set (reference only):
 - `N=80`: seeds `8, 12, 13`
 - `N=100`: seeds `2, 3, 13`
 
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the source-side repair on main in a72c3832c, with generated audit/publication outputs refreshed in 9a191251c. I reviewed this as a source/auditability repair, not an audit verdict; branch-local state/certs were not landed.

PR #1697, opened 2026-05-23
Admitted-external-import carve-out + UHF/Naimark textbook imports

JONbridgerapps:
## Summary

**PR 3b of 5 — companion to PR 3 (#1696), stacked on it** (base is the PR 3 branch, so this diff shows only the incremental changes; retarget to `main` after PR 3 merges). Adds the **admitted-external-import** carve-out so vetted standard textbook theorems can serve as retained-grade upstream — the way a math paper cites a textbook.

## Why it's needed

Confirmed against the ledger: pure textbook-import notes currently land `audited_conditional` (class E, "named non-derivation import" with hidden imported assumptions — e.g. `mermin_wagner_bogoliubov_textbook_import_note`). So today there is **no** way for a framework note to cite a settled textbook result as retained-grade upstream. This blocks:
- `powers_uhf_tracial_uniqueness` (wants UHF unique-trace as admitted)
- `lsp_projective_derivation` (wants Naimark dilation as admitted)

Both auditors explicitly offered "treat the standard import as admitted" as an acceptable repair.

## Design choice

Deliberately **not** a new audit verdict threaded through `apply_audit` / `compute_effective_status` / RANK / lint (heavy, risky). Instead it mirrors PR 3 exactly — an **allowlist + render flag + rubric carve-out**:

- **`external_import_nodes.json`** — allowlist of named, precisely-cited, scope-limited textbook imports that make no framework-specific claim.
- **`AUDIT_AGENT_PROMPT_TEMPLATE.md`** — an authority flagged `admitted_external_import: true` is retained-grade upstream **for its stated scope only**. Not a free pass: the citing note's own load-bearing step must still close, and may use the import only within scope. The import note's own audited status is irrelevant — the carve-out governs whether a *dependent* may cite it.
- **`codex_audit_runner.py`** — `render_prompt` surfaces the `admitted_external_import: true` flag (reads the registry; inert if absent).
- **Two import notes** — UHF type 2^∞ unique tracial state (Powers 1967; Bratteli–Robinson) and Naimark dilation / canonical projective Kraus form (Naimark 1940; Holevo; Watrous; Stinespring). Each states the exact theorem + hypotheses and an explicit "what this does NOT do" (no framework claim, no instrument selection).

## Safety

- Inert when the registry is absent (same as PR 3).
- Allowlist, not a wildcard; each entry is scope-limited and flagged `makes_framework_claim: false`.
[...]

PR #1698, opened 2026-05-23
Re-route derivations to cite axiom premise + admitted external imports

JONbridgerapps:
## Summary

**PR 4 of 5 — the consumer PR.** Re-routes the dependent derivations onto the new upstream so the audit lane can re-audit them to retained-grade once the carve-outs land. **Depends on PR 1 (#1694), PR 2 (#1695), PR 3 (#1696), and PR 3b (#1697).**

Small diff (14 insertions, 8 deletions across 4 notes) — the heavy lifting is in the mechanism PRs; this just points the notes at it.

## Changes

- **`powers_uhf`** (positive_theorem) — cites the UHF tracial-uniqueness textbook import as load-bearing upstream. Covers its `dependency_not_retained` repair note ("treat UHF unique-trace as admitted").
- **`lsp_projective`** (bounded_theorem) — cites the Naimark dilation textbook import. Covers its `missing_dependency_edge` blocker (the canonical Naimark/Lüders construction).
- **`prr_local`** (bounded_theorem) — adds the retained narrow theorems `cl3_complexification_split` + `cl3_faithful_irrep_dim_two` as load-bearing upstream for the per-site M₂(ℂ)/H_x=ℂ² content. Auditor already found **chain_closes: True**; this + the axiom-premise carve-out should flip it to `retained_bounded`.
- **`pwc`** — makes the `e^{-J}` sign convention explicit (auditor flag): order-`n` derivatives give `(-1)^n` times the connected cumulants; even orders carry the textbook sign, odd orders the explicit `(-1)`.

## Notes for the reviewer

- **No link renames for the canonical node.** The notes still link `MINIMAL_AXIOMS_2026-05-20.md`; PR 2's path-based canonicalization resolves that to the `minimal_axioms` axiom-premise node automatically. The carve-out applies without editing the links.
- **Import-note links resolve once PR 3b lands.** The new links to `UHF_TRACIAL_UNIQUENESS_TEXTBOOK_IMPORT` / `NAIMARK_DILATION_TEXTBOOK_IMPORT` point at notes created in PR 3b; until that lands they're dangling (the graph builder simply skips unresolved targets — no crash). After PR 3b + this land and the pipeline regenerates, the dep edges form.
- **`gleason` needs no edit** — the auditor explicitly named the axiom-premise carve-out as an acceptable repair, so PR 3 alone addresses it; left untouched to minimize churn.

## Expected re-audit outcomes (auditor-owned)

| Row | Expected after full stack |
|---|---|
[...]

PR #1699, opened 2026-05-23
Prove UHF + Naimark inline; drop external-import scaffolding

JONbridgerapps:
## Summary

Replaces the "admit textbook theorems on citation" approach with **inline proofs at framework rigor** (the auditor can then verify correct usage, not just existence), and cleans up the now-unused external-import scaffolding. Builds on the merged axiom-premise carve-out (#1696).

**Supersedes and closes #1697 (external-import mechanism) and #1698 (import-citation re-routes).**

## Changes

- **`powers_uhf` Step 4** — proves UHF unique-trace **directly**: every tracial state agrees with the canonical trace on the dense local subalgebra (finite-dim trace uniqueness, proved in Step 1), then norm-continuity of states (`‖φ‖ = φ(𝟙)`) extends the agreement to the norm closure. Powers 1967 becomes **provenance, not a load-bearing import**. Remaining inputs are the framework's elementary standard-math floor.
- **`lsp_projective` Step 1** — constructs the canonical projective dilation explicitly and **verifies it is an isometry**: `‖U_int(|ψ⟩⊗|0⟩)‖² = Σ_r ⟨ψ|P_r†P_r|ψ⟩ = Σ_r ⟨ψ|P_r|ψ⟩ = ⟨ψ|ψ⟩` using `P_r†P_r = P_r` and `Σ_r P_r = 𝟙`, hence extends to a unitary. Naimark 1940 becomes provenance; the projective case is re-proved from the defining properties of a projective measurement.
- **`premise_nodes.py`** — scoped to **axiom premises only** (removes the unused external-import handling that merged via #1696, since `external_import_nodes.json` was never landed). Textbook theorems now reach retained-grade through the normal audit path, so no external-import carve-out is needed.
- **`prr_local`** — adds the retained `cl3_complexification_split` / `cl3_faithful_irrep_dim_two` siblings as load-bearing upstream.
- **`pwc`** — makes the `e^{-J}` sign convention explicit (`κ_n = (-1)^n ∂^n W/∂J^n|₀`).

## Why this is better than admitting-by-citation

- **No policy widening.** These become genuine `positive_theorem`/`bounded_theorem` rows that close via the normal path — no new "trust the citation" assumption.
- **Verifies correct usage.** Writing the proof forces the hypotheses to be stated and shown to hold here, catching the "right theorem, wrong preconditions" failure that citation-only can't.
- **Honest floor.** Proving doesn't eliminate all external math — it lowers the import from a named high-level theorem (Powers/Naimark) down to the universally-accepted elementary floor (finite-dim linear algebra, `‖φ‖=φ(𝟙)`, density, isometry-extends-to-unitary).

The **axiom-premise** carve-out (#1696, merged) is unaffected — axioms genuinely can't be proven, so that remains the correct tool for the A1+A2 dependency.

## Test plan

- [x] `premise_nodes` scoped to axiom-only; all four consumers compile; `clean_status` still promotes a row whose only dep is the meta axiom premise
- [x] Cited sibling paths resolve
- [ ] Audit lane re-audits powers_uhf / lsp under the inline proofs (expect clean/retained-grade on their own merits)

https://claude.ai/code/session_01NHXv18f8EV4iZuxKoXBfhx

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHXv18f8EV4iZuxKoXBfhx)_

### DIFF
--- a/docs/LSP_PROJECTIVE_DERIVATION_FROM_NAIMARK_FRAME_NARROW_THEOREM_NOTE_2026-05-22.md
+++ b/docs/LSP_PROJECTIVE_DERIVATION_FROM_NAIMARK_FRAME_NARROW_THEOREM_NOTE_2026-05-22.md
@@ -36,28 +36,50 @@ Any alternative `K_P^{twist} = U · P` for a unitary `U ∈ U(H_sys)` gives the 
 
 ## Proof
 
-### Step 1 — Naimark dilation of projective measurement
+### Step 1 — Explicit canonical dilation of a projective measurement
 
-The **Naimark dilation theorem** (Naimark 1940; Holevo *Probabilistic and Statistical Aspects of Quantum Theory* Ch. III; Watrous *The Theory of Quantum Information* §2.4):
+A projective measurement is a family of orthogonal projections `{P_r}`
+on `H_sys` with `P_r P_{r'} = δ_{r r'} P_r`, `P_r† = P_r`, and
+`Σ_r P_r = 𝟙_sys`. We **construct and verify** the canonical
+system–apparatus dilation explicitly, rather than importing Naimark's
+theorem as a black box. (Naimark 1940; Holevo *Probabilistic and
+Statistical Aspects of Quantum Theory* Ch. III; Watrous *The Theory of
+Quantum Information* §2.4 are cited as provenance for the general POVM
+case; the projective case used here is the explicit construction below.)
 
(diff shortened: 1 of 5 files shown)


jonathonreilly:
Landed via review-loop directly to main after rebasing over the paused audit state. Main commits: 0a3e6d4e3 (inline proof source), 9425267ca (review boundary fixes), 1fe5ddb28 (pipeline refresh). The PR branch itself was not force-pushed because it was stale relative to intervening audit commits.

PR #2129, opened 2026-05-28
review-loop: retype axiom-first RP to no_go; settle Berezin convention

jonathonreilly:
## Summary

Clears the critical (td=915) audited_conditional on [`axiom_first_reflection_positivity_theorem_note_2026-04-29`](docs/AXIOM_FIRST_REFLECTION_POSITIVITY_THEOREM_NOTE_2026-04-29.md) by retyping it to **no_go**, AND corrects the record on a convention dispute I got wrong last session.

### 1. The Berezin-convention record correction (I was wrong last session)

Last session I opened PR #2084 attempting to **retract** this no-go, claiming the original runner used a wrong Berezin sign convention and that with a "corrected" convention the Gram is PSD (20/20 Θ-symmetric configs). **Review-loop rejected #2084, and the rejection was correct.**

I have now verified from a **first-principles brute-force Berezin integral** (where the measure normalization cancels in the ratio, so no convention is assumed) on `M = [[2,3],[5,7]]`:

```
<chi_b bar(chi)_a> =  (M^-1)[b,a]
<bar(chi)_a chi_b> = -(M^-1)[b,a]
```

All four entries match this. My #2084 "correction" used `M^-1[a,b]` — a transpose plus a dropped sign, two errors that conspired to spuriously produce PSD. **The original single-step no-go is genuine** (free U=1 min eigenvalue −0.80 is real); the on-main runner uses the correct convention. The note now documents this settled convention inline so it can't be re-litigated by mistake.

Memory has been corrected accordingly.

### 2. The audit repair (clears the conditional)

Audit verdict: the runner closes the single-step counterexample, but the note also *claimed* 2-step `T̂²` positivity without deriving it in-repo (only literature-cited). Repair offered: derive 2-step positivity, OR keep the single-step no-go as a bounded negative witness.

This takes the second path:
- `claim_type`: `bounded_theorem` → **`no_go`**. Load-bearing content is the proven negative result.
- 2-step `T̂²` positivity demoted to explicitly **non-load-bearing literature context** (STW 1981 / Palumbo 2002 / Smit / Caracciolo-Palumbo 2013). A future companion row can derive and promote it.
- Scope / Statement / Honest-Status sections rewritten so the no_go is load-bearing and 2-step positivity is context only.

A `no_go` is terminally auditable — it does not need the 2-step positivity bridge the verdict flagged as missing.

## Test plan

- [x] First-principles Berezin calc confirms `<bar_a chi_b> = -(M⁻¹)[b,a]` (all 4 entries on M=[[2,3],[5,7]])
- [x] On-main no-go runner reproduces free U=1 min eig −0.80 with the correct convention
- [x] `bash docs/audit/scripts/run_pipeline.sh` runs clean
- [x] Row: `audit_status=audited_conditional → unaudited`, `claim_type=bounded_theorem → no_go`
- [x] No new axioms / imports (literature stays context); deps + runner unchanged
- [ ] Audit lane re-picks and confirms terminal no_go (out of band)

## Note for reviewer

This supersedes the rejected #2084. I am **not** trying to restore positive_theorem — the opposite: I'm confirming your rejection was right and formalizing the row as the no_go it actually is. If you'd prefer I instead attempt the in-repo 2-step `T̂²` positivity construction (the other repair path), say so and I'll scope that as a separate companion row.


jonathonreilly:
Review-loop salvaged the landable part in 805e2e3ec and did not land the PR's no_go retype. Main had already taken the other audit-repair path before this PR was reviewed: the two-step staggered-KS transfer-matrix positivity certificate is now the primary bounded theorem surface, with the single-step no-go retained as a narrow secondary negative witness.\n\nWhat landed: the Berezin convention correction, after an independent finite-Grassmann sign check. The runner implementation already had the sign via Wick pairing parity; I corrected the misleading comment, added the settled convention note to the source, refreshed the single-step runner cache, and regenerated the affected audit graph/ledger hashes. The row remains bounded_theorem / unaudited for the independent auditor.

jonathonreilly:
Closed after partial salvage landed in 805e2e3ec. The proposed no_go retype would overwrite already-landed two-step positive science, so it was not landed.

PR #2161, opened 2026-05-29
Narrow theorem: det unique via multiplicative-character form selection (P2 det-vs-tr piece)

JONbridgerapps:
## What this ships

A narrow `bounded_theorem` closing the **det-vs-tr form-selection** half of the observable-principle **P2** admission, on the **multiplicative-composition axis** `chi(A . S) = chi(A) . chi(S)` (operator product) — the one axis the prior P1-bridge campaign never tested.

- `docs/OBSERVABLE_PRINCIPLE_DET_UNIQUE_MULTIPLICATIVE_CHARACTER_FORM_SELECTION_NARROW_THEOREM_NOTE_2026-05-28.md`
- `scripts/audit_companion_observable_principle_det_multiplicative_character_form_selection_2026_05_28.py`

## The theorem

On the composition axis, motivated by the Berezin source-insertion structure `D + J = D . (1 + D^{-1} J)` and the multiplicativity of the Berezin weight over independent source patches:

- **(i)** By the GL(n) abelianization theorem (commutator subgroup of `GL(n,C)` is `SL(n,C)`, so every continuous `chi: GL(n,C) -> C*` is `det^k`), `det` is the **unique** multiplicative character up to integer power. `tr`, the power traces `tr(M^s)` (`s != 1`), and the elementary symmetric polynomials `e_1..e_{n-1}` are all **excluded** — each fails the character property outright.
- **(ii)** Additivity over independent (block-diagonal) sectors then fixes the exponent: Cauchy gives `W = c . log|det|`, with the character power `k` absorbed into `c`.
- **(iii)** `c = 1` is a pure convention (cancels in source-derivative ratios; retained scale-invariant decoration).

## Why this is new content, not a relabel of P1

The prior campaign (≈12 routes, 6 now `retained_no_go`) operated on the **direct-sum additivity axis** `F(A (+) B) = F(A) + F(B)`, where `tr` **also passes** (`tr(A (+) B) = tr A + tr B`) — so that axis provably cannot exclude `tr`, and selecting `log` from `{|det|^p}` there is the documented Pattern-L circularity. Every prior route already **assumed** the readout was a function of `det` and debated only the exponent.

This note operates on a **different axis** where `tr` **fails** (`tr(A . S) != tr(A) . tr(S)`), and the character classification ties the survivor to `det`. The det-vs-tr selection is complete *before* the Cauchy/log step, which only fixes the exponent within the already-selected `det` family. A `Relation to the direct-sum no-go portfolio` section cites all six `retained_no_go` rows + the consolidated Route D no-go and states explicitly: those foreclose the direct-sum axis; this is a different axis they never tested — not a re-run, not a contradiction.

## Scope discipline (honest residual)

- **Does NOT** close P2 on the physical determinant parent: the "the readout *is* `det(D+J)`" identification rides on the Berezin/Grassmann integral, **gated by AC_φλ** (`staggered_dirac_realization_gate_note_2026-05-03`, admission #2 of the canonical 4). Cited explicitly.
- **Does NOT** eliminate the multiplicative-source requirement `(M)`: it is grounded in the Berezin structure but is the honest residual (thinner than admitting P2 free, but still an admission).
- **Does NOT** claim "additivity + continuity ⟹ log" (false — `tr` is additive over direct sums).
- **Does NOT** reduce the admission count below 4 (P1, AC_φλ, S, θ unchanged); it shows the det-vs-tr piece is a *theorem*, not a separate admission.
- Does not promote/demote any row; `observable_principle_from_axiom_note` stays `audited_conditional`.

## Companion runner — 30 PASS / 0 FAIL (exact sympy)

Crux confirmed symbolically on generic `2x2`/`3x3`: `det(A . S) = det(A) det(S)`; `tr`, `tr(M^s)` (s=2,3), and `e_1..e_{n-1}` all fail the character property; `e_n = det` and `det^k` pass; Cauchy ODE `r W'(r) = c` -> `W = const + c.log r`; Pattern-L guard documents `tr(A (+) B) = tr(A) + tr(B)` (tr passes the direct-sum axis, so it is not the axis used for det-vs-tr).

## Compliance

- `Status authority: independent audit lane only`, `Type: bounded_theorem`, `proposal_allowed: false` (in the audit-disposition sense; no audit data landed).
- Audit data reverted per PRs-cannot-land-audit-data policy (`git checkout -- docs/audit/data/ docs/audit/AUDIT_*.md docs/publication/`); diff is the two source files only.
- No new repo vocabulary (standard math terms: "GL(n) character", "abelianization", "multiplicative under composition", "Cauchy classifier"); no new axiom.
- Load-bearing citations as markdown links (GL(n) abelianization cited to Fulton-Harris / Goodman-Wallach inline); see-also backticked.
- Local pipeline validation clean: citation graph parses 13 dependency edges, runner attached, `effective_status: unaudited`, `open_dependency_paths: []`.


jonathonreilly:
Review-loop result: landed source-only bounded salvage on main in 3e76ff700. I narrowed the character claim to the algebraic/rational character class, kept the physical readout premise explicit as an admission, ran vocab lint, py_compile, the companion runner, runner-cache precompute, pipeline, strict audit lint, and diff-check. No submitter audit verdict/status was applied.

PR #2164, opened 2026-05-29
fix(audit): reclassify assumption_derivation_ledger bounded_theorem → meta (path b)

jonathonreilly:
## Summary
Resolves the **highest-leverage** `audited_conditional` row, `assumption_derivation_ledger` (transitive descendants: 903, criticality: critical), by correcting a type misclassification. The verdict offered two paths; this takes **path (b): split into metadata**.

## Root cause
The file is, by its own explicit claim boundary, a **bookkeeping roadmap** — it *"does not promote or demote any row"* and exists *"so that downstream readers can find the authorities."* Authoritative status for each ingredient lives in the named authority note, never in the ledger's prose. Typed `bounded_theorem`, a roadmap will **always** read as conditional because it asserts package-wide statuses it does not itself certify.

## Change
- Reclassify `**Type:** bounded_theorem` → `**Type:** meta` and document the reclassification.
- The independently-audited ingredient claims already live in their own authority rows (e.g. exact `F_adj = 8/9` wired one-hop to `RCONN_DERIVED_NOTE`; hierarchy / strong-CP / CKM / `y_t` / Higgs each have their own lane authorities). No ingredient is certified *here*.

## Blast-radius audit (per the careful path)
- The 8 direct citers: 4 are meta/methodology/work-history; the 4 science notes (CPT, spin-statistics, Higgs-mass) reference the ledger as a *"Related"* / *"recorded-in"* / *"admitted-context"* roadmap pointer — **verified not load-bearing**.
- After reclassification, **only this one row changes** vs `origin/main` (`audited_conditional → meta`); **all 903 transitive descendants are untouched**, and cycles stay at 0. The high descendant count was publication-index / co-citation aggregation, not load-bearing dependence.

## Boundary
No new axiom, import, or retained bridge. The ledger remains a faithful roadmap to the named authority notes; reclassifying it as `meta` simply stops a bookkeeping surface from being audited as a theorem. Status authority is the independent audit lane only.


jonathonreilly:
Landed source-only review-loop repair in 9a157affd. The ledger is now meta roadmap/index material rather than a theorem row; PR generated audit/status deltas were not used, and local pipeline regeneration sets the row to meta.

PR #2167, opened 2026-05-29
fix(audit): narrow audit_backlog_note to purely textual campaign index (path b)

jonathonreilly:
## Summary
Clears the high-criticality `audit_backlog_note_2026-05-02` (`audited_conditional`). The verdict flagged that this row asserts aggregate claim-states (LHCM *"fully covered"*, *"coherent SM gauge structure now flows"*, per-cycle demotions) backed only by a substring-checking runner — without the 19 cycle-level `CLAIM_STATUS_CERTIFICATE.md` files or convention/bridge authorities needed to verify them.

## Root cause
It's a **campaign log**, not a derivation — §6 already says `proposal_allowed: false`, *"campaign-level synthesis, not a new derivation."* As an untyped/meta row whose prose reads like aggregate certification, it audits as conditional.

## Change (verdict path b: narrow to a purely textual campaign index)
- Retype to `meta` and scope explicitly as a textual index: every aggregate status statement is a **historical campaign-log pointer** to the per-cycle certificate that actually certifies it; the runner is an **index-integrity check** (log well-formedness), not science verification.
- Wiring all 19 per-cycle certificates + their authorities (the other path) wouldn't change the file's nature and is a large structural task — out of scope.

## Blast-radius audit
- The 6 citers reference it as a *"canonical synthesis template"* / *"§2.3 cluster identification"* / campaign-record pointer — **verified not load-bearing**.
- Only this row changes vs `origin/main` (`audited_conditional → meta`); cycles unchanged at 0.

## Boundary
No new axiom, import, or retained bridge. Authoritative status for any cycle lives in that cycle's certificate and the lane authority note, never in this index's prose. Status authority is the independent audit lane only.

### DIFF
--- a/docs/AUDIT_BACKLOG_NOTE_2026-05-02.md
+++ b/docs/AUDIT_BACKLOG_NOTE_2026-05-02.md
@@ -1,11 +1,56 @@
 # Audit-Backlog Retained Campaign — Progress Synthesis (Cycles 1-19)
 
-**Date:** 2026-05-02
-**Status:** campaign-level progress synthesis covering cycles 1-19 of
-audit-backlog-campaign-20260502. Documents cumulative claim-state
-movement, identifies remaining Nature-grade targets, and provides
-handoff for review backlog.
-**Primary runner:** `scripts/frontier_audit_backlog_campaign_synthesis.py`
+**Date:** 2026-05-02 (2026-05-28: narrowed to a purely textual campaign
+index per audit path (b)).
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Landed source-only review-loop repair in 9a157affd. This is now a meta textual campaign index, not an authority surface; PR generated audit/status deltas were not used, and local pipeline regeneration sets the row to meta.

PR #2177, opened 2026-05-29
rp: combined U-integrated RP for mixed gauge+fermion observables (single transfer matrix)

JONbridgerapps:
## Summary

Closes the **root gate of the continuum-QFT program** at fixed lattice spacing: U-integrated combined reflection positivity (RP) for the interacting staggered + SU(3) lattice theory, for observables that **mix** gauge links and staggered fermion fields in the positive-time half.

The standing audit objection was that per-config fermion positivity + `det>0` + an abstract gauge-half Cauchy–Schwarz "do not supply the combined two-step transfer/OS bridge for gauge-coupled staggered fermion observables." This PR resolves it by assembling the **single** Osterwalder–Seiler transfer operator rather than a product of factors:

```
<Theta(F) F> = <Omega| O^dag T_full O |Omega> = || W (O Omega) ||^2 >= 0,
T_full = (T_gauge^{1/2} (x) I)(oplus_U B[U]^dag B[U])(T_gauge^{1/2} (x) I) = W^dag W.
```

This is `>= 0` for **every** observable, mixed or factorized, because it is a manifest norm-square of one vector. The three previously-separate facts are the *ingredients* (assembled, not multiplied): the per-config fermion two-step positivity is the `oplus_U B[U]^dag B[U]` block core; `det>0` is the positive `U`-average weight; the gauge-half is the positivity of `T_gauge`.

## New load-bearing content vs main

- **Single-transfer-matrix `W^dag W` framing** — the correct OS reflected-correlator form. Handles mixed observables automatically; supersedes the three-separate-factors framing (valid only for *factorized* observables, where the fermion factor is `M[U]^{-1}`-weighted and does not factor out for mixed `F`).
- **R1 — gauge-measure RP instantiated on the actual Wilson plaquette boundary.** The temporal-gauge transfer kernel `K(U,U') = exp(-beta(1 - Re Tr(U'U^dag)/N_c))` is positive-definite (Bochner / Peter–Weyl; `I_n(beta)>0`), superseding the abstract Cauchy–Schwarz identity that explicitly disclaims the Wilson boundary.
- **Mixed-observable combined Gram runner** (no existing runner checks mixed observables): Route T (operator) + Route C (literal U-integrated, real `det>0` weight, Haar), with a `-0.80` single-step no-go negative control, a teeth test, and an exact Berezin per-block decay match.

## Runner

`scripts/rp_combined_mixed_observable_u_integrated_2026_05_29.py` — `SCORECARD: PASS=6 FAIL=0`, deterministic. Mixed-Gram min eigenvalue `>= -3e-16` (machine zero) across U(1)/SU(3) in both routes; operator-Schmidt rank `>1` confirms genuine gauge–fermion entanglement (so PSD is not a trivial factorized consequence).

## Scope (honest)

- **Fixed lattice spacing only** — not continuum / OS-reconstruction (Wightman) RP.
- **Two-step transfer-matrix object** — single-step Lagrangian RP is a genuine no-go (reproduced as the negative control).
- SU(3) backgrounds sampled beyond the analytic anti-Hermitian-hop reduction (which is general); the `W^dag W` factorization is an exact analytic identity. Route C uses the coincidence-limit gauge factor; Route T is load-bearing.
- Does not close the staggered-Dirac realization substeps (1)+(2).

## Policy

Science note + audit-companion runner + cache only. No audit-lane data (`docs/audit/data/`, `AUDIT_*.md`, `docs/publication/`) touched. Deps cited as markdown links (retained det-positivity + gauge-half); parent RP / gauge-extension rows referenced as plain-text per the citation-graph convention.


jonathonreilly:
Review-loop result: landed a narrowed source-only salvage on main in 2011b959b. I kept the useful single-transfer W^dag W mixed-observable assembly, but removed the root-gate/closure framing: the landed note is finite-carrier bounded support, treats SU(3) Wilson-boundary positivity as an audit premise/standard input rather than something proven by sampling, and does not promote the parent RP theorem. Checks run: vocab lint, py_compile, runner PASS=6/0, refreshed runner cache with matching hash, full audit pipeline, strict audit lint, and diff-check. No submitter audit verdict/status was applied.

PR #2189, opened 2026-05-29
feat(koide): NCG order-one condition does NOT force the Koide ratio (diagnostic)

jonathonreilly:
## Summary
You asked me to run the **order-one test** — the one decisive fork the chirality-gate panel had flagged as untested. Result, verified two independent ways (runner PASS=9): the NCG order-one condition `[[D,π(a)],π°(b)]=0` does **not** pin the Koide ratio `r=|b|²/a²=1/2`.

## Findings
- **Circulant generation algebra ⇒ order-one VACUOUS.** Circulant matrices commute, so the inner commutator `[D,π(a)]=0` *identically* (before `J`/grading enter) — it constrains nothing. Explicit distinct-`r` witnesses (`r∈{0.05,0.2,0.5,1.0,2.0,5.0}`) **all** satisfy order-zero + order-one to machine precision → a continuum of `r`, not a unique value.
- **Full `M₃` algebra ⇒ moduli-free** (corroborating Cacic 2009, arXiv:0902.2068): Yukawa directions remain free parameters of an order-one Dirac operator.
- **Why:** order-one is a *bimodule/texture* condition — it constrains *which* generation-pair entries may be nonzero, never the *ratio* of nonzero coefficients. Also: a circulant at "`r=½`" still commutes with `R`, so the genuine Koide point is the *chiral non-circulant* `{H,Γ_χ}=0` operator, which a C₃-covariant order-one condition cannot select.

## Disposition
A clean **bounded negative**: order-one is the 7th lens to leave `r=½` unforced (after kinematic / dynamical / quantum / chiral / records / spectral-action). The charged-lepton chirality/`r` gate — shared with generation-ID and the signed-gravity Origin obligation — **stays open**, awaiting a genuinely new forcing principle. Documenting this closes the order-one sub-route so it isn't re-tried.

## Boundary
What-if on the Connes–Lott / product triple; **does not adopt** the `e₄`/P2 L/R-factor import (shows that even granting it, order-one isn't the selector). No new axiom, import, or retained bridge. Cacic 2009 cited as parallel literature only.

## Verification
Source note (`bounded_theorem`) + runner + cache. Row seeds `unaudited`, deps=[], no new markdown links. The pre-existing `origin/main` cycle is unrelated — my row has zero edges, **cycle delta = 0**.


jonathonreilly:
Review-loop landed the route-specific diagnostic in main commit 3689f9be3, narrowed from this PR. Accepted: the exact C_3-circulant order-one result that [D, pi(A)] vanishes identically and distinct r values all satisfy order-one, so this bounded route does not select r=1/2. Rejected from landing: the broader full-M3/moduli-free corroboration, because the submitted runner and prose did not match cleanly enough for source authority. Landed row remains unaudited; no order-one primitive or NCG import was approved.

PR #2190, opened 2026-05-29
docs(flavor): frontier map — form unified on one C₃ scaffold, values imported (~20→~20)

jonathonreilly:
## Summary

Honest **frontier map** of the flavor problem in the Cl(3)/Z³ framework, continuing the charged-lepton Koide campaign (which established Koide = the flavor problem localized). 8-lens panel; **7/8 reparametrization-trap, 0/8 share the Koide import.**

**Verdict:** the framework **unifies the FORM** of the whole flavor sector on one shared, genuinely-derived C₃-corner scaffold, but does **not reduce the continuous VALUES** — ~4–6 independent S₃-breaking imports, net **~20 → ~20** continuous parameters. The Koide pattern writ large: re-parametrization, not reduction.

## What is real (retained)
- One shared C₃ coordinate-cycle scaffold (projected onto hw=1) underlies all three sectors.
- Koide biconditional `Q=2/3 ⟺ r=½`; CKM sum rules + Wolfenstein identities + moduli-only Jarlskog certificate; PMNS *native* oriented-cycle basis + antiunitary CP locus.
- **Discrete** predictions: `n_gen=3`, `n_color=3`.
- **Strongest single result:** CKM CP angle `δ = arccos(1/√6) = 65.9°`, r²-independent, matches PDG `γ = 65.7 ± 3°` to **0.07σ** from one integer.

## What is not real
- Any continuous-parameter cut: the substrate pins **zero** of the ~20.
- The recurring 2/3 is **two distinct objects** (Koide `2/d` weight-ratio vs CKM `n_pair/n_color` count ratio), coincident only at (2,3).
- The Koide grading does **not** propagate to quarks (`retained_no_go`).

## Unifying obstruction
Counting-vs-splitting generalized — **one wall, six masks**: the C₃ orbit that gives the discrete wins forces equivariant operators that cannot pin any continuous modulus.

## Next step
Test a single **product-grading** `R³(gen) ⊗ taste ⊗ (H_L⊕H_R)` that breaks the orbit and might fix values in ≥2 sectors at once — converting form-unification into genuine reduction.

No new axioms/imports asserted as derived. Conditional/mapping analysis. Sets no retained status.

### DIFF
--- a/docs/FLAVOR_FRONTIER_MAP_FORM_VS_VALUE_NOTE_2026-05-29.md
+++ b/docs/FLAVOR_FRONTIER_MAP_FORM_VS_VALUE_NOTE_2026-05-29.md
@@ -1,0 +1,109 @@
+# Flavor Frontier Map — form unified, values imported (Cl(3)/Z³)
+
+**Date:** 2026-05-29
+**Claim type:** frontier map / honest status synthesis (NOT a derivation, NOT
+a promotion). Imports nothing as derived; sets no retained status (audit
+lane decides). Conditional/mapping analysis.
+**Runner:** `scripts/flavor_frontier_form_vs_value_2026_05_29.py`;
+cache `logs/runner-cache/flavor_frontier_form_vs_value_2026_05_29.txt`.
+**Source:** 8-lens flavor frontier panel (`wf_80271397`), 7/8
+reparametrization-trap, 0/8 share the Koide import. Continuation of the
+charged-lepton Koide campaign (PR #2162), which established Koide = the
+flavor problem localized.
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a narrowed meta frontier map in main commit 3689f9be3. Accepted: the form-vs-value synthesis with current-main retained-grade ledger checks and the Koide-vs-CKM 2/3 distinction. Narrowed: removed the PR's stronger retained/form-unification prose and PR-2162 dependency framing; this is a meta map, not a theorem, not a flavor primitive admission, and not a retained-status promotion.

PR #2193, opened 2026-05-29
fix(audit): break citation-graph cycle (cycles_detected 1→0)

jonathonreilly:
## Summary

The audit pipeline reported one citation-graph cycle on `origin/main`
(`cycles_detected: 1`). This PR breaks it, restoring `cycles_detected: 0`.

The cycle is a length-2 loop between two Higgs notes:

- `higgs_mass_from_axiom_note` (`docs/HIGGS_MASS_FROM_AXIOM_NOTE.md`)
- `higgs_channel_effective_ntaste_boundary_bounded_note_2026-05-08`
  (`docs/HIGGS_CHANNEL_EFFECTIVE_NTASTE_BOUNDARY_BOUNDED_NOTE_2026-05-08.md`)

## Root cause

`git blame` shows the back-edge was reintroduced by automation:

- **2026-05-28** a human added a backticked context-reference (line 524 of the
  parent note) *specifically to break this length-2 cycle*, with the rationale
  inline.
- **2026-05-29** `audit-bot` commit `0997ece243` ("nightly repair and pipeline
  refresh, automated") appended a **duplicate live markdown link** to the same
  boundary note (line 526), re-creating the exact cycle the human had removed.

The repair step wires named dependency paths without a cycle/dedup guard, so it
re-added the back-edge it was meant to avoid.

## Which edge, and why

The boundary note is **provably the downstream consumer**. Its own
*Dependencies* section cites `HIGGS_MASS_FROM_AXIOM_NOTE.md` as the load-bearing
parent and states it "bounds the parent's `N_taste` slot from below; it does not
promote, replace, or extend the parent." So:

- `boundary → parent` is the **load-bearing** direction → **kept**.
- `parent → boundary` is the **reversed / non-load-bearing** back-edge → **broken**.

Breaking the back-edge follows the repo's established convention (backtick the
non-load-bearing reference so it drops from the citation graph while remaining
human-readable). Here the parent note **already** carries that canonical
backticked entry at line 524 — so the fix simply removes `audit-bot`'s stray
duplicate live link. The resulting file is byte-for-byte the pre-bot blob
(`891ec4771`).

## Changes

- `docs/HIGGS_MASS_FROM_AXIOM_NOTE.md` — remove the duplicate live link (1 line).
- Regenerated audit data (`citation_graph.json`, `cycle_inventory.json`,
  `effective_status_summary.json`, `audit_ledger.json`, `audit_queue.json`) and
  rendered views (`AUDIT_LEDGER.md`, `AUDIT_QUEUE.md`) via
  `bash docs/audit/scripts/run_pipeline.sh`.

## Test plan

- [x] `bash docs/audit/scripts/run_pipeline.sh` exits 0 (incl. `audit_lint.py`)
- [x] `cycles_detected` 1 → 0; `cycle_inventory.cycle_count` 0
- [x] `citation_graph` `edge_count` 9948 → 9947 (exactly one edge removed)
- [x] **0 `effective_status` changes and 0 verdict changes across all 2502 claims**
- [x] both cycle nodes remain `unaudited` `bounded_theorem`
- [x] `audit_queue` entry set unchanged (1196 ids; reordered only, from updated
  load-bearing scores)

## Follow-up (out of scope)

`audit-bot`'s repair step has no cycle/dedup guard and may re-add this back-edge
on a future nightly run. A durable fix would teach the repair script to skip
edges already present as a backticked context-reference (or that would close a
cycle). Flagged separately.


jonathonreilly:
Review-loop landed the source fix in main commit 3689f9be3 and regenerated audit metadata from the combined current-main source state. Accepted: remove the duplicate live parent-to-boundary markdown link in HIGGS_MASS_FROM_AXIOM_NOTE.md while preserving the existing backticked context reference and the load-bearing boundary->parent edge. Verification after regeneration: cycle_inventory has 0 cycles; strict audit lint passes with notices only.

PR #2218, opened 2026-05-29
fix(audit): guard repair-edge wiring against backtick refs and cycles

jonathonreilly:
## Summary

The nightly `audit-bot` (`.github/workflows/audit.yml`, the `repair_missing_dependency_edges.py --apply` step) re-introduced citation-graph cycles by appending **live** markdown links to a note's *"Audit dependency repair links"* section — even when the same target was already present as a **backticked plain-text** pointer that a human deliberately added to break a length-2 cycle.

Concrete instance (manually fixed in [#2193](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/2193)): in `docs/HIGGS_MASS_FROM_AXIOM_NOTE.md`, a human backticked `HIGGS_CHANNEL_EFFECTIVE_NTASTE_BOUNDARY_BOUNDED_NOTE_2026-05-08.md` (a downstream note that cites this one as its load-bearing parent). The next nightly run appended a live link to the same target, re-closing the cycle and pushing `effective_status_summary.json` `cycles_detected` to 1. This is the **durable** fix for that recurrence.

## Root cause

`apply_repairs` only deduped against an existing `](path)` markdown link (repair_missing_dependency_edges.py:152). A backticked filename is **not** a markdown link, so it was invisible to the dedup — and there was **no** cycle check at all. The auditor names targets in `open_dependency_paths` ("what's missing from the chain") without knowing graph direction; resolving direction is the wiring step's job.

## Changes

`docs/audit/scripts/repair_missing_dependency_edges.py` — two guards in `apply_repairs`, both erring toward **not** writing an edge (a missed auto-wire is recoverable; a re-created cycle causes nightly churn):

- **(a) backtick guard** — skip if the target filename already appears inside an inline-code span in the source note (`has_backticked_reference`).
- **(b) cycle guard** — skip if adding `source -> target` would close a directed cycle (`edge_would_close_cycle`), **reusing `build_cycle_inventory.detect_cycles`** (the audit lane's canonical cycle walk) via a canonical before/after cycle-set diff.

The cycle guard builds adjacency from the **citation graph** deps. I verified ledger deps ⊆ graph deps for all 2506 rows on `main`, so guarding the graph superset protects both `cycle_inventory.cycle_count` and `effective_status_summary.cycles_detected`. The dry-run (PR-advisory) path reports the guards too, so its counts now match what `--apply` would do.

Scope check: only `repair_missing_dependency_edges.py` writes that section — `apply_audit.py` / `seed_audit_ledger.py` never touch note bodies — so the fix is localized to one script.

## Test plan

- [x] New `RepairMissingDependencyEdgesTest` (6 cases): cycle guard, backtick guard, legitimate non-cyclic add, existing-live-link dedup, plus pure-logic unit tests for `edge_would_close_cycle` (length-2/length-3/forward-shortcut/existing-edge/unknown-node) and `has_backticked_reference` (bare name / repo-relative path / live-link-is-not-a-backtick / bare prose).
- [x] Full `python3 -m unittest docs.audit.scripts.tests.test_audit_pipeline` — my class passes. (2 pre-existing `AuditLintTest` failures exist unchanged on clean `main`; the audit CI runs the linter script, not this unittest suite.)
- [x] `bash docs/audit/scripts/run_pipeline.sh` keeps `cycles_detected: 0` and `cycle_inventory.cycle_count: 0`, with **no** `effective_status_summary.json` change (untracked cross-lane docs moved aside for a clean baseline).
- [x] Against live `main` data, both guards flag the real HIGGS edge (`edge_would_close_cycle` and `has_backticked_reference` → `True`) while an unrelated non-cyclic edge (`higgs -> plaquette_self_consistency_note`) → `False` (not over-blocked).

### DIFF
--- a/docs/audit/scripts/repair_missing_dependency_edges.py
+++ b/docs/audit/scripts/repair_missing_dependency_edges.py
@@ -10,6 +10,23 @@ open_dependency_paths, but that note is not a direct dependency, the script can
 append an explicit markdown link to the source note. The normal audit pipeline
 then picks up the new edge, changes the source note hash, and resets that row
 for fresh re-audit.
+
+Two guards stop the script from fighting deliberate human cycle-breaks (an
+auditor names what is missing from the chain; it does not know the graph
+direction, so the wiring step must resolve it):
+  (a) backtick guard — if the target note's filename already appears as a
+      backticked plain-text reference in the source note, no live link is
+      added. Authors backtick a downstream note's filename to record a
+      sideways/back pointer without creating a load-bearing citation edge,
+      usually to keep the graph acyclic. A backtick is not an edge to the
+      graph builder, so a fresh markdown link would silently override that
+      decision and recreate the cycle.
+  (b) cycle guard — if adding the source->target edge would close a directed
+      cycle in the citation graph, the edge is skipped (reusing
+      build_cycle_inventory.detect_cycles, the audit lane's canonical cycle
+      walk).
+Both guards err toward NOT writing an edge: a missed auto-wire is recoverable
+(a human can add it), but a re-created cycle causes nightly bot churn.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed through review-loop on main in 06915b180. The repair-edge script now skips backticked non-edge references and candidate edges that would close a citation cycle, with targeted tests passing. Pipeline verification reports cycle_count: 0.

PR #2263, opened 2026-05-30
fix(docs): replace hard-coded /Users/jonreilly paths with relative links

JONbridgerapps:
## Summary

Sweep across the repo to replace **absolute path link targets** of the form `(/Users/jonreilly/Projects/Physics/<rest>)` with the correct repo-relative `(../...)` chain. These broke on every other developer's machine and on CI; the link text was already relative — only the target was wrong.

**Result: 276 files, 1259 link replacements.** All purely mechanical: prefix substitution based on the source file's depth.

## Why now

The hard-coded path pattern surfaced during today's draft-PR triage (one of the substantive-KEEP findings — see e.g. the existing [docs/SOURCE_RESOLVED_EXACT_GREEN_SCALING_NOTE.md](docs/SOURCE_RESOLVED_EXACT_GREEN_SCALING_NOTE.md) that was on main). The autopilot has been emitting these for months; a sibling PR (separate review-loop guard) will prevent new instances from landing.

## Method

For each `.md` file under the repo (excluding pipeline-output files):
- `docs/audit/data/` (audit-lane authoritative outputs)
- `docs/audit/AUDIT_LEDGER.md`, `docs/audit/AUDIT_QUEUE.md`, `docs/audit/AUDIT_DISPATCH_QUEUE.md`, `docs/audit/MISSING_DERIVATION_PROMPTS.md` (pipeline-regenerated)

Replace every occurrence of `/Users/jonreilly/Projects/Physics/<rest>` (inside a markdown link target only — `](/Users/jonreilly/Projects/Physics/...)`) with `<../-prefix>/<rest>`, where `<../-prefix>` is computed from the source file's depth.

Examples:
- `docs/SOURCE_RESOLVED_EXACT_GREEN_SCALING_NOTE.md` (depth 1 → `../`): `(/Users/jonreilly/Projects/Physics/scripts/X.py)` → `(../scripts/X.py)`
- `docs/work_history/repo/backlog/OVERNIGHT_WORK_BACKLOG.md` (depth 4 → `../../../../`): `(/Users/jonreilly/Projects/Physics/docs/INTEREST_MAP.md)` → `(../../../../docs/INTEREST_MAP.md)`

## Verification

```bash
# No remaining hard-coded link targets in markdown sources:
grep -rE '\]\(/Users/jonreilly/Projects/Physics/' . \
  --include='*.md' \
  --exclude-dir='docs/audit/data' \
[...]

PR #2264, opened 2026-05-30
review-loop: gate against non-portable absolute-path links in markdown

JONbridgerapps:
## Summary

Adds a new **Repository-portable links PASS gate** (§8) to the Audit-System Compatibility Gate section in `docs/ai_methodology/skills/review-loop/SKILL.md`. The gate scans every branch-modified `.md` file for link targets starting with `/Users/`, `/home/`, `/private/`, `/tmp/`, `/var/`, `/opt/`, or `file://` and blocks PASS until they're rewritten to repo-relative form.

## Why

The autopilot has been emitting markdown links of the form `(/Users/<author>/Projects/Physics/scripts/X.py)` for months. The link text was already correctly relative; only the target was absolute, so the bug was invisible to anyone using the rendered docs on the author's machine. A manual sweep of `main` on 2026-05-30 found **1259 such link targets across 276 files** — the companion PR fixes them all.

This gate exists so the next instance does not require another manual sweep. The fix is mechanical: rewrite the link target to a `../`-chain based on the source file's depth.

## Companion PR

PR #2263 lands the actual sweep on `main`. This PR adds the guard so the autopilot can't re-introduce the pattern.

## Renumbering

The new §8 sits between the existing §7 (Pipeline-output-stripped PASS gate, added in commit 7072614bd) and the existing §8 (No-Go Discipline) — so the existing §8 → §9, §9 → §10, §10 → §11. There are no cross-references to those numbers elsewhere in the file, so this is a clean shift.

## Verification

```bash
# The gate's own scan command (must produce no output on a clean branch):
git diff --name-only origin/main...HEAD -- '*.md' \
  | xargs -I{} grep -nE '\]\((/Users/|/home/|/private/|/tmp/|/var/|/opt/|file://)' {} 2>/dev/null
```

### DIFF
--- a/docs/ai_methodology/skills/review-loop/SKILL.md
+++ b/docs/ai_methodology/skills/review-loop/SKILL.md
@@ -648,7 +648,39 @@ sole channel for landing the regenerated outputs. Framework PRs that ship
 these files force a destructive overwrite of ratified audit state at merge
 time and have been an active source of broken-row regressions.
 
-8. **No-Go Discipline PASS gate (hard).** Before issuing review-loop PASS,
+8. **Repository-portable links PASS gate (hard).** Markdown link targets
+   on any branch-modified `.md` file must be repo-relative. Absolute paths
+   like `/Users/<name>/...`, `/home/<name>/...`, or `/private/tmp/...`
+   inside a link target are non-portable: they break for every other
+   developer, every CI runner, and every fresh clone. The autopilot
+   historically emitted these as `(/Users/<author>/Projects/.../scripts/X.py)`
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the portable-link guard on main in 2b141fca7, with one scope adjustment from review: the guard now blocks/fixes non-portable link targets on branch-modified markdown, but explicitly defers broad standalone rewrites over audited claim notes to the new audit-hash churn guard. That keeps the useful prevention rule without re-authorizing the #2263 failure mode.

jonathonreilly:
Landed via review-loop in 2b141fca7.

PR #2152, opened 2026-05-28
[physics-loop] conditional-backlog-batch2 bounded-support

JONbridgerapps:
## Summary

This physics-loop batch repairs six uncovered audited-conditional rows by
narrowing the source/support surface and regenerating the audit projection.

It does not promote any row to retained and does not manually retag the audit
ledger. The pipeline reset is generated from source/support hash changes.

## Target rows

- `audit_backlog_note_2026-05-02`: narrowed to meta campaign index/handoff.
- `beyond_lattice_qcd_note`: narrowed to bounded runner diagnostics.
- `born_rule_from_gleason_busch_derivation_note_2026-05-20`: narrowed to a repair-route map over imported premises.
- `cluster_decomposition_spatial_slab_bridge_theorem_note_2026-05-17`: narrowed to H1/H2 conditional spectral estimate.
- `cross_sector_a_squared_koide_vcb_bridge_promoted_via_v8_theorem_note_2026-04-29`: narrowed to conditional algebraic composition.
- `dimension_selection_lower_bound_bridge_v2_2026-05-20`: narrowed to runner-faithful eikonal sign support.

## Loop pack

- `.claude/science/physics-loops/conditional-backlog-batch2-20260528/HANDOFF.md`
- `.claude/science/physics-loops/conditional-backlog-batch2-20260528/TRACE_GATE.md`
- `.claude/science/physics-loops/conditional-backlog-batch2-20260528/CLAIM_STATUS_CERTIFICATE.md`

## Generated audit projection

After `bash docs/audit/scripts/run_pipeline.sh`, all six changed rows are:

- `audit_status: unaudited`
- zero open dependency paths

The meta row remains effective `meta`; the five theorem/support rows are back
to `unaudited` pending independent audit.

## Verification

```text
PYTHONPATH=scripts python3 scripts/frontier_audit_backlog_campaign_synthesis.py
PYTHONPATH=scripts python3 scripts/frontier_beyond_lattice_qcd.py
PYTHONPATH=scripts python3 scripts/cluster_decomposition_spatial_slab_bridge_check.py
PYTHONPATH=scripts python3 scripts/frontier_cross_sector_a_squared_koide_vcb_bridge_promoted_via_v8.py
PYTHONPATH=scripts python3 scripts/frontier_dimension_selection.py
python3 scripts/cached_runner_output.py --refresh --tail-chars 2000 scripts/frontier_audit_backlog_campaign_synthesis.py
[...]

PR #2273, opened 2026-05-30
[physics-loop] born-rule Gleason-Busch route narrow bounded support

JONbridgerapps:
## Summary

Narrows `born_rule_from_gleason_busch_derivation_note_2026-05-20` to a repair-route map over explicitly imported Gleason-Busch, Luders/update, pre-record identification, and persistent-record premises. The note no longer reads as a Born-rule derivation from the current framework surface; it documents what would close the lane if every imported premise were independently supplied, and adds an explicit `Status authority: independent audit lane only.`

This is **1 of 6 narrow PRs** split out from the original bundle #2152 (per repo policy: ship one identity at a time). The 6 rows in that bundle do not share a common-cause repair; each gets a different narrowing.

- Target row: `born_rule_from_gleason_busch_derivation_note_2026-05-20`
- This PR does not apply audit verdicts. Effective retained status remains owned by the independent audit lane.

## Sibling PRs

The remaining five splits of #2152 are being opened in parallel under `physics-loop/*-narrow-2026-05-30` branches. Numbers will be cross-linked in a follow-up edit.

## Verification

This row is a repair-route note; it has no dedicated runner. Static lint:

```
python3 scripts/vocab_lint.py --report-only target docs
```

### DIFF
--- a/docs/BORN_RULE_FROM_GLEASON_BUSCH_DERIVATION_NOTE_2026-05-20.md
+++ b/docs/BORN_RULE_FROM_GLEASON_BUSCH_DERIVATION_NOTE_2026-05-20.md
@@ -2,7 +2,7 @@
 
 **Date:** 2026-05-20
 **Status:** proposal — pre-audit bounded support / repair-route note
-**Type:** bounded_theorem
+**Type:** bounded support
 **Status authority:** independent audit lane only.
 **Repair route for:** the `audited_failed`
 `BORN_RULE_ANALYSIS_2026-04-11.md` lane (the gravitational Hartree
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop result: landed by salvage in main commit 97547b2c2, then closing this stale PR.

I did not apply the branch wholesale because it would have removed the stricter on-main 2026-05-28 audit-repair section. I salvaged the useful narrowing instead:

- changed the persistent-record dependency from "not retained" to "not effective authority";
- changed the closure language so even all imported premises only justify a future audit proposal, not a status change from this note;
- renamed the local section from "What this note derives" to "What this route formalizes";
- narrowed the Born-rule language so it is explicitly conditional on admitted inputs and not a retained derivation from A1+A2 alone.

Validation run before landing:

- python3 scripts/vocab_lint.py --fix docs/BORN_RULE_FROM_GLEASON_BUSCH_DERIVATION_NOTE_2026-05-20.md
- bash docs/audit/scripts/run_pipeline.sh
- python3 docs/audit/scripts/audit_lint.py --strict
- git diff --check

Generated audit/publication outputs from the validation run were restored; only the source note was committed.

PR #2128, opened 2026-05-28
[physics-loop] conditional-batch-repair bounded-support

JONbridgerapps:
## Summary

Science-fix batch for the May 28 audited-conditionals. This edits source
surfaces, not main ledger verdicts by hand.

- repairs 33 May 28 `audited_conditional` rows in ledger order after rebasing
  past the landed Mermin-Wagner and Koide APS review-loop splits
- narrows broad/imported claims to executable algebra, bounded support,
  runner-local diagnostics, or metadata where appropriate
- adds a PMNS projector degeneracy-boundary runner check (`PASS=17 FAIL=0`)
- repoints the wave direct-dM H=0.25 runner default/cache to the Fam1 seed-1
  control point used by the source note
- runs the audit pipeline so changed rows are queued mechanically as
  `unaudited`/meta-effective where source-demoted

The earlier Mermin-Wagner and Koide APS items from this batch are no longer
carried here because they landed upstream through review-loop splits. This
branch preserves those main versions.

## Loop pack

- Handoff: `.claude/science/physics-loops/conditional-batch-repair-20260528/HANDOFF.md`
- Trace gate: `.claude/science/physics-loops/conditional-batch-repair-20260528/TRACE_GATE.md`
- Claim status certificate: `.claude/science/physics-loops/conditional-batch-repair-20260528/CLAIM_STATUS_CERTIFICATE.md`

## Verification

- `PYTHONPATH=scripts python3 scripts/frontier_dm_leptogenesis_pmns_projector_interface.py`
- `PYTHONPATH=scripts python3 scripts/wave_direct_dm_h025_point_runner.py`
- `python3 scripts/cached_runner_output.py --check-only scripts/frontier_dm_leptogenesis_pmns_projector_interface.py`
- `python3 scripts/cached_runner_output.py --check-only scripts/wave_direct_dm_h025_point_runner.py`
- `python3 scripts/vocab_lint.py --report-only <34 target docs>`
- `git diff --check`
- `bash docs/audit/scripts/run_pipeline.sh`

## Generated audit result

After the latest rebase onto `origin/main@be771e4fe`, the 33 changed source
rows are `audit_status: unaudited` with zero open dependency paths. Independent
audit still owns every effective-status decision.

PR #2353, opened 2026-05-30
[physics-loop] su3 wigner l3 cube haar mc bounded support

JONbridgerapps:
## Summary

Narrow per-row split from bundle #2128 (`physics-loop/conditional-batch-repair-20260528`). The bundle was an explicit batch-by-date of May 28 `audited_conditional` source-row repairs; per repo policy (one identity at a time), each row gets its own PR.

This PR carries only the scope edits for `su3-wigner-l3-cube-haar-mc-negative-result` from the bundle, with audit-lane / publication-status surfaces stripped.

- This PR does not apply audit verdicts. Effective retained status remains owned by the independent audit lane.

## Verification

`bash docs/audit/scripts/run_pipeline.sh` (run by the bundle); per-row runner if applicable.

### DIFF
--- a/docs/SU3_WIGNER_L3_CUBE_HAAR_MC_NEGATIVE_RESULT_2026-05-04.md
+++ b/docs/SU3_WIGNER_L3_CUBE_HAAR_MC_NEGATIVE_RESULT_2026-05-04.md
@@ -1,27 +1,11 @@
 # SU(3) L_s=3 PBC Cube Haar Monte Carlo: Honest Negative Result
 
 **Date:** 2026-05-04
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Closing without landing. The PR removes audit-repair prose, but it also weakens the finite Monte Carlo significance wording to "within ±1 standard error" for values that are about 1.1-1.2 SE from zero, and restores a precise 1e200 / 150-orders sample-count framing that the previous repair intentionally demoted to non-binding interpretation. That is a claim-boundary regression, not a safe cleanup.

PR #2372, opened 2026-05-30
free Dirac Poincaré generators: essential self-adjointness (closes residual T1 of gap G2)

JONbridgerapps:
## What this delivers (closes residual T1 of program gap G2)

The on-main Poincaré-representation note (`FREE_DIRAC_POINCARE_REPRESENTATION_BOUNDED_NOTE_2026-05-30.md`, commit `72e4e7fa1`) constructs the ten free Dirac Poincaré generators explicitly and verifies the algebra closes — but, as landed, **explicitly does not claim** essential self-adjointness / group integration. This note supplies exactly that missing step (T1): the free Dirac Poincaré generators — especially the unbounded boost `Kⁱ` — are **essentially self-adjoint on a common invariant Schwartz core**, exponentiating to a strongly-continuous unitary representation of the (cover of the) Poincaré group.

**The load-bearing step (derived, not invoked):** on the mass shell `p = m·sinhζ`, the invariant measure is flat (`dp/2E = dζ/2`) and the orbital boost is *exactly* `K_orb = −iE∂_p = −i∂_ζ` — the textbook essentially-self-adjoint momentum operator on all of ℝ (deficiency `(0,0)`); the bounded `SU(2)`-Wigner-carrier term is dispatched by Kato–Rellich. Nelson analytic vectors are **exhibited** with the explicit bound `‖Kⁿψ‖ ≤ C Rⁿ n!`. Standard theorems (Nelson 1959, Kato–Rellich, Stone) are invoked at the level of their *statements*, with their *hypotheses verified here* — so the honest status is **CLOSED modulo cited textbook theorems**.

## Conditional scope

`bounded_theorem`, free field only. Conditional on the explicit generators of the on-main Poincaré note (referenced plain-text, audit-pending — it has no ledger row yet). No statistics selection, no emergent-Lorentz, no interacting claim. The non-triviality controls genuinely fail (half-line `−i d/dx` has deficiency `(1,0)`; a non-Hermitian perturbation fails the battery) — the boost's self-adjointness genuinely uses that rapidity runs over all of ℝ.

## Verification

Adversary gate: NEEDS-FIX → **SHIP**, P=0.78 — independently verified the exact rapidity reduction (`∂_ζ = E∂_p` symbolically), confirmed the Nelson bound is *exhibited* (digit-for-digit), confirmed the N4 controls genuinely fail (non-vacuous), and confirmed "CLOSED modulo cited theorems" is correctly calibrated. The one blocking item — a stale "companion not on main" reference (the companion landed mid-build) — is fixed. Runner `scripts/free_dirac_poincare_generators_selfadjointness_2026-05-30.py`: single-seed deterministic, `SCORECARD PASS=5 FAIL=0`. No audit-lane data.

Together with the on-main Poincaré note, this closes gap G2 (the boost/Poincaré sector of the free-field reconstruction) at the full unitary-group level, conditional on the cited rungs.

### DIFF
--- a/docs/FREE_DIRAC_POINCARE_GENERATORS_ESSENTIAL_SELFADJOINTNESS_BOUNDED_NOTE_2026-05-30.md
+++ b/docs/FREE_DIRAC_POINCARE_GENERATORS_ESSENTIAL_SELFADJOINTNESS_BOUNDED_NOTE_2026-05-30.md
@@ -43,7 +43,7 @@ Live ledger `effective_status` (`docs/audit/data/audit_ledger.json`, verified
 
 | Input | Source | Live `effective_status` (2026-05-30) |
 |---|---|---|
-| The ten explicit free Dirac Poincare generators (`H`, `P^i`, `J^i`, `K^i`) in closed form on the one-particle space, **and** the verified closure of the full Poincare algebra `iso(3,1)` on the Schwartz core (incl. `[K^i,K^j]=-i eps^{ijk}J^k`); the boost preserves `H_m^+` and the invariant measure `d^3p/(2E)`; the spin-1/2 Wigner-rotation carrier `W^i` is the smooth `SU(2)`-valued little-group cocycle | `FREE_DIRAC_POINCARE_REPRESENTATION_BOUNDED_NOTE_2026-05-30.md` (plain text -- on a separate not-yet-merged branch; **not** a citation-graph edge) | not on main |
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed reviewed/narrowed self-adjointness bounded theorem in bf9f49b63.

Review-loop changes applied before landing:
- Added audit metadata to the parent free Dirac Poincare representation note so the dependency graph can see it.
- Made the parent companion a real markdown dependency.
- Fixed the transverse rapidity reduction to use M_perp = sqrt(m^2 + |p_perp|^2).
- Fixed the Wigner carrier convention: the little-group infinitesimal is anti-Hermitian, while the self-adjoint boost uses the Hermitian i times infinitesimal term.
- Removed stale branch-local wording.

Checks run: both free Dirac runners PASS, independent transverse-rapidity / Wigner-Hermiticity / Nelson-domain checks, py_compile, vocab_lint, audit pipeline + strict lint, diff check. Generated audit/publication outputs were restored before commit. The landed rows remain unaudited bounded_theorem candidates; independent audit owns the verdict.

PR #2373, opened 2026-05-30
[physics-loop] source-measure record-intervention proposed bridge

JONbridgerapps:
## Summary

Stacked on #2370. This block attacks the single bridge left by the source-measure P-cal synthesis:

```text
physical source is a smooth sharp-record probability intervention
```

It adds a finite sharp-record representation theorem:

- the sharp-record history algebra is a finite commutative algebra of functions on histories;
- a record-facing physical source is operationally determined by its probability law over those histories;
- with full-support reference, every small smooth intervention has a Radon-Nikodym density and score tangent;
- independent histories compose by product measure, so RN densities multiply and log densities add.

## Claim Status

Source-side status: `proposed_retained` for the finite sharp-record source sector only, pending independent audit. Bare/effective retained status is not claimed.

If accepted, this closes the residual named in #2370 on the finite sharp-record source sector and supplies the source-measure bridge needed for P-cal/Y_T source-scale retirement. Generic non-record dynamics, non-smooth interventions, and support-changing interventions remain out of scope.

## Verification

- `python3 scripts/frontier_source_measure_record_intervention_theorem.py` -> PASS=33 FAIL=0
- Full five-runner stack also passes:
  - RN-cocycle -> PASS=56 FAIL=0
  - Cumulant/Mobius -> PASS=33 FAIL=0
  - Sharp-record tangent -> PASS=38 FAIL=0
  - P-cal synthesis -> PASS=36 FAIL=0
  - Record-intervention theorem -> PASS=33 FAIL=0
- `python3 -m py_compile scripts/frontier_source_measure_record_intervention_theorem.py` -> PASS
- `git diff --check` -> PASS

## Non-Claims

This PR does not claim effective retained status before audit, does not cover generic non-record dynamics, does not provide a strict top/W pole-response certificate, and does not use `H_unit`, `yt_ward_identity`, `y_t_bare`, PDG values, `alpha_LM`, plaquette/u0, Planck inputs, or fitted selectors.

PR #2379, opened 2026-05-30
[draft] source-measure dependent observable-principle P-cal map

JONbridgerapps:
## Summary

Draft dependent impact map for the observable-principle P1/P-cal repair surface.

This PR is intentionally draft. It must not exit draft unless PR #2373 receives a positive audit for the finite sharp-record record-intervention source-measure theorem.

## Dependency

- Depends on: #2373 positive audit
- Base branch: `physics-loop/source-measure-pcal-retirement-block02-20260530`
- No retained/proposed-retained movement is claimed here

## Scope

Maps the parent observable-principle P1/P-cal rows that can be reconsidered if the #2373 source-measure bridge is accepted. Row-level audit is still required before any status update.

## Non-claims

Does not retire determinant realization, action selection, Sommer scale, plaquette, gauge Haar measure, or non-record source premises.


JONbridgerapps:
Closing stale downstream draft map. The source-measure P-cal retirement parent path closed unmerged, so this conditional cascade map should not remain open as an active review surface.

PR #2381, opened 2026-05-30
[draft] source-measure dependent DM neutrino map

JONbridgerapps:
## Summary

Draft dependent impact map for DM/neutrino rows that may share the P-cal/source-measure blocker.

This PR is intentionally draft. It must not exit draft unless PR #2373 receives a positive audit for the finite sharp-record record-intervention source-measure theorem.

## Dependency

- Depends on: #2373 positive audit
- Base branch: `physics-loop/source-measure-pcal-retirement-block02-20260530`
- No DM/neutrino status movement is claimed here

## Scope

Maps likely finite source-surface / Schur / selector rows that need per-row audit after #2373. This is not a bulk promotion.

## Non-claims

Does not close DM abundance, neutrino masses, PMNS ordering, Majorana phases, bosonic normalization, or transport premises.


JONbridgerapps:
Closing stale downstream draft map. The source-measure P-cal retirement parent path closed unmerged, so this conditional cascade map should not remain open as an active review surface.

PR #2397, opened 2026-05-30
fermion-sector OS transfer representation on the 2-step blocked staggered surface (Berezin = operator)

JONbridgerapps:
## What this establishes

The **fermion-sector Osterwalder–Schrader transfer representation equality** on the 2-step blocked staggered surface (reflection `θ(t,x⃗)=(−1−t,x⃗)`, phase `η₁(t)=(−1)ᵗ`): the time-reflected **Berezin/Grassmann** block metric of the staggered fields equals the **operator** two-step transfer eigenvalue `c_block · e^{−2E_j}`, `E_j = asinh√(m²+λ_j²)`.

This is a genuine **representation equality** — *not* the automatic `‖WÔΩ‖² ≥ 0` assembly (that is the prior RP_MIXED lemma). The two sides are computed by **genuinely independent code paths**: the Berezin side via real `M⁻¹` Grassmann contractions (no operator-side reference), the operator side via the Fock transfer algebra. Verified at three levels:
- **per spatial mode** (worst residual ~7e-14);
- the **free many-field position-basis Gram** with full cross-contractions, which independently reproduces `2·e^{−2E_j}` and fixes the block normalization `c_block = 2` *a priori* (mode/mass-independent — principled, not a fit);
- **per fixed gauge background** (U(1)/SU(3), ~1e-16).

The **single-step negative control genuinely fails** (indefinite reflected metric; −0.80 Caracciolo–Palumbo Lagrangian Gram) — the 2-step `η₁` reflection-sign bookkeeping is the load-bearing in-repo derived content; the standard transfer-matrix construction is cited (Lüscher 1977; Osterwalder–Seiler 1978; Sharatchandra–Thun–Weisz 1981; Palumbo 2002; Smit §6).

## Scope (honest — fenced after an adversary gate)

An adversary gate found the genuine core solid (P≈0.93) but flagged that an earlier headline over-reached. This note is narrowed to exactly what the runner establishes. It **does not** establish: the full **gauge-fermion-entangled** (mixed-observable) representation equality (that is RP_MIXED's assembly target and the harder open step — a naive shared-wrapper dual computation reduces identically to the per-mode scalar, so it is *not* an independent test of mixing); the `det(M[U])` weight / Haar `U`-average (separate gauge measure); the continuum / OS-reconstruction (Wightman) limit; Euclidean rotational invariance; or any full interacting-RP closure. `bounded_theorem`.

## Verification

[...]

PR #2399, opened 2026-05-30
gauge-fermion-entangled OS transfer representation on a finite 3+1 staggered carrier (genuine, det-weighted; scoped)

JONbridgerapps:
## What this establishes (the genuine brick #2397 left open)

The **gauge-fermion-entangled** OS transfer representation equality on a finite **3+1** staggered carrier: the time-reflected, **`det(M[U])`-weighted, Haar-U-averaged** Berezin path-integral form of a genuinely-mixed (Wilson-line-transported) observable equals the operator transfer correlator, via **genuinely independent** code paths (Berezin Grassmann/Wick through `M[U]⁻¹` vs Fock second-quantization of `e^{−2Ĥ[U]}`, sharing only the lattice action). Worst `|Berezin − operator| = 7.2e-13` over the mixed basis.

This **fixes both defects** that made #2397's mixed claim vacuous (and was narrowed there):
- the prior Berezin side was per-mode-factorized so the entanglement cancelled identically (`Tber:=Top → 0`); **here** the non-uniform spatial links give a genuinely non-diagonal position-color Gram, so the per-mode-factorized object differs and the control **breaks** (gap 0.23/0.19);
- the prior `det(M[U])` weight was computed but unused; **here** it is the actual `U`-average weight, and dropping it **breaks** the equality (gap 0.10/0.055).

**All four non-vacuity controls fire:** per-mode-factorized breaks; flat (no-det) breaks; single-step breaks (the −0.80 Caracciolo–Palumbo no-go); operator-Schmidt rank 6/3/6 (genuine entanglement) with the equality holding. The decisive anti-tautology check: perturbing **only** the operator-side `E₀` breaks the match — the two sides are not reading a shared magnitude.

## Scope (honest — three explicit limitations, after an adversary gate)

- **Operator side is the classically-`U`-averaged fermion-Fock Gram, not a quantum `H_gauge⊗H_ferm` sandwich** — physically correct on a finite lattice (the gauge field is a c-number integration variable); the tensor-product `T_full` is exhibited separately as the Schmidt-rank diagnostic.
- **SU(3) color-mixing is exhibited on the 1-spatial-dimensional `2×1×1` carrier only**; the multi-site `2×2×1` sheet runs at N_c=1 (abelian) for Fock-tractability. No multi-dimensional SU(3) entangled result is claimed.
- **`F` is a single-creation transported bilinear**, not the number-conserving four-fermion `χ̄Uχ` (which annihilates the OS vacuum — a diagnosed convention obstruction).

Continuum / OS-reconstruction / Lorentz restoration is out of scope (the framework is 3+1; no continuum claim either way).

## Verification

Adversary gate: **SHIP after the three disclosures, P=0.78** — AST-verified the two code paths share only the action; confirmed all four controls are real discriminators (uniform links → off-diagonal exactly 0, proving the entanglement is genuinely link-sourced); confirmed `det` load-bearing and PROBE B anti-tautology. Runner `mixed_entangled_os_transfer_representation_2026-05-30.py`: single-seed deterministic, `SCORECARD PASS=27 FAIL=0`. Sole load-bearing dep: Case-A determinant positivity (`retained`, genuinely consumed). No audit-lane data.

### DIFF
--- a/docs/MIXED_ENTANGLED_OS_TRANSFER_REPRESENTATION_BOUNDED_NOTE_2026-05-30.md
+++ b/docs/MIXED_ENTANGLED_OS_TRANSFER_REPRESENTATION_BOUNDED_NOTE_2026-05-30.md
@@ -76,6 +76,32 @@ gate). This is the **lattice / transfer-matrix** representation, exhibited on a
 finite carrier and **illustrating** the cited general transfer-matrix construction;
 **no continuum claim is made either way**.
 
+## Scope (honest — three explicit limitations)
+
+- **Operator side is the classically-U-averaged fermion-Fock Gram, not a quantum
+  `H_gauge ⊗ H_ferm` sandwich.** On a finite lattice the gauge field is a c-number
+  integration variable, so the verified equality (checks P0/P1) is between the
+  Berezin path and the **per-configuration fermion-Fock vacuum correlator**
+  `<Ω| c_b e^{-2 Hhat[U]} c_{b'}^† |Ω>`, **identically classically U-averaged** on
+  both sides with the weight `e^{-S_G[U]} det(M[U])` and the Wilson amplitudes
+  `W_b(U)`. The genuine tensor-product `T_full` on `H_gauge ⊗ H_ferm` (with `What_b`
+  a quantum multiplication operator) is exhibited **separately** as the
+  operator-Schmidt-rank diagnostic (C4, rank 6); it is **not** the object the P0
+  equality numerically evaluates, and no literal `H_gauge ⊗ H_ferm`
+  operator-sandwich claim is made.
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed source-only salvage on main in 0832e5df9. Review-loop narrowed the branch from exact-Haar / literal gauge-Hilbert sandwich wording to the runner-supported finite Haar-sample/quadrature statement, stripped audit-status authority language, kept the determinant-weight dependency as a markdown edge, and refreshed the SHA-pinned runner cache. Checks run: vocab_lint --fix, py_compile, runner PASS=27 FAIL=0, independent one-particle operator check, audit pipeline + strict lint for validation, generated audit/publication outputs stripped before commit.

PR #2409, opened 2026-05-31
feat(koide): generation reality-type permitted-not-forced; reality of D necessary-not-sufficient

jonathonreilly:
## Summary

The substep4 reality-type attack (8 agents) resolves the last Koide pin cleanly and honestly: the generation reality-type — and hence the measure and `Q` — is **permitted-not-forced**. The whole charged-lepton value reduces to one scalar `b/a`. Runner (10/10):

- **Kill-shot.** `cpt_exact`'s "real anti-Hermitian `D`" is real *matrix* on an explicitly complex `C^N`. **One** such `D` admits *both* the symplectic bilinear `u^T D v` (Majorana/Pfaffian → per-block → **Q=2/3**) **and** the Hermitian `conj(u)^T (iD) v` (Dirac/det → per-dim → **Q=1**). Reality of the matrix selects **neither** — the selector is the field-reality (`chi-bar` independent vs `chi-bar=chi^T s`), not `D`. "Real `D` forces Q=2/3" conflates real-matrix with real-field (**dissolved**, without closing Q=2/3).
- **Native lemmas.** `Jcs` is the unique `C_3`-equivariant real-antisymmetric operator on `R^3` (nullspace dim 1); `(Jcs|doublet)^2=-I_2` (the per-block "1" is *computed*); Pfaffian degree 1 vs det degree 2 (per-block vs per-dim counting **theorem**); `[H,Jcs]=0` with **3 distinct real masses** (the real face is self-consistent with the Dirac-looking spectrum).
- **The default.** The retained inventory *computes* Dirac/Q=1 (substep1 independent pair, substep2 central-`i`, cl3 complexified module); Q=2/3 is native and non-circular **except** the one Wick-face selection.

## Boundary

Not a closure. Pins the residual to a single Wick-face / coupling-ratio scalar `b/a` (`r=1/2` vs `r=1`) = the campaign's already-isolated single handle (`koide_frobenius_isotype_split_uniqueness`, retained_no_go), shared with generation-ID. **Next path** (dodges the `C^3=I` obstruction that forbids continuous `U(1)_b`): a `C_3`-invariant first-order Berry/Wess-Zumino term `B^T Jcs B-dot`, writable without `U(1)_b` — is it native-generated? (Obvious sources `D_KD` and the central pseudoscalar are verified silent/mislocated.) Complements #2406, #2407. Corrects a stale citation: the signed-vs-singular readout note is **audited_FAILED** on origin/main.

## Test Plan
[...]

PR #2407, opened 2026-05-31
feat(koide): Kähler-Dirac is silent on the measure; value reduces to one substep4 binary

jonathonreilly:
## Summary

The 8-agent P1 measure attack asked whether emergent-time quantization counts the generation factor per-complex-mode (`-> r=1/2 -> Q=2/3`) or per-real-dimension (`-> r=1 -> Q=1`). Verdict: **`Q=1` is the native default on the retained tier** — but the residual is now localized maximally to **one discrete open input**. Runner (10/10) certifies:

- **Kähler-Dirac is silent.** `D_KD = d - delta` on `Lambda*(R^3)` is real-antisymmetric with `i D_KD` Hermitian, but its grade-diagonal blocks **all vanish** — in particular `Lambda^1 -> Lambda^1` (the generation triplet) `= 0`. The retained dynamics supplies **no** within-generation kinetic operator.
- **`det_C` is a red herring.** `det(aI+b(C+C^2)) = (a+2b)(a-b)^2` and `Pf(M (x) eps) = +-det(M)` both weight the doublet to **dimension 2** `-> Q=1`. The per-block `(1,1)` measure is **not** a determinant/Pfaffian output.
- **Neither native structure is a full-space `J`.** Central `i*I` (eigs `{+i,+i,+i}`, generation-blind `-> Q=1`); `Jcs` (`det 0`, doublet once `-> Q=2/3`).
- **The one binary.** The only `C_3`-equivariant **real-antisymmetric** operator on `R^3` is `span{Jcs}`; symmetric ones are measure-blind. So the value = *does the generation action carry a within-generation real-antisymmetric (symplectic) bilinear* (`-> Jcs -> Q=2/3`) *or not* (`-> central-i -> Q=1`)? `= ` the open **substep4** reality-type.

## Boundary

Not a closure. **Sharpens** `koide_frobenius_isotype_split_uniqueness` (retained_no_go) from "rep theory ranks neither weight" to "and the retained Kähler-Dirac dynamics is **provably silent**, so the deciding authority is precisely substep4" — a *dynamical* selector, exactly the type the *static* no-go cannot supply or pre-refute. Complements #2406 (the two-bit decomposition + native Kähler triple). Two adversarial cautions verified: the `arg(b)`-dynamical framing is a redundancy (`[Jcs,C]=0`, retained); the OS-Born-metric route leans on an audited_conditional object.

## Test Plan
[...]

PR #2406, opened 2026-05-30
feat(koide): chiral import dissolves into native Kähler triple + two independent bits

jonathonreilly:
## Summary

A 14-agent **find-the-derivation** panel on the shared chiral import (the one object behind Koide `Q=2/3`, generation-ID, strong-CP, signed-gravity) found **no** native derivation from `A1+A2+`retained+emergent-time — but two **durable, numerically-definitive** structural advances, verified here from scratch (11/11).

### (1) The Kähler triple is native — corrects the "missing-ω" framing
`Jcs=(C-C^2)/sqrt3` is **exactly** the `so(3)/Cl(3)` generator of the `C_3` rotation about `(1,1,1)` (`exp((2pi/3)Jcs)=C`, residual 0). With native `g` and `J`, the 2-form `omega=g.Jcs` is **forced native** (a Kähler 2-form *is* `g∘J`), the Weyl projector `P_weyl=(1/2)(P_doublet-i Jcs)` is native, and the doublet is complex-type (FS-indicator 0). This **corrects** `KOIDE_GENERATION_HERMITIAN_NOT_KAHLER_NOTE` (PR #2403): the geometry is Kähler; that note's conjugation-even `det M` is a **dynamical** statement (no first-order kinetic term), not a missing geometric 2-form.

### (2) The residual import = two provably-independent bits
- **bit(i)** orientation of `Jcs` (`det_C` vs `det_R`; signed-`√m` sign) — a `Z_2` readout datum.
- **bit(ii)** the modulus `r=1/2` (equal `C_3`-block energy; `Q=(1+2r)/3`) — a per-block-vs-per-dimension **measure** choice, **not** chirality.

**Provably independent:** `[H,Jcs]=0` for every circulant mass operator at **all** `r` (`max|[H,Jcs]|=0`), so orienting `J` never forces `r=1/2`. Hence **no orientation principle closes `Q=2/3` alone**. The import is ~80% a measure choice + ~20% an orientation sign.

### (3) The anticommuting-operator class is physically excluded
Every operator in `anticomm(Gamma_chi)` over `Sym(R^3)` has spectrum `{-s,0,+s}` — a **massless generation** (excluded for charged leptons). This closes the `W1` escape hatch (drop `C_3`-equivariance) on **physical** grounds.

## Boundary

[...]

PR #2405, opened 2026-05-30
feat(koide): qubit-factor Berry holonomy is import-sourced and r-non-selective

jonathonreilly:
## Summary

Runs the **never-executed** qubit-factor Berry holonomy of the `delta=arg(b)` loop on `R^3 (x) C^2` — the *one open seam* the equivariant-`eta` route named for sourcing the missing symplectic form `omega` (and hence `r=1/2`, `Q=2/3`) on the **native qubit factor**, where the `eta`-index is blind (the tensor-coin spectrum is `±`-symmetric). The verdict is **import-confirmed, not the derivation**: the holonomy is zero for the faithful embedding, nonzero only via an imported chiral coin, and r-non-selective even then.

This is **not** a closure — it maps **where** the `r=1/2`-selecting principle is missing (the single chiral import shared with Koide-`Q` and generation-ID) and opens the next path.

## Runner certificate (10/10, gauge-invariant, three-method)

`scripts/frontier_koide_qubit_berry_holonomy_import_2026_05_30.py`:
- **F1** — C₃-equivariant anticommutant of `B (x) I2` is **dim 0** (bare-tensor wall) while `B (x) sigma_x`, `B (x) sigma_y`, `B (x) sigma_z` are **each dim 6** → `sigma_z` is unprivileged (qubit-`SU(2)`-related).
- **F2** — the **faithful** embedding `(cos d * B + sin d * iJcs) (x) sigma_z` has `sigma_y` component **identically 0** per Fourier mode → planar loop → Berry holonomy **= 0 exactly** (Wilson-loop + Uhlmann + non-abelian lower-3-band all trivial). The bare-`R^3` circulant flatness persists verbatim into the tensor seam.
- **F3** — nonzero holonomy requires the non-collinear coin `B (x) (cos d * sigma_z + sin d * sigma_y)`, where `B (x) sigma_y` is HS-**orthogonal** to the `iJcs` sector — the C₃-orbit-splitting chiral **import** (the inter-axis relative-`i`), not a faithful `arg(b)` image.
- **F4** — even granting the import, `Gamma(r) = -pi(1 - 1/sqrt(1+4r))` is monotone, `dGamma/dr != 0` at `r=1/2`; hits `2/9` at `r ~ 0.0395`, `2pi/9` at `r ~ 0.163`, `-2pi/3` at `r=2` — **never** `r=1/2`.

## Boundary / tier honesty

[...]

PR #2412, opened 2026-05-31
feat(koide): Q=2/3 is the K0-real (real-Wedderburn-block) reading of A1+A2; qubit i doesn't fix it

jonathonreilly:
## Summary

Answers the reverse-engineering question *what type of axiom construction natively gives Koide Q=2/3 instead of Q=1?* — and finds the wall **moves**: Q=2/3 is a **native convention-derivation** within the existing A1+A2, not a different theory. Runner (10/10):

- **The fork is `K0`-real vs `K0`-complex (Frobenius-Schur).** `Z_3` has FS `(+1,0,0)`, so `R[Z_3]=R(+)C` has **2** real blocks (`K0`-real `=Z^2`; the complex-type doublet is **one** block) and `C[Z_3]=C^3` has **3** (`K0`-complex). Count each real block once → equal block energy → `r=1/2 → Q=2/3` (verified end-to-end: `sum lambda=3`, `sum lambda^2=6`); dimension-count → `r=1 → Q=1`.
- **The qubit `i` is a generation scalar.** `Cl(3)`'s pseudoscalar `s1 s2 s3 = i*I_2` is forced on the **qubit** factor but acts on the generation index as the **scalar** `i*I_3`, not the doublet `J=diag(0,+i,-i)` the per-block count needs. So the complexification that would push `Q=1` **doesn't reach** the generation factor — the doublet measure is a **free convention slot**.
- **A qubit is natively a coherent-state object** (`(2/4pi) int |n><n| dOmega = I_2`), so the per-block (one-holomorphic-mode) reading is arguably the **more faithful** reading of A1 → `Q=2/3`.
- **Correction:** `C^3=I` forbids `U(1)_b` as a *symmetry* but **not** the Schur complex structure `J` the `det_C` *measure* uses — so it does **not** make `Q=1` the forced default (corrects an over-stated local claim). Neither measure forced.

## Boundary

`Q=2/3` uses **no new axiom and no different substrate**; the generation carrier is the real `R[Z_3]`, a qubit is a coherent-state object, and the real/block reading is already retained-supported. But A1+A2 **also** admit the trace-default `Q=1`, so neither is **uniquely forced**; committing to `det_C` is the one user-approval import (unit-convention precedent). The observed `2/3` is the canonical idempotent/`K`-theoretic count of the real `Z_3` structure `A2` supplies for free — not a coincidence. Sharpens `koide_q23_block_weight_frontier` (retained_bounded). Complements #2406, #2407, #2409.

## Test Plan
- [x] Runner 10/10; FS fork, qubit-`i`-as-generation-scalar, coherent-state completeness, end-to-end `Q=2/3`, `C^3=I` correction all verified
- [x] Cache SHA-pinned and fresh; 1 note + 1 runner + 1 cache; sets no audit verdict

### DIFF
--- a/docs/KOIDE_Q23_K0_REAL_BLOCK_EQUIVALENCE_NOTE_2026-05-30.md
+++ b/docs/KOIDE_Q23_K0_REAL_BLOCK_EQUIVALENCE_NOTE_2026-05-30.md
@@ -1,0 +1,97 @@
+# Koide: Q=2/3 is the K0-real (real-Wedderburn-block) reading of A1+A2; the qubit `i` does not fix it
+
+**Date:** 2026-05-30
+**Claim type:** equivalence-characterization / bounded structural localization (positive)
+**Status:** structural result. Approves no axiom and no import; sets no audit verdict.
+The audit lane sets status and the per-block-vs-per-dimension convention tier.
+**Primary runner:**
+`scripts/frontier_koide_q23_k0_real_block_equivalence_2026_05_30.py`
+with cache
+`logs/runner-cache/frontier_koide_q23_k0_real_block_equivalence_2026_05_30.txt`.
+
+## Result (one sentence)
+
+The charged-lepton Koide value is a single **reality-structure bit** on the generation
+matter measure: `Q=2/3` **iff** the **real-Wedderburn-block** (`K0`-real, count each
+block once / coherent-state) quantization of the A1+A2 generation carrier, and `Q=1`
+**iff** the **complex-dimension** (`K0`-complex / trace-default) quantization — and the
+`Cl(3)` central pseudoscalar `i`, which would push `Q=1`, is **forced only on the qubit
+factor** and acts on the generation index as a **scalar**, so it does **not** fix the
+choice.
+
+## The fork is `K0`-real vs `K0`-complex (Frobenius-Schur)
+
+`Z_3` has Frobenius-Schur indicators `(+1, 0, 0)` (F1): the **real** group algebra
+`R[Z_3] = R (+) C` has **two** real-irreducible blocks (`K0`-real `= Z^2`) — the
+complex-type doublet `(omega, omega-bar)` is **one** real block of real-dimension 2 —
+while the complexified `C[Z_3] = C^3` has **three** blocks (`K0`-complex `= Z^3`). With
+`Q = (1+2r)/3`, `r = |b|^2/a^2`, and block energies `E_+ = 3a^2`, `E_perp = 6|b|^2`:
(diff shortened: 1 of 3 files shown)


jonathonreilly:
**Consolidation map for the charged-lepton Koide-value PRs** (this PR is the canonical/final statement; posting so the review-loop sees one coherent direction, not a drifting sequence).

The understanding narrowed across the session; the open set now tells **one** story, with this PR as the consolidated result:

| PR | Verified contribution | Status |
|----|----------------------|--------|
| ~~#2403~~ | "missing ω = the import" | **closed** — wrong framing (ω is native) |
| #2405 | the qubit-factor Berry-holonomy route does **not** source the measure (import-sourced, r-non-selective) | keep — a negative that holds |
| #2406 | the import dissolves into a **native Kähler triple** + two independent bits (orientation Z₂ + modulus); anticommutant excluded (massless generation) | keep — foundational |
| #2407 | `D_KD`'s within-generation block ≡ 0 → the retained **dynamics is silent** on the measure | keep — distinct fact |
| #2409 | `cpt_exact`'s real **D** is **necessary-not-sufficient** (one D admits both Wick faces) | keep — distinct fact |
| **#2412** | the measure is the **K₀-real vs K₀-complex** (Frobenius-Schur) bit; Q=2/3 is the native coherent-state reading; the qubit `i` is a generation **scalar**, proven irrelevant | **canonical** |

**Terminology reconciliation (the one cross-PR clash):** "det_C" is overloaded. #2407 correctly shows the **ordinary** spectral determinant `det(M)=(a+2b)(a−b)²` weights the doublet by its dimension → **Q=1** (a red herring for per-block). This PR's "det_C / K₀-real" is the **reduced/Dieudonné** determinant (Nrd) over the doublet's endomorphism field ℂ, which counts the block **once** → **Q=2/3**. Different determinants; both statements correct; the per-block measure is the reduced one. The two PRs **agree** (the ordinary det does not give per-block; the K₀/reduced count does).

Net coherent direction: route after route (qubit-Berry, D_KD dynamics, reality of D, Berry/WZ) fails to **force** the measure → it is a free native slot = the K₀-real-vs-complex bit → Q=2/3 is the natural (coherent-state) reading, with the Q=1-forcing central `i` proven to live on the wrong factor.

jonathonreilly:
Review-loop landed as commit 3029c7d83 with the claim narrowed to a bounded equivalence/localization. I kept the real-Wedderburn/per-block vs trace/dimension result and the qubit-i irrelevance check, stripped the A1+A2/approval/import overclaiming, refreshed the runner cache, and seeded the row as unaudited/ready for the independent audit lane.

PR #2418, opened 2026-05-31
Repair observable principle log baseline theorem

JONbridgerapps:
## Summary
- remove the overstated universal `+ const` from the exact multiplicative-to-additive theorem
- separate the exact solution class `W(r)=c log r` from the zero-source baseline subtraction convention
- add a runner check showing a universal additive constant violates the functional equation while baseline subtraction stays additive on direct sums

## Verification
- `python3 scripts/frontier_hierarchy_observable_principle_from_axiom.py` (passes 27/27; existing NumPy determinant warnings still print)
- `git diff --check`

No audit ledger/results/status retagging is included. P1/P2 remain admitted exactly as source-scoped.


jonathonreilly:
Closing without landing, superseded: the no-universal-additive-constant repair landed via 219a625b9 (2026-06-04) — main now states 'A universal additive constant is not part of the exact Cauchy solution' with the zero-source-subtracted representative — and the headline this PR patches was rewritten again on 2026-06-08/06-10 (P2 is now eliminated, not admitted), so this diff no longer applies.

PR #2430, opened 2026-05-31
[physics-loop] sm-gstar-higgs-scope conditional-support

JONbridgerapps:
## Summary
- narrow the Higgs-sector g_* note from bounded theorem to conditional support
- name the missing H_unit -> one-complex-EWSB-doublet thermal-census bridge instead of treating D17 scalar-singlet uniqueness as closure
- keep the exact one-vs-two-doublet g_* arithmetic and executable flavor-vs-Fock support checks, with refreshed cache

## Verification
- `python3 scripts/frontier_sm_gstar_higgs_sector_count_2026_05_29.py`
- `python3 scripts/cached_runner_output.py --check-only scripts/frontier_sm_gstar_higgs_sector_count_2026_05_29.py`
- `git diff --check`

No audit ledger/results/status retagging is included.


jonathonreilly:
Closing in favor of #3552 — the same three-file set re-cut verbatim on a fresh branch off current origin/main (the old base was verified byte-identical to main for all three files, so nothing was lost or changed). Runner re-verified on the new branch: PASS=53 FAIL=0, cache fresh. Re-cut as a ready (non-draft) PR so the review loop can pick up the narrowing before the queued audit hits the missing-bridge blocker.

PR #2456, opened 2026-05-31
observable-principle P1: exponent-fixing irreducibility (sharpened no-go pinning the admitted atom)

JONbridgerapps:
## What this is

A **sharpened no-go** that *pins* the irreducible atom of the admitted **P1** premise (scalar additivity ⇒ `W = log|det(D+J)|`), the highest-leverage Tier-A input. It does **not** retire P1 — it locates the admission precisely and proves the surrounding structure. Math independently re-verified before this PR was opened.

## Result

The scalar-readout selection splits on two algebraically distinct axes:

- **Stage FORM (theorem, not admission):** on the operator-**product** axis, the multiplicative-character requirement `χ(A·S) = χ(A)·χ(S)` selects `det` and excludes `tr` (GL(n) abelianization). This is content the direct-sum no-go portfolio could never supply — because `tr` *passes* direct-sum additivity (verified: `det(A·S)=det A·det S`; `tr(A·S)≠tr A·tr S`; `tr(A⊕S)=tr A+tr S`). Credited to the existing det-character note; this note proves its residual exponent step is P1.
- **Stage EXPONENT (the irreducible admitted atom):** the product axis leaves the exponent **free** — `|det|^p` is a product-character for *every* real `p` — so fixing the exponent so the real generator is additive over independent sectors **is** P1.

**Load-bearing Lemma (reproven, exact):** on smooth `W` with `W[0]=0`, the exponent-fixing selectors **(Add)** additivity, **(Loc)** cross-block 2nd-derivative vanishing, **(Pot)** Gibbs-potential locality are **logically equivalent** — so Cauchy, locality, cumulant, and Gibbs-potential selectors are all P1 in different vocabulary. The one selector that is *not* additivity — the normalized/Born gradient `(1/p)Z^{-p}∂(Z^p)/∂j = ∂log Z/∂j` — returns the same field for **every** `p`, so it selects nothing. That closes the last escape hatch.

This **resolves** the live §6 disagreement in the 2026-05-28 two-stage synthesis (whether `P-cal`/Gibbs-potential is a new primitive or P1) — with proof, in favor of the skeptical reading: `P-cal = (Pot) = (Add) = P1` relocated to differential form.

## Circularity (flagged honestly)

The Lemma is the same multiplicative→additive content as the Cauchy classifier. It is **not** used to "derive log" (that would be circular — additivity is the hypothesis); it is used only to prove the admission is irreducible.

## Discipline

Independently re-verified the load-bearing math (det/tr product-character, `|det|^c` free, Born-selector exponent-blindness, (Add)⇔(Loc) separability). `claim_type: no_go`; pins, does not retire P1; status authority = independent audit lane only. Companion runner `audit_companion_observable_principle_p1_exponent_irreducibility_2026_05_31.py` reproves every load-bearing fact from primitives — **PASS=17 FAIL=0**. No audit-lane data, no new axiom, no new vocabulary (builds on the existing two-stage framing), literature cited as comparator only. Ledger note: the parent `observable_principle_from_axiom_note` is `unaudited` on the live ledger (memory/CLAUDE.md said `audited_conditional` — stale); this note does not depend on any cited status.


jonathonreilly:
Review-loop landed the salvage on main in fe3b33594. I narrowed the no-go boundary to the tested selector families, stripped expected-audit/status authority, kept context-only references out of the dependency graph, refreshed the SHA-pinned runner cache, and verified the row seeds as no_go / unaudited / deps=[] / ready=true. Checks: py_compile, runner SCORECARD PASS=17 FAIL=0, vocab_lint, audit pipeline validation, audit_lint --strict, git diff --check. Generated audit/publication outputs were restored before commit.

JONbridgerapps:
Refinement (from the FORM-premise integrity check, #2503): this note's FORM-stage characterization — "the form step (det-vs-tr) is a theorem on the · axis and is **not admitted**" (§4.1, and the §2 table cell "THEOREM (not admitted)") — should read **"theorem *given* the admitted product-character premise (M)"**. #2503 proves (M) (`χ(A·S)=χ(A)·χ(S)`) is a **co-admission**: motivated by the Berezin source-insertion structure but not entailed by it (non-sequitur witnessed by `tr`, which respects the motivations yet violates (M)). So the scalar-readout chain carries **two** independent admitted atoms — (M) on the product axis and P1 on the direct-sum axis — not one. This note's core result (the exponent-fixing **is** the irreducible atom of the ⊕ stage; the Lemma; Born-too-weak) is unaffected; only the 'sole admission' framing is corrected. No new Tier-A item is banked.

PR #2462, opened 2026-06-02
docs(koide): P1 collapses the two carrier-frame residuals to one (faithfulness) — modulo audit; scalar admitted

jonathonreilly:
## Summary

The convergent kill — both carrier-frame levers hand off to one P1 constraint (microcausality / spin-statistics / reflection-positivity on M). **Verdict: P1 reduces the two residuals to one (faithfulness) as correct physics, but tier-conditionally; and never to zero.** (runner 10/10, non-circular.)

### (A) The collapse is correct physics — `L1 ⟸ G1`
Given the faithful spin-½ Weyl rep (G1), Bose-quantizing the negative-energy mode is **unbounded below** while CAR is bounded → CAR is the unique positive-energy quantization. So **statistics (L1) follows from the faithful rep (G1)** — two frame posits collapse to one. (Statistics is the conclusion, not an input.)

### (B) But tier-conditional — P1 doesn't exclude the hard-core boson on the retained surface
The retained cardinality core excludes only the **free/soft CCR** boson (`[a,a†]=I` needs infinite dim); the **hard-core** boson `b=σ_+` has `[b,b†]` traceless (evades it) and on a single site is the *same* 2×2 matrix as the fermion. So the carrier forcing rides **unaudited** rows (`axiom_first_spin_statistics`, `free_field_os_wightman_reconstruction`) — `L1⟸G1` is **real-modulo-audit**, not retained-load-bearing.

### (C) P1 does not force G1 — the scalar is admitted, never zero
The trivial scalar `J=K=0` is **positive-energy + microcausal + reflection-positive** (the canonical RP theory). RP can't exclude a rep whose 2-point function is itself RP. So **faithfulness is the lone irreducible frame posit**.

### Honesty
Does **not** claim the Lieb–Robinson commutators are byte-identical (they aren't — flagged and excluded); only the NN-bilinear *spectrum* is unitary-identical across frames.

### Net
Carrier frame: **two posits → one (faithfulness)**, with `L1⟸G1` correct physics, **modulo auditing the spin-statistics/OS step**. Next: (1) audit that step → `L1⟸G1` retained-clean; (2) **faithful-Weyl-over-scalar** is then the single auditable frontier — untouched by P1, to be pursued via M's spin content, not microcausality.

## Changes
- `docs/KOIDE_P1_COLLAPSES_FRAME_RESIDUALS_NOTE_2026-06-01.md`
- `scripts/frontier_koide_p1_collapses_frame_residuals.py`
- `logs/runner-cache/frontier_koide_p1_collapses_frame_residuals.txt`

## Test Plan
- [x] Runner 10/10; non-circular (Dirac forcing is the spin-½ c-number spectrum; statistics derived)
- [x] Reproduced: Bose-unbounded/CAR-bounded; cardinality blind to hard-core boson; scalar RP-PSD + positive-energy; NN-bilinear spectrum unitary-identical
- [x] Tier-honest: collapse rides unaudited spin-statistics/OS rows (flagged, not cited as retained); G1 not forced
- [x] Does NOT repeat the self-flagged LR-byte-identical overstatement
- [x] Retained anchors verified on `origin/main`

### DIFF
--- a/docs/KOIDE_P1_COLLAPSES_FRAME_RESIDUALS_NOTE_2026-06-01.md
+++ b/docs/KOIDE_P1_COLLAPSES_FRAME_RESIDUALS_NOTE_2026-06-01.md
@@ -1,0 +1,112 @@
+---
+claim_id: koide_p1_collapses_frame_residuals_note_2026-06-01
+claim_type_author_hint: bounded_theorem
+status_authority: independent_audit_lane_only
+direct_effective_status_change_allowed_from_this_note: false
+---
+
+# P1 collapses the two carrier-frame residuals to one (faithfulness) — modulo auditing spin-statistics; and does not force faithfulness (the scalar is admitted)
+
+**Date:** 2026-06-01
+**Claim type:** bounded collapse-theorem + no-go-companion. Adds no axiom and no
+import. `Q=2/3` never enters.
+**Status authority:** independent audit lane only.
+**Primary runner:**
+`scripts/frontier_koide_p1_collapses_frame_residuals.py`
+with cache
+`logs/runner-cache/frontier_koide_p1_collapses_frame_residuals.txt`
+(10/10 checks).
+
+## The convergence
+
+Two just-derived carrier-frame residuals point at one constraint:
+
+- **G1 (boost lever, `KOIDE_ONSITE_WEYL_BOOST_FROM_BIVECTORS_NOTE_2026-06-01`):**
+  faithfulness — matter in the boost-acting Weyl rep vs the trivial scalar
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the durable bounded science on main in 54b246cf1 as part of the Koide matter-attachment cascade. I normalized this to bounded_theorem source surfaces, removed branch-local status/audit-authority scaffolding, kept unaudited/conditional rows as context rather than dependencies, refreshed runner caches from the edited runners, and ran the audit compatibility pipeline plus strict lint and diff check. The new rows are unaudited and ready for the independent audit lane.

PR #2464, opened 2026-06-02
feat(koide): matter-attachment reduces to the Kawamoto-Smit audit

jonathonreilly:
## Summary

Front 2 of the carrier-frame faithfulness attack, one level deeper. The faithfulness note (matter `M` in a faithful boost-acting Weyl rep vs the trivial `J=K=0` scalar) reduced to a single upstream **matter-attachment** posit: *matter field index = the per-site `C^2` qubit STATE carries the `j=1/2` spinor rep of the physical spatial rotation as its transformation law.* This note settles whether the native first-order real anti-Hermitian `D` + the merger + emergent Lorentz **force** that attachment WITHOUT the unaudited Kawamoto-Smit reconstruction.

**Verdict: it does not — the matter-attachment REDUCES TO THE KS AUDIT.** The only forcing route rides the unaudited `staggered_dirac_kawamoto_smit_forcing`, with an admitted-not-forced elementary fallback. So faithfulness does **not** collapse to the `(3,1)` signature + carrier-id.

## The three facts (runner, 10/10 PASS, non-circular)

- **A** — native `D` is single-component on the lattice (no per-site spinor index), spin-blind on the `C^2` (`[H⊗I₂, I⊗σ_i/2]=0`), `-D²` a scalar mass-shell `{0,4,8,12}`.
- **B** — the merger is OPERATOR-FRAME (adjoint): `U(R) σ_i U(R)† = R_ij σ_j` (conjugation on the gamma *operators*), one level below the matter-STATE spinor law, which `per_site` explicitly disclaims (first carve-out unrestricted).
- **C** — the **KS bridge**: `Ω(x)=σ₁^{x₁}σ₂^{x₂}σ₃^{x₃}` diagonalizes the spinor away into the staggered phase, `Ω(x)† σ_μ Ω(x+ê_μ) = η_μ(x) I`. The spin-rotation↔sign-field translator IS `Ω` = the KS reconstruction (unaudited). The native `D`'s rotation symmetry is restored by a single-component diagonal sign field `W∈{±1}` — a `C²` spin rotation is **not** required.

## Residual (localized)

[...]

PR #2465, opened 2026-06-02
feat(koide): matter-attachment reduces to the cross-site graded-statistics gate

jonathonreilly:
## Summary

Consolidation of a **four-probe parallel fan-out** on the matter-attachment pin localized by PR #2464 (the per-site C² qubit STATE carrying the `j=½` spinor rep of physical spatial rotation, one level above the merger's operator-frame adjoint covariance). The probes fired one lever each at forcing that attachment from A1+A2+retained. **All four fail — for a common reason — and the pin reduces to a single cross-site graded-statistics gate.**

## The four levers (runner-verified, non-circular)

| Lever | Verdict |
|---|---|
| **(4) Selection principle** | The spin-blind compensator `W(R)` closes into an **honest untwisted** octahedral rep (`V(R₁)V(R₂)=+V(R₁R₂)`, **0/576** pairs twisted); the genuine spin lift `U(R)` is **projective** (**208/576** need the `2π=−1` sign). Both are `D`-covariant ⇒ rep-closure does **not** force the spinor. |
| **(3) Boost** | A scalar 2-point is `SO(3,1)`-covariant with no spinor (already retained: `lorentz_boost_covariance_3plus1d`); covariance does not create the index. |
| **(2) State-level theorem** | The adjoint→fundamental upgrade is an **identification, not a theorem**: Space 2 (matter field index) = Space 1 (qubit C²), an **open gate** (`MINIMAL_AXIOMS` excludes particle sectors). |
| **(1) KS-audit** | KS-forcing is forcing only at the *consistency* level; rides the statistics horn + a `k=1` multiplicity horn. |

## Common root

`D`'s spatial covariance factors through `SO(3)≅O` and is **blind to the SU(2) cover**. The native on-site `2π=−1` spinor sign DOES live on the qubit C² (`binary_octahedral_discrete_spinor_sign`, retained_bounded) but is **decoupled** from everything `D`'s covariance sees. The discriminator is **graded (fermionic) locality** — a cross-site property — not single-site covariance. That gate is the retained no-gos `staggered_dirac_substep1_statistics_agnostic_no_forcing` + `fs_rotation_exchange_discrete_insufficiency`, and the same gate as generation-ID chirality.

## Tiers (verified on origin/main)
lorentz_boost_covariance_3plus1d `retained_bounded` · binary_octahedral `retained_bounded` · u4_conditional `retained_bounded` · pre_record_tracial `retained` · statistics_agnostic_no_forcing **`retained_no_go`** · fs_rotation_exchange **`retained_no_go`** · KS-forcing **`unaudited`** · boost_free_staggered_fermion / dirac_weyl_admission_bridge **`audited_conditional`** (off the retained surface).

## Test plan
- [x] Runner — 10/10 PASS; octahedral cocycle reproduced (208/576 vs 0/576)
- [x] All cited tiers re-verified on `origin/main` ledger
- [x] Non-circular (faithful rep / `Q=2/3` never assumed)

### DIFF
--- a/docs/KOIDE_MATTER_ATTACHMENT_GRADED_STATISTICS_GATE_NARROW_THEOREM_NOTE_2026-06-02.md
+++ b/docs/KOIDE_MATTER_ATTACHMENT_GRADED_STATISTICS_GATE_NARROW_THEOREM_NOTE_2026-06-02.md
@@ -1,0 +1,160 @@
+# Matter-Attachment Reduces to the Cross-Site Graded-Statistics Gate
+
+**Date:** 2026-06-02
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only; effective status is
+pipeline-derived after audit.
+**Primary runner:** [`scripts/frontier_koide_matter_attachment_graded_statistics_gate.py`](../scripts/frontier_koide_matter_attachment_graded_statistics_gate.py)
+
+## Context
+
+The companion note `KOIDE_MATTER_ATTACHMENT_REDUCES_TO_KS_AUDIT_NARROW_THEOREM_NOTE_2026-06-02`
+localized the charged-lepton faithfulness pin to a single matter-attachment
+posit: the per-site `C^2` qubit STATE carrying the `j = 1/2` SPINOR rep of
+the physical spatial rotation as its transformation LAW (action on the ket),
+one level above the merger's OPERATOR-FRAME (adjoint) covariance
+`U(R) sigma_i U(R)^dag = R_ij sigma_j`. Four independent levers were then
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the durable bounded science on main in 54b246cf1 as part of the Koide matter-attachment cascade. I normalized this to bounded_theorem source surfaces, removed branch-local status/audit-authority scaffolding, kept unaudited/conditional rows as context rather than dependencies, refreshed runner caches from the edited runners, and ran the audit compatibility pipeline plus strict lint and diff check. The new rows are unaudited and ready for the independent audit lane.

PR #2468, opened 2026-06-02
feat(koide): Hermitian-records target is import-required; does not remove bounded

jonathonreilly:
## Summary

Determines the status of the RP/spectrum cascade's antecedent (PR #2467 flagged "records are Hermitian → trace-symmetric dissipator → `T` positive Hermitian" as an *unbuilt* target). **Verdict: import-required, not forced from A1+A2+retained.** This *refines* PR #2467 — the route it called importless actually **trades** the staggered/Wilson import for a smaller, more local one.

## The counterexample (runner 10/10)

A phase-twisted diagonal channel `K_r = √p_r·diag(e^{iα_r}, e^{iβ_r})` is:
- ✅ valid CPTP (Choi PSD), trace-preserving, **unital** (fixes `ρ_ref=I/2`)
- ✅ **persists a classical record** (Z-diagonal populations preserved)
- ❌ but **non-Hermitian Kraus**, non-trace-symmetric dissipator, **`spec(T) = {1, 1, 0.5±0.5i}`** → not positive Hermitian.

So **persistence does not force Hermiticity.** The exact residual: `K=U·P` is self-adjoint **iff `U=I`** (no relative branch phase). The retained `lsp_projective_derivation` *proves* `K_P=P` is a frame **convention** (the `U`-twist is admitted), and the framework's decoherence *carries* phases `e^{ikS}`. So the import is `NO-RELATIVE-BRANCH-PHASE` (`U=I`).

## Your bounded question, answered: **NO**

Closing this would not remove `bounded` down the full chain:
1. **It rides retained_bounded inputs** (persistent_record_as_kraus, decoherence, lsp, luders) → inherits bounded.
2. **The matter-attachment leg wouldn't even reach retained** — siblings unaudited (spectrum_condition, RP, OS, microcausality) + a `retained_no_go` (statistics_agnostic).
3. **The VALUE chain is orthogonal** — dominated by `AC_phi_lambda` (δ=2/9 + species bridge, **audited_renaming**), `r=1/2` (block weight, **retained_bounded**), and the signed-readout class (**audited_failed**).

## Two stale-premise corrections surfaced
- the `axiom_first` RP / spectrum rungs are **unaudited**, not a retained ladder;
- `koide_signed...readout` is **audited_failed** on main (any repair is in an open PR).

## The convergence lead (next path)
[...]

PR #2469, opened 2026-06-02
feat(koide): reality favors the signed readout; shared mechanism (not object) with records U=I

jonathonreilly:
## Summary

Fires the convergence probe from PR #2468: does one CPT/reality condition force **both** the records-side `U=I` (→ CAR) and the signed `√m` Koide readout? **Answer: real but as a shared *mechanism*, not a shared *object*** — and it lands a positive result on the readout pin.

## Findings (runner 12/12, non-circular)

- **(A)** `cpt_exact` (D real anti-Hermitian → `H=iD` Hermitian, retained_bounded) → real **signed** spectrum `λ_k = a + 2|b|cos(θ+2πk/3)`.
- **(B) Reality favors the signed side.** The signed readout `√m_k=λ_k` (the operator's *own* spectrum) gives `Q=2/3` **θ-independently** at `r=1/2` (retained value); the singular-value reading needs an **extra modulus** `|H|=√(H²)` *not* in the spectrum (θ-dependent, `Q∈[0.409, 0.667]`).
- **(C) The shared mechanism:** self-adjointness ⇒ real spectrum ⇒ eigen-phase collapses `U(1)→Z₂ sign`. This is what would force `U=I` on the records side too.
- **(D) But different factors:** the retained records-side signed object is the site-qubit `σ_z` sign (`C²`, yt_lsp retained_bounded); the `√m` sign lives on the generation `C³`. Same *type*, different operators/factors → **not one shared object**. The `C²↔C³` bridge is the open generation-ID gate.

## Residuals (named, NOT adopted)
- records-side: "records are of `H=iD`" is an **import** (no ledger row identifies the decoherence pointer with the mass operator).
- readout-side: "feed signed `λ_k` to Q" is an **unaudited internal identification** (`koide_readout_lane_demarcation`) — natural from self-adjointness, *not* a foreign import.
- the `C²↔C³` factor bridge = open generation-ID gate.

Correction surfaced: `koide_signed...readout` is `audited_failed` only for a boundary-wording defect (θ=π/12 window); core mechanism sound, signed VALUE rests on retained `koide_circulant_q_two_thirds_algebraic`.

## Why this matters
Reality is retained ground that **favors the signed readout** (the singular reading needs extra structure) and is the common mechanism behind both pins. The smallest next step that puts the signed value on fully retained ground is auditing `koide_readout_lane_demarcation` (an internal identification, not an import).

## Test plan
- [x] Runner 12/12 PASS (`Q_signed=2/3` θ-indep to 3e-16; `Q_singular` θ-dep; eigen-phase Z₂ collapse)
- [x] All cited tiers verified on `origin/main`
- [x] Non-circular (`Q=2/3` only as check target); no import adopted

### DIFF
--- a/docs/KOIDE_REALITY_FAVORS_SIGNED_READOUT_SHARED_MECHANISM_NARROW_THEOREM_NOTE_2026-06-02.md
+++ b/docs/KOIDE_REALITY_FAVORS_SIGNED_READOUT_SHARED_MECHANISM_NARROW_THEOREM_NOTE_2026-06-02.md
@@ -1,0 +1,137 @@
+# Reality Favors the Signed Readout, and Is a Shared Mechanism (Not a Shared Object) With the Records-Side U = I
+
+**Date:** 2026-06-02
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only; effective status is
+pipeline-derived after audit.
+**Primary runner:** [`scripts/frontier_koide_reality_favors_signed_shared_mechanism.py`](../scripts/frontier_koide_reality_favors_signed_shared_mechanism.py)
+
+## Context
+
+Two open pins of the charged-lepton program were conjectured to converge under
+one principle: the records-side `U = I` (no relative branch phase, needed for
+transfer positivity -> CAR; companion note
+`KOIDE_HERMITIAN_RECORDS_IMPORT_REQUIRED...2026-06-02`) and the value-side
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed narrowed salvage in 3bcb2a1c2. The landed note keeps the reality-favors-signed-readout mechanism and 12/12 runner, but frames the records link as an analogous phase-collapse mechanism rather than a shared object or closure; signed-readout remains source-repaired and re-audit pending.

PR #2475, opened 2026-06-02
science(koide): F1 residual — channel-counting needs a non-tracial (finite-β) modular structure; reduces value scoring to the dynamics

jonathonreilly:
## Summary

Attacks the **final residual** of the tracial-standard-form carrier (PR #2472): is channel-counting (→ `r=1/2`) forced over direction-counting (→ `r=1`, Plancherel)? **Decisive: no, not at the tracial level — and it reduces to the dynamics.**

## The result (runner 8/8, two independent methods)
The carrier state is the group trace `τ`. Its Tomita-Takesaki **modular operator is trivial**: `Δ=1`, modular Hamiltonian `K=−log Δ=0`, modular flow `σ_t=id` (a trace is KMS at `β=0` / infinite temperature — verified via the explicit GNS Tomita `S`-operator *and* the superoperator `Δ(X)=ρXρ⁻¹` with `ρ=I/N`). So the GNS/modular structure of the tracial carrier supplies **no scoring weight**.

- The carrier's *own* distinguished weight (`ρ_τ=I/N`, maximally mixed) is **direction-counting → `r=1` → `Q=1`**.
- **Channel-counting (`r=1/2`) ⟺ a finite-β KMS weight** — explicit `ρ=diag(1/5, 2/5, 2/5)`, modular gap `=ln2 → r=1/2`. The trace (gap 0) cannot supply it.
- Adversarial sweep: **no** non-import principle forces channel-counting (rep-theory → `r=1`; charge-number needs a circular spectrum-`{0,1,1}` Hamiltonian; entropy/records → `r=1`; Kähler prefactor `(1,2)` *is* the channel choice in disguise).

## Why it matters — the unification
F1's final residual **reduces to a non-tracial / finite-β / dynamical structure = the emergent-time dynamics.** This unifies the value lane with the carrier/emergent-time lane: the value scoring isn't a kinematic measure choice, it's a *dynamical* (finite-temperature/modular) input — exactly *why* the records-flow notes had to invoke a dynamics. Relocates, does not close: *which β / which 2-sector coarse-graining*, non-circularly, is the live dynamical object.

## Test plan
- [x] Runner — **SCORECARD PASS=8** (Δ=1 two ways; carrier weight → Q=1; channel ⟺ finite-β; adversarial sweep; falsifiable `r=1/(N−1)`)
- [x] Cache regenerated (sha-matched); tiers verified on `origin/main`
[...]

PR #2481, opened 2026-06-02
science(koide): F1 finite-β weight = the 2-sector partition bit reparametrized (a retained no-go), not a new dynamical freedom

jonathonreilly:
## Summary

Grounds the just-landed carrier-scoring modular note (PR #2475). **The finite-β "temperature" it invokes is not a new continuous degree of freedom — it is the same partition bit, renamed, and that bit is already a `retained_no_go` on main.**

## The exact identity (runner 11/11)
With per-direction weights `(w0,w1,w1)` and the channel-counting balance `w0·a²=w1·b²`: `r = w0/w1 = t` and `t = exp(−β·gap)`, so **`β·gap = −ln t`**. Therefore:
- `β·gap = ln2` (finite-β on 3 directions) ⟺ `t=1/2` ⟺ **β=0 / uniform on the 2 SECTORS** ⟺ `r=1/2`
- `β=0` (tracial) ⟺ `t=1` ⟺ uniform on 3 directions ⟺ `r=1`

So **temperature and 2-sector-vs-3-direction partition are two names for one free bit.** The carrier-scoring residual and the partition residual are *literally the same*, not two reductions.

## The bit is a retained no-go
`r=1/2` (`t=1/2`) and `r=1` (`t=1`) are two rays of the **one-parameter Ad-invariant isotype-weight freedom** that `koide_frobenius_isotype_split_uniqueness` (**`retained_no_go`**) already proves is not forced. The finite-β KMS weight is that retained-open freedom in modular dress.

## All four forcing principles fail to pin it
- **(a) entropy/1-bit `ln2`** — genuinely the 2-channel entropy, maximized at `r=1/2`; *but* the unconstrained Jaynes MaxEnt state is `ρ=I/3 → r=1`. The 1-bit reading *presupposes* the partition.
- **(b) C₃/n_gen=3** — gives `r=1/(N−1)` *given* channel-counting; doesn't select it.
- **(c) self-consistency** — circular (the cyclic-vector split isn't an `H`-eigenspace; `β` absorbs the target).
- **(d) records Lüders flow** — `r=1/2` is an *unstable* separatrix, stable only on the 2-sector functional.

## Why it matters
Closes off the false hope that the dynamics supplies a *new* temperature freedom on top of the partition gate. It must deliver exactly the discrete 2-sector coarse-graining — no more. This strengthens PR #2475 by grounding its modular claim in an existing `retained_no_go`.

## Test plan
- [x] Runner — **SCORECARD PASS=11** (the `β·gap=−ln t` identity; `H`-eigenbasis ≠ cyclic split; MaxEnt `I/3 → r=1`; 2-outcome entropy maxed at `r=1/2`)
- [x] Cache sha-matched; tiers verified on `origin/main` (Frobenius freedom = `retained_no_go`)
- [x] Non-circular (`r=1/2`/`Q=2/3` only as outputs/check targets)

### DIFF
--- a/docs/KOIDE_FINITE_BETA_WEIGHT_IS_THE_PARTITION_BIT_NOTE_2026-06-02.md
+++ b/docs/KOIDE_FINITE_BETA_WEIGHT_IS_THE_PARTITION_BIT_NOTE_2026-06-02.md
@@ -1,0 +1,108 @@
+# Koide F1: the finite-beta weight giving r=1/2 is the 2-sector PARTITION bit reparametrized, not a new dynamical freedom -- and that bit is a retained no-go
+
+**Date:** 2026-06-02
+**Claim type:** bounded_theorem
+**Claim boundary:** structural reduction + an exact reparametrization identity. This note proves
+the surviving channel-vs-direction *scoring* residual of the tracial-standard-form carrier is a
+single free bit, and that the "finite-beta / temperature" form of that residual
+(`KOIDE_CARRIER_SCORING_NEEDS_NONTRIVIAL_MODULAR_NOTE_2026-06-02`) is the *same* bit as the
+2-sector partition choice, by the exact identity `beta*gap = -ln t` with `t = w0/w1 = r`. It does
+**not** derive `r=1/2`, does **not** supply a dynamics, does **not** approve any import, and does
+**not** set an audit verdict. It is a diagnosis that *collapses* three framings of F1's residual
+into one already-open object and grounds that object in a **retained no-go**.
+**Primary runner:** [`scripts/koide_finite_beta_weight_is_the_partition_bit_2026_06_02.py`](../scripts/koide_finite_beta_weight_is_the_partition_bit_2026_06_02.py) (SCORECARD PASS=11).
+
+## What this adds
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as 1882c3754.

Review-loop kept the exact beta*gap=-ln(t) reparametrization and the runner-backed checks, but narrowed the boundary before landing:
- removed bare F1 / over-promoted retained-no-go wording;
- framed the result as one weight-ratio choice, not a second independent temperature freedom;
- made clear this row is not itself already audited or retained and does not derive r=1/2;
- added an explicit N1-N8 no-go discipline gate with a steelman for a future dynamics derivation;
- added markdown-linked load-bearing authorities so the audit dependency graph sees the carrier-scoring row, Frobenius no-go, retained Koide anchors, Luders flow, flow notes, and chirality decoupling.

Checks: py_compile; direct runner SCORECARD PASS=11/11; vocab_lint --fix; cached_runner_output with matching runner hash; full audit pipeline; strict audit_lint; row seeded as bounded_theorem/unaudited with explicit deps; git diff --check; staged no-audit-output/no-local-link/no-banned-status-language guards. Generated audit/publication outputs were restored before commit.

PR #2485, opened 2026-06-02
source_resolved exact-Green pocket kernel reconciliation (Tier 1 runner_artifact_issue; Lemma R inline-proved)

jonathonreilly:
Narrow bounded_theorem companion closing the audited_conditional auditor's named runner_artifact_issue repair on SOURCE_RESOLVED_EXACT_GREEN_POCKET_NOTE (score 8.2). Lemma R proves K_impl = exp(-mu*eps) * K_doc inline (machine-precision verified); Corollary R.1 absorbs the equivalence into a gain rescaling. PASS=11/0. Does NOT modify parent; does NOT claim parent lifts. Drafted via workflow wf_f7eb27a7-3f3.

### DIFF
--- a/docs/SOURCE_RESOLVED_EXACT_GREEN_POCKET_KERNEL_RECONCILIATION_NARROW_NOTE_2026-06-02.md
+++ b/docs/SOURCE_RESOLVED_EXACT_GREEN_POCKET_KERNEL_RECONCILIATION_NARROW_NOTE_2026-06-02.md
@@ -1,0 +1,230 @@
+# Source-Resolved Exact Green Pocket — Kernel Reconciliation Companion
+
+**Date:** 2026-06-02
+**Type:** bounded_theorem (narrow companion; runner_artifact_issue repair)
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as c15a96ee0.\n\nReview-loop kept the source repair and stripped audit-metadata/process wording before landing:\n- proved and verified K_impl(rho) = exp(-mu eps) * K_doc(rho);\n- verified gain_doc = exp(-mu eps) * gain_impl, so the as-documented and as-implemented Green fields agree pointwise after recalibration;\n- reproduced the parent frozen hard-bar observables under the documented convention;\n- kept parent/source/cache/ledger untouched and made no parent-lift claim.\n\nChecks: py_compile; verifier PASS=11 FAIL=0; refreshed runner cache with matching sha; vocab_lint --fix; full audit pipeline; strict audit_lint; row seeded as bounded_theorem/unaudited with parent + minimal source probe deps; git diff --check; staged no-audit-output/no-status-proposal/no-local-link guards. Generated audit/publication outputs restored before commit.

PR #2487, opened 2026-06-02
gauge_vacuum plaquette retained class-sampling scalar-support reconciliation (Tier 1 runner_artifact_issue)

jonathonreilly:
Narrow bounded_theorem companion closing the auditor's named runner_artifact_issue repair on GAUGE_VACUUM_PLAQUETTE_RETAINED_CLASS_SAMPLING_INVERSION_NOTE (score 6.9). Companion Lemma proves three reformulations (R1, R2, R3) are mathematically equivalent on retained restricted surface (N=4 witness sector). Runner PASS=14/0. Does NOT modify parent. Drafted via workflow wf_f7eb27a7-3f3.


jonathonreilly:
Reviewed under review-loop. Closing without landing; head branch retained because no source content landed.

This PR targets an older `runner_artifact_issue` posture for `gauge_vacuum_plaquette_retained_class_sampling_inversion_note_2026-04-17`, but current `origin/main` no longer has that parent row in the conditional state described by the PR; the live ledger currently has it as unaudited. That makes the companion's stated repair target stale.

The proposed note also carries generated/cache output and status-heavy language around old retained/conditional surfaces. Review-loop cannot land that as source, and there is no clean smaller source repair here: the runner mostly reconciles a stale phrase check against an old parent-runner support certificate rather than fixing a current source theorem or live pipeline defect.

Salvage result: no source-only content landed. The branch is left intact as the handle for the unlanded artifact-reconciliation work.

PR #2503, opened 2026-06-03
observable-principle: the det FORM-selection premise (M) is a co-admission, not derived

JONbridgerapps:
## What this is

The **integrity check** (task A of the P1 follow-up) on the premise of the composition-axis FORM theorem. It finds the FORM stage is **not free**: its premise is itself admitted, so the scalar-readout chain rests on **two** independent atoms, not one. Math independently re-verified before opening.

## The finding — `form_premise_is_co_admission`

The FORM theorem (`...DET_UNIQUE_MULTIPLICATIVE_CHARACTER_FORM_SELECTION...2026-05-28`) selects `det` over `tr` from the **product-character** requirement `χ(A·S)=χ(A)·χ(S)` (M) via GL(n) abelianization — a theorem **given (M)**. This note asks whether (M) is derived or admitted:

- (M) is **motivated** by the Berezin source-insertion structure (`D+J = D(1+D⁻¹J)` is exact; `Z=det(D+J)` factorizes over independent patches) but **not entailed** by it.
- Both motivations are properties of the operator **weight Z**, true regardless of the readout `χ`. The bridge "operator composition is multiplicative ⟹ the *readout* must be a product character" is a **non-sequitur**, witnessed by **`tr`**: a continuous readout that respects both motivations (`tr(A⊕S)=tr A+tr S`) yet violates (M) (`tr(A·S)≠tr A·tr S`).

So **(M) is a co-admission**, logically independent of P1: (M) lives on the product (`·`) axis, P1 on the direct-sum (`⊕`) axis, and `tr` separates them.

## Consequence

The chain folded under the single Tier-A label **P1** (`scalar additivity ⇒ W=log|det(D+J)|`) hides **two** admitted steps: the **FORM** admission (M) and the **EXPONENT** admission P1. This **confirms and locates** the FORM note's own honest §7 residual (with proof), and **refines** the exponent-fixing note (#2456): its "FORM is a theorem (not admitted)" holds only as "theorem *given* the admitted premise (M)."

**No new Tier-A item is banked** (count unchanged) — what changes is the documented internal anatomy of P1.

## Discipline

Independently re-verified the non-sequitur (`(Fac)∧(Mul) ⇏ (M)`, witnessed by `tr`). `claim_type: no_go`; pins, does not retire. Companion runner **18 PASS / 0 FAIL** (exact sympy). No audit-lane data, no new axiom, no new vocabulary; literature as comparator only. Status authority: independent audit lane only.


jonathonreilly:
Landed source-only salvage in 98822db8d: narrowed to the scoped no-go that source/product factorization facts do not derive the product-character readout law chi(AS)=chi(A)chi(S). The landed version does not register a new admission or Tier-A item, strips status/provenance surfaces, keeps the determinant form-selection theorem conditional on the product-character premise, and leaves the verdict to the independent audit process.

PR #2506, opened 2026-06-03
science(koide): the C²↔C³ bridge is SELF-IMPOSED — the two terminal Z₂ bits are independent (disjoint commuting factors)

jonathonreilly:
## Summary

Attacks the **necessity** assumption behind "the carrier reduces to one bridge import." **Result: the two terminal Z₂ bits are *independent*, and the `C²↔C³` bridge is *self-imposed* — it over-couples two separable residuals.**

## The factorization
The carrier factorizes as **(VALUE bit on generation C³) ⊗ (CARRIER bit on site C²)** over **disjoint commuting tensor factors**: `[H⊗I₂, I⊗σ_i/2]=0` makes the site-indexed `D` **spin-blind on C²**. No single operator is forced to be both bits. Each closes by its **own factor-local argument**:
- **VALUE (C³):** the `Γ_χ`-anticommuting chiral mass — an `O_h`/Frobenius question (z3 no-go is C₃-scoped, off-generation route open).
- **CARRIER (C²):** cross-site graded/CAR statistics over the native hard-core boson — the `statistics_agnostic` `retained_no_go`.

Neither residual mentions the other factor.

## The welding links are posited, not proven
- **(a) records-Z₂ = sign(β):** appears in *exactly one place* on main (`carrier_locus`), explicitly labeled **"open, not a theorem."** Link (a) *is* the bridge.
- **(b) sign(β) = Hodge/value bit:** rides an unaudited note **past what `cpt_exact` carries** (retained_bounded, only C1+C2 — no `β`, no Pfaffian, no Hodge).

Every capstone note asserting the unification is **unaudited**; the P1 note states the *opposite* on the retained tier ("two posits remain").

## Net
Closing the bridge is **sufficient but not necessary**. The carrier closes when both factor-local residuals close — **two parallel lanes, no bridge needed.** This corrects the "single terminal import" framing.

## Test plan
- [x] Runner — **15/15 PASS** (disjoint factors; value-on-C³; carrier-on-C²; welding-links-posited audit; no forced coupling)
- [x] Cache via `precompute_audit_runners.py --push-mode=none`, sha-matched
- [x] Non-circular (`Q=2/3` never enters); tiers verified on `origin/main`

### DIFF
--- a/docs/KOIDE_CARRIER_BRIDGE_SELF_IMPOSED_TWO_INDEPENDENT_RESIDUALS_NOTE_2026-06-02.md
+++ b/docs/KOIDE_CARRIER_BRIDGE_SELF_IMPOSED_TWO_INDEPENDENT_RESIDUALS_NOTE_2026-06-02.md
@@ -179,32 +179,155 @@ corrects) remains plain text:
 and
 `koide_berry_monopole_bridge_reduction_note_2026-05-31`.
 
-## No-Go Discipline Gate
+## No-go discipline gate (N1-N8)
 
-- **N1 - Alternative routes.** Three weldings were tested: (a) `records = sign(beta)`
-  (open bridge), (b) `sign(beta) = Hodge bit` (asserted past CPT's C1/C2), and a
-  forced shared operator on the joint space (commutes / independent). All fail to
-  force one object.
-- **N2 - Wall independence.** The VALUE wall (`koide_z3_equivariant_anticommuting_no_go`)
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed source-only salvage in 64946be9d: narrowed the branch to the scoped no-go that product-factor algebra alone does not force a C2-to-C3 carrier-value bridge. The landed note keeps value and carrier derivations open, leaves future welding theorems/admissions open, strips status/prose assertions and cache output, and leaves the verdict to the independent audit process.

PR #2507, opened 2026-06-03
science(koide): carrier import is TWO bits on retained (STAT+FAITH), one only modulo auditing 3 spin-statistics rows

jonathonreilly:
Resolves the COUNT assumption behind 'the carrier reduces to one Z₂ bit.' Carrier import = **STAT** (CAR vs hard-core boson) + **FAITH** (faithful Weyl vs scalar). **On strict-retained the count is TWO** (the retained statistics authorities are blind to the hard-core boson; FAITH never collapses into STAT). 'One bit' = FAITH only *modulo* auditing 3 unaudited rows (axiom_first_spin_statistics, free_field_os_wightman, free_sector_spin_statistics_level1). Residual either way = FAITH. Runner 16/16; non-circular; tiers verified on origin/main. Source-only.

### DIFF
--- a/docs/KOIDE_CARRIER_BIT_COUNT_IS_TWO_ON_RETAINED_ONE_MODULO_AUDIT_NOTE_2026-06-02.md
+++ b/docs/KOIDE_CARRIER_BIT_COUNT_IS_TWO_ON_RETAINED_ONE_MODULO_AUDIT_NOTE_2026-06-02.md
@@ -1,0 +1,153 @@
+# The charged-lepton carrier import is TWO bits on the retained tier, ONE only modulo auditing named spin-statistics rows
+
+**Date:** 2026-06-02
+**Claim type:** bounded_theorem
+**Claim boundary:** a bit-count audit of the charged-lepton **carrier** import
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop result: not landed. The durable algebraic content is already represented on current main by KOIDE_CARRIER_LOCUS_DECOMPOSITION_NOTE_2026-06-01.md, KOIDE_P1_COLLAPSES_FRAME_RESIDUALS_NOTE_2026-06-01.md, and QUANTUM_LOCAL_ALGEBRA_DOES_NOT_FORCE_BOOST_ACTION_FAITH_NO_GO_NOTE_2026-06-02.md: hard-core vs CAR is not decided by the single-site/cardinality surface, the conditional FAITH -> STAT collapse is kept conditional, and Quantum-alone does not force boost-action faithfulness. This PR's added note/runner instead counts exact current ledger tiers and names unaudited rows as the count boundary, which is not a durable source theorem surface for review-loop to land, and the cache log is excluded by the source-only rule. No additional salvage is needed on top of the current-main surfaces. Branch left undeleted because this PR content was not landed.

PR #2509, opened 2026-06-03
science(koide): octahedral site-symmetry O_h does NOT supply the value bit — it over-constrains the triplet to degeneracy

jonathonreilly:
Attacks the substrate assumption that the value bit (r=1/2) could be fixed by the lattice site-symmetry **O_h** instead of staying a C₃-internal question. **Result: O_h over-constrains.**

The hw=1 generation triplet carries the 3-dim irrep T₁ of O_h. An O_h-equivariant mass is by **Schur a scalar** `μ·I₃` → fully **degenerate** spectrum (m_e=m_μ=m_τ), no splitting at all. O_h equivariance is strictly *stronger* than C₃ — it forbids the very circulant non-scalar part `{a on diagonal, b off}` that the `r=|b|²/a²` family lives in.

Breaking O_h **down to C₃** is exactly what reintroduces the one-parameter Frobenius weight freedom; **r=1/2 vs r=1 stays the same internal C₃ isotype-weight question** (retained Frobenius no-go). The lattice's own symmetry over-determines, rather than fixes, the flavor splitting.

## Test plan
- [x] Runner — **28/28 PASS** (|O_h|=48; T₁ irrep; Schur scalar; degeneracy; C₃-breaking restores weight freedom)
- [x] Cache via `--push-mode=none`, sha-matched
- [x] Non-circular (Q=2/3 never enters); tiers verified on `origin/main`

### DIFF
--- a/docs/KOIDE_OCTAHEDRAL_OVERCONSTRAINS_VALUE_BIT_NARROW_NOTE_2026-06-02.md
+++ b/docs/KOIDE_OCTAHEDRAL_OVERCONSTRAINS_VALUE_BIT_NARROW_NOTE_2026-06-02.md
@@ -1,0 +1,178 @@
+# Koide Value Bit: Octahedral O_h Equivariance Over-Constrains, It Does Not Derive r = 1/2
+
+**Date:** 2026-06-02
+**Claim type:** no_go (bounded over-constraint)
+**Status authority:** independent audit lane only; effective status is
+pipeline-derived after audit. This note approves no new import, sets no audit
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed source-only salvage in a68c78be1: narrowed to the scoped O_h overconstraint no-go on the R^3 generation-axis carrier, kept C3-internal/split-preserving selection routes open, stripped cache/generated/status surfaces, and left the verdict to the independent audit process.

PR #2512, opened 2026-06-03
science(koide): carrier import is NOT foreign — it relocates to matter-attachment, where SO(2) framing-writhe and the SU(2) spinor Z₂ provably decouple

jonathonreilly:
Attacks the **foreignness** assumption behind "the carrier needs one external import." **Result: the import is native** — it relocates onto the matter-attachment (KS) step, and there the only available geometric Z₂ (framing writhe of the embedded worldline in Z³) is the **wrong** Z₂.

An embedded fermion worldline carries a **self-linking / framing writhe** valued in the SO(2) normal-bundle framing — an **integer** (`π₁(SO(2))=Z`). The CAR/spinor sign the carrier needs is the **SU(2)=Spin(3) center Z₂**. The obstruction is explicit: the exchange/framing class has base-edge projection `P(t)=0` — pure **2-torsion**, orthogonal to every continuous SO(2) writhe and every link sign. `Z` (infinite order) and `Z₂` (order 2) are different groups; the embedding forces no `Z→Z₂`, and the natural mod-2 writhe is **not** the canonical CAR sign — they **decouple**.

So (i) the carrier import is **not** parachuted in — it *is* the matter-attachment posit, internal to A1+A2's own KS realization; (ii) the lattice does offer a Z₂ (mod-2 writhe) but it provably fails to **be** the statistics bit. Relocation **sharpens** the residual: pins it to the attachment map, certifies the nearest geometric candidate is the wrong class.

## Test plan
- [x] Runner — **24/24 PASS** (`P(t)=0`; SO(2) writhe ∈ Z; spinor Z₂; no forced `Z→Z₂`; decoupling)
- [x] Cache via `--push-mode=none`, sha-matched
- [x] Non-circular; tiers verified on `origin/main`; source-only

### DIFF
--- a/docs/KOIDE_EMBEDDING_FRAMING_WRITHE_SO2_VS_SPIN_Z2_DECOUPLING_NARROW_NO_GO_NOTE_2026-06-02.md
+++ b/docs/KOIDE_EMBEDDING_FRAMING_WRITHE_SO2_VS_SPIN_Z2_DECOUPLING_NARROW_NO_GO_NOTE_2026-06-02.md
@@ -248,74 +248,176 @@ second-quantized graded-locality route remain open; this note localizes the
 residual to the matter-attachment pin and identifies the precise group-theoretic
 reason the abelian-framing family cannot reach it.
 
-## 6. No-Go Discipline Gate
+## 6. No-Go Discipline Gate (N1-N8)
 
-**N1 -- Alternative route enumeration.**
+**Status:** PASS for the narrow scoped no-go only. The claim being closed is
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed source-only salvage in b5296cece: narrowed the branch to the SO(2)/U(1) embedding-framing no-go for the tested UD_2 witnesses, kept spinor-state and second-quantized graded-locality routes open, removed native/import-relocation framing, removed audit-status/YAML/cache surfaces, and left generated outputs to the independent audit process.

PR #2504, opened 2026-06-03
observable-principle P1: exponent-selector dichotomy (closes path-(a) for a precise class)

JONbridgerapps:
## What this is

Task B of the P1 follow-up: it **closes #2456's open forward-path (a)** — "could a genuinely new selector *outside* the (Add)/(Loc)/(Pot) class fix the exponent?" — as a **theorem**, for a precisely-defined class 𝒞 strictly larger than #2456's four-selector enumeration. Math independently re-verified before opening.

**Verdict: `irreducible_against_precise_class`** (deliberately *not* "all selectors" — that's ill-posed; an over-claimed "ALL" would be worse than a precise strengthening).

## The dichotomy (theorem)

The scalar generator is `W = Φ(r)`, `r = |det(D+J)| > 0`; independent direct-sum sectors compose by the **product** on `r` (value group `(ℝ₊,×)`). The candidate family `{F_p = |det|^p}` is a **single orbit** of the reparametrization group `{r ↦ r^p}` (`F_{p'} = (F_p)^{p'/p}`), and the composite law it induces carries **no `p`**. So every **sector-composition selector** (the class 𝒞 — cross-block conditions statable without pre-encoding a privileged exponent) is exactly one of two exhaustive, mutually-exclusive faces:

- **(BLIND)** orbit-invariant ⇒ same object for all `p` ⇒ fixes nothing (the Born/normalized gradient is the instance);
- **(ADD)** orbit-breaking ⇒ references the bare value; the unique such law is bare additivity `Φ(r_A r_B)=Φ(r_A)+Φ(r_B)` = **P1**, whose unique measurable/monotone solution on `(ℝ₊,×)` is `c·log r` (Cauchy/Shannon–Khinchin, reproven).

**No third option pins a finite nonzero `p` via a non-additive principle.** Both steelman escapes #2456 flagged are closed: (i) extensivity at a nonlinear gauge `h(W)` reduces to bare additivity for `G=h∘Φ` ⇒ Cauchy ⇒ face ADD; (ii) prescribing a nonzero cross-block 2nd-derivative `K≠0` *pre-encodes* `p` (`∂²(Z^p)/∂j_a∂j_b = 4 j_a j_b p² Z^p/(Z_A Z_B)`), so it's out of class — only `K=0` (=Loc=Add) is in-class. Single-sector conditions (positivity/monotonicity/convexity/analyticity) are gauge-stable ⇒ orbit-invariant ⇒ BLIND ⇒ fix nothing.

[...]

PR #2517, opened 2026-06-03
observable-principle P1: symmetry-type/energy readout reduces to P1 (energy path-a' instance closed)

JONbridgerapps:
## What this is

Tests the deep reframe: *additive and multiplicative readouts aren't exclusive — map each observable to its symmetry type* (additive = continuous-symmetry generators / Lie algebra; multiplicative = discrete-symmetry characters / Lie group; bridged by the same `exp`/`log` as `Z=exp(log Z)`). The framework genuinely uses **both** ledgers — `log|det|` (additive action/free-energy) and `Z₃` characters (multiplicative Koide/color/generation). The question: does this instantiate #2504's open **path-(a′)** and *retain* P1 for the v-chain? Math independently re-verified.

## Verdict: `circular_reduces_to_p1`

The symmetry-type **law is a theorem** (reproven: tensor generators add; characters multiply; `exp`/`log` is the algebra↔group bridge). But the *energy route to retaining P1* collapses into #2504's `{ADD=P1, BLIND=nothing}` dichotomy — it is **not** an out-of-class path-(a′) input.

**Decisive, framework-specific reason:** A1–A2 commit **no dynamics**; the realized free sector *builds* the Hamiltonian as `Ĥ = −log(T̂²)/(2a_τ)` with `T̂²` multiplicative. So "energies are additive" ⟺ "`T̂²` multiplicative" is the **same `exp`/`log` move one level up** (transfer matrix instead of partition function). `{(T̂²)^s}` is a single `{r↦r^s}` orbit:
- intensive (per-site/Born) energy `−log((T̂²)^s)/(2s·a_τ) = E` for **all** `s` → face **BLIND** (fixes nothing);
- the bare additive energy references the bare value → face **ADD = P1**.

Pivot 2 ("the v-chain scalar *is* the free energy") is content (II.b) of the structural-reframing no-go, already `(II.b) ⇔ P1`, and the framework's free energy is *defined* by the log-det convention — so it presupposes `log|det|`.

## Honest boundary

Closes the **energy/free-energy instance** of path-(a′) and explains *why* (the dynamics' additive generator is the same `exp`/`log` object as the additive partition-function readout — circularity flagged). It does **not** foreclose all of path-(a′): a primitive introducing a privileged scale by a mechanism that is *neither* the energy ledger *nor* a sector-composition selector remains open. **P1 stays admitted.**

So: your physics intuition is *confirmed* (both ledgers are real and the framework uses both — the symmetry-type law is a theorem), but it doesn't retain P1, because in a Euclidean lattice framework the energy is *itself* a log of the multiplicative transfer matrix.

## Discipline

Independently re-verified (tensor-additive eigenvalues; multiplicative characters; `Ĥ=−log(T̂²)`; orbit-invariance of the intensive energy; the two-level `exp`/`log` identity). Runner **PASS=27 FAIL=0** (exact sympy). `claim_type: no_go`. No audit-lane data, no new vocabulary; literature as comparator only. Builds on #2456 (on main); references #2504/#2503 textually until they land.

### DIFF
--- a/docs/OBSERVABLE_PRINCIPLE_P1_SYMMETRY_TYPE_ENERGY_READOUT_NARROW_NOTE_2026-06-02.md
+++ b/docs/OBSERVABLE_PRINCIPLE_P1_SYMMETRY_TYPE_ENERGY_READOUT_NARROW_NOTE_2026-06-02.md
@@ -1,0 +1,776 @@
+# Observable-Principle P1 Bridge — Symmetry-Type / Energy-Readout Narrow Note
+
+**Date:** 2026-06-02
+**Claim type:** no_go
+**Type:** no_go
+**Claim scope:** narrow no_go that tests the **symmetry-type readout reframe**
+as a candidate instance of the open forward path **(a′)** recorded in
+`OBSERVABLE_PRINCIPLE_P1_EXPONENT_SELECTOR_DICHOTOMY_NARROW_NOTE_2026-06-02.md`
+(#2504): a "genuinely new retained primitive introducing a separate physical
+**structure**, **outside** the sector-composition selector class `𝒞`." The
+reframe proposes that the v-chain's **additive** readout `W = log|det(D+J)|` is
+**forced** (a representation-theory theorem, not a selection) by the fact that
+the VEV `v` is an **energy** — i.e. an eigenvalue of a generator of a
+**continuous** (time-translation) symmetry, which is additive over a tensor
+product `H = H_A ⊗ I + I ⊗ H_B` — while **multiplicative** readouts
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed source-only salvage as 95ad1b613. Kept the narrow no-go only: the symmetry-type law is a theorem, but the energy/free-energy route reduces to the existing P1 additivity premise and does not close P1 or promote any row. Updated stale A1/A2 and retained-status wording to the Lattice/Quantum/Record framework. Checks: runner PASS=27 FAIL=0, py_compile, independent algebra spot-check, no-go N1-N8 visible, audit pipeline, strict lint, staged guards.

PR #2531, opened 2026-06-03
feat(flavor): Door A findings — native hw=1 projection gives democratic r=0; every action condition reaching r=1/2 unifies to equal-block HS norm

jonathonreilly:
## Summary
Two derived findings from the Door A investigation (does any framework-internal lattice action geometrically fix the charged-lepton ratio `r=|b|²/a²` at `r=1/2`?). Both packaged source-only: 1 note + 1 runner + 1 cache each. **Neither adds an axiom or import; neither sets or changes any audit status** (independent audit lane owns all grading).

## Findings
1. **`FLAVOR_HW1_STAGGERED_PROJECTION_DEMOCRATIC_R0`** (runner 5/5) — With one M₂(ℂ) qubit per corner and the native staggered single-bit-flip generators on 3 corner qubits (ℂ⁸), the Schur-complement onto the hw=1 generation sector induces a diagonal but **exactly-zero** generation hopping (max induced |b| ≈ 1e-14 over a 401-point reference-energy sweep) by staggered-sign path cancellation. So projection geometry alone forces the **democratic endpoint r=0 (Q=1/3)**, not the charged-lepton r=1/2. A clean derived geometric fact + a value-axis negative.

2. **`FLAVOR_DOOR_A_UNIFIES_TO_EQUAL_BLOCK_HS`** (runner 4/4) — No framework-internal action geometrically fixes `a:|b|`; and **every** route that *reaches* r=1/2 (Wilson, hw=1 projection, Cl(3)/heat-kernel, self-dual) collapses algebraically to the single native condition **3a²=6|b²** = equal total Hilbert–Schmidt norm of the mass block vs the hopping block = `AC_φλ`. The 1:2 multiplicities are forced geometric trace-counts; the dimension/Plancherel reading gives r=1 (Q=1). **Door A = the measure axis**, same wall, new direction. Two strengthenings: the gap is **native** (not Wilson-import-dependent), and it is one precise coordinate-free object.

## Relationship to existing science
Both **confirm and sharpen** the standing chain-of-custody note already on `main` (structure derived; value modulo the single Tier-A admitted input `AC_φλ`). They do not overclaim closure: the value r=1/2 remains unforced; these notes localize exactly *where* and record the action-axis = measure-axis coincidence.

## Test Plan
- [x] `python3 scripts/flavor_hw1_staggered_projection_democratic_r0_2026_06_02.py` → SCORECARD 5/5
- [x] `python3 scripts/flavor_door_a_unifies_to_equal_block_hs_2026_06_02.py` → SCORECARD 4/4
- [x] Source-only (notes + runners + caches); no audit-language; status authority = independent audit lane

### DIFF
--- a/docs/FLAVOR_DOOR_A_UNIFIES_TO_EQUAL_BLOCK_HS_2026-06-02.md
+++ b/docs/FLAVOR_DOOR_A_UNIFIES_TO_EQUAL_BLOCK_HS_2026-06-02.md
@@ -1,0 +1,68 @@
+# Flavor — the "geometry-fixed action" axis (Door A) is the same wall as the measure axis: every action-level condition reaching r=1/2 reduces to the native equal-block Hilbert-Schmidt weight 3a²=6|b|²
+
+**Date:** 2026-06-02
+**Claim type:** a unification / localization result (no action fixes the value; the residual is one named native object). Not closure, not an import adoption.
+**Status authority:** independent audit lane only. This note sets no audit status and assigns no grade.
+**Runner:** `scripts/flavor_door_a_unifies_to_equal_block_hs_2026_06_02.py` (SCORECARD 4/4).
+
+## Question
+Does any framework-internal lattice **action** geometrically fix the charged-lepton ratio
+`r = |b|²/a²` at `r=1/2` (`|b|/a = 1/√2`), rather than leaving the on-site mass a free Yukawa? Four
+action-level routes were examined: a Wilson term, the hw=1 induced diagonal, the Cl(3) grade /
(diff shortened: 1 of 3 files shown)


jonathonreilly:
review-loop landed a narrowed source-only salvage as f3513cc33 on main. I landed only the specified hw=1 staggered-projection no-go note and runner; I did not land the runner caches or the broader Door-A/equal-block packet, because current main already carries the action-axis/equal-block result and the PR wording overreached beyond its runner. Validation: runner PASS=5/0, py_compile, independent Schur-induced cancellation check, audit pipeline, strict audit lint; generated audit/status outputs were restored before commit.

PR #2534, opened 2026-06-03
science(koide): the canonical forced native metric (HS on the C₃-commutant) gives r=1, not r=1/2 — no native variational principle forces r=1/2

jonathonreilly:
> ### ⚠️ Framing correction (2026-06-02)
> `Q=1` and `Q=2/3` **coexist** as different operator classes — non-chiral (`r=1`) vs chiral (`r=1/2`) — per retained `koide_anticommuting_operator_derivation_theorem`. **`r=1/2` is a chirality *sector label*, not a variational outcome.** So "no native principle forces `r=1/2`" is the *expected* result, **not a wall and not evidence against `Q=2/3`**: the canonical metric correctly sits at the non-chiral `Q=1` endpoint — the **complement** of the charged-lepton sector. This PR *characterizes that complement*; `Q=2/3` is fixed by the chirality source, not by metric optimization.

---

## Summary

**VALUE LANE attack — does any A1+A2-native variational principle force `r=1/2`?** **Verdict: NO (high confidence)** — and the sharpening is the payload: the **canonical forced native metric gives `r=1`**, not `r=1/2`.

## Argument
On C³, `H = aI + b(J−I)` gives `Q = Tr(H²)/(Tr H)² = 1/3 + (2/3)r` exactly, `r=1/2 ⟺ Q=2/3`. Every genuinely native operator-spectral functional — spectral/von-Neumann entropy, relative entropy, purity, the Born/tracial state, and **decisively the canonical HS metric on the C₃-commutant `span{I,C,C²}`** (forced, isotropic) — lands at `r=0` or `r=1` (`Q=1/3` or `Q=1`), **never `r=1/2`**. The functionals that *do* land at `r=1/2` first **fold** the two complex modes `{C,C²}` into one real doublet — precisely the block-count / `det_C` / Frobenius-`β=0` choice. RP/T-positivity is an inequality (`e^{−tH}` PD on the whole `r`-line) and pins nothing.

Locates + re-derives the prior `retained_bounded` result `flavor_r_half_is_a_stationary_point_not_forced` (2-sector power entropy: stationary at `r=1/2`, but per-DOF entropy peaks at `r=1`) from the **energy side**, and makes it explicit against `koide_frobenius_isotype_split_uniqueness` (`retained_no_go`): the Ad-invariant PD metric family `αTr(AB) + β tr(A)tr(B)` has a **free `β`**, and `r=1/2 ⟺ β=0` is one unforced point of the PD cone. **The variational route re-derives the gate, it doesn't evade it.** Missing ingredient: a native principle selecting the 2-real-block (`β=0`) grouping.

> **Import flags** (in the note): positing a flow (thermalizing vs sharpening) to make `r=1/2` an attractor; declaring block-count to be *the* mass measure.

## Test plan
- [x] Runner — **17/17 PASS**; cache exit 0; non-circular; tiers verified on `origin/main`; source-only

### DIFF
--- a/docs/FLAVOR_R_HALF_VARIATIONAL_FORCING_NOTE_2026-06-02.md
+++ b/docs/FLAVOR_R_HALF_VARIATIONAL_FORCING_NOTE_2026-06-02.md
@@ -176,6 +176,172 @@ energy-functional side. Candidate next paths (NOT closures, and the two that are
 
 ---
 
+## No-go discipline gate (N1–N8)
+
+**Status:** PASS for the narrow scoped claim only. The claim under discipline is
+**not** "Q=2/3 is excluded" and **not** "the charged-lepton Koide relation fails."
+It is the single negative metric statement: *no native operator-spectral variational /
+measure functional on the C₃-circulant weight family makes `r=|b|²/a²=1/2` a forced
+extremum; every such functional lands at `r=0` or `r=1`, and the only functionals that
+reach `r=1/2` are built on the block-fold / Frobenius β=0 / det_C 2-sector grouping.*
+Per the **Framing correction** (and retained `koide_anticommuting_operator_derivation_theorem`),
+`r=1` (Q=1) and `r=1/2` (Q=2/3) **coexist** as distinct operator classes on
+`Q = 1/3 + (2/3)r`; the canonical metric correctly characterizes the **non-chiral Q=1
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the durable narrowed source result on main as 68543e1a8. The landed version closes only the choice-free operator-spectral/HS variational route: those functionals do not force r=1/2; the functionals that reach r=1/2 first insert the two-sector block fold.

PR #2535, opened 2026-06-03
science(koide): the framework's DERIVED tracial reference state induces the dimension measure (1,2) → Q=1; (1,1)→Q=2/3 is admissible but non-tracial

jonathonreilly:
> ### ⚠️ Framing correction (2026-06-02)
> `Q=1` and `Q=2/3` **coexist** (non-chiral vs chiral sectors), per retained `koide_anticommuting_operator_derivation_theorem`. The derived tracial reference landing on `(1,2)→Q=1` **characterizes the non-chiral complement sector** — this is a *sector-localization*, **NOT evidence against `Q=2/3`**. The charged leptons are the chiral, **non-tracial** sector; their selector is **chirality**, not a measure/positivity extremum. The (correct) computations below show only which sector the *tracial reference* is.

---

## Summary

**VALUE LANE attack — does any native principle break the `(1,2)→Q=1` vs `(1,1)→Q=2/3` measure tie?** **Verdict: NO — and the framework's own derived reference lands on `Q=1`.**

## Argument
Two canonical measures: (i) dimension/trace (Plancherel) → weights `(1,2)` → `Q=1`; (ii) block/idempotent-count on the 2 central idempotents of `ℝ[Z₃]=ℝ⊕ℂ` → `(1,1)` → `Q=2/3`. Testing every mandated native principle, *without privileging either direction*:

| principle | outcome |
|---|---|
| Reflection positivity (OS) | **agnostic** (PSD at both points; selects readout class only) |
| Emergent-time T-positivity | **agnostic** |
| On-site qubit unit-trace `Tr=1` | **forces (1,2)→Q=1** |
| KMS / Tomita–Takesaki modular | **forces (1,2)→Q=1** (trace is `β=0`, `Δ=1`) |
| Locality / cluster | **forces (1,2)→Q=1** (all region sizes) |
| `(1,1)→Q=2/3` | **posit** — needs an imported non-tracial finite-`β` weight |

The framework's **derived** reference state is the unique trace `ρ=⊗I/2` (retained: `pre_record_reference_state_tracial_derivation`, `powers_uhf_tracial_uniqueness`, `tomita_tensor_trace`), whose Born weight on the two idempotents is `Tr(ρe₀):Tr(ρe₁) = 1/3:2/3 = (1,2)`. The observed `(1,1)→r=1/2` is admissible but **non-tracial**, needing `β·gap = ln 2` that no native principle supplies — mild evidence *against* `Q=2/3` being native-forced.

This does **not** rank the abstract measures (the `retained_no_go`s still decline to); it shows which measure the **derived reference state** induces. Anti-overreach: I do not call either measure "the consistent one."

> **Import flag** (in the note): a finite-`β` (non-tracial) reference state — the single ingredient `(1,1)→r=1/2` needs.

## Test plan
- [x] Runner — **38/38 PASS** (explicit unit-trace Born weight `1/3:2/3`; `Δ=1` Tomita; RP PSD at both points); cache exit 0; tiers verified on `origin/main` (RP/OS rows unaudited, **not** load-bearing); source-only

### DIFF
--- a/docs/FLAVOR_KOIDE_MEASURE_TIEBREAK_NATIVE_PRINCIPLES_NOTE_2026-06-02.md
+++ b/docs/FLAVOR_KOIDE_MEASURE_TIEBREAK_NATIVE_PRINCIPLES_NOTE_2026-06-02.md
@@ -191,38 +191,190 @@ derivation.
 
 ## No-Go Discipline Gate (N1–N8)
 
-- **N1 — Alternative route enumeration.** Five native principles checked:
-  (1) RP — agnostic (PSD for both counts); (2) emergent-time T-positivity —
-  same as RP; (3) unit-trace normalization — forces `(1,2)`/Q=1 (derived
-  reference Born weight); (4) KMS/Tomita–Takesaki modular — trace `Δ=1`, no
-  reweighting, `(1,2)`/Q=1; (5) locality/cluster — product/trace structure
-  `(1,2)` for all `|Λ|`. None forces `(1,1)`/Q=2/3.
-- **N2 — Wall-independence audit.** The principles collapse to two independent
-  facts: RP/T-positivity blindness, and the derived-reference-is-the-trace fact.
-  Closing one does not close the other; neither yields `(1,1)`.
-- **N3 — Hidden-wall scan.** The tracial reference is the retained
-  `pre_record_reference_state_tracial_derivation` chain; the finite-β witness is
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the durable narrowed source result on main as 478bc7289. The landed version does not claim Record/native principles derive a physical flavor reference; it only closes the tracial/product/modular route as a selector for Q=2/3, leaving the non-tracial/chiral-sector selector open.

PR #2536, opened 2026-06-03
science(koide): A1 forces STATE-SPACE faithfulness (qubit=spinor, scalar excluded by dimension) but boost-action FAITH reduces to the matter-attachment pin

jonathonreilly:
## Summary

**CARRIER LANE attack — does A1 force FAITH (faithful Weyl boost over the scalar)?** **Verdict: split** — A1 forces **state-space** faithfulness but **not** boost-action faithfulness; the latter reduces to the matter-attachment pin.

## Argument
Two independent things live on the site C²: the **state space** and the **dynamical (boost/mass) action**. A1 (qubit = C² = Cl(3,0) spinor) forces the **state space** to be the faithful `j=½` spinor — the scalar `j=0` is 1-dimensional and so *isn't a qubit*, **excluded by dimension**. But FAITH is about how the reconstructed **boost/mass operator** acts on that fixed C² — faithful `K=iσ/2`, or a scalar `K=c·I` on the *same* C². The qubit's 2-dimensionality constrains the state space, **not** the boost generator; a scalar boost `S(η)=e^{ηc}I₂` is a legitimate non-faithful rep that A1 alone does not exclude.

What excludes it is an **extra ingredient** — the relativistic spin-½ kinetic kernel (`m·I − iσ·p` is traceless-nonzero, so covariance requires the faithful intertwiner; the scalar fails covariance). That selector is **matter-attachment content, not** a consequence of "C² = spinor."

**So the carrier does *not* reduce to STAT-only at the A1 level.** The available reduction: **state-space FAITH is an A1-consequence**, and **boost-action FAITH reduces to the single matter-attachment pin** ("matter index = C² spinor under the reconstructed Lorentz action").

Reconciles the live-ledger cluster: `koide_onsite_boost_reconstruction` (retained_bounded — forced via the propagator numerator, massless-chiral, massive doubling a residual); `koide_faithfulness_rotation_scalar_excluded` (retained_bounded — dominant open piece = the matter-attachment pin); `koide_onsite_weyl_boost_from_bivectors` (audited_conditional — dynamics `H=iD` is **spin-blind**, so the dynamical lever can't force it).

> **Import flag** (latent, in the note): lifting `per_site_su2_spin_half`'s disclaimed operator-frame action to the physical matter rep — no A1+A2+retained derivation.

## Test plan
- [x] Runner — **18/18 PASS** (state-space faithful by dimension; scalar boost conditional on operator-frame triple; dynamics spin-blind; kernel-not-A1 is the selector); cache exit 0; verdict leans on no `unaudited` row; source-only

### DIFF
--- a/docs/FAITH_FORCED_BY_A1_ANGLE_NOTE_2026-06-02.md
+++ b/docs/FAITH_FORCED_BY_A1_ANGLE_NOTE_2026-06-02.md
@@ -215,6 +215,161 @@ row above and does not lean on any `unaudited` row.
 
 ---
 
+## No-go discipline gate (N1–N8)
+
+**Status:** PASS for the narrow scoped claim only. The negative core being
+gated is *not* "FAITH is false" and *not* "boost-action faithfulness can never
+be derived." It is the single scoped statement: **A1 alone does not force the
+reconstructed boost/mass operator to act faithfully on C^2** — because the
+scalar boost `S(eta) = exp(eta·c) I_2` is a legitimate (non-faithful) rep on the
+*same* qubit C^2, and the operator that excludes it (the relativistic spin-1/2
+kinetic kernel) is matter-attachment content, not A1. The positive companion —
+*state-space* FAITH (j=0 is 1-dim, excluded by dimension) — IS an A1-consequence
+and is NOT being negated here.
+
+### N1 — Alternative route enumeration
+
+Routes by which A1 (or A1 + native dynamics) might be claimed to *force*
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the durable narrowed source result on main as fdb746c18. The landed version replaces the old A1/dimension framing with named Quantum language: the local Clifford algebra is faithful, but Quantum alone does not force boost-action faithfulness without a matter-attachment or kinetic-kernel selector.

PR #2537, opened 2026-06-03
science(koide): reflection/T-positivity is BLIND to the cross-site statistics sign (frame-identical transfer operator) — yet the sign is a genuine generator-level invariant, surfacing only on closed loops

jonathonreilly:
## Summary

**CARRIER LANE attack — does reflection/T-positivity force CAR over the hard-core boson?** **Verdict: NEUTRAL (high confidence)** — positivity is *blind* to the sign — but the sign is a **genuine generator-level invariant**, and the note isolates exactly where it becomes physical.

## Argument
The object these principles certify is the positive transfer operator `T = exp(−τH)`. On the local/open chain this is **frame-identical**: the same nearest-neighbour hopping built in the hard-core-boson and JW-fermion frames gives `H_HCB = H_fermion` as literally the **same Hermitian operator** (each bond differs by the string sign, but the Hermitian *combination* is equal), so `T` has identical positive spectrum (min eig `0.3269` both). Bare `⟨a₀†a₁⟩` sign-flips (`±0.4472`) but gauge-invariant `⟨n₀n₁⟩` is identical (`0.05`). **Positivity is blind to the sign.**

**Physical vs convention (DHR):** the sign is convention where **local** (it lives in the JW string — DHR's gauge); it becomes **physical only on a closed loop** (the ring wrap-around carries `(−1)^Q`; ring spectra differ: HCB `−2.828` vs fermion `−2.0`) — but **both** ring theories stay Hermitian and bounded below, so both give a positive transfer operator. The loop sign distinguishes two theories; it doesn't make either violate positivity.

**Self-correction (recorded, not used as support):** built and **refuted** in-suite the over-claim "the JW string is a trivial unitary gauge" — *false* at the generator level: the intertwiner `{W : W bᵢ = cᵢ W}` has **null dimension 0**, because `[b₁,b₂]=0` vs `{c₁,c₂}=0` is an isomorphism-invariant relation no `*`-automorphism can cross. So the sign is a genuine distinction; neutrality comes from **positivity-blindness, not gauge-triviality**.

**Next path this opens:** a fermion-parity-superselection / graded-locality principle on closed spatial loops would select CAR — and the note hands it a concrete witness (the **ring ground-state splitting**).

## Test plan
- [x] Runner — **31/31 PASS** (`H_HCB=H_fermion`; identical `T` spectrum; `⟨n₀n₁⟩` equal; ring splitting; null intertwiner dim 0); **no imports**; tiers verified on `origin/main`; source-only

### DIFF
--- a/docs/CAR_FROM_POSITIVITY_NEUTRALITY_NOTE_2026-06-02.md
+++ b/docs/CAR_FROM_POSITIVITY_NEUTRALITY_NOTE_2026-06-02.md
@@ -211,6 +211,161 @@ lens that has now been ruled *neutral*, sharpening — not closing — the searc
 
 ---
 
+## No-go discipline gate (N1–N8)
+
+**Status:** PASS for the narrow positivity-neutrality claim only. The claim being
+gated is *not* a repo-wide "fermions are impossible" no-go and *not* a claim that the
+cross-site sign is unphysical. It is the single negative statement that **reflection
+positivity / emergent-time T-positivity / single-clock Stone reconstruction do not
+force CAR over the hard-core boson, nor force the HCB** — because the object these
+principles certify (the positive transfer operator `T = exp(-τH)` and its spectrum,
+equivalently `H_gen = -(1/τ) log T ≥ 0`) is **frame-identical** on the local
+structure (§8a–c), so positivity is **blind** to the bare exchange sign. The gate's
+"alternative routes" are therefore the *positivity-flavored* mechanisms that could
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop landed the durable narrowed source result on main as 40d2de534. The landed version is route-local: finite transfer positivity / Stone readout is neutral on the tested open-chain and ring witnesses; it does not claim a global reflection-positivity no-go or close graded-locality / parity-superselection routes.

PR #2543, opened 2026-06-03
observable-principle P1: record/recorded-information route reduces to P1 (measurement path-a' instance closed)

JONbridgerapps:
## What this is

Tests the most promising P1-retention idea: the framework is **"qubit (probability) until record (information)"** — and the multiplicative→additive switch (the `−log`) happens *at the record* (`amplitude × → Born p=|ψ|² × → record → info −log p +`). Does that **derive** P1 from A1's measurement structure (observables are records; records' information is additive)? Math independently re-verified.

## Verdict: `circular_log_reintroduced`

It **re-describes** P1 but re-introduces the same `−log` free choice. Two decisive findings (runner `PASS=19 FAIL=0`):

1. **A1 doesn't force "observable = record."** `MINIMAL_AXIOMS` commits only the per-site qubit operator algebra (`M₂(ℂ)≅Cl(3,0)`) + the `Z³` substrate, and says *records / Born probabilities are **not** additional primitives in A1–A2* (the Born rule is `audited_conditional`). So "the observable is the recorded outcome" is an *extra identification*, not an A1 consequence — the route can't derive P1 *from A1*.
2. **The `−log` at the record is the same choice as P1.** The framework's record object (a Kraus instrument, `retained_bounded`) labels branches by the Born weight `p_r`, **multiplicative** on independent branches. Any *additive* recorded scalar satisfies `Φ(p_A p_B)=Φ(p_A)+Φ(p_B)` → `Φ=c·log` = **(Add)=P1**; the record's own Born/normalized quantification is exponent-**BLIND** (#2504). No third face.

## The decisive `−log` test, resolved

Free-monoid concatenation length **is** genuinely additive and unique-up-to-scale (reproven) — but it's **inert**: to make record-*size* equal the continuous v-chain observable `log|det|`, the `branch↦word` map must assign size `−log_b p` (Shannon/Kraft) = **the `−log` of P1**. The only alternative — a bare integer mark-count — is A1-unpinned *and* integer-quantized, so it can't equal a continuous/irrational `log|det|` (`−log₂(1/3)=log₂3` is irrational). So single-record physical size is **not** additive-matching-the-observable independently of the log-of-probability choice. (Averaged recorded entropy `−Σ p log p` = the structural-reframing `(II.b)` = P1, already `retained_no_go`.)

**Time:** `Ĥ=−log(T̂²)` is the #2517 circularity; time-as-recording-sequence is free-monoid concatenation, whose per-click mark-count is again `−log p`. Both reduce to the `−log` choice.

## Honest boundary

Closes the **measurement/record instance** of path-(a′) — the framework's actual Born/Kraus record carries the *multiplicative* weight, and making the recorded scalar additive *is* selecting `log` = P1. Does **not** foreclose all of path-(a′). **P1 stays admitted.**

So the "qubit-until-record" intuition is *physically* right about where the `log` sits — but the switch *relocates* the choice to the branch→scalar map; it doesn't remove it.

`claim_type: no_go`. No audit-lane data, no new vocabulary; Shannon/Boltzmann/free-monoid length cited as comparator only.

### DIFF
--- a/docs/OBSERVABLE_PRINCIPLE_P1_RECORD_INFORMATION_ROUTE_NARROW_NOTE_2026-06-03.md
+++ b/docs/OBSERVABLE_PRINCIPLE_P1_RECORD_INFORMATION_ROUTE_NARROW_NOTE_2026-06-03.md
@@ -1,0 +1,785 @@
+# Observable-Principle P1 Bridge — Record / Recorded-Information Route Narrow Note
+
+**Date:** 2026-06-03
+**Claim type:** no_go
+**Type:** no_go
+**Claim scope:** narrow no_go that tests the **recorded-information reframe** as
+a candidate derivation of the admitted P1 premise (scalar additivity on
+independent subsystems) of
+[`OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md`](OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md)
+from the framework's own measurement structure. The reframe proposes the
+pipeline "qubit (probability) until record (information)": amplitude `psi`
+(multiplicative for independent tensor systems) → Born `p = |psi|^2`
+(multiplicative) → a **record** forms → recorded **information** `I = -log p`
+(**additive**: `I_AB = I_A + I_B`). The claim is that the physical scalar
+observable is the recorded information (additive) because (i) A1 is a
+measurement framework so observables **are** records/measurement-outcomes, and
+(ii) a physical record is a **sequence of definite marks** whose size is
+additive **by concatenation** (a free-monoid structural fact), so P1 would
+reduce to A1 + record-additivity = **derived**, not admitted.
+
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed the durable narrowed result on main as cb8b9fc02.

What landed: OBSERVABLE_PRINCIPLE_RECORD_SCALAR_MAP_NO_GO_NOTE_2026-06-05 plus its exact runner. The landed form updates the PR to the approved Lattice, Quantum, Record framework: Record supplies finite scalar additivity after a scalar record surface is specified, but it does not derive the branch-to-scalar map, log-det generator, Born weight, determinant amplitude, P1 exponent selection, or arbitrary physical-observable identification.

What did not land: the original stale A1/A2 wording, status/audit-verdict language, and the broader record-route packet. Checks run: paired runner PASS=25 FAIL=0, py_compile, vocab_lint --fix, audit pipeline validation, audit_lint --strict, generated-output restore, diff checks, generated-output guard, and raw added-line hygiene scan.

PR #2552, opened 2026-06-03
observable-principle: the det FORM atom (M) reduces to fermionic statistics (consolidation, not dissolution)

JONbridgerapps:
## What this is

The culminating check of the P1/(M) thread. Tests the "single statement" hope: `W = log|det(D+J)|` = **"the recorded information of the fermion field"** — with det from Berezin (a theorem) and log from records→information (P1) — hoping (M) *dissolves* so records→information is the lone admission. Math independently re-verified.

## Verdict: `M_reduces_to_fermionic_statistics` — two atoms, not one

The Berezin reframe genuinely **improves** (M) but **relocates** it onto statistics rather than dissolving it (runner `PASS=29 FAIL=0` + independent 7/0):

1. **Berezin `Z_F = det(M)` is a theorem** — *conditional* on the antisymmetric/Grassmann measure (`{χ,χ}=0`). det multiplicative (`·` and `⊕`) ⟹ `−log det` additive.
2. **#2503's specific `tr`-witness is defused.** In `W=−log Z` the weight is fixed by the *integration measure*, so `tr` is **not** a candidate weight (no Gaussian has `Z=tr`; `tr≠det`, `tr≠permanent`). The abstract "product-character readout vs tr" choice no longer arises — **the real improvement.**
3. **But (M) doesn't dissolve.** `det`-vs-`tr` is replaced by **`det`-vs-permanent** (antisymmetric vs symmetric measure; bosonic Gaussian → permanent, structurally `≠ det`). And A1/A2 do **not** force fermionic statistics — the ratified `retained_no_go` `STATISTICS_AGNOSTIC_NO_FORCING`: JW-fermion and hard-core-boson frames are the **same** ungraded algebra `M_{2^N}(ℂ)` (2-site: bare ladders commute, JW-dressed anticommute, both span `M₄`, dim 16). So statistics is an **admission**.
4. **Orthogonal axes.** The permanent is **also** `⊕`-multiplicative, so **P1 alone cannot pick det over permanent**; statistics says nothing about the exponent. **Net = two atoms:** P1 (records→information; det a theorem *given* fermions) + **FS** (fermionic statistics).

## The genuine win — consolidation

The FS atom that (M) relocates onto **is substep 1 of the staggered-Dirac realization gate** (AC_φλ). So the FORM atom (M) and the staggered-Dirac substep-1 residual are the **same admission viewed twice** — this makes that identity explicit. (M) is **reframed** from the abstract "det-character readout" to the *physical* "the matter sector is fermionic," and **identified with an existing admission** rather than standing as a separate one.

So the scalar observable rests on **two clean physical founding principles** — *spin-statistics* (the field is fermionic ⇒ det) and *records→information* (⇒ log) — fused in "the observable is the recorded information of the fermion field." Not one axiom, but two *physical* ones, with (M) shown to be the staggered-Dirac fermionization admission, not a new thing.

`claim_type: no_go`. No new axiom/Tier-A item, no new vocabulary; Berezin/spin-statistics/JW cited as comparators only. No audit-lane data. References #2503 textually pending its landing.


jonathonreilly:
Review-loop closed without landing.\n\nReason:\n- The useful core is superseded by the cleaner mainline surfaces landed during this pass: 97ec65bed localizes Z=det to FS as an owner-approved fermionic-statistics admission, and 379a828f3 separates Record-readout form from Koide sector weight.\n- This branch's source surface is stale under the current axiom/admission framework: it uses bare A1/A2 language, raw retained/status labels, and a no-go wall frame that now needs to be stated in terms of Lattice / Quantum / Record plus explicit owner-approved admissions.\n- It also references #2503 as pending and tries to preserve old det-form vocabulary that has been narrowed by the newer log-det factor work.\n\nNo source files were landed from this PR. The branch should be rebuilt, if still needed, as a compact follow-up that cites the FS admission locator and avoids treating axioms/primitives or prior status labels as grade sources.

PR #2559, opened 2026-06-04
Force spatial dimension d_s=3 from the complex-qubit algebra (axiom-only route)

JONbridgerapps:
## What

A narrow `bounded_theorem` that **forces `d_s = 3` from axiom A1 alone** (the local algebra is the complex qubit `M_2(C) = Cl(3,0)`), reducing axiom A2. The integer `3` is currently stipulated **twice** — in A1's `Cl(3)` and again in A2's `Z^3` — and treated as independent. This proves they are not: the spatial dimension is forced to equal the vector-grade dimension of the per-site algebra, which is `3` because A1 is the *complex* qubit.

- `docs/QUBIT_ALGEBRA_FORCES_SPATIAL_DIMENSION_THREE_NARROW_THEOREM_NOTE_2026-06-03.md`
- `scripts/audit_companion_qubit_algebra_forces_spatial_dimension_three_exact.py` (12/12 sympy PASS)

## The forcing (import-free core)

- **`d_s <= 3`:** the traceless-Hermitian subspace of `M_2(C)` is 3-dimensional; the only matrix anticommuting with all three Paulis is `0` — no fourth direction.
- **`d_s >= 3`:** `{I, s_1, s_2, s_3}` spans the Hermitian observables, so the full qubit realizes all three directions.
- **the complex unit is the third axis:** `omega = s_1 s_2 s_3 = i*I` — the QM imaginary unit *is* the volume element of the 3-frame, so a *complex* qubit needs all three (`d=2` is the real `M_2(R)=Cl(2,0)`; `d=4` is `M_2(H)`, dim 16).
- **uniqueness:** `dim_R Cl(d,0) = 2^d = 8` only at `d = 3`.

## Reconciliation with the two prior routes (this is the third, axiom-only one)

This PR is **not** the first d=3 forcing — it is the one that uses **only A1**:

- **`native-su2-tightness-forces-ds3`** (scratch, `.claude/science/derivations/`) forces `d_s=3` from the **empirical** weak SU(2) via `dim(bivectors)=dim su(2)=3 => n=3`, and carries an `n=6` / `spin(6)=su(4)` twin. This note needs no empirical input and has **no n=6 twin** (`2^d=8` is uniquely `d=3` — check C11). Shared lever (`su(2)=bivectors`) appears here only as corollary P5.
- The retained **`_forces_d_four`** family (e.g. `yang_mills_coupling_marginality_forces_d_four`, `klein_four_product_bz_corners_forces_d_four`) forces **spacetime d=4** and **presupposes `d_s=3`** — so this note is **upstream** of it, supplying the `d_s=3` they assume.

## Scope / honesty

The pure-algebra content is unconditional. The **spatial-dimension reading** is conditional on premise `(GR)`: *"spatial directions are the vector grade of the local algebra"* — a strictly weaker, principled replacement for A2's dimension stipulation (the framework's own "space emerges from the qubit" stance made precise). A free integer becomes a theorem.

**Does NOT claim:** the lattice's discreteness / integer / cubic / `Z` structure (only the dimension count); `d_t`; that `M_2(C)` itself is forced; to be the only d=3 route; any change to a numerical prediction. `claim_type: bounded_theorem` — `(GR)` is named as the admitted premise.

## Discipline

Reproves all algebra from the Pauli generators at exact precision; Lawson & Michelsohn Clifford classification cited as comparator only. All cited links (Cl(3) backbone, the `_forces_d_four` family, native-su2) confirmed present. No audit-lane data committed.

### DIFF
--- a/docs/QUBIT_ALGEBRA_FORCES_SPATIAL_DIMENSION_THREE_NARROW_THEOREM_NOTE_2026-06-03.md
+++ b/docs/QUBIT_ALGEBRA_FORCES_SPATIAL_DIMENSION_THREE_NARROW_THEOREM_NOTE_2026-06-03.md
@@ -6,12 +6,14 @@
 qubit operator algebra `M_2(C)` — equivalently `Cl(3,0)` (axiom A1) —
 intrinsically carries **exactly** a 3-dimensional Euclidean vector grade,
 so that the spatial-dimension count `d_s = 3` is **forced** and **unique**
-from A1 alone, once spatial directions are identified with the
-mutually-anticommuting Hermitian directions of the local algebra. The
-single premise that carries the word "spatial" — `(GR)` below, "spatial
-directions are the vector grade of the local algebra" — is offered as the
-**replacement** for the current independent stipulation in axiom A2 (`the
-sites form Z^3`). The pure-algebra content (P1)–(P5) is unconditional; the
+from A1 alone, once a spatial direction is identified with an independent
+observable of the local qubit (a mutually-anticommuting Hermitian
+measurement axis — a Bloch axis). The single premise that carries the word
+"spatial" — `(GR)` below, "a spatial direction is an independent qubit
+observable" — is offered as the **replacement** for the current independent
+stipulation in axiom A2 (`the sites form Z^3`). It is an identity, not a
+projection: the spatial directions **are** the qubit's observables, and a
+two-state system has exactly `2^2 - 1 = 3` of them. The pure-algebra content (P1)–(P5) is unconditional; the
 spatial-dimension reading (T) is conditional on (GR). This is the
 **axiom-only route**: it uses no empirical input. No anomaly content, no
 staggered realization, no gauge group, no lattice numerics, and no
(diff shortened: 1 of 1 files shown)


JONbridgerapps:
Closing as superseded by an honest correction. On working the d=3 derivation through to the connectivity layer (as the review panel demanded), the load-bearing premise (GR) — 'a spatial direction is an independent qubit observable' — turns out to be FALSE in general: the Ising qubit model on Z^d (same sigma_3 bond reused in all d directions, distinguished by the graph not the algebra) is a valid {Locality, Quantum} model for ANY d, so the qubit's three observables do NOT fix the lattice dimension. The '3' caps d_s only for an anticommuting (Dirac) spatial frame, where a 2-component spinor admits at most 3 anticommuting Hermitian gammas (maximal: the 2-component Weyl spinor of 3+1D). That relativistic structure is the admitted staggered-Dirac gate, so d=3 is downstream of the matter-dynamics admission, NOT a theorem of {Locality, Quantum, Record}. Superseded by SPATIAL_DIMENSION_NOT_FIXED_BY_LOCALITY_QUANTUM_DIRAC_SPINOR_CAP_NARROW_NO_GO_NOTE_2026-06-04 (6/6 sympy). The composition results (#2563/#2573) are unaffected.

PR #2563, opened 2026-06-04
Tensor composition requires local tomography beyond Locality (narrow no-go)

JONbridgerapps:
## What

A narrow `no_go`: **Locality (Axiom II) does not entail the C\*-tensor composition rule** for per-site qubits. The precise additional content is **local tomography**, which is *strictly stronger* than Locality — witnessed irreducibly by real-vector-space QM (rebits).

- `docs/TENSOR_COMPOSITION_REQUIRES_LOCAL_TOMOGRAPHY_BEYOND_LOCALITY_NARROW_NO_GO_NOTE_2026-06-03.md`
- `scripts/audit_companion_tensor_composition_requires_local_tomography_exact.py` (8/8 sympy PASS)

## The no-go (import-free core)

- **rebits satisfy Locality but fail local tomography:** the rebit composite has commuting tensor factors (no-signalling = Axiom II) yet `dim_sa = 3·3 = 9 ≠ 10 = dim_sa(M₄(ℝ))`. So local tomography is **not** entailed by Locality.
- **complex qubits are locally tomographic:** `4·4 = 16 = dim_sa(M₄(ℂ))`.
- **given local tomography (+ M₂ nuclearity):** the two-qubit composite is uniquely `M₂(ℂ) ⊗ M₂(ℂ) = M₄(ℂ)`.
- **the gap is the complex unit:** `dim_sa(M₂(ℂ)) − dim_sa(M₂(ℝ)) = 4 − 3 = 1` — the central `i`. This is the **same** `i = ω = σ₁σ₂σ₃` the companion d_s=3 note reads as the third spatial direction, so the qubit's complex structure fixes *both* `d_s=3` and the composition rule.

## Why it matters (minimum-axiom program)

`MINIMAL_AXIOMS` records the C\*-tensor composition as *"standard mathematics, not an axiomatic commitment."* That's correct **given local tomography** — but this note proves local tomography is itself a physical premise *not* supplied by A1+A2. So the composition rule carries exactly **one named, precise admission** rather than none. This **refines** (does not contradict) the axioms memo, and surfaces a hidden admission so it can be minimised, derived, or accepted on the record.

## Scope / honesty

`claim_type: no_go` + a conditional corollary. **Does NOT** claim local tomography is false (complex QM has it); **does NOT** derive the tensor rule (the corollary is conditional on local tomography); **does NOT** resolve whether to accept vs derive local tomography (deriving it from A1's complex unit is a flagged open lane); **does NOT** address the graded/superselected composite.

## Discipline

[...]

PR #2586, opened 2026-06-04
Spatial dimension is NOT fixed by Locality+Quantum (corrective no-go, supersedes #2559)

JONbridgerapps:
## What

A corrective `no_go`. Working the d=3 derivation through to the connectivity layer (as the review panel demanded) shows the qubit **does not** fix the spatial dimension — superseding #2559's "d=3 from the qubit."

- `docs/SPATIAL_DIMENSION_NOT_FIXED_BY_LOCALITY_QUANTUM_DIRAC_SPINOR_CAP_NARROW_NO_GO_NOTE_2026-06-04.md`
- `scripts/audit_companion_spatial_dimension_not_fixed_by_locality_quantum_exact.py` (6/6 sympy PASS)

## The result

- **(NG)** `{Locality, Quantum}` do **not** determine `d_s`. The Ising qubit model `H = Σ σ_z^x σ_z^y` on `Z^d` is a valid local-qubit model for **any** `d`: the *same* bond operator σ_z is reused in every direction (directions distinguished by the **graph**, not the algebra), the bonds commute, no algebraic cap. The qubit's three observables impose nothing on the lattice dimension. **This kills #2559's premise** "spatial direction = independent qubit observable" (false: the Z^5 Ising model has qubits with 3 observables and 5 directions).
- **(CAP)** The "3" caps `d_s` **only** for an anticommuting **Dirac** frame: a 2-component spinor admits at most 3 anticommuting Hermitian γ's. Maximal case: the 2-component **Weyl spinor of 3+1D** (`σ^μ = (I, σ_i)`).
- **(SIZE)** The cap is set by the **spinor size** (a 4-spinor admits a 4-element frame), i.e. by the *matter dynamics*, not by the axioms.
- **(CONS)** So `d_s = 3` is **downstream of the relativistic-matter admission** (the staggered-Dirac gate), **not a theorem of `{Locality, Quantum, Record}`**.

## Why this is the honest correction

The panel was right: a single qubit has no neighbours, so geometry can't come from one site's algebra. It comes (if at all) from the *dynamics* — and the dynamics is admitted. This note also flags the real **su(2) double-use** the panel raised: `σ_i σ_j = i σ_k`, so the spatial frame (vectors) and the weak su(2) (bivectors) are the *same* operators — unresolved by the algebra (clause E).

## Scope

`claim_type: no_go`. Does NOT claim `d_s` is intrinsically undetermined (the dynamics fixes it); does NOT claim the Dirac dynamics is wrong; does NOT derive that dynamics (the open staggered-Dirac gate); does NOT resolve the double-use; changes no number. The composition results (#2563/#2573) are unaffected. Reproves all facts from the Paulis; no imports. No audit-lane data.


JONbridgerapps:
Closing — the 'd=3 is downstream of the matter dynamics' framing is wrong, and backwards (space's dimension should not depend on the matter placed on it). The honest position is the original one: d=3 is a structural PRIMITIVE of the lattice axiom (Z^3), where the framework has always had it (Cl(3) x Z^3). The bare mathematical sub-result (a qubit can live on any Z^d, so the #2559 qubit-derivation of d=3 was doomed) is true, but the correct conclusion is simply 'd=3 stays a lattice-axiom primitive', not 'it rides on the Dirac admission'. Reverting to Cl(3) x Z^3 with d=3 as a spatial-structure primitive. The composition results (#2563/#2573) are unaffected.

PR #2591, opened 2026-06-04
Koide r=1/2 (Q=2/3) is NOT symmetry-protected — a dynamical norm-balance (narrow no-go)

JONbridgerapps:
## What

A clarifying `no_go` from attacking the core Koide link: **r = |b|²/a² = 1/2 (Q = 2/3) is not protected by any unitary symmetry** of the C₃-circulant Yukawa — it's a *dynamical norm-balance*.

- `docs/KOIDE_R_HALF_NOT_SYMMETRY_PROTECTED_DYNAMICAL_NORM_BALANCE_NARROW_NO_GO_NOTE_2026-06-04.md`
- `scripts/audit_companion_koide_r_half_not_symmetry_protected_exact.py` (6/6 sympy PASS)

## The result

**Q=2/3 sharpened** — three equivalent clean statements: √m at **45°** to the democratic (1,1,1) axis ⟺ **equal energy** in the C₃ singlet/doublet channels (3a²=6|b|²) ⟺ equal diagonal/off-diagonal Frobenius weight.

**No symmetry protects it:**
- the full axis-permutation **S₃** forces **two degenerate masses** (multiplicities {1,2}) — excluded for charged leptons;
- **C₃** leaves r free (Q=(1+2r)/3 is non-constant);
- the singlet (1-dim) and doublet (2-dim) are different-dim irreps — no unitary can swap them.

So r=1/2 is a **norm-balance condition**, not a symmetry fixed point.

## Why it matters

- **Explains the 30-probe BAE campaign's failure** — there's no symmetry to find.
- **Redirects** the problem: r=1/2 must be set by the **dynamical corner coupling** |b|/a = 1/√2 on the three hw=1 cube corners — gated on the staggered-Dirac realization (the same gate the generations ride on). Sharpens the existing Schur / Z₃-equivariant foreclosures.

## Scope

`claim_type: no_go`. Does NOT claim r=1/2 is false (leptons sit at it); does NOT compute |b|/a (the open dynamical question); does NOT settle the quark/neutrino sectors. Reproves from primitives; PDG values are comparators only. No audit-lane data.


jonathonreilly:
Landed a narrowed review-loop salvage on main as 1ed92e738. I did not land the original overbroad claim as-is: the merged note is scoped to the tested C3/S3 unitary symmetry routes, keeps dynamical/extra-structure mechanisms open, fixes the graph-visible dependency links, and leaves the claim unaudited in the audit queue.

PR #2594, opened 2026-06-04
feat(flavor): log-det FORM from the Record axiom + det-character; three-factor provenance reduces the 59-row log-det blocker to a precise residual

jonathonreilly:
## Summary
The genuine post-Record unlock target, packaged as **new files only** — zero overlap with the dependency-rewrite audit (#2593): it does not re-cite/edit the 91 dependents, `observable_principle_from_axiom_note`, the axiom-premise allowlist, or any audit data file. 2 notes + 2 runners (6/6, 5/5) + 2 caches. No audit-language; status authority = independent audit lane.

## Context
#2593's exhaustive audit found the Record axiom moves **0 of 91** existing observable-principle dependents (all need broader content), with the **log-det generator `W = log|det(D+J)|` the dominant blocker (59 rows)**. The Record axiom's real contribution is to *new* additivity content and to **decomposing that blocker** — which is what this PR does.

## Findings
1. **`FLAVOR_LOGDET_FORM_UNDER_RECORD_AXIOM`** (6/6) — Form theorem: Record axiom (additivity over disjoint collections) + det multiplicative-character ⟹ **W = c·log|det|** (scale c = additive-baseline convention). `det` is the unique multiplicative character (`det(AB)=det(A)det(B)`; `Tr` is not; family `det^k`); only its additive image `log|det|` survives additivity over disjoint blocks. The Record axiom supplies the additivity that was the historical sticking point (the `p1_exponent_fixing` no-go and `extensivity_primitive` two-slope hole — both factor-1 content).

2. **`FLAVOR_LOGDET_GENERATOR_THREE_FACTOR_PROVENANCE`** (5/5) — `W = log|det(D+J)|` decomposes into: **(1) additivity** [closed by the Record axiom], **(2) det multiplicative-character** [`DET_UNIQUE_MULTIPLICATIVE`, unaudited — separate], **(3) source/action coupling** `dW/dj_x = Re Tr[(D+J)⁻¹P_x]` [admission — separate]. All three verified numerically. **The Record axiom discharges exactly factor 1**, reducing the 59-row blocker to the precise residual **{det-character (needs audit), source-coupling (admission)}** — the actionable path to cascade those rows.

## Scope discipline / relationship to #2593
Complementary and strictly non-overlapping: #2593 classifies *which* existing rows are blocked; this PR supplies the *positive* form theorem + the precise decomposition of the dominant blocker. New files only; no row re-cited, no audit data touched.

## Honest boundary
[...]

PR #2599, opened 2026-06-04
feat(flavor): audit-ready backbone for det-character selection (log-det factor 2) — det unique via composition-axis, Pattern-L honest bound, residual collapses to the fermionic origin of Z=det

jonathonreilly:
## Summary
Hardens **factor 2** of the log-det generator (`det` multiplicative-character selection) into an audit-ready, verified object, with the honest two-premise bound. New files only — does **not** touch the existing `DET_UNIQUE_MULTIPLICATIVE` note, the 91 dependents, `observable_principle_from_axiom_note`, the axiom-premise allowlist, or any audit data file. 1 note + 1 runner (7/7) + 1 cache. No audit-language; status authority = independent audit lane.

## What it verifies
- **Step (i) — composition-axis selection (clean math):** `det(A·S)=det(A)det(S)`; `tr`, power-traces `tr(M^s)`, elementary symmetric `e_k` **all fail** composition-multiplicativity and are excluded; the full family is `det^k` (GL(n) abelianization, computationally confirmed).
- **Step (ii) — additive-of-multiplicative → log (Record axis):** `Z=det` multiplies over patches; an observable additive over patches (**Record axiom**) is uniquely `W=c·log|det|` (Cauchy; `|det|^p` multiplies, only `log|det|` adds).

## The honest bound (Pattern L)
Verified explicitly: `tr` **is** additive over direct sums (`tr(A⊕B)=tr(A)+tr(B)`), so the Record axiom's additivity axis **cannot exclude `tr`** — selecting `det` requires the separate **composition-multiplicativity premise (M)**, which is supplied by the Berezin/Grassmann amplitude `Z=det(D+J)` multiplying over patches (a **fermionic-frame** property, overlapping factor 3).

## Net effect on the chain
Factor 2's *mathematics* (det-selection + log form) is now verified/audit-ready. The remaining open content of factor 2 collapses onto premise (M) — *why `Z=det` is the partition amplitude (the fermionic/Berezin frame)*. Combined with PR #2594, the **59-row log-det blocker now reduces to a single remaining open premise** (the fermionic origin of `Z=det` = the source-coupling/Grassmann-realization admission already tracked downstream), with additivity (Record) and the det-selection/log math both supplied.

## Honest boundary
Closes the **math** of factor 2; does not close premise (M) (fermionic frame) or the Koide **value** (Record gives form, not weight).

## Test Plan
- [x] `flavor_det_character_selection_audit_ready_2026_06_04.py` → 7/7
- [x] New files only; existing det-character note untouched; no audit-language

### DIFF
--- a/docs/FLAVOR_DET_CHARACTER_SELECTION_AUDIT_READY_2026-06-04.md
+++ b/docs/FLAVOR_DET_CHARACTER_SELECTION_AUDIT_READY_2026-06-04.md
@@ -1,0 +1,61 @@
+# Flavor — audit-ready verification of the det multiplicative-character selection (log-det factor 2): det is the unique composition-character (clean abelianization math) and additive-of-multiplicative fixes W = c·log|det|; the residual is the fermionic origin of Z=det, not the det-selection math
+
+**Date:** 2026-06-04
+**Claim type:** an audit-ready verification companion — supplies the verified computational backbone for the det multiplicative-character form-selection step (factor 2 of the log-det generator), with the honest two-premise bound. Not a value derivation; does not retag any row.
+**Status authority:** independent audit lane only. This note sets no audit status and assigns no grade. It does not edit, re-cite, or promote any existing row — in particular it leaves `OBSERVABLE_PRINCIPLE_DET_UNIQUE_MULTIPLICATIVE_CHARACTER_FORM_SELECTION` (the claim it backs) untouched.
+**Runner:** `scripts/flavor_det_character_selection_audit_ready_2026_06_04.py` (SCORECARD 7/7).
+**Depends:** `minimal_axioms` (the Record axiom) for the additive-over-patches step.
+
+## Purpose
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a narrowed version as b24e322ab.\n\nKey review changes before landing:\n- fixed the GL(n) character statement: regular algebraic characters are integer determinant powers; fractional complex branches such as det^(1/2) are explicitly rejected as global characters\n- made the log step depend on continuous/smooth positive-amplitude readouts, so Record supplies finite additivity only and does not by itself select det\n- kept the fermionic/Berezin origin of Z=det as the remaining premise (M), not as something supplied by Record\n- seeded one new unaudited ready row with deps ['minimal_axioms']; axioms are not treated as a retained_bounded source\n\nChecks passed: runner 8/8, cache check-only, py_compile, vocab_lint, audit pipeline, strict audit_lint, and git diff --check.

PR #2601, opened 2026-06-04
koide: corner fermion determinant does not select r=1/2 (dynamical-route pruning)

JONbridgerapps:
## Physics-loop dirac-corner-coupling — block 1

Prunes the Koide no-go ledger's **#1 OPEN dynamical lead**: does the staggered
corner **fermion determinant** re-weight the C3 isotype blocks to the balanced
lepton value `r = |b|²/a² = 1/2` (Q = 2/3)? **It does not.**

### Result (exact, sympy, 10/10)

On the three hw=1 cubic-BZ corners (kinetic staggered-Dirac `D = 0` there), the
mass matrix is the C3-circulant `M = aI + bC + b̄C²` with
`det(M) = a³ − 3a|b|² + 2|b|³ cos 3δ`. At fixed Frobenius scale `3a² + 6|b|² = 1`:

- `r = 1/2` is **not** a stationary point of `det(M)` (`d/dx ≠ 0` at `x = 1/2`, both δ-branches).
- The shape-stationary points are `r = 1` (det = 0, two-fold degeneracy) and `r = 4`; extrema otherwise at the boundaries `r → 0, ∞`.
- **Cycle-2 upgrade to the full one-loop effective potential:** the fermion CW potential `V_ferm = −log det(M)` has the same stationary points (monotonicity); any **scale-only** scalar addition is `x`-flat on the slice, so the combined `V_eff` has no `r = 1/2` vacuum.
- Contrast: `r = 1/2` IS the max-sector-entropy point (`S = log 2`) — so the **dynamical** (determinant) and **balanced** (entropy) criteria disagree.

### Synthesis — the three leading dynamical ingredients all fail r=1/2

| route | result | status |
|---|---|---|
| free Gaussian measure on Herm_circ(3) | (1,2)/F3 → κ=1, r=1 | ruled out (Probe 25) |
| fermion determinant `det(M)` | stationary at r=1, r=4 | ruled out (**this block**) |
| Z3 scalar potential `V(Tr K)` | V_eff min ≠ physical point | ruled out (framework's own Section 5 concession) |

This extends Probe 29's partial falsification (framework dynamics → κ=1 vs empirical κ=2) from the measure routes to the **dynamical** routes.

### Honest scope

- **Narrow route-pruning, NOT a universal no-go.** Two routes remain **UNTESTED**: taste-breaking scalar normalization; multi-factor Connes-Lott. The N1–N8 no-go discipline (route enumeration, wall-independence, steelman, residual) is recorded in the note.
- **Conditional** on the open staggered-Dirac realization gate (mass/Yukawa is gated on main).
- Does **not** claim Q=2/3 is empirically wrong — the lepton masses fit it at the admitted physical point `m_*` (<0.05% on √mass); it claims the **selection** of that point is not produced by these dynamics.
- `actual_current_surface_status`: exact negative boundary on the fermion-determinant route, conditional on the gate. No bare `retained`. No PDG values as derivation inputs. No audit-lane data.

### Files (2)
- `docs/CORNER_FERMION_DETERMINANT_DOES_NOT_SELECT_KOIDE_R_HALF_NARROW_OBSTRUCTION_NOTE_2026-06-04.md`
- `scripts/audit_companion_corner_fermion_determinant_not_r_half_exact.py` (10/10 exact)

### DIFF
--- a/docs/CORNER_FERMION_DETERMINANT_DOES_NOT_SELECT_KOIDE_R_HALF_NARROW_OBSTRUCTION_NOTE_2026-06-04.md
+++ b/docs/CORNER_FERMION_DETERMINANT_DOES_NOT_SELECT_KOIDE_R_HALF_NARROW_OBSTRUCTION_NOTE_2026-06-04.md
@@ -1,0 +1,130 @@
+# Corner Fermion Determinant Does Not Select Koide r=1/2 (Narrow Dynamical-Route Obstruction)
+
+**Date:** 2026-06-04
+**Type:** no_go
+**Claim type:** no_go (narrow, route-specific) — prunes the fermion-determinant
+corner-coupling route; the broader dynamical class stays open (two untested routes).
+**Claim scope:** on the three hw=1 cubic-BZ corners (where the kinetic staggered-Dirac
+operator vanishes), the corner Grassmann integral gives the fermion weight `det(M)` for
+the C3-circulant mass `M = a I + b C + b-bar C^2`. At fixed Frobenius scale the
+balanced Koide value `r = |b|^2/a^2 = 1/2` (Q = 2/3) is **not** a stationary point or
+extremum of `det(M)`. So the corner fermion-determinant route — flagged as the #1 OPEN
+dynamical lead in the Koide no-go ledger — does **not** select r=1/2.
+**actual_current_surface_status:** exact negative boundary on the fermion-determinant
+route; **conditional** on the open staggered-Dirac realization gate; bounded obstruction
+for the dynamical class with two named untested routes. Not retained on the current surface.
+**bare_retained_allowed:** false
+**Status:** independent audit required.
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed a narrowed/cleaned version as 6c3e8cf76.\n\nKey review changes before landing:\n- made the open staggered-Dirac gate and prior Koide route-pruning surfaces graph-visible dependencies\n- kept the claim as narrow route pruning, not a universal no-go; two routes remain explicitly untested\n- fixed the CW upgrade to use branchwise -log|det(M)| on both determinant sign branches instead of only the positive det branch\n- seeded one unaudited no_go row, dependency-blocked on the open/unaudited parent surfaces\n\nChecks passed: runner/cache 10/0, cache check-only, py_compile, vocab_lint, no-go N1-N8 review, audit pipeline, strict audit_lint, and git diff --check.

PR #2604, opened 2026-06-04
science(koide): reflection positivity + the closed-loop (−1)^Q monodromy does NOT force CAR — the carrier-side FS grading stays an input (statistics_agnostic, ring-witness exercised)

jonathonreilly:
## Summary
**Carrier-lane chirality, strongest angle:** does RP + T-positivity + the closed-loop `(−1)^Q` monodromy (where the statistics sign is physical) **force** a Z₂-graded (CAR / fermion-parity) local algebra — deriving graded locality? **Verdict: NO; the neutrality is derived, the forcing is a posit.**

The two ring algebras are built explicitly; the sign IS physical on the loop (ring ground states split). But: both rings are Hermitian/bounded → **both T-positive in every parity sector**; the hopping bond is parity-**even** → single-valuedness holds for both; qubit-tensor **locality favours the hard-core boson** (CAR's seam carries the non-local JW string); the ground-state parity flow selects CAR **only after** imposing FS; and the "HCB-is-inconsistent-on-a-loop" steelman is **refuted in-suite**. The only selector is **positing** the graded loop boundary condition = FS.

This **exercises** the ring-monodromy lens (previously only *named* in the carrier no-go) and rules it neutral, relocating the search onto the explicit ring parity-flow witness.

> **Import flag:** graded locality / fermion-parity superselection (FS) as the graded loop BC — the precise missing input. (A separate literature scout corroborates that the grading is gauge-underdetermined by observables, so RP alone cannot fix it; that citation would itself be an import.)

## Test plan
- [x] Runner — **29/29 PASS** (ring positivity, parity-sector RP, locality, monodromy, spectral flow, refuted steelman); full N1–N8 gate
- [x] Cache exit 0; non-circular; retained anchors verified on `origin/main` (`statistics_agnostic` flagged unaudited, re-derived independently); source-only


jonathonreilly:
review-loop landed the source-only salvage on main in 0d6a0cffd. I narrowed the note to the Lattice+Quantum+RP ring-monodromy no-go, removed stale A1/A2/status/provenance language, refreshed the runner/cache, and verified the generated audit row as no_go/unaudited with runner scripts/ring_monodromy_car.py before stripping generated outputs.

PR #2593, opened 2026-06-04
Rewrite eligible P1 dependencies to Record axiom

jonathonreilly:
## Summary

Audit and migration of all **91 direct dependents** of the old `observable_principle_from_axiom_note` after the 2026-06-04 adoption of `MINIMAL_AXIOMS_2026-06-04.md` as the framework's three-axiom baseline (Lattice, Quantum, Record).

**Method**: each of the 91 direct dependents was read in full; for every row we assessed whether the row uses only narrow Record additivity (REWRITE eligible — cite `minimal_axioms` instead, update prose from "P1" to "Record"), broader old-parent content (LEAVE — keep old parent), or both (SPLIT — cite both with explicit split).

## Counts

| Decision | Count |
|---|---:|
| REWRITE | **0** |
| SPLIT | **0** |
| LEAVE | **91** |
| Total | **91** |

**No rows in the current direct-dependent set rely *only* on narrow Record additivity.** Every direct dependent of the old parent uses broader content (log-det, P2/modulus, source/action, measurement/Born/decoherence, dynamics, or observable bridge). The migration is therefore a no-op on the source-rewrite front; the audit's value is the per-row content characterization and the explicit blocking-content category breakdown.

## Blocking-content distribution (LEAVE rows by primary category)

| Category | Count |
|---|---:|
| **log-det generator** (`W = log\|det(D+J)\|` and source-derivative algebra) | **59** |
| **observable bridge** (hierarchy `v` identification, observable selectors, P2-style observable wiring) | **18** |
| **P2/modulus** (phase-blind scalar-generator selection) | **5** |
| **source/action** (source-coupled local action admission, source-response coefficients) | **4** |
| **dynamics** (staggered-taste blocking transport, Theorem-4 substrate) | **3** |
| **measurement/Born/decoherence** (Kraus instruments, records sharpening, T-positivity) | **2** |
| **normalization/scale** | **0** |
| **Total** | **91** |

Each row is assigned exactly one **primary** blocking category corresponding to the most load-bearing broader-content dependence; secondary dependences (when present) are recorded in the verifier's cached output. The per-category headline lists of all 91 rows are in `docs/RECORD_P1_DEPENDENCY_AUDIT_NOTE_2026-06-04.md` §"Per-row decisions".

## Before / after counts on `origin/main`

| Quantity | Before (this branch base) | After (post-this-PR projection) |
|---|---:|---:|
| Old-P1 direct dependents remaining | **91** | **91** (UNCHANGED — no rewrites) |
[...]

PR #2619, opened 2026-06-04
feat(strong-cp): joint-basis bridge fails — theta_bar not forced (reflection-composed, anomaly-blind); residual = holomorphic generation coupling shared with Koide Q=2/3 + generation-ID (triple convergence)

jonathonreilly:
## Summary
Evaluates the load-bearing Prong-B blocker for strong-CP (and the AC_φλ mass-orientation reduction). **The joint-basis bridge fails — θ̄ is not forced to zero — and the single residual is one brick shared with two other major gates.** New files only; off #2593's surface. 1 note + 1 runner (6/6) + 1 cache. No audit-language; status authority = independent audit lane.

## Why it fails (all verified)
- **Sector-disjoint:** the gauge-OS reflection `Θ_OS` and the generation conjugation-parity `P` act on different tensor factors (`Θ_OS`=identity on generation, `P`=identity on gauge). A global `Θ=Θ_OS⊗P` is an **admitted splice**, not forced.
- **Reflection-composed → θ̄ survives:** parity rule `Θ(iQ)Θ⁻¹=−i(RQR)` — **pure-K** (no reflection) would force `θ̄=0`, but it's **unavailable** (`conj(M)≠M` for complex `b`; the Koide radius `|b|²/a²=1/2` is off the real axis). Only reflection-composed `P` is a symmetry, under which `θ̄` is even.
- **Anomaly-blind:** the axial rotation acts on fermions, not gauge links; the Wilson measure stays real; the Fujikawa Jacobian is `Θ_OS`-odd and cancels in the reflected expectation. The two reality conditions don't co-move.

## The payoff — triple convergence
The residual is a **global pure-K = a genuinely complex/holomorphic generation coupling** breaking `coeff(C²)=conj(coeff(C))`. That is *exactly* the holomorphic polarization `koide_emergent_time` flags as open, and the **same chiral-grading gate** as **Koide Q=2/3** and **generation-ID**. So **three of the framework's hardest gates reduce to one unbuilt brick** — a chiral/holomorphic grading on the generation R³ factor (which spacetime γ₅ can't transport over). The residual is *shared*, not three separate imports.

## Honest standing
The framework does **not** force `θ̄=0` (strong-CP stays a genuine admission, shared with the SM). The value is the conclusive evaluation of the bridge + the high-leverage identification that the residual is one shared brick.

## Test Plan
- [x] `strong_cp_joint_bridge_holomorphic_residual_2026_06_04.py` → 6/6
- [x] New files only; no existing row or audit data touched; no audit-language

### DIFF
--- a/docs/STRONG_CP_JOINT_BRIDGE_FAILS_HOLOMORPHIC_RESIDUAL_2026-06-04.md
+++ b/docs/STRONG_CP_JOINT_BRIDGE_FAILS_HOLOMORPHIC_RESIDUAL_2026-06-04.md
@@ -1,0 +1,74 @@
+# Strong-CP — the joint-basis bridge fails: θ̄ is not forced to zero (the emergent-time antiunitary is reflection-composed and anomaly-blind), and the single residual is a holomorphic generation coupling — the same unbuilt brick as Koide Q=2/3 and generation-ID
+
+**Date:** 2026-06-04
+**Claim type:** a no-go / load-bearing-blocker evaluation (the strong-CP joint-basis bridge fails; θ̄ survives) + a triple-convergence identification (the residual is one shared brick). Not a strong-CP solution.
+**Status authority:** independent audit lane only. This note sets no audit status and assigns no grade. It does not edit, re-cite, or promote any existing row.
+**Runner:** `scripts/strong_cp_joint_bridge_holomorphic_residual_2026_06_04.py` (SCORECARD 6/6).
+
+## The bridge and why it fails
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed durable content to main as e2cb6c687 (strong-cp: narrow joint-basis bridge failure). The landed version deliberately narrows the PR: it keeps only the runner-backed bridge-specific no-go, strips the broad triple-convergence/same-brick claim, marks holomorphic/chiral generation structure as an open lead, and leaves audit status to the independent audit lane. Closing this stale head after landing.

PR #2441, opened 2026-05-31
feat(koide): Berry-monopole bridge-reduction; value = chiral mass (nonzero Berry -> Q=2/3) vs non-chiral (Q=1)

jonathonreilly:
## Summary

Building the B-coupling→B-field bridge (which would force Q=2/3) builds a native, choice-free skeleton and **derives** the native dynamical answer, collapsing the whole charged-lepton Koide value into **one computable operator criterion**. Runner (8/8):

- **Index-map native/choice-free.** The `C₃` DFT diagonalizes `C`; `(1/3)Tr(C⁻¹H)=b` identifies `b` as the `C₃`-doublet Fourier amplitude of `H=aI+bC+b̄C²`.
- **Native effective-action order is SECOND (Q=1) — derived.** The Berry curvature of the filled band on the `b`-plane is **zero** by **eigenvector rigidity** (`H_circ`'s eigenvectors are the `b`-independent Fourier modes; `b` shifts eigenvalues without rotating eigenvectors) → second-order Ginzburg-Landau → per-dim → Q=1.
- **The criterion (unification).** A `Γ_χ`-**anticommuting** (chiral) coupling `Re(b)σ_x + Im(b)σ_y` has a **nonzero Berry monopole** → first-order → per-block → **Q=2/3**; a `Γ_χ`-**commuting** coupling has **zero** Berry → second-order → per-dim → **Q=1**. This unifies the **action-order**, the **det_C-vs-det_R reality type**, and the **equal-block-vs-trace measure** into ONE operator pin: chiral vs non-chiral generation mass. Non-circular (derives `F=0→Q=1`, `F≠0→Q=2/3` from bare `H[b]`).
- **The native mass is non-chiral → Q=1.** Every native circulant `Λ¹` mass commutes with `Γ_χ` (`[H,Γ_χ]=0`) → zero Berry → Q=1. `Q=2/3` = the single **chiral import**, shared with generation-ID, forbidden on the generation `R³` by `C³=I`.

## Boundary

Not a derivation of Q=2/3 — a bridge-reduction (native dynamics bounded to Q=1; the import pinned to one criterion). **Next path** (open, not foreclosed): `C³=I` blocks a chiral grading sourced from the generation `R³`, but not one from a tensor factor *distinct* from it — can an **off-generation** factor supply `Γ_χ`-anticommuting → nonzero Berry → Q=2/3 without breaking `C³=I`? Complements #2412 (canonical) and #2425.

## Test Plan
[...]

PR #2621, opened 2026-06-04
science(koide): the eigenstate quantum geometric tensor is identically ZERO (circulants share the fixed DFT basis) — information geometry does NOT escape the r=1/2 metric wall

jonathonreilly:
## Summary
Does **information geometry** (Fisher-Rao / quantum metric) or the **Foot-45°** geometry force the equipartition point `r=1/2` (`β/a=1/√2`, Q=2/3) on the non-chiral commuting-circulant route — where the canonical HS metric and dimension/trace counting both give `r=1`? **Verdict: NO — and the reason is new and structural.**

**New result:** the **quantum geometric tensor** (Fubini-Study metric **and** Berry curvature) of the circulant eigenstate bundle over the doublet plane is **identically zero** — because every circulant is diagonalized by the **fixed DFT basis** (the C₃ characters), so the spectral projectors are `b`-constant (`‖F†HF−diag‖~2e-16`; FD QGT `~2e-20`). This **extends the Berry=0 result (#2441) to the full QGT**, and pins *why* every state-weighting metric inherits the HS-metric neutrality: the eigenstate bundle is **flat in `b`**, so no quantum/Bures/FS extremum in `b` exists. Classical Fisher/Bures land at `r ∈ {0, 0.0147, 1/16, 1}`, never `1/2`.

Both prongs reduce to the **same single unforced bit**: `(1,1)` block-count (`r=1/2`) vs `(1,2)` dimension-count (`r=1`) on `ℝ[Z₃]=ℝ⊕ℂ` — which `koide_frobenius_isotype_split_uniqueness` (**retained_no_go**) proves underdetermined. This narrows the search from "find the right metric" (now closed by QGT=0) to "find a non-metric **state-selection** principle."

> **Honesty flag (surfaced, load-bearing):** the tracial-default chain (`pre_record_reference_state_tracial_derivation`, `powers_uhf_tracial_uniqueness`, `tomita_tensor_trace`) is **all unaudited** on `origin/main` — so the `(1,1)`-vs-`(1,2)` fork is **symmetric** on the retained surface, **not** tilted toward `r=1` as some recent notes state.

## Test plan
- [x] Runner — **31/31 PASS** (QGT≡0 exact; circulant DFT-diagonal; Fisher/Bures landing points; (1,1)-vs-(1,2) underdetermined); full N1–N8 gate
- [x] Cache exit 0; non-circular; retained anchors verified on `origin/main`; source-only


jonathonreilly:
Review-loop landed a narrowed source-only version in b73ee8936. The landed note keeps the runner-backed native-circulant QGT/Fisher/Bures/Foot-45 no-go, removes stale A1/A2 and exact-status language, adds a visible N1-N8 no-go gate, and leaves audit status to the independent audit lane.

PR #2622, opened 2026-06-04
science(koide): the √2 amplitude (Q=2/3) and the δ=2/9 phase are INDEPENDENT data — the charged-lepton closure factorizes into two separate gates (no shortcut from one to the other)

jonathonreilly:
## Summary
Does **one** native Z₃ structure produce **both** the Brannen amplitude `√2` (⟺ `β/a=1/√2` ⟺ Q=2/3, phase-independent) and the phase `δ=2/9`? **Verdict: NO — they are independent data; the closure *factorizes* into two separate gates.**

Three converging, verified reasons:
1. **Polar vs azimuth.** On the Fisher-Rao √m sphere, `cos²θ_p = 1/(3Q)`, so amplitude `√2` is the **colatitude** `θ_p=π/4` and `δ` is the **azimuth**; `g_φφ` is φ-independent → orthogonal coordinates. One can't fix the other.
2. **Every `(N−1)/N²` object is real-valued** (modulus `2/9`, argument 0); the only object with **argument** `2/9` (a Z₃ CFT character) has modulus **exactly 1** by unitarity. **Zero** native objects pair argument `2/9` with modulus `√2` (η-scan over p≤12 confirms).
3. **Amplitude is N-independent, phase is N-dependent.** Amplitude = Koide cone Q=2/3 (45°, all N); phase = `(N−1)/N²`. A common N-object forcing amplitude `~√(N−1)` breaks Q=2/3 at N=4 (gives 5/6).

So the closure **factorizes into two independent gates** — (1) amplitude/equipartition `r=1/2` and (2) phase `δ=2/9` — each to be derived on its own terms. Forecloses only the *shortcut*; both gates stay open.

> **Import flag:** `δ=2/9` is Tier-A **admitted** (`charged_lepton_brannen_bae_delta_tier_a` = retained_pending_chain), not derived — the test is conditional on it (flagged); `b=(1/√2)e^{i·2/9}` is exhibited to show it's a **posit** (modulus ⊥ argument).

## Test plan
- [x] Runner — **27/27 PASS** (orthogonal Fisher coords; real-valued (N−1)/N² objects; unit-modulus character; N-dependence break; η-scan); full N1–N8 gate
- [x] Cache exit 0; non-circular; retained anchors verified on `origin/main`; source-only


jonathonreilly:
Review-loop landed a narrowed tested-class no-go on main as 804946a09. I scoped it to the enumerated Z3/C3 shortcut classes, kept both amplitude and phase gates open, removed /tmp companion references and broad native-exhaustion wording, and verified runner 27/0, cache freshness, vocab lint, run_pipeline.sh, and strict audit lint before stripping generated outputs.

PR #2628, opened 2026-06-04
observable: P2 phase-blindness is sector-resolved; residual routes to AC_φλ

JONbridgerapps:
## Summary

After the Record axiom absorbed the P1 (finite scalar additivity) premise, **P2** (phase-blindness — the scalar generator depends on `|Z|=|det(D+J)|` alone, not `arg Z`) is the sole remaining scalar-selection premise of [`OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md`](../blob/main/docs/OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md) (live `unaudited`). The parent itself names "derive ... CPT-even phase-blindness from retained bridge work, or narrow" as its re-audit trigger; the additivity half landed via Record on 2026-06-04, leaving P2.

This narrow bounded note shows **P2 is not an independent admission — it is sector-resolved, and its only residual is the `(M)`/Berezin determinant identification, which is already `AC_phi_lambda`-gated.** P2 therefore collapses into `AC_phi_lambda`; the genuine Tier-A admission count is unchanged (`AC_phi_lambda`, `theta`).

## The two complementary sectors

| Sector | `det(D+J)` | Why P2 holds | New? |
|---|---|---|---|
| mass-like / self-adjoint (the consumed sector) | `∈ R_{>0}` | phase-blind **by retained positivity** (`STAGGERED_ONLY_DET_POSITIVITY_CASE_A`, `retained`) | no |
| generic non-self-adjoint | genuinely complex | continuous single-valued real readout kills the compact `U(1)` factor | no — **prior** observation (two-stage synthesis §2.3) |

## Honesty / scope (what this is NOT)

- **Not a novel closure.** The compact-`U(1)`-admits-no-map-to-`R` fact is standard and already stated in `OBSERVABLE_PRINCIPLE_P1P2_TWO_STAGE_SYNTHESIS...` §2.3 (credited, not claimed new). The contribution is the **residual placement** — the analogue, for the phase-blindness half, of the form-selection note's §7 routing of det-vs-tr into `AC_phi_lambda`.
- **Continuity is attributed to finite-block analyticity**, not to Record (Record supplies additivity only).
- Honest that the compactness lemma is **vacuous on the mass-like sector** (where the phase is already 0) and the positivity route covers it there.
- Not circular with the exponent `retained_no_go`: that foreclosed the `R_{>0}` (exponent) factor; the phase content is on the orthogonal `U(1)` factor.
- Does **not** retag the parent or set any audit status (audit lane is sole authority).

## Validation

Companion runner reproves, from primitives, at exact `sympy`/machine precision — **19 PASS / 0 FAIL**:
- compact-phase lemma (single-valuedness forces `c=0`; `arg` is `R/2πZ`-valued; `c·log|z|` is phase-blind);
- mass-like positivity (real-antisymmetric `det = ∏(m²+λ²) > 0`, symbolic `n=2,4` + 200 numeric seeds + ε-graded cross-check), with an explicit honest note that a *generic complex* anti-Hermitian would **not** give a real det;
- generic-sector witness, vacuity/scope guard, and the residual documentation.

```
python3 scripts/audit_companion_observable_principle_p2_phase_blindness_sector_resolved_2026_06_04.py
```

No new axiom, no new repo vocabulary, no fitted/PDG/MC inputs. All cited `effective_status` values verified against the live ledger.


jonathonreilly:
review-loop landed a narrowed source-only version on main in 436c26ca1. I kept this as a bounded residual-placement theorem: P2 has no separate residual only under the named determinant-readout regularity hypotheses, and the remaining package routes through (M)/AC_phi_lambda. I removed context-only dependency edges, added a bounded-wall N1-N8 gate, strengthened the runner's final checks against repo data, refreshed cache, and verified the audit row as bounded_theorem/unaudited before stripping generated outputs.

PR #2632, opened 2026-06-04
science(koide): the native emergent-time arrow is SHARPENING (heat-kernel) — it flows r→1, so Q=2/3 is anti-favored; resolves the r=1/2 stability contradiction

jonathonreilly:
## Summary
Resolves the live contradiction between two `audited_conditional` flow notes on main (records-flow-separatrix vs stable-under-thermalizing-arrow). **The native arrow is sharpening → `r` flows to 1 (Q=1), so Q=2/3 is anti-favored — a clean negative.**

The two maps `S(r)=2r²` (sharpening) and `g(r)=√(r/2)` (thermalizing) are **exact functional inverses** (`g∘S=S∘g=id`, verified) — the same flow in opposite time — so the question collapses to **which direction is native**. The **retained** `single_clock_stone_finite_dim_uniqueness` fixes the native step as the imaginary-time **contraction** `T^n=exp(−nτH_gen)` (heat-kernel, `H_gen⪰0`), whose generation-sector flow is `r(t)=tanh⁴t`: strictly increasing `r=0`(UV)→`r=1`(IR), with `r=1/2` a **transit value** pushed upward. So `r=1/2` is the **unstable separatrix**; the native dynamics flows away from it toward **Q=1**.

Corroborated: native RG/blocking fixed points sit only at `r=0`/`r=1` (retained_bounded `generation_degeneracy_minimal_symmetry_breaking`); the Born equilibrium `ρ=I/3` weights blocks by **dimension (1:2) → r=1**. The thermalizing-stable note failed twice (anti-native arrow + block-count measure). `r=1/2` is the balanced **stationary** point, reachable only by being *placed* there via the separate det_C input, not flowed to.

**Next path:** realizing `r=1/2` *dynamically* needs a **T-odd / chiral** ingredient (CP-violation or the chiral grading) installing the det_C coarse-graining — the **same** pin shared with generation-ID and the Koide chirality gate.

> **Honesty flag:** `luders_rule` (the `r→2r²` foundation) is **unaudited** on main — this note re-grounds the arrow on the **retained** single-clock step instead.

## Test plan
- [x] Runner — **24/24 PASS** (exact inverse maps; `tanh⁴t` heat-kernel flow; `β_r≠0` at 1/2; RG fixed points r=0/1; Born→r=1); full N1–N8 gate
- [x] Cache exit 0; non-circular (Q derived from spectrum); retained anchors verified on `origin/main`; source-only


jonathonreilly:
Landed reviewed salvage to main as cb5acb417. I did not land the original unconditional native-arrow no-go: the runner hard-coded the tanh^4 flow and did not derive it from the retained single-clock/native-beta surface. The landed source is narrowed to a bounded route-pruning theorem for the supplied heat-kernel path, with the native-beta/coarse-graining bridge left explicit for independent audit.

PR #2624, opened 2026-06-04
koide: FRONTIER CORRECTION -- modulus gives r=1, chirality is phase-only (retracts blocks 4/5)

JONbridgerapps:
## Physics-loop dirac-corner-coupling — block 6 (frontier correction)

**Real-QFT review refutes the block-4 (#2614) / block-5 (#2617) "chiral → r=1/2" mechanism.** Both landed on main; this corrects them. This is frontier science self-correcting: external physics killed my own headline lead.

### The refutation (real QFT, cross-checked vs the literature)
The fermion-induced fluctuation-determinant **modulus** that sets the Koide magnitude `r = |b|²/a²` is an `|b|²`-type function:
- `V ~ Tr log(M†M)` (**Coleman-Weinberg**) → the C3 doublet contributes **two** real modes `(Re b, Im b)` (doublet Hessian **rank 2**) → `(1,2)` → **r = 1**.
- For a **chiral (Weyl)** fermion the effective action splits as `Re W = ½·(Dirac modulus)` (still `|b|²`, r=1) + `Im W = η`-invariant (a **phase**) — **Alvarez-Gaumé**. The modulus is unchanged by chirality.
- Holomorphy (`b` once → r=1/2) is a property of the **SUSY superpotential** only (**Seiberg**); even the SUSY *Kähler* potential is non-holomorphic. The framework has no superpotential.
- For **Kähler-Dirac/staggered** fermions the chiral structure is a `Z₄` anomaly on the **phase** (**Catterall-Butt**); the modulus is real.

So **chirality moves only the determinant phase** → it governs `δ = arg(b)` (the Koide phase ≈ 2/9), **not** the magnitude `r`. Blocks 4/5 mis-assigned it.

### Verified (sympy+numpy, 5/5)
1. `M = aI+bC+b̄C²` Hermitian → vector mass has real det (no chiral phase).
2. `V_mod = Tr log(M†M)` depends on both Re b, Im b; doublet Hessian **rank 2** → non-holomorphic → (1,2).
3. modulus extremum → **r = 1** (matches block 1, #2601) from Coleman-Weinberg.
4. the chiral η-phase needs a *non-Hermitian* M and moves `δ`, not `r`.
5. r=1/2 would need a rank-1 (holomorphic) doublet = a SUSY superpotential the framework lacks.

### Corrected verdict
[...]

PR #2685, opened 2026-06-05
flavor/gauge: one chirality grading underlies both weak parity violation and the flavor phase (no qulink)

JONbridgerapps:
## One chirality ε, two roles — and it does *not* stand on qulinks

The chirality push, with the owner's check answered up front: **this needs no qulink ontology.** `ε=(-1)^{x+y+z}` is a per-site sign, the weak `su(2)` is the on-site qubit's own, the Yukawa is site-based — all on the existing `{Quantum, Locality, Record}`.

### Result (numpy/sympy, 6/6)
The single chirality grading `ε` is the common root of two distinct SM features:
- **(A) Parity violation** — the chiral coupling `T^a ⊗ P_L` (`P_L=(𝟙+ε)/2`) is parity-violating; the vector `T^a ⊗ 𝟙` is not. SU(2)_L's chirality *is* `ε`.
- **(B) The η / δ mechanism** — the graded trace `Str(O)=Tr(ε O)` gives the **index/η** (`ε`-dependent), the determinant-phase that #2624 ties to the Koide phase `δ`. The vector/modulus potential is **even in δ** (pinned at the CP-conserving `δ=0`); only the chiral odd η selects `δ≠0`.
- **Vector limit** `ε→𝟙`: parity restored **and** η mechanism off — *both vanish together*.

So weak parity violation and the charged-lepton phase are **one chirality, not two inputs** — no new axiom, no qulink.

### Honest scope
- **Structural map**, not a value derivation. **Not** done here (gated/untried): the value `δ=2/9` (η on the gated staggered-Dirac mass), *why* `su(2)` couples chirally, and the magnitude `r=1/2` (the irreducible flavor admission, orthogonal).
- `actual_current_surface_status`: structural map; no bare `retained`; standard chirality/index math reproven from Pauli primitives; no PDG values; no fitted parameters; no audit-lane data.

### Files (2)
- `docs/ONE_CHIRALITY_GRADING_UNDERLIES_WEAK_PARITY_VIOLATION_AND_THE_FLAVOR_PHASE_NARROW_NOTE_2026-06-04.md`
- `scripts/audit_companion_one_chirality_grades_weak_su2_and_flavor_phase_exact.py` (6/6 exact)

### DIFF
--- a/docs/ONE_CHIRALITY_GRADING_UNDERLIES_WEAK_PARITY_VIOLATION_AND_THE_FLAVOR_PHASE_NARROW_NOTE_2026-06-04.md
+++ b/docs/ONE_CHIRALITY_GRADING_UNDERLIES_WEAK_PARITY_VIOLATION_AND_THE_FLAVOR_PHASE_NARROW_NOTE_2026-06-04.md
@@ -1,0 +1,84 @@
+# One Chirality Grading ε Underlies Both Weak Parity Violation and the Flavor-Phase Mechanism (Narrow Structural Note)
+
+**Date:** 2026-06-04
+**Type:** structural_map
+**Claim type:** narrow structural note — a single chirality grading `ε` is the common root of two
+distinct features; the VALUES (δ=2/9, the chiral origin of SU(2)_L) are **not** derived here (gated).
+**Claim scope:** the framework carries one chirality grading `ε = (-1)^{x+y+z}` — a **per-site** sign
+(the staggered chirality, `{ε,D}=0`). The same `ε` underlies **(A)** the chiral weak coupling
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop did not land this PR. The runner passes 6/0 for the elementary finite-algebra subset, but the source note overclaims the supported surface: it depends on open or not-landed surfaces including the electroweak qubit-link note and the #2624 eta-to-delta lead; it uses stale baseline axiom language with Locality instead of the current Lattice + Quantum + Record framework; and the runner hard-codes the no-qulink/site-matter independence conclusion rather than deriving it. The durable runner-backed subset is only the projector/parity-flip/graded-trace toy algebra, which does not by itself establish that one chirality underlies both weak parity violation and the flavor phase. No source was landed, and the branch is intentionally kept for a narrowed rewrite.

PR #2694, opened 2026-06-05
deriv(koide): Koide value Q=1/3+(2/3)r of a C3-equivariant generation mass operator

jonathonreilly:
## Summary

A clean, exact symbolic derivation (no fitting, no measured masses) of the Koide **structure** from the adopted axioms.

For any **C3-equivariant, K/CPT-real** generation mass operator
`Y = a·I + b·C + conj(b)·C²` (a∈R, b∈C, C the C3 cyclic shift), the Koide ratio is

```
Q = (Σ m_k)/(Σ √m_k)² = (Σ λ_k²)/(Σ λ_k)² = 1/3 + (2/3) r,    r = |b|²/a²
```

**exactly and independently of arg(b)**. The two real Wedderburn blocks carry powers `a²` (singlet) and `2|b|²` (doublet); their swap acts as `r → 1/(4r)` with unique fixed point `r = 1/2` (equal block power), at which **Q = 2/3**. The other distinguished settings are `r=0 → Q=1/3` and `r=1 → Q=1`.

## Proof (all exact in the runner, 14/14 PASS)

1. Eigenvalues `λ_k = a + 2|b|cos(θ + 2πk/3)` (verified against `a + b·ω^k + conj(b)·ω^{2k}`).
2. `Σλ_k = 3a`, `Σλ_k² = 3a² + 6|b|²`.
3. `Q = (3a²+6|b|²)/(9a²) = 1/3 + (2/3)(|b|²/a²)`; `dQ/dθ = 0`.
4. Equal block power `a²=2|b|²` ⟺ `r=1/2`; swap `r→1/(4r)` fixed point `r=1/2`.
5. `r = 1/2, 0, 1 → Q = 2/3, 1/3, 1`.

## Scope (honest boundary)

Derives the Koide **structure** (the family `Q=1/3+(2/3)r` and that the symmetric setting is `Q=2/3`). It does **not** claim any particular sector sits at `r=1/2` — the dial position `r` is a free flavor input (`a,b` unconstrained), handled in the companion no-go `GENERATION_DIAL_OCCUPANCY_FREE_INPUT`. No measured masses are used.

## Assumptions / provenance

- **A2 (Quantum)** + the emergent 3-generation carrier (hw=1 BZ-corner orbit, C3 regular rep; taste-generation provenance cited).
- **A3 (Record):** durable K/CPT-orbit readout fixes `a` real and fuses the two faithful characters into the real doublet (companion `RECORD_GENERATION_READOUT_TWO_SECTORS`).

[...]

PR #2695, opened 2026-06-05
deriv(generation): generation Koide data reduces to a single derived modulus per sector

jonathonreilly:
## Summary

The **positive companion** to the Koide value theorem (#2694), treated the same way: the value isn't forced, but the *structure* is derived.

For a C3-equivariant K/CPT-real generation mass operator `Y = a·I + b·C + conj(b)·C²`, the Koide observable is governed by **exactly one real modulus**. Up to scale, `Y` has two real parameters (`r = |b|²/a²` and `θ = arg b`), and `Q = 1/3 + (2/3)r` depends on `r` alone (θ flat). So:

> The framework **derives** that the per-sector generation Koide data collapses to a single real modulus `r` on the derived dial `Q = 1/3 + (2/3)r` — a dimensional reduction from the SM's free per-sector Yukawa to one number on a derived axis with distinguished settings r=0,1/2,1.

"r is free" is the **positive** statement "r is the *one* derived residual parameter per sector" — the reduction, not a gap.

## Proof (50/50 PASS, exact)

1. One modulus controls Q (scale- and θ-invariant; only `r=|b|²/a²` matters).
2. The map `(a,|b|) → r` is onto `[0,∞)` — every dial value realized (explicit preimages).
3. A1/A2 force the circulant form, A3's K/CPT fixes `a` real + the two-block readout; none constrain `|b|/a` (A3 supplies no weighting/occupancy rule).
4. Hence: per-sector generation flavor = one real modulus on the derived axis. (Independent confirmation: `koide_frobenius_isotype_split_uniqueness`, `action_normalization`, both retained_no_go.)

## Scope

The parameter-count companion to #2694: derives that the residual generation freedom is **one real modulus per sector** on the derived axis. It isolates that modulus as the single input; assigns no value. No measured masses used.


jonathonreilly:
Closed by review-loop: salvaged conditional single-modulus reduction landed on main with the runner path/guard fixed and audit dependencies explicit.

PR #2696, opened 2026-06-05
deriv(generation): color and generation are derived as independent Z3 structures (SM color ⊥ generation)

jonathonreilly:
## Summary

The **positive** form of the color/generation question, treated like Koide: the identification isn't forced — but the *derived structure* is the win, and here it reproduces the Standard Model.

The framework derives **two genuinely independent Z3 structures**:
- **Color Z3** — `SU(3)_c` center acts as `ω·I₃` on the color triplet → character `(3, 3ω, 3ω²) = 3·χ_ω`.
- **Generation Z3** — regular rep on the hw=1 orbit → character `(3, 0, 0) = χ₀+χ_ω+χ_ω²`.

They are character-inequivalent, hence **independent** — a fermion carries a color *and* a generation label independently (3×3), exactly as in the SM. The inequivalence **is** the correct `color ⊥ generation` physics; reading it as "failure to identify" inverts the physics (the SM requires them distinct).

## Proof (21/21 PASS, exact)

1. Two characters: color `(3,3ω,3ω²)`, generation `(3,0,0)`.
2. Inequivalent: agree at `e` (3), differ at `g`/`g²` (`3ω,3ω²` vs `0,0`); `g↔g²` relabel doesn't fix it.
3. `χ_ω` multiplicity 3 (color) vs 1 (generation).
4. Schur: `dim Hom = 3` but every intertwiner has rank ≤ 1 → no equivariant isomorphism.
5. Independence = SM structure (3 colors × 3 generations).

## Scope

Derives color and generation as distinct, independent Z3 structures (positive, SM-matching). It does **not** identify them — that would require the import `scalar-generation-action`, which *discards* the derived cubic axis cycle to force an identification the SM does not want. Color carrier `cl3_color_automorphism_theorem` is retained; generation carrier cited as provenance with the character reconstructed independently. No measured inputs.


jonathonreilly:
Closed by review-loop: salvaged conditional color/generation Z3 independence result landed on main with carrier provenance linked for audit dependency tracking.

PR #2702, opened 2026-06-05
[frontier] Born from envariance: p_k=|a_k|² derived from {Quantum + Record} non-circularly, conditional on ONE named premise (a state-functional probability exists) — the record-scalar no-go does NOT wall off Born

JONbridgerapps:
**Context / correction.** The frontier-map agent had relayed the new `RECORD_SCALAR_MAP_NO_GO` as "Born-from-Record is walled off." That over-reads it. The no-go is explicitly narrow — it blocks *only* "Record additivity **alone** ⟹ the branch measure" (the (ℝ₊,×)→(ℝ,+) homomorphism gives −c log p, not which measure), and states it "does not foreclose future ... admissions." Born via **state-symmetry** (envariance) or **frame-function additivity** (Gleason) is untouched. This PR goes through that door.

## Result (verified)
Born `p_k = |a_k|²` is derived from {Quantum + Record} on the record state, **non-circularly about the value**, conditional on a single named premise.

**Route 1 — envariance (the record-state route).** On `|ψ⟩=Σ_k a_k|s_k⟩|E_k⟩`:
- Equal amplitudes: a system swap `U_S` is undone by an environment swap `U_E`, `(U_S⊗U_E)|ψ⟩=|ψ⟩` exactly → equal outcomes are envariant → equal probability.
- **Anti-circularity (the key check):** for *unequal* amplitudes the same swap is **not** a symmetry (state moves by 0.239) — so "equal amp ⇒ equal prob" is **derived**, not assumed (rules out the strong Schlosshauer–Fine smuggle).
- General case: rational |a_k|² fine-grains into equal sub-records → counting + Record additivity → `p_k=|a_k|²` (verified on 2/3,1/3 etc.).

**Route 2 — Gleason/Busch backstop.** Records form a PVM; additivity over orthogonal projectors + Quantum (Hilbert) forces the measure **form** `Tr(ρ·)` (Busch covers dim 2). Outside the no-go (orthogonal-projector frame function, not the branch homomorphism).

## The honest residual — pinned to ONE premise
A full circularity ledger isolates the lone admission: **A3 = "a state-functional probability exists."** Everything else is physical invariance or the Record axiom; the value-equality A4 is *forced* by A1+A2 once A3 is granted. A3 is **not** in {Quantum, Record} — it is exactly the Schlosshauer–Fine (2005) / Barnum (2003) premise, now pinned to a single named assumption for this framework. (Route 2's value rides on the separate ρ=I/d identification.)

## Why this matters
- **Born-from-record is not walled off** — it is derivable conditional on A3, with the no-go intact.
- It pins the framework's Born residual to its *minimal* core ("probabilities exist"), arguably cleaner than the existing routes' residuals (non-contextual additivity / P-cal); given A3, the |amp|² **value** is forced.
- It is **complementary** to the prior QD result (`FLAVOR_QD_OBJECTIVITY_FIXES_BASIS_NOT_WEIGHT`): redundancy/objectivity + record-formation fix the **basis** (#2701); envariance fixes the **weight**. A3 is exactly the previously-flagged "measure-selection gap," now named.

## Scope
Claim type conditional/support. Does NOT claim unconditional Born from {Quantum, Record} alone; does NOT enlarge the Record axiom (A3 recorded as an admission); does NOT contradict the no-go. Runner PASS=44. Forbidden-import clean (Gleason 1957 / Busch 2003 cited as standard math). Independently re-verified by the parent (equal-case envariance exact; unequal not a symmetry; fine-grain → |a|²).

### DIFF
--- a/docs/BORN_FROM_ENVARIANCE_CONDITIONAL_ON_STATE_FUNCTIONAL_PROBABILITY_NOTE_2026-06-05.md
+++ b/docs/BORN_FROM_ENVARIANCE_CONDITIONAL_ON_STATE_FUNCTIONAL_PROBABILITY_NOTE_2026-06-05.md
@@ -135,6 +135,23 @@ plus additivity over **disjoint** records (+ A3); Route 2 uses frame-function
 additivity over **orthogonal projectors**. Both are different structures from the
 barred homomorphism. The no-go is unaffected.
 
+## Relation to the prior Quantum-Darwinism basis-not-weight result (no contradiction)
+
+[`FLAVOR_QD_OBJECTIVITY_FIXES_BASIS_NOT_WEIGHT_2026-06-02.md`](FLAVOR_QD_OBJECTIVITY_FIXES_BASIS_NOT_WEIGHT_2026-06-02.md)
+showed that the **redundancy / objectivity** machinery of quantum Darwinism
+(plateau, mutual-information redundancy `R_delta`) fixes the pointer **basis** but
+**not** the sector **weight**. This note does **not** contradict it: envariance is a
+distinct mechanism — a **state-swap symmetry**, not a redundancy/plateau argument —
+and it bears on the **weight**, not the basis. The two results are complementary:
+the redundancy route (and the pointer-non-demolition record-formation result) fix
+the **basis**; envariance fixes the **weight** `|a_k|^2`, but only after `A3`. The
+"measure-selection gap" flagged by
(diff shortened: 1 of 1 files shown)


jonathonreilly:
review-loop did not land this PR: Quantum + Record do not supply a Born probability/state functional by themselves. The branch is kept as the un-landed work handle.

PR #2704, opened 2026-06-05
[frontier] Born: form Tr(ρ·) + value |a_k|² are UNCONDITIONAL from {Quantum + Record}; the only residual is the operational/typicality identification (sharpens #2702: A3 shrinks)

JONbridgerapps:
Follow-up to #2702, triggered by an owner observation: *"doesn't the qubit just give us probability anyhow?"* — it does, and that collapses most of the Born residual.

## The sharpening (verified)
The Quantum axiom is `M₂(ℂ)`, a C*-algebra. Its states are positive normalized **linear** functionals, forced by the finite-dim structure theorem to be `ω(A)=Tr(ρA)` — so `ω(P)∈[0,1]`, additive over orthogonal projectors, and `ω(P_k)=|a_k|²` for a pure state, which **is Born**. So the "a state-functional probability *exists*" content (#2702's premise A3) is a **theorem of the algebra**, not an external admission. #2702 flagged A3 as external only by reading the axiom memo's prose rather than the algebra.

## Record discharges the Gleason "additivity input" residual
The existing `BORN_RULE_FROM_GLEASON_BUSCH` note's own G1 finding was "non-contextual additivity is an **input**, not derived" — but that note (2026-05-20) **predates the Record axiom (2026-06-04)**. The Record axiom (I additive over disjoint records = orthogonal outcomes) **is** exactly Gleason's non-contextual frame-function additivity. So:
```
{Quantum (Hilbert + algebraic state) + Record (additivity)}  →  Busch (dim 2)  →  ω=Tr(ρ·)  →  Born |a_k|²
```
Busch does real work (it *upgrades* Record's frame-additivity to the unique linear Tr-form — the form is **earned**, not assumed; verified: ρ uniquely recovered from additive effect-values on the qubit).

## The one genuine residual
The only thing left is the **operational identification** — that this `ω` equals the **empirical frequencies you record**. The frequency-operator route (Hartle) gives `⟨f̂_P⟩=Tr(ρP)` exactly (non-circular, but merely *restates* ω); its *convergence* "recorded freq → |a_k|²" is **circular**: the deviation weight is the binomial Born measure (uniform weighting gives ½, not ⅔). Verified across N=4..20. So the operational piece is **conditional on typicality** — the universal measurement-link residue that no Born derivation (any interpretation) escapes.

## Verdict (honest)
**Born's form `Tr(ρ·)` and value `|a_k|²` are unconditional from {Quantum + Record}** (modulo a small normalization input: Record gives additivity into (ℝ,+), the [0,1]/sum=1 bound is a minor separate input). The **only** irreducible residual is the operational/typicality identification. This **sharpens #2702**: the residual A3 does not vanish but **shrinks** from "a probability functional exists" (now shown to be in Quantum) to just "ω = empirical frequency." The envariance route (#2702) and this Gleason route land on the *same* irreducible residual via different paths.

So: **Born-from-record is not walled off** — the form and value are derived from {Quantum + Record}; what remains is the universal typicality/operational link, not a framework-specific gap. The narrow additivity no-go is intact and distinct throughout. Claim type conditional/support; no over-claim of "Born unconditional"; runner PASS=29; forbidden-import clean. Independently re-verified by the parent (algebraic state → |a|², Busch recovery exact, frequency circularity Born-vs-uniform).

### DIFF
--- a/docs/BORN_QUANTUM_RECORD_UNCONDITIONAL_FORM_VS_OPERATIONAL_RESIDUAL_NOTE_2026-06-05.md
+++ b/docs/BORN_QUANTUM_RECORD_UNCONDITIONAL_FORM_VS_OPERATIONAL_RESIDUAL_NOTE_2026-06-05.md
@@ -1,0 +1,309 @@
+# How Much Of Born Is Unconditional From {Quantum + Record}
+
+**Date:** 2026-06-05
+**Claim type:** conditional / support (with a sharpened residual)
+**Status authority:** independent audit lane only. This source note does not set,
+predict, or assert an audit verdict and does not claim "retained" or "promoted"
+standing.
+**Primary runner:**
+[`scripts/frontier_born_quantum_record_unconditional_2026_06_05.py`](../scripts/frontier_born_quantum_record_unconditional_2026_06_05.py)
+with cache
+[`logs/runner-cache/frontier_born_quantum_record_unconditional_2026_06_05.txt`](../logs/runner-cache/frontier_born_quantum_record_unconditional_2026_06_05.txt)
+(PASS=29, FAIL=0; peak RSS ~85 MB).
+
+---
+
+## Scope and honesty (read first)
+
+This note sharpens the prior conditional result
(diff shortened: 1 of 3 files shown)


jonathonreilly:
review-loop did not land this PR: Quantum + Record do not unconditionally supply the Born form/value. The branch is kept as the un-landed work handle.

PR #2713, opened 2026-06-06
science: strong-CP residual collapses into the flavor-CP phase (θ̄ = θ_QCD + arg det M_q)

JONbridgerapps:
## Summary

A **narrow structural note + exact runner (6/6 sympy/numpy)** showing that the strong-CP admission's *physical residual* is not independent of the flavor-CP phase already admitted as part of `AC_φλ`.

The physical, basis-invariant strong-CP angle is `θ̄ = θ_QCD + arg det(M_q)`. Two framework facts fix where it lives:

1. **Gauge side:** the real-Wilson gauge action carries no `θ` slot → `θ_QCD = 0` (the repo's already-selected surface, `STRONG_CP_THETA_ZERO_NOTE`; RP *cannot* force it — 2026-05-16 no-go).
2. **Matter side:** the generation Yukawa is the **Hermitian** C₃-circulant `M = aI + bC + b̄C²`, so `det M` is **real** and `arg det M = 0` — the **CP-conserving default** at which `θ̄ = 0` **and** the Koide phase `δ = 0` coincide.

So on the framework's surface the entire physical `θ̄ = arg det(M_q)` — a CP-odd phase of the *same* generation mass matrices whose CP-odd phase is `δ`. Both vanish at the real/Hermitian default and turn on together from **one reality-breaking**. The strong-CP problem here **is** the flavor-CP (Nelson-Barr) problem.

## Scope (honest — this is a collapse, not a solution)

- **Not a strong-CP solution.** It does not derive `θ̄ ≈ 0`; the controlled smallness (`θ̄ ≈ 0` while the CKM phase is `O(1)`) is exactly the **Nelson-Barr** problem, also unsolved in the SM.
- **Not, by itself, a Tier-A count reduction to 1.** The gauge-side real-Wilson selection (`θ_QCD = 0`) remains, and RP provably cannot force it.
- It *does* tighten the Tier-A bookkeeping: `θ̄` is shown to be the overall phase of the already-admitted generation mass matrices, not an independent gauge input.

## Verification

`scripts/audit_companion_theta_bar_residual_collapses_into_flavor_cp_phase_exact.py` — **6/6 exact**: Hermitian circulant ⇒ real det (arg = 0); `θ̄` chiral-rotation invariance; non-Hermitian (CP-violating) deformation ⇒ `arg det ≠ 0`; structural collapse; source-note boundary tokens.

**Discipline:** all algebra reproven from the C₃-circulant primitive; Nelson (1983)/Barr (1984)/Vafa-Witten (1984) are comparators only; no PDG values as derivation inputs (`δ≈2/9`, CKM phase named only as the targets *not* derived). No audit-lane data touched.

### DIFF
--- a/docs/THETA_BAR_RESIDUAL_COLLAPSES_INTO_THE_FLAVOR_CP_PHASE_NARROW_NOTE_2026-06-05.md
+++ b/docs/THETA_BAR_RESIDUAL_COLLAPSES_INTO_THE_FLAVOR_CP_PHASE_NARROW_NOTE_2026-06-05.md
@@ -1,0 +1,98 @@
+# The Strong-CP Angle's Physical Residual Collapses Into the Flavor-CP Phase (θ̄ = θ_QCD + arg det M_q; the Hermitian Circulant Is the CP-Conserving Default)
+
+**Date:** 2026-06-05
+**Type:** narrow_theorem
+**Claim type:** narrow_theorem (structural, computable-side) — the physical strong-CP angle
+`θ̄ = θ_QCD + arg det(M_q)` is, on the framework's real-Wilson gauge action (`θ_QCD = 0`), equal to
+the **generation mass-matrix phase** `arg det(M_q)`, a CP-odd phase of the *same* mass matrices whose
+CP-odd phase is the Koide flavor phase `δ`. The framework's generation Yukawa is the **Hermitian**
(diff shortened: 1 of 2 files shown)


jonathonreilly:
review-loop did not land this PR: the strong-CP collapse claim needs independent support and was not salvageable as-is. The branch is kept as the un-landed work handle.

PR #2719, opened 2026-06-06
[frontier] Arrow of time from record formation: the DIRECTION is record-derived (away from the low-record boundary); the residual is exactly the past hypothesis (universal-floor, distinct from typicality + ρ=I/d)

JONbridgerapps:
Attacks the arrow-of-time open gate (MINIMAL_AXIOMS lists "arrow … record-production dynamics" as outside the axioms; the smuggle audit notes the framework has *zero* past-hypothesis notes — genuinely un-attacked). The sequel to #2701 (record formation = irreversible writing).

## Result (verified, on an explicit 6-qubit system)
- **Record monotonicity = a candidate arrow:** under the #2701 redundant-broadcast record-writes, from a low-record initial the redundancy/record content R(t) increases monotonically (0→5 bits).
- **The decisive control — the arrow is in the initial condition, not the dynamics.** The microdynamics is **time-symmetric** (`H` real-symmetric, `ΘUΘ=U⁻¹`, verified non-trivially), so the *same* operators give: low-record initial → R(t) **increases** (forward arrow); high-record GHZ initial (built from scratch, not from the forward orbit) → R(t) **decreases** (arrow reversed); I/d equilibrium → **flat** (no arrow); generic high-entropy → records don't form (vanishes). The arrow's **sign is an output of the boundary condition.**
- **Residual = the past hypothesis.** Record formation derives the arrow's **direction** ("away from the low-record boundary"); the irreducible admission is the **existence of a low-entropy initial** = the past hypothesis (Boltzmann/Penrose). It is **universal-floor** — every time-symmetric microdynamics (CM/QM/QFT/GR) needs it — not a framework-specific gap.
- **Collapse check:** the past hypothesis is **distinct** from the other named residuals: ρ=I/d is *max-entropy* (the no-arrow fixed point), the past hypothesis is *low-entropy* (the arrow's source) — opposite extremes; and Born typicality is a within-sector weight, not a global initial-state selection. A third, distinct admission.

## Verdict
The arrow's **direction is record-formation-derived modulo the past hypothesis**; its **existence remains admitted** (the past hypothesis, universal-floor). Reconciled with T's time-symmetry: the symmetry is precisely *why* the arrow cannot live in the map — the same T gives increase, decrease, or flat purely by boundary condition. Not a from-nothing derivation; an honest pinning.

Bounded_theorem (structural/boundary-condition result on the explicit redundant-broadcast model; the "symmetric map ⇒ sign from boundary" logic is generic). Forbidden-import clean; runner PASS=20. Independently re-verified by the parent (low-record R=[0,1,2,3] vs GHZ R=[3,2,1,0] vs I/d flat, same dynamics).

### DIFF
--- a/docs/ARROW_FROM_RECORD_FORMATION_PAST_HYPOTHESIS_RESIDUAL_NOTE_2026-06-05.md
+++ b/docs/ARROW_FROM_RECORD_FORMATION_PAST_HYPOTHESIS_RESIDUAL_NOTE_2026-06-05.md
@@ -1,0 +1,142 @@
+# Arrow Of Time From Record Formation, Modulo The Past Hypothesis (Bounded)
+
+**Date:** 2026-06-05
+**Claim type:** bounded_theorem (boundary-condition / structural result; honest
+pinning, **not** a from-nothing derivation of the arrow).
+**Status authority:** independent audit lane only. This source note does not set,
+predict, or estimate any audit verdict. Effective status is pipeline-derived
+after independent audit and dependency closure.
+**Primary runner:**
+[`scripts/frontier_arrow_from_record_formation_2026_06_05.py`](../scripts/frontier_arrow_from_record_formation_2026_06_05.py)
+**Cached log:**
+[`logs/runner-cache/frontier_arrow_from_record_formation_2026_06_05.txt`](../logs/runner-cache/frontier_arrow_from_record_formation_2026_06_05.txt)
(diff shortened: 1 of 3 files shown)


jonathonreilly:
review-loop salvaged the durable source-level content to main in 1f2f29bb8; generated audit/status authority was stripped. Independent audit lane owns verdict/effective status.

PR #2732, opened 2026-06-06
science(koide): the charged-lepton Koide value reduces to ONE counting-measure bit; the order-4 J_cs is native (not missing), a static J is measure-neutral

jonathonreilly:
## Summary
Drafts the convergent diagnosis of the charged-lepton value as a clean A1+A2-native narrow theorem (no #2667/A3 dependence, no new axiom; `Q=2/3` is only a downstream check).

**The framework natively fixes everything except one binary choice.** It supplies the masses `m_k=λ_k²`, the form `Q=1/3+(2/3)r` (retained), **and** the order-4 complex structure `J_cs=(C−C²)/√3` on the doublet — `J_cs²=−P_doublet`, `J_cs⁴=+P_doublet`, `[J_cs,H]=0`, **real**, native on the **even** 2-dim doublet, needing **no** 4th dimension. This corrects the tempting "`R³` is odd → the order-4 `J` is missing" framing: det_C only needs `J` on the doublet (even); the odd direction is the irrelevant singlet.

**Why `J_cs` doesn't fix the value:** a *static* complex structure is **measure-neutral** — `exp(s J_cs)` has `det=1` (preserves the real volume) *and* `|det_C|=1` (preserves the complex volume). To set the value it would have to act as the **measure** (promote the operator to a mode-**count**); it doesn't.

**So the value = one counting bit** on `ℝ[Z₃]=ℝ⊕ℂ`:

| measure | r | Q | role |
|---|---|---|---|
| block-count `(1,1)` = det_C | 1/2 | **2/3** | observed |
| dimension `(1,2)` = det_R | 1 | 1 | over-determined default |

`koide_frobenius_isotype_split_uniqueness` (**retained_no_go**) forces neither; the masses are identical either way, only the count (the `√m`-sign readout) differs. Every route attacked (variational, RP, dynamics/einselection, gauge-invariant footing, 4th-dimension/emergent-time) reduces **time-independently** to this same bit. The one forcing-shaped open handle is a **grading** (not a static `J`) — the chirality-graded supertrace / equivariant index — conditional on the gated staggered-Dirac mass structure.

## Test plan
- [x] Runner — **22/22 PASS** (exact sympy: the family; `J_cs²=−P_d`/order-4/`[J_cs,H]=0`; `det(exp sJ)=1` measure-neutral; the two idempotents; block-count→r=1/2, dimension→r=1); full **N1–N8** gate
- [x] Cache `--push-mode=none` (exit 0); non-circular (`Q=2/3` a downstream check); all 7 cited anchors verified retained on `origin/main`; source-only

### DIFF
--- a/docs/CHARGED_LEPTON_VALUE_REDUCES_TO_ONE_COUNTING_BIT_SYNTHESIS_NOTE_2026-06-05.md
+++ b/docs/CHARGED_LEPTON_VALUE_REDUCES_TO_ONE_COUNTING_BIT_SYNTHESIS_NOTE_2026-06-05.md
@@ -1,0 +1,187 @@
+# Charged-Lepton Koide Value Reduces to a Single Counting-Measure Bit; the Order-4 J_cs Is Native, Not Missing — Narrow Synthesis
+
+**Date:** 2026-06-05
+**Type:** narrow theorem (structural reduction) + honest-status synthesis
+**Claim scope:** the standalone, A1+A2-native fact that the charged-lepton Koide
+value `Q` reduces — after the framework natively fixes the masses **and** the
+order-4 complex structure — to one binary measure choice (block-count vs
+dimension) on `R[Z_3] = R (+) C`, which the retained no-go
+`koide_frobenius_isotype_split_uniqueness` forces neither way. Does **not** claim
+to force `Q=2/3`; characterizes the residual.
+**Status authority:** independent audit lane; no audit outcome set; no new axiom,
+no import, no fitted value (`Q=2/3` is a downstream check, never an input).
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Closing as duplicate — the diagnosis is already landed at better tiers on main: `koide_import_two_bit_decomposition` (retained_bounded) already proves J_cs=(C−C²)/√3 is the native order-4 complex structure ("not a missing two-form"), that it commutes with H for all r (measure-neutral), and that the residual = the orientation bit + the block-count(r=1/2→Q=2/3)-vs-dimension(r=1→Q=1) measure bit; `flavor_split_the_brick_doublet_complex_structure` (retained) covers the same. This PR re-derives that content without adding new science.

PR #2907, opened 2026-06-06
science(meta): sharpen record principle to register-not-read ontology

jonathonreilly:
## Summary

Follow-up to **#2898** (closed = cherry-picked to main). Corrects a realist slip in the canonical record-outcome observable principle note and sharpens guardrail **G3**.

**The correction:** the record **registers** which central sector is realized — observables are *constituted in the record stack*, not observed/read from a pre-record value. "Read" smuggled realism back in (an `r` sitting out there to be measured); the framework has no such thing.

## Changes (delta vs the note on main)
- **New Ontology section** — register not read; observables constituted not observed; pre-record operators/states are **reconstructions**, not reality.
- **Sharpened G3** — the *partition* is **delivered** by the record; the *per-sector weight* (e.g. Koide `r`) is the **registered pattern** of the realized state — matched like the masses, neither imposed (axiom disclaims weighting) nor read (no pre-record value). **Worked case:** `r=1/2` is the charged-lepton registered pattern; the `I/3 → r=1` reference is a *reconstruction*, **not** the lepton record (the realist slip to avoid).
- **New runner check** — a sector weight depends on the realized state, not the partition (`w_ref=0.667` vs `w_realized=9.0`), making the partition-delivered / weight-registered separation concrete. Runner **PASS=13**.

## Test plan
- [x] Runner PASS=13 FAIL=0; cache regenerated (status ok)
- [x] Clean delta off origin/main; source-only triple; no audit-data files; no new axiom/import

### DIFF
--- a/docs/RECORD_OUTCOME_OBSERVABLE_PRINCIPLE_CANONICAL_PROPOSAL_NOTE_2026-06-05.md
+++ b/docs/RECORD_OUTCOME_OBSERVABLE_PRINCIPLE_CANONICAL_PROPOSAL_NOTE_2026-06-05.md
@@ -13,57 +13,65 @@ by the independent audit lane, not self-declared here.
 ## Purpose
 
 To state, once and citably, the reading of the RECORD axiom that several lanes have
-been using implicitly: **physical observables are recorded outcomes, i.e. central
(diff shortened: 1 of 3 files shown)


jonathonreilly:

PR #2925, opened 2026-06-06
science(koide): Dirac singular-value operator gives Q=2/3 at observed masses; readout pin moot

jonathonreilly:
## Summary

Final step of the chirality-pin attack. Resolves the **readout half** for the *physical* masses. Source-only: 1 note + 1 runner + 1 cache. Status: `unaudited`.

**Grant the steelman** (`KOIDE_DIRAC_MASS_FORCES_R_ONE...`): charged leptons are Dirac → physical masses are singular values (sign-blind). The fact it overlooks:

```
observed sqrt(m) = [0.715, 10.279, 42.153]  (all positive)
Q = (sum m)/(sum sqrt m)^2 = 0.66666 = 2/3   ->   r = 1/2
r=1/2 circulant at Brannen phase: eigenvalues (2.379, 0.040, 0.580) -- ALL POSITIVE
=> signed == singular-value here. The signed-vs-singular distinction is MOOT for the physical masses.
```

The steelman's fact-4 ("at r=1/2 *some* sqrt(m)<0, so you need the signed reading") is **δ-dependent** — true at δ=0.9, **false at the observed Brannen δ**. So the physical Dirac/singular-value masses already give Q=2/3. **No signed reading needed.**

The steelman's "**r=1**" is its *derivation* of r (12 routes: modulus, det, Berry-flat, …). Under **register-not-read** (G3), r is a *registered* per-sector weight (r=1/2 for leptons, like the masses), not a derivation target — so "12 routes → r=1" **confirms** r is non-derivable, consistent with the registered r=1/2, not a contradiction.

## Changes
- `docs/KOIDE_DIRAC_SINGULAR_VALUE_GIVES_Q_TWO_THIRDS_AT_OBSERVED_MASSES_NARROW_THEOREM_NOTE_2026-06-06.md`
- `scripts/dirac_singular_value_gives_koide_runner.py` (class A; **PASS=9**)
- `logs/runner-cache/dirac_singular_value_gives_koide_runner.txt`

## Boundary (honest)
- Does **not** derive r=1/2. The steelman's bounded no-go ("all derivation routes → r=1") **stands on the derivation frame** and is *consistent* with this note (different frames). The "r=1" resolution is the reframe (register, don't derive), which rests on the **unaudited** record-outcome principle — not a refutation of the steelman's algebra.
- Does **not** derive the Brannen δ=2/9 (separate; registered under G3).
- Grants the steelman's facts; adds the observed-δ all-positivity that makes its readout half moot.

## Test plan
- [x] Runner PASS=9 FAIL=0; cache regenerated (status ok)
- [x] Source-only triple; no audit-data files; no new axiom/import

### DIFF
--- a/docs/KOIDE_DIRAC_SINGULAR_VALUE_GIVES_Q_TWO_THIRDS_AT_OBSERVED_MASSES_NARROW_THEOREM_NOTE_2026-06-06.md
+++ b/docs/KOIDE_DIRAC_SINGULAR_VALUE_GIVES_Q_TWO_THIRDS_AT_OBSERVED_MASSES_NARROW_THEOREM_NOTE_2026-06-06.md
@@ -1,0 +1,120 @@
+# The Dirac (Singular-Value) Operator Gives Q=2/3 at the Observed Masses — the Chirality Pin's Readout Half is Moot — Narrow Bridge Theorem
+
+**Date:** 2026-06-06
+**Claim type:** positive_theorem (narrow; resolves the readout half of the chirality pin for the physical masses)
+**Status:** unaudited candidate. Graph-visible only so the independent audit lane can decide.
+**Primary runner:** [`scripts/dirac_singular_value_gives_koide_runner.py`](../scripts/dirac_singular_value_gives_koide_runner.py)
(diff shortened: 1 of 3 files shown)


jonathonreilly:

PR #2953, opened 2026-06-06
science(flavor): emergent C3 coupling scale constrained to a 9-order window; neutrino lightness explains small-CKM/large-PMNS

jonathonreilly:
## Summary

The last quantitative piece of the flavor decoherence-pointer-basis account (#2948). Source-only: 1 note + 1 runner + 1 cache. Status: `unaudited`.

The predictability sieve makes a generation sector **corner** if its mass spread ≫ `|K|`, **C₃** if ≪ `|K|`. The computation:

```
neutrino spread     ~ 0.05 eV
smallest other splitting (m_s - m_d) ~ 92 MeV
=> a ~9-order GAP. Window for |K|: [0.05 eV, 92 MeV].
```

- **Robust (no fine-tuning):** *any* `|K|` in that 9-order window gives neutrino → C₃ (large PMNS) and charged leptons + up + down quarks → corner (small CKM, `U_e=I`).
- **Explanation:** the mechanism predicts *only sub-`|K|` sectors are C₃* → **only the neutrino** (uniquely light). The framework's noticed-but-underived small-CKM/large-PMNS anti-correlation is **explained** by the neutrino's position in the gap.

## Changes
- `docs/EMERGENT_C3_COUPLING_SCALE_PREDICTABILITY_SIEVE_WINDOW_NARROW_THEOREM_NOTE_2026-06-06.md`
- `scripts/emergent_coupling_scale_window_runner.py` (class A; **PASS=7**)
- `logs/runner-cache/emergent_coupling_scale_window_runner.txt`

## Boundary (honest)
- A **constraint + robustness + explanation**, **not** a precise `|K|` prediction. The precise value (the emergent double-shift scale ~ `t²/E_gap`) is the open LATTICE+QUANTUM computation; the result is robust to it.
- Uses observed PDG masses (empirical comparator); the charged corner-basis (`U_e=I`, trichotomy) is unaudited; "why the neutrino is light" is a separate question.

## Test plan
- [x] Runner PASS=7 FAIL=0; cache regenerated (status ok)
- [x] Source-only triple; no audit-data files; no new axiom/import

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2965, opened 2026-06-06
science(flavor): emergent C3 coupling |K| -- naive 2nd order cancels; sourced by the hw-asymmetry

jonathonreilly:
## Summary

Pins the **structure and source** of the emergent C₃ coupling `|K|` (the last quantitative thread, after the 9-order-window note #2953). Source-only: 1 note + 1 runner + 1 cache. Status: `unaudited`.

`|K|` is the coefficient of the native double-shift C₃ coupling (`J − I`) on the generation triplet. The computation:

```
1st order: single-hop vanishes on hw=1 (P V P^T = 0)
2nd order, SYMMETRIC energies (E_n = n·ε): H_eff ∝ I  -> the J-I CANCELS -> |K|_naive = 0
   (hw=0 path +t²/ε·J  and  hw=2 path −t²/ε·(I+J)  cancel; same as the r=0 staggered no-go)
hw-ASYMMETRY (interaction shift δ, E_2 = 2ε+δ): cancellation broken -> |K| ~ t²·δ/ε², J-I form
```

So the precise `|K|` is **not** a naive hopping amplitude — it's **sourced by the hw=0-vs-hw=2 energy asymmetry** (interaction nonlinearity `δ`), the open emergent coupling. The load-bearing finding is the **cancellation**: don't estimate `|K|` from the naive single-hop (it's zero).

## Changes
- `docs/EMERGENT_C3_COUPLING_AMPLITUDE_SOURCED_BY_HW_ASYMMETRY_NARROW_THEOREM_NOTE_2026-06-06.md`
- `scripts/emergent_coupling_amplitude_source_runner.py` (class A; **PASS=7**)
- `logs/runner-cache/emergent_coupling_amplitude_source_runner.txt`

## Boundary (honest)
- A **structure + source** result, not a precise `|K|`: it shows `|K| ~ t²·δ/ε²` and that the naive (symmetric) computation cancels; it does **not** compute `δ`, `t`, `ε`.
- The flavor pattern is **robust** to the precise `|K|` (the 9-order window, #2953), so the open `δ` doesn't threaten the qualitative result.

## Test plan
- [x] Runner PASS=7 FAIL=0; cache regenerated (status ok)
- [x] Source-only triple; no audit-data files; no new axiom/import

### DIFF
--- a/docs/EMERGENT_C3_COUPLING_AMPLITUDE_SOURCED_BY_HW_ASYMMETRY_NARROW_THEOREM_NOTE_2026-06-06.md
+++ b/docs/EMERGENT_C3_COUPLING_AMPLITUDE_SOURCED_BY_HW_ASYMMETRY_NARROW_THEOREM_NOTE_2026-06-06.md
@@ -1,0 +1,91 @@
+# The Emergent C3 Coupling |K|: the Naive Second Order Cancels; |K| is Sourced by the hw-Asymmetry — Narrow Theorem
+
+**Date:** 2026-06-06
+**Claim type:** positive_theorem (the structure and source of |K|; precise value left open)
+**Status:** unaudited candidate. Graph-visible only so the independent audit lane can decide.
+**Primary runner:** [`scripts/emergent_coupling_amplitude_source_runner.py`](../scripts/emergent_coupling_amplitude_source_runner.py)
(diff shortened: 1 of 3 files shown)


jonathonreilly:

PR #2972, opened 2026-06-06
science: register-not-read — the RECORD registration map filters the S3-breaking direction out of the generation orbit

jonathonreilly:
## Summary
A bounded theorem (class-A finite, C³) that **corrects the "chirality import" framing** via the RECORD axiom (register-not-read): the framework's registration does not *carry* the S₃-breaking direction — it **filters it out**, through two independent, convergent channels.

## Theorem
- **T1** — `D(M)=Σ Pₖ M Pₖ` on the C₃-character decomposition registers only C₃-equivariant (circulant) content → `[D(M),S]=0`.
- **T2** — for any T-odd Hermitian `M` (`i·so(3)`), `D(M) ∝ A=i(C−C²)` (the unique C₃-invariant K-odd direction); the registered part of *every* T-odd operator is the S₃-symmetric A-line.
- **T3** — the S₃-breaking content `M−D(M)` is exactly the inter-sector coherence `P₁MP₂+h.c.`, which `D` **annihilates** (registration is blind to the S₃-breaking direction).
- **T4** — `K` (complex conjugation) swaps the doublet characters (`conj(P₁)=P₂`); with the C₃ cycle it generates the full S₃ (order 6), so `{χ₁,χ₂}` is **one K/CPT orbit** — the S₃-breaking direction is within-orbit data the RECORD K/CPT-orbit clause excludes. The realized coarsening `{χ₀}|{χ₁,χ₂}` **is** the `(1,1)` Koide partition.

**Conclusion:** the S₃-breaking direction is **unregistrable** — not a pre-record operator to derive (the realist slip), and **not an import**; the realized value is a **registered pattern** (matched like masses, G3). Convergence (not coincidence): the same K/CPT coarsening underlies the registered Koide weight `r`.

## Scope / discipline
- Corrects only the loose word "import" in the operator-route no-gos #2927/#2935 (those remain correct as operator-route statements).
- Does **not** force `r`/`Q`; does **not** derive the partition predicate (the standing G2 input); scoped to the registration map (does not foreclose other structure).
- Source-only: 1 note + 1 runner + 1 cache. Runner **14/14 PASS** (incl. 3×5000 random-trial checks). 3 one-hop deps verified at retained tiers. No new axiom/import; no audit-status promotion.


jonathonreilly:

PR #3023, opened 2026-06-07
[physics-loop] lensing-source-packet-inline exact-support

JONbridgerapps:
# Summary

Repairs the active `lensing_finite_path_explanation_note` restricted-packet blocker by inlining long-path source/cache checks into the primary finite-path runner.

The latest audit asked for `scripts/lensing_long_path_test.py` and its fresh runner cache/output so the `T_phys=7.5` measured slope `-1.4356` and finite-path prediction `-1.7336` can be verified inside the restricted packet. The primary runner now verifies those artifacts directly and reports `INLINE SOURCE PACKET: PASS=31 FAIL=0`.

# Scope

This is exact support for a packet-completeness blocker. It does not retag the audit ledger, does not close the layer-weighted detector-centroid derivation, and does not claim standard `1/b` lensing closure.

# Verification

```bash
python3 -m py_compile scripts/lensing_analytical_finite_path.py scripts/lensing_finite_path_centroid_packet_manifest_2026_06_04.py
PYTHONPATH=scripts python3 scripts/cached_runner_output.py --check-only scripts/lensing_analytical_finite_path.py
PYTHONPATH=scripts python3 scripts/cached_runner_output.py --check-only scripts/lensing_finite_path_centroid_packet_manifest_2026_06_04.py
git diff --check
```

### DIFF
--- a/.claude/science/physics-loops/lensing-source-packet-inline-20260606/ARTIFACT_PLAN.md
+++ b/.claude/science/physics-loops/lensing-source-packet-inline-20260606/ARTIFACT_PLAN.md
@@ -1,0 +1,12 @@
+# Artifact Plan
+
+Completed artifacts:
+
+- add inline source-packet checks to `scripts/lensing_analytical_finite_path.py`;
+- update `docs/LENSING_FINITE_PATH_EXPLANATION_NOTE.md` with the primary-runner inlining repair;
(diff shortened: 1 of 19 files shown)


jonathonreilly:
Landed/superseded by review-loop commit 86723e422 on main. Branch deleted where GitHub allowed.

PR #3115, opened 2026-06-07
science: register-not-read RE-INSTATES the Koide chirality no-go (corrects #2972)

jonathonreilly:
## Summary

The rejected route #2972 (`science/record-registration-filters-s3-breaking-2026-06-06`) claimed the RECORD registration map `D(M)=Σ P_k M P_k` *filters the S₃-breaking direction out*, dissolving the Koide chirality requirement. **This no-go proves the opposite.**

The C₃-character central-sector partition is **exactly the Γ_χ-eigenpartition** (Γ_χ=(2/3)J−I has eig {+1,−1,−1}; P₀→+1, P₁/P₂→−1). Therefore:
- **D commutes with Γ_χ** (both 3-block and 2-block partitions), so
- **D annihilates every Γ_χ-anticommuting operator**: {M,Γ_χ}=0 ⟹ D(M)=0.

The Q=2/3 carrier anticommutes with Γ_χ (retained `koide_anticommuting_operator_derivation_theorem`), so registration **never registers it** — it registers only the circulant Q=1 family. **Register-not-read re-instates the chirality no-go; it does not dissolve it.**

Corrects #2972's *inverse realist slip*: it correctly dismissed the pre-record operator as a reconstruction, then reified the C₃-character *partition* as A3-delivered. A3 supplies no decomposition (G1/G2) — the partition is a per-lane input. The residual is pinned to the K-reality partition-coarseness predicate (retained_bounded `flavor_einselection_2sector_modulo_kreality` GAP A), not a missing operator.

## Claim type
`no_go`. Sharpens, does not close — the T-odd-emergent route to K-reality (N1.6) is untouched and is the live derivation direction.

## Artifacts
- `docs/REGISTRATION_REINSTATES_CHIRALITY_NO_GO_NOTE_2026-06-07.md` — note with full N1–N8 discipline.
- `scripts/frontier_registration_reinstates_chirality_nogo.py` — class-A runner, **TOTAL: PASS=23 FAIL=0**, with adversarial controls (CTRL1: D registers circulant ops; CTRL2: a straddling non-C₃ partition *does* register the carrier — annihilation is partition-specific, not vacuous).
- paired cache.

## Discipline
- Deps verified retained fresh on origin/main: anticommuting_derivation (retained), z3_equivariant_no_go (retained_bounded), flavor_einselection (retained_bounded), circulant_character_bridge (retained).
- `r` untouched: H block-diagonal ∀r, D constrains it not at all (G3 dial).
[...]

PR #3124, opened 2026-06-07
science: K-reality (δ=0) ≡ C₃-symmetry of the realized past-hypothesis boundary (bounded, relocation)

jonathonreilly:
## Summary

Path-2 result from the max-agent T-odd attack on the Koide chirality gate. **No derivation of δ=0 survived** — but the gate was genuinely **relocated and sharpened** (all numpy-verified, runner 14/14).

The chirality gate reduces to: *why is the realized partition the K-real 2-block one (δ=0)?* (= `flavor_einselection` GAP A). The standing obstruction is that the emergent-time arrow's η-conjugation is conjugation-**even** (`koide_emergent_time_eta_conjugation_parity`, retained_bounded) — no spectral functional can select δ.

**This note relocates the gate to a non-spectral, registration-level statement the even no-go does not reach:**

> **K-reality (δ=0) ⟺ C₃-symmetry of the realized past-hypothesis boundary** — and that boundary symmetry is **not forced** from A1+A2+retained.

### Theorem (runner-verified)
- **T1/T2 relocation:** the T-odd arrow carrier Aₕₐₘ=i(T−Tᵀ)/2 is forced into the partition-blind line span{A=i(C−C²)} on a C₃-symmetric history ([A,S]=0 → K-reality on the whole cone, even no-go re-derived); partition-selective on a C₃-breaking history.
- **T3–T5 arrow-vs-symmetry dichotomy:** I/3 is C₃-symmetric but the no-arrow fixed point (Aₕₐₘ=0); the A1-reachable single-axis |e_k⟩ bears an arrow but is C₃-broken & selective; **register-collapse** D(|e_k⟩⟨e_k|)=I/3 (registration symmetrizes but discards the arrow). Not the forbidden S₃ average.
- **T6 parity-mismatch:** every T-odd functional is null at δ=0 (Tr(A·M)=6|b|sin δ), so the full weight sits on the conjugation-**even** sub-target — the unbuilt boundary-symmetry datum.

## Claim type
`bounded_theorem` (relocation + no-forcing). **Sharpens, does not close** — the new residual is a sharp arrow-vs-C₃-symmetry tension on the one realized boundary (a past-hypothesis question). Derives neither δ nor r (G3 untouched).

## Artifacts
- `docs/KOIDE_KREALITY_EQUALS_BOUNDARY_C3_SYMMETRY_BOUNDED_THEOREM_NOTE_2026-06-07.md`
- `scripts/frontier_kreality_equals_boundary_c3_symmetry.py` — class-A runner, **TOTAL: PASS=14 FAIL=0**, with 3 adversarial controls (teeth).
- paired cache.

## Discipline
- Deps verified retained fresh on origin/main: anticommuting_derivation (retained), z3_equivariant_no_go (retained_bounded), emergent_time_eta_conjugation (retained_bounded), flavor_einselection (retained_bounded), arrow_from_record_formation (retained_bounded), circulant_character_bridge (retained).
- Companion: `registration_reinstates_chirality_no_go` (PR #3115).


jonathonreilly:
Review-loop reject as-is: the K-reality equals boundary C3-symmetry relocation uses branch-local A1/A2/A3/G* vocabulary and imports a past-hypothesis/record-boundary move that is not cleanly supplied by the current Record baseline. It needs a fresh theory-layer rewrite after the Record guardrail landed in bd74d87c6. No source salvage landed.

PR #3182, opened 2026-06-08
science: K/CPT-orbit does not supply the temporal factor of 2 (narrow no-go)

jonathonreilly:
## Summary

A **narrow, negative-route-pruning no-go**. The Record axiom's **K/CPT orbit** does **not** supply the temporal factor of 2 in the staggered determinant magnitude — the `8 -> 16` that lifts the native `alpha_LM^8` to the Euclidean `alpha_LM^16` in `v = M_Pl (7/8)^{1/4} alpha_LM^16`.

This closes the tempting escape opened by the P2 fork (#2748) and the native-Lorentzian test (#2744, NATIVE-GIVES-DIFFERENT): *"maybe the K/CPT orbit's native 2-fold is the missing temporal corner."* It is not.

## Four independent obstructions (each individually sufficient)

| | obstruction | basis |
|---|---|---|
| **A** | K/CPT is a **spatial, count-PRESERVING** Z2 (P, C, K permute the 8 BZ corners among themselves); the temporal factor is a **count-DOUBLING** Z2 (`k4 in {0,pi}`: 8→16) | retained: `staggered_dirac_substep3_bz_corner_hamming_orbit`, `naive_lattice_fermion_two_power_d` |
| **B** | the K/CPT involution is **antiunitary**; the temporal-corner exchange is **unitary** | retained_bounded: `cpt_exact_real_anti_hermitian_d` |
| **C** | A3's finite-additive readout over the realized **orbit-as-one-record** gives multiplicity **1**; the ×2 needs the **occupancy rule A3 disclaims verbatim** | A3 axiom text (`MINIMAL_AXIOMS_2026-06-05`) |
| **D** | the only native-Lorentzian temporal Z2 (**Keldysh / thermofield**) **cancels** in `\|det\|` — `det·det⁻¹` = the Keldysh `Z=1` — opposite to the Euclidean Matsubara product | re-derived first-principles in the runner; Kamenev / Bisognano–Wichmann as corroboration |

## What it does NOT foreclose (scope)

- **Not** the factor of 2, the `16`, or the `v`-match — the Euclidean temporal product supplies them; the open Wick primitive **P2** remains the live route.
- **Not** a future native d=3+1 mechanism furnishing a *temporal, unitary, count-doubling* Z2 surviving in `|det|`. The note **specifies exactly what such a mechanism must be** and prunes only the K/CPT-orbit route to it.
[...]

PR #3193, opened 2026-06-08
science: magnitude temporal factor of 2 is a COUNT not a RATE (wall moves)

jonathonreilly:
## Summary

**Walls move.** A 3-lens panel had parked the "native transfer-matrix" route to the magnitude's temporal factor of 2 behind the retained no-go `POST_RECORD_CLOCK_RATE_INTERFACE` (records give order/count, not a clock rate). This PR shows that no-go is **scoped to the clock RATE/metric**, and the magnitude's temporal factor is a **transfer-step COUNT** — which the same no-go **explicitly supports** records carrying.

The no-go's own "What this unlocks" section: *"Rows that only need event order, prefix preservation, finite length, **counts** ... can cite the exact post-record stack"*, and names **"transfer-step count"** / **"lattice step"** as legitimate denominators. The magnitude exponent `16 = 8 (spatial) × 2 (temporal)` is exactly such a count.

## Bounded theorem (runner 15/15)

| | claim |
|---|---|
| **T1** | the exponent `8·Lₜ` is the staggered-determinant **mode count** (matrix dimension): scales with `Lₜ`, **independent of the hopping rate `u₀`**. The per-mode value `α_LM` (rate) depends on `u₀` — that's the separate `DELTA0` gate, not at issue. |
| **T2** | a transfer-step **count** is in the no-go's **supported** zone; the **rate** is what it forbids. So the no-go does not block this route. |
| **T3** | the minimal reflection-positive temporal block is **2** (staggered phase `η₁(t)=(−1)ᵗ` has period 2; 2-step `T̂²` is positive — `axiom_first_rp_two_step_transfer_matrix_positivity`, retained_bounded). Native reason the temporal count is 2. |
| **T4** | the **count-2 survives** the OS energy normalization `H=−ln(T̂²)/(2a_τ)` that divides the **rate-2** out (rescaling eigenvalues ≠ changing their number). This reconciles the panel's "extent-2 dissolves" (the rate-2) with the surviving count-2. |

## Reconciliation with the three prior negatives (all consistent)

- `P2_KCPT_ORBIT_TEMPORAL_FACTOR_NO_GO` (#3182): untouched — uses the staggered transfer count, not the K/CPT orbit.
- the transfer-matrix panel: T4 reproduces its rate-2 dissolution exactly; the **count**-2 is a different object the panel did not separate out.
- `POST_RECORD_CLOCK_RATE_INTERFACE` (retained_no_go): consistent and load-bearing — the magnitude needs the count it supports, not the rate it forbids.

## Scope (honest)

- Does **not** derive the magnitude `v` or close the hierarchy gate. `α_LM` (the `DELTA0` magnitude) is a separate open gate.
- The named residual — that the magnitude is read at the per-record/UV minimal-block scale (`Lₜ=2`) rather than the OS continuum — is a **readout-scale** question (UV-per-record vs emergent-continuum), **not** a clock-rate question, so the retained no-go does not bear on it. The frontier moves from "behind a retained no-go" to this tractable, unblocked residual.
- Does **not** contradict or demote the clock-rate no-go; it uses that no-go's own supported zone.

## Files
- `docs/MAGNITUDE_TEMPORAL_FACTOR_IS_COUNT_NOT_RATE_2026-06-06.md` (claim_type=bounded_theorem)
- `scripts/magnitude_temporal_factor_is_count_not_rate_2026_06_06.py` — **15/15 PASS**
- `logs/runner-cache/magnitude_temporal_factor_is_count_not_rate_2026_06_06.txt`

All cited rows verified on the live ledger. No import. **Status authority: independent audit lane only.**

### DIFF
--- a/docs/MAGNITUDE_TEMPORAL_FACTOR_IS_COUNT_NOT_RATE_2026-06-06.md
+++ b/docs/MAGNITUDE_TEMPORAL_FACTOR_IS_COUNT_NOT_RATE_2026-06-06.md
@@ -1,0 +1,168 @@
+# The Magnitude's Temporal Factor of 2 is a Transfer-Step COUNT, Not a Clock RATE
+
+**Date:** 2026-06-06
+**Type:** boundary correction / bounded positive route
+**Claim type:** bounded_theorem
+**Status:** branch-local bounded; the magnitude-count temporal factor is shown to
+sit in the SUPPORTED (count) zone of the retained clock-rate no-go, not its
+forbidden (rate) zone. Sets no audit status; audit lane owns final classification.
+`audit_required_before_effective_retained=true; bare_retained_allowed=false`.
+**Runner:** [`scripts/magnitude_temporal_factor_is_count_not_rate_2026_06_06.py`](../scripts/magnitude_temporal_factor_is_count_not_rate_2026_06_06.py)
+(`TOTAL: PASS=15 FAIL=0`).
+**Cached log:** `logs/runner-cache/magnitude_temporal_factor_is_count_not_rate_2026_06_06.txt`
+
+## Background
+
+The electroweak/lepton magnitude `v = M_Pl (7/8)^{1/4} alpha_LM^16` carries the
+exponent `16 = 8 (spatial Z^3 corners) x 2 (temporal)`. The native-Lorentzian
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed manually via review-loop to main (commits a0ff66c6e, e0b0eefa8, 8ddc1c52f, 5a68397c9). Closing the source PR.

PR #3214, opened 2026-06-08
science: scalar generator metric-Hessian is rank-1 longitudinal, TT graviton block in exact kernel (sharpens universal-GR polarization-frame blocker)

jonathonreilly:
## Summary

Went after the **live universal-GR blocker** (the operator-identification turning the localized Hessian `D²W` into the Einstein/Regge tensor law). A focused, memory-safe opus investigation (5 angles + adversarial verify) found **no escape** but produced a decisive **sharpening** with a clean positive theorem.

**Decisive theorem (verified):** a metric source enters `det(D+J)` only through the `O_h`-scalar `s(q)=g_ij q̂^iq̂^j` (linear in `g`), so the per-mode metric-Hessian of the scalar generator `W=log|det(D+J)|` is **rank-1 longitudinal** `(q̂q̂)⊗(q̂q̂)` — hence the **transverse-traceless (spin-2 graviton) block is in its exact kernel** (`|⟨h_TT|H|h_TT⟩| < 1e-16` over thousands of modes; holds for *both* the scalar lattice-Laplacian symbol and the `Cl(3)`-Dirac det `m₀²−q̂ᵀgq̂`).

So the scalar `W`-route **provably cannot couple** to the spin-2 polarization where the GR trace/shear split lives → the supermetric trace/shear degeneracy **cannot be lifted from `W` alone**. This pins the *structural reason* for the blocker.

**Why the native lattice doesn't rescue it:** the graph-Laplacian York/TT projector *does* split spin-0/spin-2 but is **frame-dependent** (`‖P_TT(z)−P_TT(111)‖=1.56`); the isotropic background has no canonical propagation direction, and the sphere-average collapses to `(2/5)P_traceless` (zero spin-2 info). `SO(3)` isotypy *does* fix the trace-vs-shear **label** (commutant dim 2) — a partial correction to the blocker's "no canonical section" wording.

**Re-scoped missing primitive:** from "a covariant polarization frame" to **a spin-2-coupled, two-derivative curvature generator** (the cubic-Coxeter Regge `δ²` supplies the structure but isn't `W`-derived, and leaves the gluing + temporal channels open).

## Claim type
`bounded_theorem` — a **sharpening** of retained_bounded `universal_gr_polarization_frame_bundle_blocker`. **Not an escape** (the blocker's conclusion holds). Runner **8/8**.

## Discipline
- Deps verified retained fresh: polarization_frame_bundle_blocker (retained_bounded), supermetric_normal_form (retained), invariant_frame_obstruction (retained_bounded), cubic_coxeter_regge_deficit (retained, method-context only).
- Regge/Lichnerowicz machinery used as method-context comparator, **not** imported. No registered scale touched. Memory-safe finite tensor algebra (no heavy harness — post-OOM discipline).


jonathonreilly:
Landed manually via review-loop to main (commits a0ff66c6e, e0b0eefa8, 8ddc1c52f, 5a68397c9). Closing the source PR.

PR #3224, opened 2026-06-08
fix(audit-graph): break cycle-0001 (reflection-positivity ↔ wilson-temporal repair) to unblock ~1038-row cascade

JONbridgerapps:
## The blocker

The audit is **globally stalled** — the dispatch queue is empty. Diagnosis: the foundational cluster (reflection-positivity → microcausality/Lieb-Robinson → single-clock → anomaly-forces-time → matter sector, ~1000 rows) is **cycle-locked**. The single biggest cycle (`cycle-0001`, `max_transitive_descendants = 1038`) is a 2-cycle:

`axiom_first_reflection_positivity_theorem` (2026-04-29) ↔ `rp_wilson_temporal_gauge_bridge_sign_and_positivity_repair` (2026-06-06)

Neither can ever be "deps-all-retained," so neither dispatches, so the whole cluster stalls.

## Why the April→June edge is spurious (not graph-falsification)

It's a **forward citation**: the 2026-04-29 theorem cites a 2026-06-06 repair note, for the **conditional Wilson-plane addendum** (added 2026-06-07, "conditional on independent audit of the companion bridge note"). The April theorem's **core** bounded claim — the staggered-only 2-step transfer-matrix positivity — is *derived in-repo from first principles* (`axiom_first_rp_two_step_transfer_matrix_positivity`) and depends on **none** of the June repair; its 4 other deps are all retained-grade. The genuine logical direction is **repair → builds-on → theorem**.

## Fix (cycle-break repair pattern)

Render the April→June reference as a plain (non-link) citation instead of a load-bearing markdown dependency, and document why (it supports only the conditional addendum, not the core). The conditional Wilson-plane claim still travels with the companion note's own independent-audit status.

## Validated (pipeline run locally, then reverted)

After rebuild: April's deps no longer include the June repair; **April is DISPATCH-READY** (deps-all-retained); `cycle_count` 20 → 19; the RP↔wilson cycle is gone. This makes the cluster **root** auditable and lets the ~1000-row bottom-up cascade proceed. **No audit-lane data is committed** (graph/seed/cycle steps run only to validate, then `git checkout` reverted).

### DIFF
--- a/docs/AXIOM_FIRST_REFLECTION_POSITIVITY_THEOREM_NOTE_2026-04-29.md
+++ b/docs/AXIOM_FIRST_REFLECTION_POSITIVITY_THEOREM_NOTE_2026-04-29.md
@@ -158,7 +158,7 @@ gauge-case reduction, plus the single-step no-go runner:
   Wilson-plaquette boundary closure.
 - The **Wilson-plane sign and character-kernel source packet** is supplied
   by
-  [RP_WILSON_TEMPORAL_GAUGE_BRIDGE_SIGN_AND_POSITIVITY_REPAIR_NOTE_2026-06-06.md](RP_WILSON_TEMPORAL_GAUGE_BRIDGE_SIGN_AND_POSITIVITY_REPAIR_NOTE_2026-06-06.md)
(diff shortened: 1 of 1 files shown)


jonathonreilly:

PR #3228, opened 2026-06-08
science: lepton phase/modulus separation no-go (r and delta on separate C3-invariants)

jonathonreilly:
## Summary

From the δ=2/9 **fresh-angle hunt** (workflow `wa0o7fje5`, explicit AVOID-list of the worked routes). **No fresh *derivation* of δ=2/9** — every exact-2/9 object collapses to the AVOID-list or fails nativeness (incl. the striking find that the Hirzebruch G-signature defect of L(3;1) is *exactly* −2/9, but it's non-native + APS-η + already on main). The genuinely-new, native, off-AVOID-list result is a **no-go**:

## The separation (exact sympy)

For `√m_k = a(1 + 2√r·cos(δ + 2πk/3))`, the elementary symmetric functions **separate r from δ**:

| | |
|---|---|
| `e₁ = 3a` | scale |
| `e₂ = 3a²(1−r)` | **δ-blind** — carries the *entire* Koide content (`r=1/2 ⟺ Q=2/3`) |
| `e₃ = a³(1−3r+2r^{3/2}cos3δ)` | the **only** carrier of δ, solely via `cos3δ` |

So any real C₃-invariant scalar action `W(|z|², Re z³)` has `∂W/∂δ ∝ sin(3δ)` → δ∈{0, π/3, 2π/3, …}. **δ=2/9 is not among them.** A scalar/variational action can fix the modulus r (→r=1/2) but **provably cannot stationarize the phase δ at 2/9.**

## Convergence + consequence

This lands on the **same object** as this session's register-not-read result, from the opposite side: the partition map delivers the weight ratio `r`, never the within-block phase `δ`. So the (r,δ) plane has exactly **two native classes** — r = modulus (scalar/variational/partition → r=1/2), δ = phase (C₃-**covariant** η/holonomy only). 

**δ's only native home is the C₃-covariant η/holonomy = the chirality/orbit-splitting gate — the *same gate* as Koide-Q and generation-ID.** δ=2/9 is **not an independent target**; the lepton-phase gate and the chirality gate are one gate, not two.

## Scope (no-go discipline, full N1–N8)

- **Prunes** the scalar/variational/partition route to δ only.
- Does **not** close δ=2/9 — the C₃-covariant η/holonomy (= chirality) route is untouched/open.
- N7 steelman refuted exactly: *any* `W(|z|²,Re z³)` has δ-dependence solely through `cos3δ`, so `∂_δ ∝ sin3δ` regardless of W's form; no scalar choice hits 2/9.

## Files
- `docs/LEPTON_PHASE_MODULUS_SEPARATION_NO_GO_2026-06-06.md` (claim_type=no_go, negative_route_pruning)
- `scripts/...py` — **13/13 PASS**, exact sympy
- `logs/runner-cache/...txt`

No import; cited rows verified on the live ledger. **Status authority: independent audit lane only.**

### DIFF
--- a/docs/LEPTON_PHASE_MODULUS_SEPARATION_NO_GO_2026-06-06.md
+++ b/docs/LEPTON_PHASE_MODULUS_SEPARATION_NO_GO_2026-06-06.md
@@ -1,0 +1,167 @@
+# Charged-Lepton Phase/Modulus Separation Narrow No-Go
+
+**Date:** 2026-06-06
+**Type:** no_go
+**Claim scope:** the standalone structural fact that, for the Brannen/circulant
+charged-lepton sqrt-mass spectrum `sqrt(m_k) = a(1 + 2 sqrt(r) cos(delta + 2 pi
+k/3))`, the Koide modulus `r = |b|^2/a^2` and the Brannen phase `delta = arg(b)`
+live on **separate `C3`-invariants** — so a real `C3`-invariant **scalar action**
+can fix `r` (the modulus, `r=1/2 <=> Q=2/3`) but **provably cannot stationarize
+`delta` at `2/9`** (its stationary phases are forced to the `C3`-rational
+directions `delta in {0, pi/3, 2pi/3, ...}`). This **prunes** the
+scalar/variational/partition route to `delta`; it does **not** close `delta = 2/9`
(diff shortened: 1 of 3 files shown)


jonathonreilly:

PR #3229, opened 2026-06-08
fix(audit-graph): break the staggered chirality-parity audit 2-cycle (unblocks the staggered branch)

JONbridgerapps:
## The cycle

Second foundational-cluster cycle-break (after cycle-0001 / #3224). The 2-cycle (315 descendants):

`staggered_dirac_chirality_parity_bridge` (2026-06-06) ↔ `staggered_dirac_kawamoto_smit_conditional_realization_rescoping_companion` (2026-06-03)

gates the **staggered branch** of the stalled foundational cluster (→ kawamoto-smit forcing → anomaly-forces-time → matter sector).

## Verified direction (opposite of cycle-0001)

- `conditional → bridge` is **genuinely load-bearing** and is **kept**: the conditional note's current "Rescoped Statement" is *given the chirality-parity bridge*, and its 2026-06-06 repair replaced a free `H_staggered_chirality` premise **with** the bridge's `ε(x)` result.
- `bridge → conditional` is the **spurious** edge: the bridge cites the conditional note only in its "Targeted Blocker" section as the **consumer it serves**. The bridge's theorem — `ε(x)=(−1)^(x₁+x₂+x₃)` is the unique scalar edge-flip grading on the Z³ coordinate graph — is self-contained and consumes none of that note.

(I verified this by reading both notes, not by the date heuristic alone — exactly as I did to *reject* cutting the genuine `observable_principle` cycle.)

## Fix + validation

Render the bridge's consumer-reference as a plain (non-link) citation. **Validated** (pipeline run locally, reverted): the bridge's deps become empty → **DISPATCH-READY**; the conditional dep is gone; the cycle is eliminated; the genuine `conditional → bridge` dependency is preserved. **No audit-lane data committed.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3235, opened 2026-06-08
fix(audit-graph): break observable_principle parent↔p2 cycle by lemma extraction (unblocks +722 rows → ~98% with the other two)

JONbridgerapps:
## The cycle (the largest remaining structural blocker)

`observable_principle_from_axiom_note` ↔ `observable_principle_positive_source_cone_p2_elimination` (2026-06-06) — a 2-cycle gating **+722 rows**. Unlike cycle-0001/staggered, this one is **genuine both ways** (neither edge is a clean cut): the parent needs p2's `det(D+J)∈ℝ₊` fact for its P2-vacuity step, and p2's proof frames itself on "the parent's derivative identities." Cutting either edge would falsify the graph.

## Fix: lemma extraction (honest, no falsification)

The parent needs only one **self-contained** fact, not the whole p2 application note. New lemma `REAL_DIAGONAL_SOURCE_DET_POSITIVITY_AND_LOG_READOUT_LEMMA` (runner **4/4**):

- **L1** `det(S+D) > 0` for D real-antisymmetric + S real-positive-diagonal (eig(B) imaginary ⇒ `det(I+B)=∏(1+λ²)>0`) — strictly positive real, **no phase**.
- **L2** sign-constant on the Neumann derivative patch (`‖D⁻¹J‖<1`) ⇒ `det(D+J)∈ℝ₊`.
- **L3** det multiplicative over disjoint blocks + Record additivity ⇒ `c·log det` readout.

Pure linear algebra + the Record premise; **consumes none of the parent** (deps = `key_terminology`, `minimal_axioms`).

`OBSERVABLE_PRINCIPLE_FROM_AXIOM` is repointed: its load-bearing det-positivity citation now → the lemma (one-directional), with the fuller p2 application kept as a **plain (non-load-bearing) reference**.

## Validated (pipeline run locally, reverted)

`from_axiom` deps become `{minimal_axioms, lemma}` (no p2 edge); the lemma is dispatch-ready; the `observable_principle` cycle is gone (cycle_count 20→18). Combined with the reflection-positivity (#3224) and staggered (#3229) breaks, the cascade simulation reaches **~2932/2993 rows (98%)** eventually auditable. **No audit-lane data committed.**

### DIFF
--- a/docs/OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md
+++ b/docs/OBSERVABLE_PRINCIPLE_FROM_AXIOM_NOTE.md
@@ -396,10 +396,13 @@ phase premise on the consumed finite source surface:
 - The runner `frontier_hierarchy_observable_principle_from_axiom.py`
   verifies the exact-algebra closure for the selected candidate generator on
   the finite staggered source block.
-- The positive-source-cone bridge
-  [`OBSERVABLE_PRINCIPLE_POSITIVE_SOURCE_CONE_P2_ELIMINATION_NARROW_THEOREM_NOTE_2026-06-06.md`](OBSERVABLE_PRINCIPLE_POSITIVE_SOURCE_CONE_P2_ELIMINATION_NARROW_THEOREM_NOTE_2026-06-06.md)
+- The det-positivity lemma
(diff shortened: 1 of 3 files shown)


jonathonreilly:

PR #3327, opened 2026-06-09
review-loop: note_hash drift is a re-audit-pending notice (non-retained), hard error only for retained rows

JONbridgerapps:
## The nag

Every fresh-from-`main` branch was hitting a growing pile of blocking `audit_lint --strict` `note_hash mismatch` ERRORs. Root cause: review-loop "Review …" repairs edit claim notes (add audit front-matter, formal-carrier repair sections), which drifts their **source-content** `note_hash` from the seeded value — and the review-loop discipline (correctly) forbids committing audit-lane ledger churn to refresh it, so the drift could only accumulate. The linter hard-errored on *every* such benign drift.

## Fix

**`note_hash` is a source-content hash, not an audit verdict.** The linter's own docstring already says re-audit-required items should be **notices** so strict lint stays useful — the `note_hash` check just never followed it.

- **`audit_lint.py`:** a `note_hash` mismatch is now a non-blocking **`note_hash_drift_reaudit_pending` notice** when the row is **not** retained-grade (`unaudited` / `audited_conditional` / pending) — re-audit is pending and the nightly audit-lane re-seed refreshes the hash. It stays a hard **error** only for a **retained-grade** row (`retained` / `retained_bounded` / `retained_no_go`): an edited *retained* note laundered past a stale ratification is a real integrity violation.
- **`review-loop/SKILL.md`:** documents the principle — non-retained drift is a benign notice that does **not** block PASS and must **not** be "fixed" by committing ledger churn; retained-grade drift is resolved by re-audit, never by refreshing the hash while keeping the retained verdict.

## Effect (on current `main`)

Converts **15 benign review-loop `note_hash` drifts → non-blocking notices** (the nag) and correctly **isolates exactly 1 genuine retained-grade drift** — `fifth_family_complex_boundary_note` (`audited_clean` / `retained_bounded`, **semantically** edited 2026-06-08 without re-audit: the load-bearing anchor row changed `drift=0.05,seed=1 → 0.20,seed=0`). Previously buried among the mass; now correctly flagged as a hard error that blocks until re-audited.

That last error is the fix **working as intended** — it surfaces a real retained-laundering. Resolving it is an audit-lane verdict reset (re-seed archives + resets it), out of scope for this tooling PR; flagged for follow-up. Strict lint goes from many blocking note_hash errors → 1 genuine one.

Tooling/skill fix only — **no audit-lane data committed**, no science change. `audit_lint.py` parses; behavior verified (15 notices + 1 retained error).

### DIFF
--- a/docs/ai_methodology/skills/review-loop/SKILL.md
+++ b/docs/ai_methodology/skills/review-loop/SKILL.md
@@ -674,6 +674,22 @@ git diff --check
 The known graph-cycle warning is acceptable. Any strict-lint error blocks a
 review-loop PASS.
 
+**`note_hash` drift is a notice for non-retained rows, an error only for
+retained-grade rows.** `note_hash` is a *source-content* hash, not an audit
+verdict. When a review fix edits a claim note, its `note_hash` drifts from the
+seeded value; `audit_lint.py` reports this drift as a non-blocking
+`note_hash_drift_reaudit_pending` **notice** when the row is not retained-grade
+(`unaudited` / `audited_conditional` / pending), because re-audit is simply
+pending and the audit-lane re-seed refreshes the hash. Such a notice does **not**
+block review-loop PASS and must **not** be "fixed" by committing audit-lane
+ledger churn (that is the forbidden pipeline-output commit). A `note_hash`
+mismatch on a **retained-grade** row (`retained` / `retained_bounded` /
+`retained_no_go`) stays a hard **error**: an edited retained note laundered past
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed to main as a review-loop tooling fix. Non-retained note_hash drift is now a re-audit-pending notice; retained-grade drift remains a hard lint error. Verified by py_compile plus full audit pipeline and strict lint after seeding/invalidation; generated audit authority data was not committed.

PR #3330, opened 2026-06-09
strong-CP theta_gauge: O_h/parity route -- measure claim corrected, gate sharpened to the color<->orientation-Z2 coupling

JONbridgerapps:
## Going after the field-grade target (θ_gauge=0)

Result (runner **7/7** + a salvaged 6-lens forcing red-team, **all 6 → NOT forced**): θ_gauge=0 is **not forced**, but the residual is now **corrected, reframed, and sharply pinned** — a much tighter, parity-framed result than the prior "minimality" gate.

## A correction (verified)
The exercise claimed O_h-invariance doesn't forbid F̃F because "the measure `det(R)` cancels the F̃F `det(R)`." **Refuted:** the global θ-term `Σ_x Q[F_x]` is P-odd (`Q[R·F]=det(R)Q[F]`, all 48 O_h), the lattice sum is a relabeling (no Jacobian), and the volume measure is **det-even** (`|det R|=1`) — so an O_h/parity-invariant action **forbids** F̃F. The landed `EPSILON_PSEUDOTENSOR_OH_SIGN` B6 is right.

## Reframing + parity-solution class
θ_gauge=0 is a **parity** question, not a diffuse "minimality" one. Color is **vectorlike/P-even** (`[Γ5,color]=0`; Wilson `Re Tr U_P` parity-invariant); only weak `su(2)_L` is chiral. The matter determinant is **K-real** (`arg det M_q ∈ {0,π}`). So the framework sits in the **parity-solution class**, with the K-real determinant blocking the usual θ̄ regeneration.

## The gate (no-go)
θ_gauge=0 needs **parity**-invariance of the color action (proper rotations leave `Q[F]` invariant; only improper/reflection elements forbid it). But the Cl(3) volume element `ω=σ₁σ₂σ₃=iI` (the i-gate's native pseudoscalar = `sign(Vandermonde)` orientation Z₂) transforms by `det(R)` under all 48 O_h → **the lattice orientation is a framework-native P-source for F̃F**. So "color has no P-source → P-even action" is the EFT-naturalness assumption *"respect substrate symmetry absent a source"* — **not derived**. θ_gauge is the un-derived color coupling to the lattice-orientation Z₂ — the same β=6-class wall, now parity-framed and unified with the i-gate orientation Z₂ (which also sets the δ-sign / arg det sign).

**Verdict:** not a closure; a **corrected + reframed + sharpened parity-gate**. No axiom, no imported value. Runner `scripts/strong_cp_theta_gauge_parity_route_2026_06_08.py` (exact numpy, 7/7).

_Note: `audit_lint --strict` shows pre-existing `note_hash`-drift errors on `main` (review-loop seeding hygiene) — unrelated to this PR; addressed by PR #3327 (linter reclassifies non-retained drift as a notice)._

### DIFF
--- a/docs/STRONG_CP_THETA_GAUGE_PARITY_ROUTE_MEASURE_CORRECTED_GATE_SHARPENED_NO_GO_NOTE_2026-06-08.md
+++ b/docs/STRONG_CP_THETA_GAUGE_PARITY_ROUTE_MEASURE_CORRECTED_GATE_SHARPENED_NO_GO_NOTE_2026-06-08.md
@@ -63,10 +63,11 @@ emergent-Lorentz-forces-parity) **unanimously failed to force `θ_gauge = 0`**, 
 one recurring reason:
 
 > **The lattice orientation is a framework-native P-source for `F̃F`.** The Cl(3)
-> volume element `ω = σ₁σ₂σ₃ = i·I` (the i-identity gate's native pseudoscalar =
-> `sign(Vandermonde)` orientation `Z₂`) transforms by `det(R)` under all 48 O_h
-> actions — the **same** det-odd character as `F̃F`. So `F̃F` is **sourced** by the
-> native lattice orientation, not "unsourced."
+> volume element `ω = σ₁σ₂σ₃ = i·I` (the i-identity gate's `K`-odd spatial/qubit
+> pseudoscalar — a *distinct* object from the `K`-even generation `sign(Vandermonde)`;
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed a narrowed review-loop salvage to main. I kept the measure-cancellation correction and the sharpened parity/orientation gate, but did not land the stronger parity-solution-class or Record-derived mass/K-real posture. The landed row is no_go/unaudited: parity-invariant color action would forbid the determinant-odd theta slot, but the minimal axioms do not derive zero coupling to the lattice-orientation character. Runner PASS=7 FAIL=0; pipeline and strict lint had no errors.

PR #3335, opened 2026-06-09
publication: admission-floor unification synthesis (structure derived, couplings admitted; one shared missing primitive)

JONbridgerapps:
## Part (b): the Nature-facing synthesis

A **manuscript-surface** synthesis (per the "meta-framings land backward" discipline — it lives on the publication surface, is backward-looking narrative, and is **not a new audit claim**: `proposal_allowed: false`, no new axiom/primitive/vocabulary, no status upgrades).

It consolidates the session's unified picture into one honest statement:

- **Derived = the SM *structure*** (gauge algebra given the carrier, generation pattern, anomaly cancellation, emergent spatial isotropy ratio) — with **honest per-component status** (retained / bounded / in-review), not a blanket "derived."
- **Admitted = the floor**: AC_φλ (`r=1/2`, `δ`), `θ` (SM-shared), `ξ=1` (full Lorentz beyond the conformal ratio), + universal-floor **scale** + **arrow**.
- **The unification:** `r`, `δ`, `θ`, `ξ` are four faces of **one shared missing primitive** — the un-derived CP-odd / coupling sector of the Record-forced action, tied to a single **orientation `Z₂`** and the **clock-rate / arrow**. (Record forces the action *form-class*, not its *coefficients*; no axiom-native CP-source.)
- **The honest claim:** *structure from three axioms; the residual is one primitive, not many, and it is the same primitive the field also lacks* (flavor texture + strong-CP + universal-floor scale/arrow).

**Honesty guards:** it explicitly flags that several per-item characterizations (Bridge-1 #3334, no-axiom-native-CP #3312, strong-CP parity #3330, i-identity gate #3323) are **proposals in active review**, not landed retained results, and that "one shared missing primitive" is a *characterization* of the admission, not a derivation of it. No row's status is changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3337, opened 2026-06-09
chiral content = the {ε,D}=0 chirality import; distinct recurring import (not the orientation Z₂) + corrects an over-unification

JONbridgerapps:
## The chiral-content admission, and a correction it forces

Attacks the chiral-content admission (why `su(2)_L` gauges `P_L` — parity violation) and tests the unification question: does it collapse into the orientation `Z₂`?

**Finding (runner 5/5):** No — and verifying that **catches an over-unification** in two open PRs.

- The chiral content is **already consolidated** to one gate `{ε,D}=0`, `ε(x)=(−1)^{x+y+z}` (`CARRIER_ATTACHMENT`: chirality + r=1/2 pin + half-integer carrier + generation-id), and Record is a *consumer*, not a chirality source (`CHIRALITY_RECORD_TYPING`).
- **Distinctness (the answer):** `ε` (K-even, *spatial*), `ω=σ₁σ₂σ₃=iI` (**K-odd**, *qubit* pseudoscalar), and `sign(Vandermonde)` (K-even, *generation*) are **three distinct objects** — different K-parity *and* sector. `ω` is K-odd while the other two are K-even, so they **cannot** be the same `Z₂`. The chiral content does **not** collapse into the gauge/flavor "orientation `Z₂`."
- **Correction:** the loose "one orientation `Z₂` = `ω` = `sign(Vandermonde)`" in the strong-CP parity note (**#3330**) and the synthesis (**#3335**) is an over-unification. The honest picture: the floor is a **small set of distinct dynamical-structure imports** `{the {ε,D}=0 chirality gate, the CP-odd/coupling sector, the scalar i, the scale, the arrow}`, unified only by the theme *"Record forces the action **form**, not the couplings/gradings"* — not literally one `Z₂`. (Companion fixes to #3330/#3335 follow.)
- The chiral **gauging** (chiral-vs-vector = the `r` binary) is the un-derived import `ε` is blind to (`[ε,T^a]=0`; dressings anticommute identically).

`claim_type=bounded_theorem` (finite-algebra distinctness) + synthesis correction. No axiom, no fitted value.

### DIFF
--- a/docs/CHIRAL_CONTENT_IS_THE_EPSILON_D_CHIRALITY_IMPORT_DISTINCT_FROM_ORIENTATION_NARROW_THEOREM_NOTE_2026-06-08.md
+++ b/docs/CHIRAL_CONTENT_IS_THE_EPSILON_D_CHIRALITY_IMPORT_DISTINCT_FROM_ORIENTATION_NARROW_THEOREM_NOTE_2026-06-08.md
@@ -1,0 +1,95 @@
+# The Chiral-Content Admission Is the `{ε,D}=0` Staggered Chirality Import — a Distinct Recurring Import, Not the Gauge/Flavor Orientation `Z₂` (Narrow Theorem + Synthesis Correction)
+
+> **Key terms used in this doc** are indexed A-Z at
+> [docs/KEY_TERMINOLOGY.md](KEY_TERMINOLOGY.md); each row points to the
+> canonical source-of-truth doc.
+
+**Date:** 2026-06-08
+**Type:** narrow finite-algebra theorem (distinctness) + correction to the unification narrative
+**Claim type:** bounded_theorem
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed the durable part of this PR on main as ec1a9ee52. I narrowed it to the finite distinctness correction: staggered chirality epsilon, qubit pseudoscalar omega=iI, and supplied generation Vandermonde handedness are distinct sector/K-parity data, so the earlier omega=Vandermonde=epsilon / single-orientation-Z2 conflation is not valid. I removed the broader action-form/unified-admission-floor language, made the generation circulant an explicit supplied generation-sector context rather than an axiom, and kept chiral gauging/r=1/2/Record consequences out of scope. Gates run: py_compile, runner PASS=5 FAIL=0, SHA-pinned cache refresh/check-only, vocab_lint --fix, docs/audit/scripts/run_pipeline.sh, and strict audit_lint (notices only).

PR #3345, opened 2026-06-09
P1 PMNS runner: set iv.prec=200 (interval parts ran at 53-bit despite the note's 200-bit claim)

JONbridgerapps:
## The bug

`scripts/frontier_pmns_theta12_theta13_dcp_predictions_narrow.py` set `mp.prec=200` (line 72) but never `iv.prec`. In mpmath the `mp` and `iv` contexts have **independent** precision — `mp.prec` does not affect `iv.prec` (default **53**). So Parts 2–5 (interval Newton, adjugate projectors, the box-Krawczyk δ_CP certificate) actually ran at **53-bit** interval precision, not the **200-bit** the source note repeatedly claims.

```
python3 -c "from mpmath import iv, mp; mp.prec=200; print(iv.prec)"   # -> 53
```

## The fix

Add `iv.prec = 200` next to `mp.prec = 200` at module top (before the module-level `iv` constants such as `SQRT_8_3`).

## Verification

- **`PASS=61 FAIL=0`** still holds.
- Part 2 eigenvalue bracket widths tighten **`4.4e-15 → 0.0`** — now genuinely honoring the note's "`<10⁻¹³` at 200-bit" claim.
- Part 5 **science numbers unchanged**: δ_CP bracket `[251.8648°, 269.9997°]`, 872 bisected sub-boxes, `cos_neg_num` range `[1.676e-7, 9.058e-3]` — all identical to the 53-bit run.

The bracket is unchanged because it is **bisection-depth-limited** (depth 6, at the chamber-edge approach to maximal CP), **not** precision-limited — exactly as the note's §8 already states (tightening route = deeper bisection, not more bits). So the fix makes the runner's self-reported precision accurate **without changing any reported result**; no note number needs updating. Cache regenerated.

The sibling X6 certificate already sets `iv.prec=200` in its own `main()`, so it is unaffected.

### DIFF
--- a/logs/runner-cache/frontier_pmns_theta12_theta13_dcp_predictions_narrow.txt
+++ b/logs/runner-cache/frontier_pmns_theta12_theta13_dcp_predictions_narrow.txt
@@ -1,11 +1,3 @@
-===== runner cache v1 =====
-runner: scripts/frontier_pmns_theta12_theta13_dcp_predictions_narrow.py
-runner_sha256: 022d4590d179c6585d682d54699ac2ae33d9640a029a810fe76270808e7f68ec
-timeout_sec: 120
-exit_code: 0
-elapsed_sec: 32.31
-status: ok
------ stdout -----
 ================================================================================
 PMNS theta_12 / theta_13 / delta_CP predictions narrow rescope
                              (Cycle 8, 2026-05-17)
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed on main as 2103d9925 after review-loop. The runner now sets iv.prec=200 next to mp.prec=200 before interval constants are created; cache regenerated canonically. Verified runner via precompute PASS=61/FAIL=0, cache freshness, py_compile, vocab lint, full audit pipeline, strict audit lint, and staged checks.

PR #3350, opened 2026-06-09
Fill #3121 Part D's open input: fixed-point velocity anomalous dimension gamma < gamma_crit for the tight LV bounds (no-go on the xi->inf flow-suppression escape)

JONbridgerapps:
## What this is

A single, standalone **computed no-go** that fills the open input the interacting velocity-RG attractor note (#3121, on main) names explicitly in its Part D: *"the anomalous dimension `γ` at the physical fixed point, hence whether the hierarchy suppression beats bounds."*

On the continuous-time (`ξ → ∞`) horn the species velocity anisotropy is suppressed by the attractor as `|δv|_IR ~ |δv|_UV (μ/M_Pl)^γ`, so beating a Lorentz-violation bound `B` needs `γ ≥ γ_crit(B)`. This note computes `γ` with the full SU(3) colour structure and finds **`γ < γ_crit` for the tight bounds, robustly** — the hierarchy suppression does not beat them; #3121's residual stands.

## The result (exact runner scope, 15/15)

- `γ = (C_F + T_F N_f) α_s = (4/3 + N_f/2) α_s ≈ 0.15–0.34` at β=6 — obtained by **diagonalizing** the coupled 2×2 velocity RG (nonzero eigenvalue `−(C_F+T_F N_f)α_s`, zero common-mode; `C_A` structurally absent from the difference channel, verified numerically rather than asserted).
- An inflated over-estimate (full `C_A` forced into the difference + `N_f=6`) gives `γ_max ≈ 0.58`, still below every tight `γ_crit` (photon 0.95, electron 1.06, nucleon 1.32).
- **Robust to the (also-open) coefficient**, stated as a margin: the one-loop estimate `δv_UV ~ 1.6×10⁻²` sits 7–14 orders above the threshold where the no-go would fail. (Honest boundary: it does not survive an unbounded downward revision of `δv_UV` below `~10⁻⁹`; the estimate is `~10⁻²`.)
- **One honest boundary:** the no-go holds *because the gauge sector is asymptotically free*. A walking/near-conformal `α*~0.3` plateau over the hierarchy *would* close the gap but is mutually exclusive with AF (Banks-Zaks needs `N_f~16.5`). Plus a bare-vs-MS-bar false-escape guard.

## Reworked to the #3331 review-loop guidance

#3331 was *not landed*; this addresses each point:
- **Stale branch →** branched off **current** `main` (not before landed review-loop commits).
- **Hard-coded `PASS=True` interpretive checks →** every `check()` is now an **independent numeric test**; interpretive/scope statements are printed narration. (The robustness test even caught and corrected an over-aggressive claim mid-rework.)
- **Over-headlined coefficient →** `δv ~ 0.2 α_s` is kept at its honest **one-loop-estimate** scope; the headline (`γ < γ_crit`) is what the runner actually computes, and the no-go is robust to the coefficient.
- **Status-demotion posture →** none; explicit no-promotion statement; this note only *adds a computed bound*.
- **Internal inconsistency between commits →** this is one standalone note; it does not depend on the contested derived-time / consumer-audit commits.
- **Cite, don't restate, B₄ →** cites the on-main B₄ boundary for the `ξ=1` surface.

## Honest scope

A no-go on the **flow-suppression sufficiency** of the `ξ → ∞` horn **only**. On the `ξ = 1` discrete surface the cited B₄ boundary gives `δv = 0`, so **both horns of `δv(ξ)` remain live**. This does **not** establish `δv = 0`, is **not** a custodial mechanism, and is **not** a framework inconsistency. Sets no audit status.

### DIFF
--- a/docs/GAMMA_FULL_VS_GAMMA_CRIT_DECISIVE_NOGO_NOTE_2026-06-08.md
+++ b/docs/GAMMA_FULL_VS_GAMMA_CRIT_DECISIVE_NOGO_NOTE_2026-06-08.md
@@ -1,0 +1,134 @@
+# The Fixed-Point Velocity Anomalous Dimension `γ = (4/3 + N_f/2)α_s` Is Below the Critical Exponent `γ_crit` for the Tight Lorentz-Violation Bounds — Filling the Open Input Named by the Interacting Velocity-RG Attractor (#3121, Part D) — No-Go Note
+
+**Date:** 2026-06-08
+**Claim type:** no_go (a computed bound on the flow-suppression sufficiency of the `ξ → ∞` horn; fills a named open input of a landed note)
+**Type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome, and does not change the status of any other note.
+**Primary runner:**
+[`scripts/frontier_gamma_full_vs_gamma_crit_decisive_nogo_2026_06_08.py`](../scripts/frontier_gamma_full_vs_gamma_crit_decisive_nogo_2026_06_08.py)
+**Cached runner output:**
+[`logs/runner-cache/frontier_gamma_full_vs_gamma_crit_decisive_nogo_2026_06_08.txt`](../logs/runner-cache/frontier_gamma_full_vs_gamma_crit_decisive_nogo_2026_06_08.txt)
+
+---
+
+## What this fills
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as commit 0643be441 after review-loop narrowing and no-go discipline checks. Closing this PR; durable content is on main.

PR #3264, opened 2026-06-08
science: the W-native induced Newton constant is positive and finite (capstone; magnitude registered G3)

jonathonreilly:
## Summary

Step 3 (capstone) of the universal-GR tensor closure: extract the continuum-extrapolated, Ward-clean, isotropic TT graviton stiffness `C_TT` — the **induced** `1/(16πG_Newton)` (Sakharov) — on the framework's native elliptic staggered Kähler–Dirac generator.

**Result:** `C_TT` is **positive** (`≈ +0.020` continuum, `m=0.7`; attractive, no ghost) and **finite** (lattice-regulated, positive for all `m`). So `1/(16πG) = C_TT > 0`: a consistent **positive, finite** `G = a²/(16π·C_TT)` exists. The **magnitude is registered** (G3) via the lattice scale `a`; only the **sign / finiteness / structure / W-nativeness** is derived.

The positive sign is non-trivial — it **requires** the elliptic pin: the non-elliptic bare-Hermitian `σ·sin` generator gives a **negative (tachyonic/ghost)** `C_TT ≈ −22` (the #3222 control). And the sign matches the **attractive** scalar Newtonian gravity already derived on the same lattice (#3184).

This assembles the program: the W-native graviton is **healthy** (#3222) + **Ward-clean/transverse** (#3242) + **isotropic** (#3257), and its stiffness is a **positive finite induced Newton constant**.

## Verification (runner 5/5, memory-safe, 12s)

- **T1** `C_TT` positive, N-convergent → continuum `+0.020` (induced `1/(16πG) > 0`).
- **T2** positive + finite for all `m∈{1,0.7,0.5,0.3}` (lattice-regulated, no pathology).
- **T3** control: non-elliptic bare-Hermitian gives **negative** `C_TT ≈ −22` (tachyonic `1/G`) — positive `G` requires the elliptic pin.
- **T4** staggered generator is elliptic (`DD†=Δ·1`, `Δ>0`, valid `Z`).
- **T5** consistency: `1/(16πG)=C_TT>0` → finite positive `G/a² ≈ 1.0` (`a` registered, G3); sign matches attractive scalar Newton.

Honest scope: **no magnitude derived** (registered, G3); `C_TT` is a regulator-dependent UV quantity (grows as `m→0`). The ~3% graviton anisotropy of #3257 is inherited.

Source-only: one note + one runner + one cache.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3184, opened 2026-06-08
science: lattice Green function 1/(4pi r) from the A1 heat-kernel resolvent (retires P1 import; emergent-gravity program)

jonathonreilly:
## Summary

**Fresh program: emergent gravity.** First deliverable — retires the single load-bearing **import** in the flagship inverse-square-gravity chain.

The chain (lattice propagator → 1/r Newtonian potential → inverse-square force) rests on `G(r)→1/(4π|r|)`. The Maradudin note (retained_bounded) pins the *constant* via the small-k Fourier route, but the **asymptotic-existence/uniqueness step P1** is carried as an accepted-premise *textbook import*. **This note supplies P1 framework-internally** via a different complete route — the A1 graph-Laplacian **resolvent**:

- `G = L⁻¹ = ∫₀^∞ e^{−tL} dt` (resolvent identity; `L=−Δ_lat` is the A1 graph Laplacian).
- `e^{−tL}` factorizes over the commuting axes into the **exact** 1D NN heat kernels `(e^{−tL₁})_{0n}=e^{−2t}I_n(2t)` → `G(x)=∫∏_μ e^{−2t}I_{x_μ}(2t)dt` (framework-internal).
- Large-r asymptotic from the **large-t continuum Gaussian limit**: `e^{−2t}I_n(2t)→(4πt)^{−1/2}e^{−n²/4t}` ⟹ `∏→(4πt)^{−3/2}e^{−|x|²/4t}` ⟹ `G→1/(4π|x|)` — **isotropic**, value `4π` fixed, lattice corrections subleading.

Verified: `4πr·G(r)→1` (axis `1.02@r=4 → 1.0001@r=64`; diagonal likewise — isotropic) and `G(0)=0.252731` (Watson integral). The only non-framework ingredient is **standard Bessel/Gaussian asymptotics** (math, reconstructed in the runner), not a physics import.

## Claim type
`positive_theorem` (framework-internal derivation; standard analysis reconstructed). Runner **9/9** (1D heat kernel exact; factorization; `4πrG→1` axis+diagonal; Watson `G(0)`; Bessel→Gaussian; continuum Green identity; isotropy + wrong-normalization controls).

## What it retires / does not claim
- **Retires** the P1 accepted-premise import (existence + uniqueness + value of the isotropic `1/(4π|r|)` asymptotic) for the flagship chain.
[...]

PR #3365, opened 2026-06-09
science: tensor (graviton C_TT) and scalar (Poisson 1/4pi r) Newton-constant routes are mutually consistent -- both propagate via the same inverted graph Laplacian

jonathonreilly:
## Summary

Mutual-consistency check of the W-native induced gravity's **two** Newton-constant routes (both G3-registered): the **scalar Poisson** route (`G(r)→1/(4πr)`, the inverted A1 graph-Laplacian heat-kernel resolvent, #3184) and the **tensor graviton** route (the induced TT graviton stiffness `C_TT = 1/(16πG)`, #3264).

**Result:** the two routes are mutually consistent — both gravity sectors propagate via the **same inverted A1 graph Laplacian** (`→1/(4πr)`). The graviton is **massless** in the TT channel (`Π_TT(k)−Π_TT(0) ~ C_TT·(2−2cos k)`, the lattice Laplacian dispersion, at small k), so its propagator is `(1/C_TT)·1/(4πr)` — the same Green function scaled by `1/C_TT = 16πG`. With the standard GR tensor factor the two routes register the **same observed G**: the framework's tensor and scalar gravity are one consistent theory.

## Runner (3/3 PASS, <1s)
- **T1** scalar `G(0)=0.252731` (Watson) and `4πr·G→1` (1.004, 1.001, 1.0002) — the `1/(4πr)` Green function.
- **T2** tensor `C_TT=+0.022>0 → G_tensor=1/(16πC_TT)=+0.90>0`.
- **T3** graviton massless: `Π_TT(k)−Π_TT(0) ~ C_TT·(2−2cos k)` at small k (15% O(a²) correction) → same inverted Laplacian → `(1/C_TT)·1/(4πr)`.

## Honest scope
**Both** Newton-constant magnitudes are **registered (G3)** — this note does not derive the absolute `G`. It derives the **structural consistency** (shared `1/(4πr)` propagator + graviton masslessness + the standard GR tensor factor), which fixes the two routes to the same `G` up to a pure GR number. The dispersion-shape match carries the expected `O(a²)` lattice correction (~15%). Strong-field/nonlinear is a retained no-go (self-gravity loop diverges).

All decisive computations independently re-derived. Memory-safe (scalar: scipy.quad; tensor: N≤10, 2×2).


jonathonreilly:
Review-loop landed a narrowed salvage in cd038399c. I did not land the submitted claim that the tensor and scalar Newton routes register the same observed G or already form one complete gravity theory. The landed source is a bounded shared-Laplacian-shape diagnostic only: scalar Z^3 Green tail plus sampled positive tensor stiffness/shape compatibility.\n\nChecks run: runner PASS=3 FAIL=0, py_compile, vocab_lint --fix, full audit pipeline, strict audit_lint, and git diff --check. Generated audit/publication authority outputs were stripped; only source, runner, cache, and front-door snapshot landed.

PR #3367, opened 2026-06-09
attack the bottom: the gravity sign G>0 is leading-order (O(k²)) — the catch-22 + naturalness gap are subleading (O(k⁴)); the real bottom is the emergent dynamical metric

JONbridgerapps:
## Attacking the bottom of the gravity-sign chain directly

R1+R2+R3 reduced `G>0` to "emergent IR-exact Lorentz" — carrying the **catch-22** and the **`LORENTZ_NATURALNESS_GAP`**. The decisive question: does the *sign* need IR-*exact* Lorentz, or only leading-order isotropy?

**Result (runner 4/4):** the gravity sign is the **leading O(k²)** kinetic coefficient; the naturalness-gap LV is **O(k⁴)** — different orders, so the LV cannot flip the sign.
- **B1:** `ω²(k)=c₂k²(1+α·A₄(k)·k²)`; for **every** LV strength `α` (incl. UV), the `k→0` leading coefficient → `c₂=+½ > 0`. The LV never flips the leading sign.
- **B2:** the leading O(k²) coefficient is isotropic (lattice O(p²) dispersion = `|p|²`, the ξ-isotropy result) and its sign is fixed by **RP**. So the sign needs leading-order SO(3) (held) + RP (theorem), **not** IR-exact Lorentz.
- **B3:** the catch-22 + naturalness gap bound the **O(k⁴)** Lorentz-violating corrections, **not** the O(k²) sign.
- **B4:** the sign bottoms out at the **emergent dynamical metric** (the edge-length DOF on which R3's Regge/EH action lives) — of which the framework has the *kinematic* half (records-conformal-class + scale), not the dynamical half.

**Verdict:** `G>0` does **not** require IR-exact Lorentz — it's a leading-order (IR; 1/r lives at `k→0`) property. **The scary walls are subleading and cannot flip the sign.** The deepest gravity atom bottoms out at the **emergent dynamical metric** — a cleaner, more tractable frontier than the naturalness-gap no-go. Downgrades + sharpens the bottom. (Honest: the LV/naturalness gap are real for the *corrections*; the emergent dynamical metric is the genuine remaining input.) `claim_type=bounded_theorem`.

### DIFF
--- a/docs/GRAVITY_SIGN_BOTTOM_IS_LEADING_ORDER_DECOUPLES_FROM_LV_REAL_BOTTOM_IS_EMERGENT_METRIC_NARROW_THEOREM_NOTE_2026-06-08.md
+++ b/docs/GRAVITY_SIGN_BOTTOM_IS_LEADING_ORDER_DECOUPLES_FROM_LV_REAL_BOTTOM_IS_EMERGENT_METRIC_NARROW_THEOREM_NOTE_2026-06-08.md
@@ -1,0 +1,97 @@
+# Attacking the Bottom: the Gravity Sign G>0 Is a Leading-Order (O(k²)) Property — the Catch-22 and the Lorentz Naturalness Gap Are Subleading (O(k⁴)); the Real Bottom Is the Emergent Dynamical Metric
+
+> **Key terms used in this doc** are indexed A-Z at
+> [docs/KEY_TERMINOLOGY.md](KEY_TERMINOLOGY.md); each row points to the
+> canonical source-of-truth doc.
+
+**Date:** 2026-06-08
+**Type:** bottom analysis (downgrades/relocates the deepest residual of the graviton-diffeomorphism chain)
+**Claim type:** bounded_theorem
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed a narrowed salvage on main in 290419782.\n\nWhat landed: a conditional order-separation theorem. If a physical TT kernel independently supplies a nonzero leading c2 and the LV term is genuinely O(k^4), then finite alpha cannot change the strict k->0 sign sign(c2).\n\nWhat did not land: the broad bottom-relocation claim, G>0, framework-has-leading-SO(3), RP-to-gravity-sign closure, IR-exact Lorentz irrelevance, catch-22/naturalness retirement, or emergent-dynamical-metric closure. The no-go discipline gate is recorded in the note and fails the broad claim; the runner explicitly depends on supplied c2.\n\nChecks run: runner PASS=4 FAIL=0, cache refresh status ok, py_compile, vocab_lint --fix, full audit pipeline, strict audit_lint, git diff --check.

PR #3399, opened 2026-06-09
Add direct fifth-family radial certificate check

JONbridgerapps:
## Summary
- expose the fifth-family radial orientation certificate as a callable computed result
- make the parent radial failure runner import that certificate source and assert the computed zero/neutral/sign/slope values directly
- refresh the SHA-pinned runner caches and document that the cache is now a freshness guard, not the load-bearing orientation proof

## Verification
- python3 scripts/fifth_family_radial_symmetry_orientation_certificate_2026_06_08.py
- python3 scripts/FIFTH_FAMILY_RADIAL_FAILURE_AUDIT.py
- python3 scripts/precompute_audit_runners.py --runners scripts/fifth_family_radial_symmetry_orientation_certificate_2026_06_08.py,scripts/FIFTH_FAMILY_RADIAL_FAILURE_AUDIT.py --check-only --allow-non-main --push-mode=none
- python3 scripts/precompute_audit_runners.py --pr-diff origin/main --check-only --allow-non-main --push-mode=none


jonathonreilly:
Landed to main as 44bb3415b. Review-loop kept the useful repair: the parent runner now imports the certificate source and checks computed values directly; the touched fifth-family row is ready for independent re-audit.

PR #3397, opened 2026-06-09
Koide r-polarization gate sharpening: the Record K/CPT-orbit quotient is the complex-slot quotient (23/23)

JONbridgerapps:
Opens the next campaign: the **AC_φλ / Koide lane** (highest-leverage Tier-A admission in the ledger). Not a derivation of r=1/2 — the gate stays open — but three computed sharpenings, built on the one ingredient that **postdates every refuted Koide attempt**: the 2026-06-05 Record refinement (outcome = K/CPT orbit of the realized central sector).

## The three advances (runner 23/23)
1. **Completeness (new classification):** the Z₃-commutant on the generation doublet is `span{1, J}`; its complex structures are exactly `{+J, −J}` (computed: solutions of `(x+yJ)²=−1` are `x=0, y=±1`), exchanged by K; quaternionic is dimensionally impossible. So the fork `{real, holomorphic}` is **exhaustive** — `r ∈ {1, ½}`, `Q ∈ {1, ⅔}` are the only cells.
2. **The orbit quotient IS the complex-slot quotient (new):** `K(e₁)=e₂` (computed), so the central-sector orbit partition `{e₀}, {e₁,e₂}` coincides with the R-block/C-block partition of `R[Z₃] = R ⊕ C` — the landed fork note's *named positive route* ("show the readout functional factors through the doublet complex-slot quotient"), exhibited. **Entailed:** record-readouts are functions on orbits ⇒ phase-resolved sector readouts are **excluded**; doublet conjugation-invariance `I(b)=I(b̄)` follows from orbit granularity (narrow — *not* full P2/modulus; the earlier "P2 not Record-derivable" result is untouched).
3. **Honest boundary (computed):** both `|b|` (degree 1) and `|b|²` (degree 2) factor through the orbit quotient — orbit granularity does **not** select the slot degree. The residual selector is therefore **exactly one atom**: the existing modulus/(M) degree question. **The Koide-r gate and P2/(M) are unified into one residual.**

## Discipline (this lane's history demands it)
- The landed fork table (`KOIDE_BEREZIN_DETC_VS_DETR_FORK_MECHANISM_NOTE`) is a **verbatim cross-check gate inside the runner** — the #3138 inverted-mapping failure mode is structurally guarded against.
- No refuted route is re-walked (chiral→r=½; Dyson/Pfaffian det_C; charge-forecloses-r=½).
[...]

PR #3138, opened 2026-06-07
koide(det_C): the r=1/2 fork is the Dyson reality class — charged-lepton mass is symmetric → r=1 (sharper obstruction)

JONbridgerapps:
## Summary

Fresh physics-loop campaign on **the one genuinely open derivation lever** for a flavor *value* — the charged-lepton Koide magnitude `r = |b|²/a² = 1/2` (`Q = 2/3`) via the multiplicity-stripped `det_C` / Kähler-Dirac index (gated on `AC_φλ`, Tier-A lev 41). Per the owner mandate — *"spend 10 physicists and logicians worth of trying to reframe the problem before attacking it from all sides"* — a **16-agent reframe→attack→synthesize workflow** (10 distinct physicist/logician lenses, then 5 adversarial derive-or-refute attacks, then synthesis) ran **before** any derivation.

## Result — sharper obstruction (the lever is foreclosed for charged leptons)

The reframe converged on a new object and the attack closed it:

**The `det_C`-vs-`det_R` fork is the Dyson reality class**, not a continuous measure choice (the `J_cs` measure-neutrality no-go forecloses only the *continuous* selectors):

| reading | object | degree | doublet | result | class |
|---|---|---|---|---|---|
| `det_R` | determinant | 2 | counted twice | `r=1` (`Q=1`) | Dirac / **symmetric** |
| `det_C` | Pfaffian `=√det` | 1 | counted once | `r=1/2` (`Q=2/3`) | Majorana / **antisymmetric** |

The K/CPT-real charged-lepton mass `M = aI + bC + b̄C²` is **Hermitian** with distinct real doublet eigenvalues `a − b_r ∓ √3·b_i` (`m_μ ≠ m_τ`), built along **`i·J_cs`** — the *symmetric* partner of the antisymmetric `J_cs`. So it has a **determinant, no Pfaffian → r=1**. No antisymmetric (`K²=−1`) structure can live on the distinct-eigenvalue doublet (any `J` commuting with `diag(λ₁,λ₂)`, `λ₁≠λ₂`, is diagonal → `J²≥0`). And **electric charge forbids** the Majorana/Pfaffian (`r=1/2`) structure for a charge `−1` lepton (`ΔQ=−2`) — independent of the staggered realization. **Three convergent independent arguments** (Coleman-Weinberg rank-2 Hessian, KO-mod-2 Dyson parity, Berezin degree) plus charge all give `r=1`.

[...]

PR #3400, opened 2026-06-09
Koide occupancy atom: independence proven + the orbit-occupancy premise candidate (24/24)

JONbridgerapps:
The measure-class shot, following #3397's named next target. Settles the **status** of the one residual atom behind the Koide r-gate — and ends the seven-attempt derivation loop with a definite answer.

## Result 1 — the atom is independent of the current premise surface
Two prongs, both computed:
- **The axiom's own clause (O2):** the landed Record axiom supplies *"no ... weighting, normalization, probability ... or occupancy rule"* — checked mechanically on the live `MINIMAL_AXIOMS_2026-06-05.md`. The doublet weight class **is** a weighting/occupancy rule. Record declines to supply it, in its own text.
- **Exhibition (O3):** two explicit models — `M_sector` (one slot per real component, `Z_d = 2π/g`) and `M_orbit` (one slot per record-outcome, `Z_d = π/g`, exact integrals) — **both** satisfy every axiom-imposed constraint (Z₃-equivariance, K/orbit-invariance, positivity, finite additivity on the two-orbit outcome algebra). They differ exactly by the occupancy factor **2 = the fiber count of the 2:1 sector→orbit covering**.

Consequence: derivation attempts from the current surface **cannot** settle r — which mechanizes the refuted-attempt history (each prior route smuggled an occupancy rule; the CW-modulus route is a sector-side choice, supplied not retained).

## Discipline
- Consequence map **only at the landed ρ bookkeeping** — no new microscopic bridge. Explicit **orientation guard**: both candidate normalizations computed, the inverted one *rejected against the landed table* (the #3138 failure mode, structurally guarded). Convention-free core: `r_sector/r_orbit = Z_sector/Z_orbit = 2` exactly.
- PDG `Q = 0.666661` is a **labeled comparator only** (sits on the orbit cell to 6×10⁻⁶; the sector cell `Q=1` is excluded empirically by ~50%).

It is the **unique granularity-matching** choice (slot-groups ↔ outcomes bijection, 2 = 2; the sector model has 3 slots vs 2 outcomes) — the readout-side analogue of the approved kinetic-isotropy primitive ("tick grained like edge" :: "statistics grained like outcomes"): dimensionless, structural, binary, no fitted number. **If approved** (owner decision via `AXIOM_MINIMALITY_POLICY` §6, the ξ=1 playbook): `r = 1/2` and `Q = 2/3` follow from the landed lever — the Koide relation becomes the statement that nature's record statistics counts outcomes, not sectors.

Runner 24/24; markdown deps all landed (the #3397 note is plain-text context); sets no audit verdict.

### DIFF
--- a/docs/KOIDE_ORBIT_OCCUPANCY_INDEPENDENCE_AND_PREMISE_CANDIDATE_NOTE_2026-06-09.md
+++ b/docs/KOIDE_ORBIT_OCCUPANCY_INDEPENDENCE_AND_PREMISE_CANDIDATE_NOTE_2026-06-09.md
@@ -1,0 +1,148 @@
+# The Koide Occupancy Atom Is Independent of the Current Premise Surface; the Orbit-Occupancy Premise Candidate
+
+**Date:** 2026-06-09
+**Claim type:** bounded_theorem (independence by exhibition + the axiom's own boundary clause) + a premise-candidate proposal (NOT adopted)
+**Type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:**
+[`scripts/frontier_koide_orbit_occupancy_independence_2026_06_09.py`](../scripts/frontier_koide_orbit_occupancy_independence_2026_06_09.py)
+**Cached runner output:**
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed to main as e0aa53252 with review-loop narrowing: bounded current-checked-surface independence, explicit no-go-discipline gate, no premise adoption, no audit verdict.

PR #3405, opened 2026-06-09
feat(interacting): Fierz singlet-channel selector is a weight, not a partition — register-not-read κ_EW=0 route demoted

jonathonreilly:
## Summary

Lands the first of the ADM-1 find-the-escape panel's **pre-verified route-pruning triples** (the panel adversarially recomputed these grounds during its `forced_finding`). A **narrow route-demote**: the `κ_EW=0`-via-register-not-read route is demoted on three exact grounds; **the κ_EW gate itself remains open.**

## The three grounds — `TOTAL: PASS=14 FAIL=0`

- **(G-A) The singlet projector is a twirl, not a partition map.** `E_sing(M)=(TrM/N_c)I` = the Haar/depolarizing **twirl** (Ad-invariant, idempotent, unital; MC-confirmed). Partition maps preserve diagonal blocks; the twirl averages them. On the SU(3)-**irreducible** triplet the only G1-compliant partition is trivial `{I}` (Schur), whose `D=id≠twirl`. The genuine register-not-read license (the central-sector partition map) does **not** cover the channel split — invoking it is the **loose** dichotomy demoted by the 2026-06-06 scope correction.
- **(G-B) κ_EW is a within-channel weight (weight-leak).** Count `8/9` fixed by the decomposition; the **weight** κ is within-sector data the partition never delivers (G3). The runner exhibits the formal isomorphism with the Koide `r`-dial (same fixed-count/free-weight shape) — *"singlet unregistered ⟹ κ=0"* is the move that would force the known-free `r`.
- **(G-C) Category:** the Fierz channel decomposes the **same-site** current trace (invariant under local rotations), not the gauge-link kernel (co-transforms) — selecting κ_EW says nothing about ADM-1.

## Honest residuals

Other κ_EW routes **not** foreclosed (no closing language); the exact Fierz algebra (`F_adj=8/9`) untouched; ADM-1 untouched; sits **beside** the existing `retained_no_go` family and the `rconn` gate note (`audited_conditional`), duplicating neither. All cited statuses ledger-verified fresh. No new axiom/import; no PDG value.

## Changes (source-only triple)

- `docs/FIERZ_SINGLET_CHANNEL_SELECTOR_IS_WEIGHT_NOT_PARTITION_NARROW_NO_GO_NOTE_2026-06-08.md`
- `scripts/frontier_fierz_singlet_selector_weight_not_partition_2026_06_08.py`
- `logs/runner-cache/frontier_fierz_singlet_selector_weight_not_partition_2026_06_08.txt`

### DIFF
--- a/docs/FIERZ_SINGLET_CHANNEL_SELECTOR_IS_WEIGHT_NOT_PARTITION_NARROW_NO_GO_NOTE_2026-06-08.md
+++ b/docs/FIERZ_SINGLET_CHANNEL_SELECTOR_IS_WEIGHT_NOT_PARTITION_NARROW_NO_GO_NOTE_2026-06-08.md
@@ -1,0 +1,89 @@
+# The Fierz Singlet-Channel Selector Is a Weight, Not a Partition — the Register-Not-Read κ_EW=0 Route Demoted
+
+**Date:** 2026-06-08
+**Type:** narrow no-go (route-demote) — prunes ONE proposed closure route; the gate stays open
+**Claim type:** no_go
+**Script:** `scripts/frontier_fierz_singlet_selector_weight_not_partition_2026_06_08.py`
+**Cache:** `logs/runner-cache/frontier_fierz_singlet_selector_weight_not_partition_2026_06_08.txt`
+**Status:** source proposal. The three grounds are exact finite algebra (runner `PASS=14
+FAIL=0`); the route assessment was independently recomputed by an adversarial panel before
+this note was written. Authority role: source proposal; audit lane sets status.
+
+## The route under demotion
+
+The EW matching rule leaves the one-parameter family `R(κ_EW) = F_adj + κ_EW(1−F_adj)`,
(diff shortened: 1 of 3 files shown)


jonathonreilly:
review-loop landed a narrowed salvage on main as d42b21230. Repairs before landing: replaced shorthand G-label language with explicit route-ground names, qualified the MC-backed twirl check instead of calling every ground exact, added the visible no-go discipline gate for the route-demotion scope, and strengthened the runner's partition-symmetry check to use the SU(3) generator basis rather than a random group sample. Scope remains narrow: only the register-not-read color-trace route is demoted; the kappa_EW gate stays open. Checks: py_compile passed, direct runner PASS=14 FAIL=0, cache refreshed and fresh, vocab lint clean, audit pipeline and strict audit_lint passed with existing notices only, generated audit/effective-status authority surfaces stripped; FRONT_DOOR_STATUS kept for the new unaudited source row.

PR #3404, opened 2026-06-09
Orbit-occupancy out-of-sample program: neutrino sector — rung-0 dichotomy + Σm_ν = 0.059–0.061 eV band (17/17)

JONbridgerapps:
Executes the panel's upgrade trigger: a **second multiplet constrained by the same counting rule**, with the stakes decided by data that doesn't exist yet. Kill conditions registered as prominently as predictions.

## Rung 0 — structural, no knobs
Conjugation maps `J → −J`, so the K-fixed locus of the generation doublet carries **no invariant complex structure**: a **Majorana** multiplet is *forced* onto the sector cell (`Q=1`); plain-Dirac with the standard circulant readout sits on the orbit cell (`Q=2/3`). (Grounded in the landed Majorana–Berezin cell — verbatim cross-check, the #3138 guard.)

**Exclusion band (oscillation comparators only):** for *any* absolute mass scale, `Q_ν ≤ 0.586` (NO) / `0.500` (IO) ⇒ **both** direct cells are excluded for neutrinos.

**Rung-0 prediction (import-free):** orbit-occupancy + oscillation data ⇒ the neutrino mass operator is structurally different — composite (seesaw-class) and/or Majorana. **K1 (bust line):** conclusive plain-Dirac *elementary* neutrino masses with a standard readout falsifies the rule. The Majorana horn is what 0νββ (LEGEND-1000/nEXO) decides.

## Rung 1 — minimal seesaw (two imports, flagged) → a sharp use-novel band
With `m_ν = m_D²/M_R`, degenerate `M_R`, and the orbit rule on the Dirac block: **derived closed form** `Q_ν(δ) = (25.5 + 6√2 cos 3δ)/36 ∈ [0.473, 0.944]` (symbolic + 12-point certification). Then:
- `δ_ν = δ_l = 2/9` gives `Q_ν = 0.894` — **excluded**: the rule forces the neutrino Koide phase to differ from the charged-lepton phase.
- Intersecting the data band: **normal ordering strongly preferred**, `m₁ ≤ 1.6 meV`, and

  **`Σm_ν = 0.059 – 0.061 eV`**

  — not yet measured, exactly at the DESI/CMB-S4 resolution frontier. **K2:** IO confirmed or `Σm_ν > 0.065 eV` kills rung 1. **K3:** `Q_ν` outside `[0.473, 0.586]` kills rung 1.

## Discipline
[...]

PR #3408, opened 2026-06-09
Occupancy non-exclusivity: mixture bound (1-p < 2.2e-5) + availability-preference formulation (16/16)

JONbridgerapps:
Answers the owner's question — *can r=½ be a valid (even preferred) outcome recorded frequently without excluding r=1?* — with computation rather than discussion.

## Within one context: the data crushes admixture, with a sign
A sector-cell admixture `(1−p)` in the charged-lepton record statistics gives `Q(p) = 2/3 + (1−p)/3` — any admixture pushes **up** (sign is aggregation-independent; the linear coefficient is a flagged assumption). The data: `Q_PDG = 0.6666605` sits **below** 2/3 with `σ(Q) = 6.8×10⁻⁶` from `m_τ`. So the best-fit admixture is **exactly zero** and the one-sided 2σ bound is **`(1−p) < 2.2×10⁻⁵`**. Exclusivity within this context is an *empirical fact at the 10⁻⁵ level* — not an axiom.

## Across contexts: the instinct is right, and already load-bearing
The independence result makes the rule **per-readout-context** — global exclusivity was never claimed. And both cells are *realized*: the sector cell is **forced** for K-fixed (Majorana) multiplets (orbit cell structurally unavailable — rung 0 of #3404); the orbit cell is empirically realized for charged leptons. Both are valid outcomes of the same framework.

One principle, zero per-context freedom (availability is decided by K-fixedness), reproducing every realized case — same predictive content as the rung-0 dichotomy, cleaner ontology, and arguably the better abstract wording for the Tier-A entry. It also sharpens the retirement route: derive the *coarse-cell preference* from record dynamics (coarser outcome ↔ more durable registration) at the measurement/record-production gates — flagged open, not derived.

Panel's Tier-A verdict unchanged. Runner 16/16; comparators labeled; sets no audit status.

### DIFF
--- a/docs/OCCUPANCY_NONEXCLUSIVITY_MIXTURE_BOUND_NOTE_2026-06-09.md
+++ b/docs/OCCUPANCY_NONEXCLUSIVITY_MIXTURE_BOUND_NOTE_2026-06-09.md
@@ -1,0 +1,99 @@
+# Occupancy Non-Exclusivity: the Mixture Bound, Per-Context Realization, and the Availability-Preference Formulation
+
+**Date:** 2026-06-09
+**Claim type:** bounded support (a computed mixture bound + a reformulation of the existing admission)
+**Type:** open_gate / support
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:**
+[`scripts/frontier_occupancy_nonexclusivity_mixture_bound_2026_06_09.py`](../scripts/frontier_occupancy_nonexclusivity_mixture_bound_2026_06_09.py)
+**Cached runner output:**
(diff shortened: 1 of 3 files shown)


jonathonreilly:
review-loop landed a narrowed open-gate salvage on main as 215196396. Repairs before landing: changed to explicit open_gate support, removed unlanded #3400 as an authority dependency, removed Tier-A/admission/primitive wording, narrowed 'both realized' to valid cells with the sector cell conditional on K-fixed/Majorana contexts, and fenced PDG masses/uncertainties as comparators used only for the charged-lepton mixture bound. Checks: py_compile passed, direct runner PASS=16 FAIL=0, cache refreshed/fresh, vocab lint clean, audit pipeline and strict audit_lint passed with existing notices only, generated audit/effective-status authority surfaces stripped; FRONT_DOOR_STATUS kept for the new row.

PR #3411, opened 2026-06-09
Durability derives the orbit granularity — provably not the weight; admission shrinks to MAXENT-R (11/11 + 16/16)

JONbridgerapps:
The retirement attempt: can record **durability** derive the coarse-cell preference and let us skip the occupancy admission? Answer, both halves computed: **half yes (a new conditional theorem), half provably no (a sharp boundary)** — and the admission shrinks to its minimal universal core.

## D1 — positive (new): durability + CPT ⇒ orbit granularity
The commutant of the K-swap is `span{1, K}` = the orbit-function algebra (computed). Every K-covariant registration channel has **zero sector distinguishability** (200 random channels, residual 0.0). Sector registers are K-frame-variant; orbit registers exactly invariant — only orbit content is *"fixed once registered"* frame-independently. With **CPT retained exact on the lattice** (`CPT_EXACT_NOTE`), the hypothesis is retained-grade: **the Record axiom's 2026-06-05 orbit clause is derived from its own durability clause**, not stipulated.

## D2 — negative (sharp): durability cannot fix the weight
An explicit family of registration processes with **perfectly durable** (absorbing, orbit-labeled) records realizes every occupancy: standing weight ratio `O_d/O_s = 1.000 → 1.400 → 2.000` as the pre-registration **source measure** sweeps outcome-fed → fiber-fed. Durability constrains *what is written*, not *how often each basin is fed*. This dynamically exhibits the independence theorem (#3400) — as it must; "skip the admission entirely" is **impossible**, proven.

## D3 — the relocation (companion runner, 16/16): MAXENT-R
The admission's minimal true core: *record statistics = maximum entropy over the **registrable** (outcome) alternatives at common stiffness.* Computed: (i) the entire inter-cell factor 2 **is** the fiber count of the 2:1 sector→orbit covering (two independent exact bookkeepings); (ii) MAXENT-R gives the landed holomorphic cell (`r=½, Q=⅔`) in Dirac contexts; (iii) in K-fixed (Majorana) contexts the coefficient is forced real and the **same** principle gives the landed sector cell (`r=1, Q=1`) — **both realized cells, zero per-context choices, and it voluntarily outputs Q=1 where structure dictates** (answering the panel's post-hoc counterfactual head-on); (iv) the sector-side rule must count provably-unregistrable distinctions with a Liouville-class measure that is **not retained** on generation space. Jaynes-vs-Liouville remains a flagged choice — supported by the retained-status asymmetry, not derived.

## Net
- Granularity half of "preferred outcome": **theorem** (conditional on CPT-covariant registration).
- Weight half: **not derivable from the current surface** (proven twice over); residual = one universal principle (MAXENT-R), candidate for re-panel against the primitive purity test the bare binary failed.

Runners 11/11 + 16/16; no comparators consumed; sets no audit status.

### DIFF
--- a/docs/RECORD_DURABILITY_DERIVES_GRANULARITY_NOT_WEIGHT_BOUNDED_THEOREM_NOTE_2026-06-09.md
+++ b/docs/RECORD_DURABILITY_DERIVES_GRANULARITY_NOT_WEIGHT_BOUNDED_THEOREM_NOTE_2026-06-09.md
@@ -1,0 +1,112 @@
+# Record Durability Derives the Orbit Granularity — and Provably Not the Weight
+
+**Date:** 2026-06-09
+**Claim type:** bounded_theorem (a conditional derivation + a sharp negative boundary) + a companion relocation
+**Type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:**
+[`scripts/frontier_record_durability_coarse_preference_2026_06_09.py`](../scripts/frontier_record_durability_coarse_preference_2026_06_09.py)
+(SCORECARD: PASS=11, FAIL=0; cached:
+[`logs/runner-cache/frontier_record_durability_coarse_preference_2026_06_09.txt`](../logs/runner-cache/frontier_record_durability_coarse_preference_2026_06_09.txt))
+**Companion runner (the relocation half):**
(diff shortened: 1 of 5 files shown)


jonathonreilly:
Landed reviewed source salvage directly on main in e4c52419e. Scope narrowed: D1 is a conditional central-record orbit-granularity theorem; D2 is a local durability-does-not-fix-weight boundary with no-go discipline; MAXENT-R is kept as support-only candidate material, not an adopted admission.

PR #3415, opened 2026-06-09
The occupancy subsumption: the factor 2 is a determinant exponent, fixed by the existing gate — zero new admissions (20/20)

JONbridgerapps:
Route 1 of the wall-breaking exercise (three slices converged independently), built against the landed durability/MAXENT salvage (e4c52419e). The re-panel ruled MAXENT-R 6-0 Tier-A-class; this makes that question **moot by deleting the candidate**.

## The chain (runner 20/20, sympy-exact + from-scratch Grassmann engine)
- **`det_R = |det_C|²` exactly** — the two fork cells differ by ONE determinant exponent (1 vs 2); `|det_C|` is the K-orbit-invariant atom. Hostile check: orbit-invariance alone permits every exponent (independence intact).
- **Berezin uniqueness (computed):** translation invariance forces the Grassmann functional unique up to scale — **no measure freedom exists in the fermionic realization**. The engine reproduces `det_C(A)` for generic 2-mode Dirac (off-diagonal cross-term included) and `Pf` with `Pf² = det_R` for Majorana.
- **The subsumption:** the cell is decided by **polarization = the existing staggered Tier-A gate** (mechanically verified in `tier_a_admissions.json`). Dirac → `det_C` → `r=½, Q=⅔`; Majorana → `Pf/det_R` → `r=1, Q=1`. Cell map cross-checked verbatim against the landed fork table.
- **Kraus closure:** arbitrary K-covariant quantum channels × all K-invariant effects → identical `e₁/e₂` statistics (2×10⁻¹⁶) — closes the landed durability note's classical-only channel gap.
- **Adversarial:** exponent 2 is obtainable only by **doubling the field content** (computed) — realization business, never a readout freedom. **Stiffness-independence:** the exponent is scale-free; the re-panel's "common stiffness" objection does not apply to this route.
- **Canonical name + new falsifier:** Frobenius–Schur, factor `2^(1−FS)`; Q₈ control (`FS=−1` computed from g² classes) ⇒ **quaternionic factor-4 kill condition** registered. Jones `√2` named negative.

## Net
```
occupancy weight  =  CONDITIONAL THEOREM under {staggered gate (existing Tier-A), C1 CPT-covariance}
MAXENT-R          =  unnecessary (stays support-only, per the landed salvage's own framing)
new admissions    =  ZERO   (Tier-A inventory unchanged)
Koide Q = 2/3     =  follows from the framework's Dirac realization, nothing new admitted
```
Independence (unconditional surface) intact — "the node is deleted, not the conditionality." Falsifiers: quaternionic context ≠ factor 4; gate closure contradicting complex-mode realization; the #3404 neutrino kill conditions.

Sets no audit status; all markdown deps resolve on main.

### DIFF
--- a/docs/OCCUPANCY_READOUT_EXPONENT_BEREZIN_SUBSUMPTION_BOUNDED_THEOREM_NOTE_2026-06-09.md
+++ b/docs/OCCUPANCY_READOUT_EXPONENT_BEREZIN_SUBSUMPTION_BOUNDED_THEOREM_NOTE_2026-06-09.md
@@ -1,0 +1,128 @@
+# The Occupancy Subsumption: the Factor 2 Is a Determinant Exponent, Fixed by the Existing Gate — Zero New Admissions
+
+**Date:** 2026-06-09
+**Claim type:** bounded_theorem (conditional subsumption of the occupancy atom into the existing premise surface)
+**Type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:**
+[`scripts/frontier_occupancy_readout_exponent_berezin_subsumption_2026_06_09.py`](../scripts/frontier_occupancy_readout_exponent_berezin_subsumption_2026_06_09.py)
+(SCORECARD: PASS=20, FAIL=0; cached:
+[`logs/runner-cache/frontier_occupancy_readout_exponent_berezin_subsumption_2026_06_09.txt`](../logs/runner-cache/frontier_occupancy_readout_exponent_berezin_subsumption_2026_06_09.txt))
+
+> **What this closes, and how much.** A six-lens re-panel unanimously ruled the
+> MAXENT-R relocation still Tier-A-class (a probability/weighting rule cannot be
+> a primitive under the registry's purity language). A five-slice wall-breaking
+> exercise (per the repo exercise skill) then found the deeper move, with three
+> slices converging independently: **the occupancy factor 2 was never a measure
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a narrowed version as commit 3b987a7cd.

Landed: a conditional occupancy-exponent/Berezin subsumption theorem with runner/cache. The durable content is the determinant-exponent mechanism: det_R = |det_C|^2, Berezin uniqueness/no measure freedom in the tested fermionic realization, generic det_C and Pf checks, K-covariant channel closure, multiplicity requiring field-content doubling, and the Frobenius-Schur factor/falsifier.

Narrowing applied before landing: removed the unbounded “Koide Q=2/3 follows / zero admissions” rhetoric. The landed note says the exponent is conditional on the existing staggered Tier-A gate plus K/CPT-covariant registration; it does not derive the gate, readout context, probability rule, physical-species bridge, or charged-lepton value from the baseline axioms. It introduces no additional Tier-A node, but it also does not change the existing gate status or audit status. Record is explicitly barred from supplying weighting/readout/selector content.

Audit boundary: source-only landing. Pipeline validation passed and seeded the row, but generated audit/status outputs were restored because this is a new source row, not a retained-source reset. Independent audit owns the row and its upstream chain.

Checks run: runner PASS=20 FAIL=0; cache refresh and check-only; precompute_audit_runners --check-only; py_compile; vocab_lint --fix; docs/audit/scripts/run_pipeline.sh; audit_lint --strict; git diff --check; staged markdown link hygiene; source overclaim scan; stale-overlap guard clean.

PR #3420, opened 2026-06-09
K/CPT-orbit shrinks of both Tier-A admissions: θ premise-2 discharged, AC_φλ orientation stripped (20/20)

JONbridgerapps:
One mechanism — the occupancy campaign's K/CPT-orbit frame-variance machinery — aimed at **named premises inside both registered Tier-A admissions**. Neither admission is eliminated; each **shrinks**, residuals named, existing no-go portfolios untouched. Registry not edited (audit-lane owned); this is a source-side proposal.

## Shrink 1 — θ admission: 2 premises → 1
The landed `STRONG_CP_THETA_ZERO` note is conditional on (P1) no bare θ/F̃F slot and (P2) the positive-real quark-mass orientation. **This PR derives (P2) from the existing surface:**
- K maps `det → conj(det)` (orbit = ±arg det). **Hostile check kept:** the orbit alone gives only evenness (`cos(arg det)` is K-invariant) — no over-claim.
- **The multiplicative lemma (exact):** continuous multiplicative functionals on `C*` are `|z|^s e^{ik·arg z}`; K-invariance forces `k = 0` — **K-invariant det-class readouts are `|det|^s`, phase-free.**
- The mass-surface readout lane is det-class (the existing `(M)` atom, consumed not added) ⇒ every determinant value is readout-equivalent to its positive-real representative ⇒ **(P2) discharged. Residual = (P1) alone.**

**Guards (all checked):** θ is *not* identified with `arg det` (refuted route stays refuted; θ = the F̃F coefficient, P1 untouched); the RP no-go respected (no RP used); observed CP violation not contradicted (lemma scoped to the multiplicative class; non-multiplicative K-invariant functionals carry phase content — exhibited).

## Shrink 2 — AC_φλ admission: {magnitude, orientation, bridge} → {magnitude}
- `conj(H(δ)) = H(−δ)` **exactly** (matrix identity).
- All elementary symmetric polynomials of the spectrum agree at ±δ (exact); the eigenvalue reshuffle under the flip is the `k → −k` species relabel **the registry already strips as convention** ⇒ the orientation is labeling content; the registrable atom is `|δ|` (`cos 3δ`).
- **Honest residual:** `|δ| = 2/9` stays — its irreducibility is separately audited (radian-bridge portfolio); not re-litigated.

## Net (proposed)
```
non-axiom surface  =  2 primitives (scale, kinetic isotropy)
                    + 1 action-form premise (θ's P1)
                    + 1 number (|δ| = 2/9)
[...]

PR #3429, opened 2026-06-09
θ's last premise via the action-form lane: the F̃F slot does not exist in the per-plaquette class (12/12)

JONbridgerapps:
The second Tier-A admission's last premise (P1: no bare θ/F̃F slot), discharged by an **f-independent combinatorial theorem** — stronger than any orientation-evenness argument, needing no symmetry premise at all.

## The theorem (exact)
`F̃F = ε^{μνρσ}F_{μν}F_{ρσ}` is a **cross-plane product** (4 distinct directions). A per-plaquette action `S = Σ_P f(U_P)` is **additive over planes** ⇒ every cross-plane coefficient is the mixed derivative of a sum of single-plane functions = **0 identically, for any f** (all three ε-pairings computed; no reality/evenness/parity assumption). Direction count: injective 4-from-2 assignments = **0**; single-plaquette tensors carry indices in {μ,ν} only.

## Boundaries respected (sources read directly, per the session discipline)
- The multiplaquette note's verdict — *"only a single-plaquette/minimality admission removes the F̃F slot"* — **T1 is that removal, now a theorem of the class.**
- The single-plaquette note's Im-part flows to single-plane `trF³` (consistent, not F̃F). The RP no-go untouched.

## The discharge
The action lane's **entire live candidate class** (Wilson, heat-kernel, Manton — the uniqueness no-go's own listing, grep-verified) is per-plaquette ⇒ **every candidate yields θ_bare = 0 automatically; the lane need not pick a winner for θ to die.** With P2 already discharged (PR #3420's K-orbit lemma), the θ admission's surviving content = **one structural class statement** — *"the gauge action is per-plaquette (minimal-loop)"* — the gauge sibling of the Lattice axiom's no-diagonal minimality: structural, no number, null-consequence profile. Grade (premise now vs future derivation from record-preservation + minimality) = owner/audit call. **No numeric or CP-specific content survives in the θ admission.**

Honest: per-plaquette not derived here (the lane's which-f/couplings selection stays open for its own purposes — only the θ question is f-independent within the class); falsifier registered (a derivation forcing multi-plaquette terms reopens the slot); θ=π moot on this route. Runner 12/12; sets no audit status; all markdown deps resolve on main.

### DIFF
--- a/docs/THETA_P1_PER_PLAQUETTE_NO_FTF_SLOT_BOUNDED_THEOREM_NOTE_2026-06-09.md
+++ b/docs/THETA_P1_PER_PLAQUETTE_NO_FTF_SLOT_BOUNDED_THEOREM_NOTE_2026-06-09.md
@@ -1,0 +1,77 @@
+# θ's Last Premise via the Action-Form Lane: the F̃F Slot Does Not Exist in the Per-Plaquette Class
+
+**Date:** 2026-06-09
+**Claim type:** bounded_theorem (an f-independent combinatorial exclusion + the discharge of θ-P1 into the lane's standing class)
+**Type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:**
+[`scripts/frontier_theta_p1_per_plaquette_no_ftf_slot_2026_06_09.py`](../scripts/frontier_theta_p1_per_plaquette_no_ftf_slot_2026_06_09.py)
+(SCORECARD: PASS=12, FAIL=0; cached:
+[`logs/runner-cache/frontier_theta_p1_per_plaquette_no_ftf_slot_2026_06_09.txt`](../logs/runner-cache/frontier_theta_p1_per_plaquette_no_ftf_slot_2026_06_09.txt))
+
+---
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a narrowed salvage on main in d47de3bf0.

What landed:
- Conditional bounded theorem: in a supplied per-plaquette action class, local cross-plane F tilde F monomials have zero mixed-derivative coefficients, so no local cross-plane F tilde F slot appears within that supplied class.
- Added a narrowed source note, runner, and cache: PASS=18 FAIL=0.
- Added No-Go Discipline coverage for the negative sentence.

What did not land from the PR framing:
- No claim that per-plaquette support is derived from the framework baseline, Record, adjacency, or a primitive.
- No claim that the full theta/strong-CP admission is retired.
- No claim that the action lane's entire live candidate class is settled, no P1/P2 discharge language, and no theta_bare=0 framework closure.
- Multi-plaquette/clover terms remain outside the theorem and reopen the slot.

Checks run: py_compile, direct runner, cache refresh/check-only, precompute check-only, vocab_lint --fix, audit pipeline compatibility, dependency inspection (only the intended prior theta-boundary deps), restored generated audit/publication/front-door outputs, strict audit_lint, staged diff check.

PR #3440, opened 2026-06-09
Tighten front-door audit status wording

JONbridgerapps:
## Summary
- clarify retained-grade effective_status vocabulary in the root README
- replace the publication README's stale package-state inventory with audit-authoritative navigation
- point readers to generated front-door status, effective-status mirrors, and divergence report before treating rows as retained

## Verification
- git diff --cached --check before commit
- targeted scan for reported stale phrases and conflict markers
- verified referenced front-door/audit files exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,12 @@ Use these entrypoints in order:
 The repo is in an audit-transition state. Source notes and publication tables
 still contain legacy `retained` / `promoted` wording, but the publication-facing
 authority is the audit-derived `effective_status` in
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed on main as a70f6de29 after review-loop. This keeps README/publication README as navigation only, makes audit-derived effective_status the authority for retained-grade use, and passed vocab lint, runner-cache check, full audit pipeline, strict audit lint, and staged diff checks.

PR #3448, opened 2026-06-10
Regge second variation on the retained cubic-Coxeter complex = linearized EH at leading order (exact, isotropic, derived normalization)

JONbridgerapps:
## What

The retained deficit-vanishing row ends at *flat is a solution*; the landed R3 target-operator row explicitly guards that `δ²S_R` of the cubic-Coxeter action is not supplied. This PR computes it **natively on the exact retained complex** (six-tetrahedra body-diagonal chain, flat lengths {1,√2,√3}; the corrected 6-tet axis edge-star is reproduced) and compares it to the in-runner-derived continuum operator.

Runner `scripts/frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.py` (PASS=10; machine-exact internal gates: per-tet Schläfli ~1e-15, Hessian symmetry = Regge's lemma realized, end-to-end action finite-difference, gauge-convention gate).

## Results

| | result |
|---|---|
| R3/R3b | vertex displacements are **exact zero modes at every momentum** (discrete diffeo family) — the geometric action does what the stencil-transcribed operator could not (~p⁵ residual there) |
| R5 | 7 edge classes vs 6 metric components: the single **non-metric breathing mode is massive** (+48 in standard Euclidean orientation) — no spurious flat directions |
| **R6** | **`Q_h(k) = c·Q_EH(k) + O(k⁴)` with ONE constant `c = −1/2` in all three lattice directions** (residual ~1e-8, spread ~1e-8, Schur-vs-projection slice-invariant). `c=−1/2` **is** the textbook `δ²S_R = ½δ²∫√gR` with `δ(√gR) = −√gG^{μν}δg_μν` — the Regge↔EH correspondence **including its ½ normalization, derived**; **exact O(k²) isotropy** |
| R7 | TT pair **equal** (exact spin-2 isotropy); transverse-trace **equal-magnitude opposite sign (ratio −1)** — the degenerate-supermetric no-go's **supplied** comparator pair (±k²/2) is hereby **derived from the framework's own retained geometry** |

**The one remaining sign** is the overall action orientation (S_R vs −S_R) — the *same* located sign residual as the arrow/stability/spectral no-go row: the geometric route adds no second sign admission.

Methodological lemma (load-bearing): the metric-sector comparison requires the **exact line-averaged edge response** (midpoint phase × sinc); a phase-free map is an O(k)-wrong slice that leaks the breathing weight into the metric channels — the contaminated first-draft numbers are reproduced and explained in-runner (R2c gate pins the conventions: continuum gauge maps exactly onto discrete vertex displacements).

## Scope (honest)

3D spatial (the retained complex's own dimension); the 4D/timelike extension (kinetic fiber metric from the geometric action itself) is the named next step. Edge-length DOF provenance and action selection remain open. No axiom/primitive/fitted value; comparators (Regge 1961, Rocek–Williams, Cheeger–Müller–Schrader) cited as context only — every number computed in-runner.

## Discipline

Source note + runner + cache only; no audit-lane data; all load-bearing deps verified present on main and markdown-linked; status authority = independent audit lane only.

### DIFF
--- a/docs/CUBIC_COXETER_REGGE_SECOND_VARIATION_EQUALS_LINEARIZED_EH_NARROW_THEOREM_NOTE_2026-06-09.md
+++ b/docs/CUBIC_COXETER_REGGE_SECOND_VARIATION_EQUALS_LINEARIZED_EH_NARROW_THEOREM_NOTE_2026-06-09.md
@@ -1,0 +1,114 @@
+# The Second Variation of the Regge Action on the Retained Cubic-Coxeter Complex Equals the Linearized Einstein-Hilbert Second Variation at Leading Order — Exactly, Isotropically, with Derived Normalization
+
+**Date:** 2026-06-09
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note writes no
+audit verdict and does not retag any ledger row.
+**Primary runner:** [`scripts/frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.py`](../scripts/frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.py) (PASS=10 FAIL=0)
+**Runner cache:** [`logs/runner-cache/frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.txt`](../logs/runner-cache/frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.txt)
+
+## Scope
+
+The retained row
+[`CUBIC_COXETER_REGGE_DEFICIT_VANISHING_NARROW_THEOREM_NOTE_2026-05-10.md`](CUBIC_COXETER_REGGE_DEFICIT_VANISHING_NARROW_THEOREM_NOTE_2026-05-10.md)
+establishes, for the standard six-tetrahedra body-diagonal Coxeter chain `T(Z³)` (flat edge lengths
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on `main` as `c0075f96f`.

Review-loop verdict: PASS TO LAND, pending independent audit. No audit verdict was applied.

Review fixes made before landing:
- Demoted the unaudited R3/no-go/sign rows from load-bearing authority to contextual targets and removed their markdown dependency edges.
- Reworded the comparator language so those prior no-go rows only name the target pair; the theorem derives its own comparator rather than consuming unaudited authority.
- Kept the retained cubic-Coxeter deficit row as the only load-bearing retained dependency.

Checks run:
- Direct runner: `frontier_cubic_coxeter_regge_second_variation_3d_2026_06_09.py` PASS=10 FAIL=0.
- Independent comparator/gauge check: PASS.
- `py_compile`: PASS.
- Runner cache precompute/check-only: fresh.
- `vocab_lint.py --fix`: clean.
- `docs/audit/scripts/run_pipeline.sh`: PASS; row seeded for independent audit.
- `python3 docs/audit/scripts/audit_lint.py --strict`: PASS with the existing 148 notices.
- Generated audit/status outputs restored before commit.

PR #3332, opened 2026-06-09
feat(interacting): minimal-coupling kinematics — link connection forced by fibre-frame independence

jonathonreilly:
## Summary

First source-triple of the **interacting-theory program** (sub-target 1: matter–gauge minimal coupling). Derives on the framework surface — rather than imports — the link connection `U_μ(x)`, its parallel-transporter gauge law, the gauge-covariant hopping `H_cov = Σ a_x† U_μ(x) a_{x+μ} + h.c.`, and the lattice covariant derivative `D_μ`.

The wall-map of the interacting sector found that **every** prior note using the link `U_μ = exp(igaA_μᵃTᵃ)` *imports* it — no note derives where the gauge field on the `Z³` links comes from. This fills that gap at the **kinematic** level.

## Mechanism (on retained foundations)

The retained nearest-neighbour hopping (`tensor_product_translation_fermion_operator_bridge`, **retained**) on the retained per-site `U(3)` fibre (`graph_first_su3_integration`, **retained**) silently chooses a cross-site fibre frame. One-qubit-per-site locality supplies **no canonical** such identification, and the `SU(3)` fibre is the **commutant** of the observable action (so the frame registers nothing). Hence frame-independent hopping **forces** a link connection transforming as `U_μ(x) → g(x) U_μ(x) g(x+μ)†`.

## Honest tier

`bounded_theorem`. The gauge-**covariance** core is unconditional algebra; the *"connection is forced"* physical reading rests on **one named premise (ADM-1)** — the local fibre frame is a gauge redundancy / unregistered — strongly supported by the commutant + record ontology but its own audit target.

## Changes (source-only triple)

- `docs/MATTER_GAUGE_MINIMAL_COUPLING_FIBER_FRAME_FORCES_CONNECTION_NARROW_THEOREM_NOTE_2026-06-08.md` — source note
- `scripts/frontier_minimal_coupling_fiber_frame_connection_2026_06_08.py` — class-A runner (`(L=2)³`, `C³` fibre; memory-safe)
- `logs/runner-cache/frontier_minimal_coupling_fiber_frame_connection_2026_06_08.txt` — SHA-pinned cache

## Verification — `TOTAL: PASS=23 FAIL=0`

- Central identity `G H_cov[U] G† = H_cov[U']` under the transporter law, to `6×10⁻¹⁵`; spectrum gauge-invariant.
- Transporter law **unique** — four wrong site-assignments all break covariance.
[...]

PR #3449, opened 2026-06-10
feat(bounded): ADM-2 color depolarization mechanisms collapse to two gauge-structure admissions

jonathonreilly:
## Summary

Block 09 of the gauge-link / color-einselection dynamics campaign. An **anti-relocation consolidation**: blocks 04-08 mapped the ADM-2 input (*does the matter dynamics depolarize `rho_color` to the color-blind `I3/3`?*) onto **three** mechanisms — twirl, primitivity, entanglement — each gated by a separately-named admission. Continuing to relocate per block is corollary churn. This note shows, with exact finite-dimensional algebra (`TOTAL: PASS=24 FAIL=0`, no Monte-Carlo fit in the logic path), that the three **collapse to two** irreducible admissions:

- **(A) a presupposed local `SU(3)` connection `V != I3` (+ a named record frame `B`)** — covers **twirl AND primitivity**. The twirl's load-bearing *uniform* averaging weight is supplied by **A2 (unitarity)**: for any unitary `U`, `S_ij = |U_ij|^2` is doubly stochastic, so its primitive stationary vector is forced uniform (`= I3/3`) — not a Record-supplied weight. The twirl residual is then exactly the primitivity admission, and a primitive `S` requires a non-diagonal link = the gauge link the campaign **induces** (circular). Free color-diagonal hopping gives `S = I3` frame-independently (no depolarization).
- **(B) a global color-singlet / Gauss-law physical-state condition** — covers **entanglement**. Schur gives `I3/3` from a global-invariant joint state, but *"observables are `SU(3)`-invariant"* does **not** entail *"the state is a singlet"*: a non-singlet state has a polarized marginal, and the total color Casimir is conserved under the global action (neutrality is an import / initial-condition admission, not dynamically forced).

Both (A) and (B) coincide with the undelivered gauge-structure objects of the campaign; **neither is supplied by A1+A2+Record**.

## Discipline

- **NO hat discharged** (ADM-1, R1 link generator, R2 link-measure delivery, blocking isometry untouched). The default A1+A2+Record configuration (free hopping + arbitrary product matter state) depolarizes on neither lane.
- **NOT a no-go** — supplying the named object (a local link) *does* depolarize (witness G7.4); three open paths named (audit #3332's connection forcing; a future structural color-Gauss-law premise; the comparatively unworked blocking-isometry hat). No closing language.
- Weight-leak guard: every `I3/3` is forced by unitarity or invariance, never by fiat (right side of the Fierz weight-not-partition reading). No partition map for the irreducible triplet; no edge native-color framing; no `r=1/2` forced; no ST1/ST2 ranking; statuses cited at live ledger tier.

## Method / verification

Source-only triple: one note (`docs/`), one class-A runner (`scripts/`, `PASS=24 FAIL=0`), one SHA-pinned cache (`logs/runner-cache/`). An adversarial default-refute review agent independently re-computed all decisive identities (disjoint seeds, max err ~1e-15) and audited circularity / hidden-import / stale-status / weight-leak / overclaim / duplication / closing-language — confirming the weight-is-QUANTUM-supplied sub-result does **not** discharge any hat (net progress toward R2 = 0, pure relocation onto the same wall) → LAND, no edits.

### DIFF
--- a/docs/COLOR_DEPOLARIZATION_ADM2_GATING_ADMISSIONS_COLLAPSE_TO_TWO_NARROW_THEOREM_NOTE_2026-06-09.md
+++ b/docs/COLOR_DEPOLARIZATION_ADM2_GATING_ADMISSIONS_COLLAPSE_TO_TWO_NARROW_THEOREM_NOTE_2026-06-09.md
@@ -1,0 +1,158 @@
+# Color depolarization (ADM-2): the three mapped mechanisms collapse to two gauge-structure admissions
+
+**Tier (branch-local, audit-pending):** bounded_theorem (consolidation). NO hat discharged.
+**Campaign:** gauge-link / color-einselection dynamics (four hats: ADM-1, R1 link
+generator, R2 link-measure delivery, blocking isometry).
+**Runner:** `scripts/frontier_color_depolarization_adm2_admission_collapse_2026_06_09.py`
+(exact finite-dimensional linear algebra on `C^3` and `C^3 (x) C^3`; `TOTAL: PASS=24
+FAIL=0`; random unitary / SU(3) witnesses for already-proven identities, no Monte-Carlo
+fit in the logic path; memory-safe).
+
+## The question consolidated
+
+The campaign reduced the R2 input of the undelivered gauge-link generator to one matter
+question (ADM-2): does the matter dynamics depolarize the single-carrier color density
+`rho_color` to the color-blind floor `I3/3` on the `C^3` fundamental carrier? Over blocks
+04-08 this input was mapped onto **three** distinct depolarization mechanisms, each gated
+by a separately-named admission:
+
+- **TWIRL** (block 05; and the landed reading
+  `FIERZ_SINGLET_CHANNEL_SELECTOR_IS_WEIGHT_NOT_PARTITION...`, unaudited): an averaging map
+  needing a `>=2`-element complementary-frame / multi-instrument average with a **uniform
+  weight** Record does not supply.
+- **PRIMITIVITY** (blocks 06/07): a single named record frame `B` interleaved with the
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as 01e9de8de after review-loop.

Review disposition: PASS WITH BOUNDED CLAIMS. I kept this as a bounded consolidation: ADM-2 color depolarization across the mapped mechanisms gates on two explicit admissions, not on Lattice + Quantum + Record alone. No hat is discharged, and this is not a no-go.

Review fixes made before landing:
- Replaced live A1/A2 shorthand with explicit Lattice + Quantum + Record / Quantum language.
- Phase-normalized random group witnesses into SU(3) where the runner/note says SU(3).
- Regenerated the paired runner cache.

No-Go Discipline gate: PASS. N1 routes considered: local connection forcing, global Gauss-law/neutrality, blocking-isometry delivery, multi-frame/twirl averaging, and record-frame formation. N2 collapsed the raw mechanisms to two independent admissions. N3 found no hidden baseline primitive/axiom wall. N4 residual citations match their residuals or are named as unaudited/open. N5 rhetoric is narrowed to single-carrier / global-action resolutions and explicitly says not a no-go. N6 partial-closure paths remain open (#3332/local connection, future Gauss law premise, blocking isometry). N7 steelman is that those paths can still close ADM-2, which the note preserves. N8 prior record/generator walls are cited only as residual boundaries, not laundered into new impossibility.

Checks: direct runner PASS=24 FAIL=0; independent formula spot-check PASS; py_compile PASS; runner cache check-only fresh; vocab_lint --fix clean; run_pipeline.sh complete; audit_lint --strict OK with existing 148 notices; git diff --check clean. Pipeline-generated audit/status outputs were restored before commit. Proposed claim remains for independent audit lane; no audit verdict was applied.

PR #3476, opened 2026-06-10
physics: substep1 grassmann bridge — consume retained U4 closure, two-candidate collapse becomes positive_theorem

JONbridgerapps:
## What this is
The first **boundary discharge** of the bounded→retained campaign: the substep-1 Grassmann forcing bridge note (2026-05-16) still declared "The U4 bridge remains open" — but U4 was closed four days after it was written (`u4_closes_under_qubit_reframe`, retained `positive_theorem`, re-audited 2026-06-05). The composition had never been written into the note, so its `retained_bounded` status could never upgrade on re-audit. This PR writes it.

## The composition (hostile interface check passed)
- Retained U4 closure delivers `H_x = ℂ²` = the single faithful complex irreducible `Cl(3,0) ≅ M₂(ℂ)` module with k(x) = 1, from the Quantum axiom.
- Composed with (C1) of the single-module note (k = 1 ⇒ dim_ℂ H_x = 2), the bridge's dimensional readout becomes a statement about the *physical* per-site space: the free-boson candidate is excluded (ℵ₀ ≠ 2) and single-pair Grassmann is the unique survivor of the two-candidate surface.
- Interface verified delivered = consumed exactly (sympy: γ_i ↦ σ_i real-algebra isomorphism, scalar commutant, ω ↦ +iI, k=1; the U4 note's licensing premise persists verbatim-in-content on the live axiom memo).

## What it does NOT claim (the honest residual)
The collapse is **not** statistics forcing. The retained no-go (`statistics_agnostic_no_forcing`, 2026-05-25) shows a hard-core-boson frame ties on per-site dimension; the spin-statistics input (`axiom_first_spin_statistics_theorem_note_2026-04-29`) is still **unaudited**. The note declares this as the single named boundary **(B-stat)** and hard-codes the distinction the older spine prose conflated (two-candidate collapse ≠ Grassmann forcing).

## Status change
`bounded_theorem` → **`positive_theorem`**, scoped to the two-candidate collapse on the physical per-site Hilbert space. Five load-bearing one-hop links (two retained, two retained_bounded consumed within scope, one axiom node).

## Runner
Extended in place: U4 composition certificate (COLLAPSE CERTIFICATE line), falsification leg (a k=2 module ρ₊⊕ρ₊ on ℂ⁴ defeats the collapse without the U4 input — the consumed authority is load-bearing), declared scope-boundary residual line. **TOTAL: PASS=38, FAIL=0**, byte-identical across runs, ~0.3 s.

## Downstream discharge
[...]

PR #3479, opened 2026-06-10
physics: substep1 statistics selection — boundary sharpened to one predicate GL(F), conditional forcing certificate

JONbridgerapps:
## What this is
The (B-stat) residual — the single remaining boundary on the staggered-Dirac substep-1 surface after the U4 discharge (PR #3476) — mined per the no-go's own proof and reduced to **one named predicate**.

## The no-go's named input (quoted from the 05-25 statistics-agnostic no-forcing note)
§6: a genuine forcing requires "an independent, retained *locality+statistics* principle (e.g. a derived spin-statistics input — currently carried by the **unaudited** `axiom_first_spin_statistics_theorem`)." N7 steelman: with such a theorem retained, "the hard-core-boson frame could be disallowed while the CAR frame survives."

## Supplier audit: the 04-29 spin-statistics note does NOT supply it
Its load-bearing Step 2 hypothesis is the canonical CCR `[a,a†]=δ` — the hard-core boson violates that on-site (defect norm 2, computed), so its bosonic exclusion covers only the *free* boson, re-proving the retained dimensional half. Its Fact 2.3 ("the only finite-dim alternative is Grassmann Fock") is the same false binary the retained no-go refutes. It cannot be honestly fixed into the supplier. No retained composition works either — each nearby retained row is computed **frame-blind** (parity grading: both candidates F-odd; RP surface: H_hardcore == H_JW exactly on L=4; the dim-2 collapse: 2 = 2).

## The theorem (new note + runner)
(B-stat) reduces to the single predicate **GL(F)**: cross-site graded locality (`{ψ_x,ψ_y} = {ψ_x,ψ_y†} = 0`, x≠y) of the F-odd site fields relative to the *retained* parity grading. Proven:
- **T1**: every retained input is constant across the tied frames (computed, per-row).
- **T2**: GL(F) is non-constant — with the retained dim-2 exclusion, **exactly one of three candidates survives `{dim=2} ∧ GL(F)`: Grassmann/CAR**. Selection forced conditionally on GL(F). (Load-bearing step, class C.)
- **T3** (falsification): remove GL(F) and the tie returns — the no-go is reproduced, not contradicted; and GL(F) is not ungraded-algebra data (the same matrix algebra contains a passing and a failing generating family).

## Runner
`TOTAL: PASS=20 FAIL=0`, deterministic, 0.4 s. Key line: `selection certificate: exactly one of three candidates passes {dim = 2} AND GL(F): ['Grassmann/CAR']`.

## What the spine boundary becomes
(B-stat′): one missing input, GL(F)/the FS exchange sign — a genuine admission candidate (not retained, not Tier-A). The un-refuted derivation route per the 06-06 exercise: multi-loop graded-net cocycle consistency. Tier-A admission is an owner decision per no-go §7.


jonathonreilly:
review-loop landed a narrowed source-only salvage on main as e49826f31.\n\nChanges made before landing:\n- Renamed the note to a canonical bounded-theorem filename: STAGGERED_DIRAC_SUBSTEP1_STATISTICS_GL_F_CONDITIONAL_DISCRIMINATOR_BOUNDED_THEOREM_NOTE_2026-06-10.md.\n- Removed overclaim that the two-candidate collapse was already retained; the landed result is an explicit three-candidate conditional discriminator.\n- Replaced the stronger 'exactly single predicate / sharpest form' posture with a tested-surface boundary: GL(F) distinguishes the tied hard-core vs Grassmann/CAR frames, while the retained inputs checked here are frame-blind.\n- Added a runner cache and kept audit verdict/effective-status outputs out of the commit.\n\nChecks run: py_compile, runner PASS=20 FAIL=0, independent CAR-vs-hard-core GL(F)/CCR check, vocab_lint --fix, docs/audit/scripts/run_pipeline.sh, audit_lint --strict, stale-language scan, whitespace and local-link checks. Pipeline seeded the row as unaudited leaf bounded_theorem with deps: substep1 statistics no-go, substep1 Grassmann bridge, fermion parity Z2 grading, and minimal axioms; generated audit/status outputs were stripped before commit.\n\nClosing because the reviewed content has landed.

PR #3483, opened 2026-06-10
physics: staggered kinetic-class forcing — P-KIN/P-SD collapse to a single bit (P-FLUX), with countermodel certifying sharpness

JONbridgerapps:
## What this is
The rank-1 boundary's remaining legs: the declared premises P-KIN (naive-Dirac kinetic form) and P-SD (site-local spin diagonalization) that bound the staggered spine (~1163 movable rows). Outcome: **both premises collapse to one bit.**

## Theorem K (P-KIN collapse)
On the adjacency-licensed, Q-conserving NN bilinear surface over the U4-closed per-site ℂ²: covariance under translations + the 24 proper cubic rotations (each up to site-local U(1) frame) + nondegeneracy forces uniform |t| and uniform **real** plaquette flux (the in-plane C2 reverses plaquette orientation, so φ = φ̄), and flux classifies up to frame on simply connected regions (H¹ = 0, certified by GF(2)/integer cohomology + exhaustive 2¹² enumeration). **Exactly two kinetic classes survive**: K0 = flux(+1) (scalar tight-binding) and K1 = flux(−1) (Kawamoto-Smit class = absorbed naive Dirac).

## Theorem S (P-SD discharged on K1)
The retained U4 closure leaves no per-site spinor room (CAR(2) needs dim 4 > 2), so scalarization is *derived* as the only site-local route; the absorbing frame exists explicitly, its image is rigidly flux(−1), unique up to gauge. **P-FLUX (φ = −1) + retained set ⇒ P-SD.**

## Sharpness — the honest one-bit residual
K0 satisfies every imposed constraint and every retained separator tested (parity, hermiticity, cubic/translation invariance): the selector **φ = −1 is not forced** — the computed countermodel certifies exactly where forcing stops (the same wall the index-pairing no-go calls the kinetic-order selector: first-order Dirac vs second-order scalar). Bonus finding: dropping U4 admits the full Wilson continuum M_μ(r) = γ_μ + rI; retained U4 collapses it to the two endpoints — U4 is precisely what makes the classification two points instead of a continuum.

## Downstream effect
The Kawamoto-Smit note's B2 narrows to P-FLUX (one bit); its B3 (P-SD) can cite Theorem S as discharged given P-FLUX; the realization gate's premise table swaps P-SD for P-FLUX. Combined with PRs #3476/#3479/#3482, the staggered spine's entire premise surface is now: **P-FLUX (one bit) + GL(F) (one predicate)**.

## Runner
`TOTAL: PASS=27 FAIL=0`, deterministic, ~1 s; falsification legs D1–D5 confirm every constraint is load-bearing. No contradiction with the retained statistics-agnostic or labeling no-gos; RP deliberately not imposed (staggered-scoped → would be circular as a selector).

### DIFF
--- a/docs/STAGGERED_DIRAC_KINETIC_CLASS_FORCING_NARROW_THEOREM_NOTE_2026-06-10.md
+++ b/docs/STAGGERED_DIRAC_KINETIC_CLASS_FORCING_NARROW_THEOREM_NOTE_2026-06-10.md
@@ -1,0 +1,410 @@
+# Staggered-Dirac Kinetic-Class Forcing — Two-Flux-Class Collapse of P-KIN and Discharge of P-SD (Narrow)
+
+**Date:** 2026-06-10
+**Claim type:** bounded_theorem (narrow), with a computed sharpness
+countermodel
+**Claim scope:** On the adjacency-licensed, charge-conserving
+nearest-neighbor bilinear surface over the U4-closed per-site `C²`
+(retained authorities cited in §4), covariance under the substrate
+automorphisms (translations and the 24 proper cubic rotations, each up
+to site-local `U(1)` frame) collapses the kinetic family to EXACTLY TWO
+frame classes on simply connected regions: `K0` = uniform plaquette
+flux `+1` (representative `t ≡ 1`, scalar tight-binding) and
+`K1` = uniform plaquette flux `−1` (representative the Kawamoto-Smit
+sign system `η⁰`) — Theorem K. On the `K1` branch the site-local
+unitary absorbing frame of premise P-SD EXISTS, lands exactly in `K1`,
+and is unique up to site-local `U(1)` gauge times one global frame —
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed the salvage on main as c4732ad16. I kept the theorem narrow: two flux classes on the licensed kinetic surface, P-SD discharged only on the flux(-1) branch, and the final selector left as a one-bit residual. I removed retained-status authority language and did not land P-FLUX as a new named premise; the landed text uses the flux(-1) selector instead. Validation: runner PASS=27 FAIL=0, independent unit-cube flux/CAR(2) check passed, no-go discipline gate added for the scoped selector-not-forced claim, pipeline seeded staggered_dirac_kinetic_class_forcing_narrow_theorem_note_2026-06-10 as bounded_theorem/unaudited/leaf and ready, and strict audit lint passed. Generated audit/status outputs were stripped before commit.

PR #3484, opened 2026-06-10
physics: P1 exponent — barrier-parameter selector escapes the irreducible class and point-selects log (conditional); Lemma R extends the no-go

JONbridgerapps:
## What this is
The N7 escape-hatch attack on the P1 exponent irreducibility (the rank-4 boundary, ~786-row cone + 149 retained_bounded rows). The 05-31 no-go proved every selector in the (Add)/(Loc)/(Pot)+normalized-gradient class is equivalent to P1 itself, and named in its own steelman what would evade it: "a positivity/convexity/analyticity condition that selects log without cross-block second-derivative vanishing or additive sector decomposition." This PR delivers exactly that object — conditionally — plus the sharpened no-go from the candidates that failed.

## The selector (conditional narrow theorem)
**(NU)** — nonvanishing constant-sign curvature on R₊ plus finite barrier parameter `ν[W] := sup W'²/|W''| < ∞` (Nesterov–Nemirovski barrier calculus). On the normalized exponent family, `W'²/|W''| = |s|·z^p/|p−1|`: finite-sup **iff p = 0**, and `ν[log] = 1` exactly (canonical barrier). Subtlety handled honestly: bare self-concordance is NOT the selector (it leaks p=2); (NU) rejects it.

## Why it escapes the proven class (the audit-critical part)
- Witness family `log z + ε·cos(ω·log z)` passes (NU) but violates the additive identity at **every** nondegenerate pair — (NU) entails no additive instance, not even on a slice.
- Structural check: (NU)'s expression closure contains only one-variable data — it cannot demand (Loc).
- Not exponent-blind (contrast with the Born/normalized-gradient selector, recomputed).

## The sharpened no-go (Route-C content in the same note)
**Lemma R (single-pair rigidity)**: ONE nondegenerate additive instance already forces p=0 — so scale-free response kernels reduce to (Add) exactly; mode-product linearization of the 06-08 lemma's L1 is additive decomposition verbatim; inversion antisymmetry is the reciprocal-pair instance. Operator/complete monotonicity and concavity are interval-selectors, not point-selectors. Lemma R is now the screening gate for all future N7 proposals.

## The honest boundary
The (NU) premise is **unlicensed** — a ledger scan (runner T10) finds zero retained-grade rows supplying barrier/curvature structure. P1/T1-d is NOT closed. What changes is the premise's *shape*: from a two-block additivity postulate to a **one-block convexity/regularity postulate** — the disjoint-blocks clause stops being load-bearing for exponent fixing, and Record additivity becomes a derived consistency check. The single named target for full retirement: any future retained row supplying bounded-curvature/convexity structure of the readout (RP/transfer-kernel rows are the natural hunting ground).

## Runner
`TOTAL: PASS=49 FAIL=0`, exact SymPy throughout, deterministic, ~0.8 s; falsification legs include compact-domain collapse (the full-R₊ domain clause is load-bearing, identical to the already-declared T1-d/L3 clause).


jonathonreilly:
Review-loop landed the salvageable source-only theorem as commit 8884cd385.\n\nLanded scope: bounded theorem / conditional selector-existence result for the P1 exponent barrier-parameter selector, with (NU) kept explicitly unlicensed and not adopted as a primitive, axiom, Tier-A admission, or P1 closure. I also converted the load-bearing citations to graph-visible markdown links, refreshed the runner cache, and validated the row through the audit pipeline/strict lint without committing generated audit outputs.\n\nValidation: runner PASS=49 FAIL=0; independent math cross-check passed; vocab lint clean; pipeline validation + strict audit_lint clean. The seeded row remains unaudited for the independent audit lane.

PR #3493, opened 2026-06-10
physics: P1 (NU) license hunt — no retained supplier (computed witnesses); demand reduced to one clause (BR)

JONbridgerapps:
## What this is
Derive-out campaign, lane 3: attempt to license the barrier-parameter selector (NU) from the retained surface, so the conditional P1 exponent theorem ("Add P1 exponent barrier selector theorem", landed today) becomes a P1 retirement. Method per the Tier-A retirement precedent (AC_φλ, θ): split the admission into mechanically checkable clauses, hunt clause-by-clause for adjacent retained structure, ship the residual strictly smaller.

## Outcome: not licensed — but the demand provably shrinks
1. **No-go with computed witnesses**: zero retained-grade rows supply the curvature clauses, proved structurally (every retained one-hop row is amplitude-side/readout-blind, probability-path-conditional, or a no-go). Witness `W_G = log z + (log z)³` satisfies every retained constraint while having sign-changing curvature (kills clause ii); `W_Q = g₂` kills the finite-barrier clause independently.
2. **Fisher route killed at full steelman**: even granting both unlicensed bridges gratis, the retained Fisher row's own chart readout `log((z+1/z)/2)` has sign-changing curvature and infinite barrier parameter — the route *violates* (NU) rather than supplying it.
3. **Lemma N (the campaign yield)**: `(NU) ⟹ sup|z·W'| ≤ ν[W]` (sharp at log: 1 = 1), and the weaker single clause **(BR) — bounded log-scale response** — *alone* point-selects p = 0 and escapes the extended irreducible class (witness family recomputed). Ladder: `(Add) ⟹ (NU) ⟹ (BR)`, all strict.

## What the P1 premise is now
Monotone shrink across today's work: disjoint-records additivity → (NU) (two curvature clauses) → **(BR)** (one clause: `z·dW/dz` uniformly bounded on the already-declared T1-d domain). Next named target: a retained (BR) license — candidate shape per the precedent method: a finite-resolution/record-capacity adjacent structure — or governance ratification of (BR).

## Firewalls
Count-probability firewall: no probability law constructed anywhere (Fisher facts appear only as comparator content inside the route-kill). Record-scalar-map no-go: no readout identified (the bridge appears only hypothetically in the steelman and fails anyway). Both stated in §5 and string-checked by the runner.

## Runner
`TOTAL: PASS=38 FAIL=0`, exact SymPy, deterministic (byte-identical double run), <1 s; expected-output block matches stdout exactly.


jonathonreilly:
review-loop landed this as 4a0e3cd58. I salvaged the durable core and narrowed the source note: NU => BR, BR alone selects the logarithmic exponent on the tested family, and the linked current supplier candidates do not supply NU/BR. I removed the stale/broad source prose and exact audit-verdict dependency in the runner; BR remains an open license and P1 is not retired. Checks: runner PASS=38 FAIL=0 with cache, vocab_lint clean, links resolve, py_compile clean, run_pipeline.sh complete, audit_lint --strict OK (notices only), diff checks clean.

PR #3495, opened 2026-06-10
physics: P-FLUX matter-content route — sharpened no-go with a surprising witness: K0 secretly carries the generation structure

JONbridgerapps:
## What this is
Derive-out campaign, lane 2: attempt to force the kinetic-class bit φ = −1 from the matter-content/observable sector (the route the index-pairing no-go's selector catalog never enumerated). Outcome: **sharpened no-go — with the most informative witness of the campaign.**

## The surprising computed witness
The scalar branch K0 does NOT lack the generation structure: at 4|L (PBC), K0's zero-mode sector contains exactly one embedded Klein-cube carrier at momenta (±π/2)³ that passes **every retained carrier requirement identically to K1** — full dressed-character cube multiplicity 1, Hamming grading 1+3+3+1, C3[111] cycling, exact hw=1 characters, commutant dimension 3, M₃(ℂ) generated (no proper quotient ⇒ observable-stable count 3), and the c-involution realized by the same sublattice chirality. So "three generations" alone cannot select the branch.

## Why the obvious route was also blocked at status level (circularity discipline held)
All discriminating-battery rows are **carrier-conditional abstract theorems** — none is a branch-neutral requirement; the realization bridge (BZ-corner forcing) is unaudited; the single-clock theorem is unaudited; retained RP is staggered-scoped (circular as a selector). No turn-1 inversion was used (nothing forced from the matched-3=3 count).

## The campaign yield: three named escape conditions with computed separators
The only computed separators between the branches are:
- **kernel extensivity**: K0's massless sector grows 20→68 with L; K1's stays 8→8;
- **carrier exactness**: K1's kernel IS the corner cube; K0's kernel strictly contains it plus 12 extra massless modes.

Hence B-BIT survives but with a precise spec (§7/N6): a retained-grade **point-like-zero-set** requirement, a retained **kernel-equals-carrier** ("no extra massless sectors") requirement, or a branch-neutral retained positivity/transfer theorem — **any one of these now selects φ = −1 immediately** given these certificates.

## Runner
`TOTAL: PASS=36 FAIL=0`, deterministic, <1 s. Non-vacuity leg (anisotropic comparator fails at the C3 stage), frame-robustness (dressed witness under generic gauge), shared wrap-sensitivity (L=6 PBC kills the carrier on both branches), residual-matching leg certifying no new wall beyond B-BIT.


jonathonreilly:
Review-loop landed the durable no-go content in f1d755622 as a narrowed source note. Changes from the PR: removed campaign/status/ledger language; scoped the claim to the finite P-FLUX matter-content route; kept the computed K0 embedded-carrier witness; made the N1-N8 no-go discipline gate explicit; added a runner cache. Verified runner PASS=36 FAIL=0, vocab lint, pipeline, and strict audit lint. Audit status remains independent/unaudited.

PR #3498, opened 2026-06-10
physics: P1 (BR) license hunt — record-capacity route fails with witnesses; demand reduced to (BR-int), supplier spec minimal: a per-e-fold registration-rate cap

JONbridgerapps:
## What this is
Derive-out campaign wave 2, P1 lane: hunt the (BR) supplier in the retained record-capacity rows. Outcome: **sharpened gap — the candidate theorem is false, with computed witnesses — but the ladder gains a strict rung and the missing clause is now a single named rate cap.**

## Results
1. **Lemma W (theorem)**: (BR) ⟹ (BR-int) — bounded e-fold *increments* — strictly weaker (exact witness with bounded increments but unbounded z·W′ along z_m = 2πm), and (BR-int) still point-selects p = 0 and escapes the extended irreducible class. Ladder now: `(Add) ⟹ (NU) ⟹ (BR) ⟹ (BR-int)`, all strict.
2. **Lemma C (conditional capacity theorem)**: finite-sector realization + magnitude cap M + per-e-fold count cap K ⟹ increments ≤ K·M ⟹ (BR-int) ⟹ log. The retained unit-record schema supplies M = 1 by its own normalization.
3. **The no-go**: the two remaining capacity sub-clauses fail independently against the retained rows — a single sector can carry an unbounded datum (the finite-sector row's own freedom identity, recomputed, proves sector data are arbitrary), and 4^k unit records per e-fold is *affirmatively licensed* by the retained unbounded-additivity schema. Route B (Busch/Gleason/local tomography) killed: the qubit effect bound is magnitude-shaped only, probability-conditional (firewall-blocked), and readout-blind.

## The minimal supplier spec (the campaign yield)
**(CAP-real) + (CAP-K)**: a retained per-e-fold registration-rate cap for the T1-d readout realization. Granting it, log selection completes via Lemma C. Notably the spec is honest about (CAP-real) being a quantitative slice of the record_scalar_map "middle arrow" — declared as part of the open premise, not smuggled.

## Firewalls
No probability law constructed (Busch hypotheses flagged as the blocked bridge); no branch-to-scalar map asserted; scoping sentence present in the note and string-checked by the runner (T10).

## Runner
`TOTAL: PASS=31 FAIL=0`, exact witnesses, expected-output block matches stdout byte-for-byte, deterministic, ~1.5 s.

### DIFF
--- a/docs/OBSERVABLE_PRINCIPLE_P1_BR_LICENSE_FROM_RECORD_CAPACITY_NARROW_NO_GO_NOTE_2026-06-10.md
+++ b/docs/OBSERVABLE_PRINCIPLE_P1_BR_LICENSE_FROM_RECORD_CAPACITY_NARROW_NO_GO_NOTE_2026-06-10.md
@@ -1,0 +1,580 @@
+# Observable-Principle P1 — The (BR) License Has No Record-Capacity Supplier (Narrow No-Go), the Demand Reduces Further to the Increment Clause (BR-int), and the Missing Record Clause Is Exactly a Registration-Rate Cap
+
+**Date:** 2026-06-10
+**Claim type:** narrow no_go (license-absence over the retained
+record/measurement surface, with computed witnesses splitting the capacity
+clause into two independently failing sub-clauses) + two narrow positive
+lemmas (Lemma W second demand reduction `(BR) => (BR-int)`; Lemma C
+conditional record-capacity theorem).
+**Claim scope (narrow):** the single open clause left by the (NU)-license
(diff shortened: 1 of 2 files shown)


jonathonreilly:
review-loop landed this as c93254961. I narrowed the source note to the durable result: BR => BR-int, BR-int still selects the logarithmic exponent, and record-capacity would close the route only with a supplied realization plus per-e-fold registration-rate cap. I removed the stale/status-heavy framing; BR/BR-int/CAP are not licensed and P1 is not retired. Checks: runner PASS=31 FAIL=0 with cache, vocab_lint clean, links resolve, py_compile clean, run_pipeline.sh complete, audit_lint --strict OK (notices only), diff checks clean.

PR #3491, opened 2026-06-10
feat(gauge-dynamics): induced holonomy — the composite link's loops are a derived gauge-covariant state functional (central on the sea, non-central off it)

jonathonreilly:
## Summary

After the orbit-converse ([#3487](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3487): frame fields supply **only** the flat sector), this note exhibits the holonomy supplier the framework already contains: **the composite link** (#3398, `U_eff = polar(M(x,y))`). Its loops are a **derived, gauge-covariant functional of the matter state**. The ST2 holonomy demand is **addressed conditionally at the kinematic/state-functional level** — dynamics slaved, identification gate open. **Panel-narrowed** (maximum-tier 5-lens panel, `land_with_edits`, all eleven edits applied). `TOTAL: PASS=12 FAIL=0` (exact, no MC).

## New content

- **The curvature scalar** `C = 1 − |tr Hol|/3` — conjugation **and** center invariant (the U(1)/det phase it quotients is a *named open thread*).
- **The two-pole dichotomy:** sea and sea-orbit states induce **central** holonomy (`C = 0` exactly — the sea induces the flat sector); off-sea states have **`C > 0` strictly**, with filling-dependent magnitude (near-half-filling mean ≈ 0.7; minimal occupancy ≈ 0.2; near-sea excitations stay ~flat — all decimals seed/filling-specific).
- **Creation from flat, recalibrated:** from non-stationary flat data the derived flow leaves the flat stratum **tangentially** (quartic liftoff, Richardson order 4.0 verified; `O(0.1)` only at `t = O(1)`; almost-periodic excursion, not relaxation). The stationary sea stays exactly flat.

H2/H3a/H4 are corollaries of the landed block-01 trajectory note (cited); H4 rebuilt with **panel-fixed valid one-body densities** (the draft's invalid perturbation replaced; `0 ≤ eig ≤ 1` asserted in-runner).

## Mandatory reconciliations

The **retained** primitivity note's object is the single-hop color **transporter** (verified in-runner: `scalar·I₃`, color-inert) — a different functional; its own steelman *explicitly leaves the composed-link route open*, and this note walks through that door. The native-plaquette `retained_no_go` concerns the **operator** loop — different object. The pack block-12 probe was scoped to fed-in links + sea-like states (pack-internal; nothing landed contradicted).

## Honest limits (panel-verified)

[...]

PR #3486, opened 2026-06-10
feat(gauge-dynamics): open-shell selection restructures — conditional on the invariant locus, neutrality with zero further admissions; no derived selector

jonathonreilly:
## Summary

**Retire-mode continuation of the admission-(B) program** (#3458 → #3474 → #3478): restructures the open-shell residual (*"which `ρ_color` spectrum on the degenerate ground manifold"*, #3474-T3b). **Panel-gated** (advances (B); walks the G3 guard): 4-lens mandatory adversarial panel, `land_with_edits`, **all five required edits applied**. `TOTAL: PASS=17 FAIL=0` (exact, no MC). **Which locus the realized state occupies is the full remaining residual and is NOT resolved here.**

## The theorems

- **(T1) Invariant-locus universality** (panel-upgraded to a rigorous decomposition): the deg-20 manifold = **4·singlet ⊕ 2·octet** (Casimir `0×4`/`12×16`; commutant dim exactly `4²+2²=20`, computed); **every** invariant density on it — full commutant basis + random invariant PSD states — is neutral at **every** site (`3×10⁻¹⁵`). The #3474-T3b obstruction is a property of non-invariant pure selections, not of the manifold. No thermodynamic limit or SSB folklore anywhere.
- **(T2) No derived selector:** exact equivariance of the named hopping, color-blind instruments, and conservation laws; the invariant locus is **dynamically stable** through interleaved Hamiltonian+record steps (`10⁻¹⁷`); color-blind records preserve a non-invariant state's per-site `ρ_color` **exactly** (preservation only — blocks 05–07 consistent). Underived in **both** directions.
- **(T3) The registrable order parameter** `D` = worst-site purity excess: manifestly SU(3)-invariant (a max-entry-norm draft form was *not* unitarily invariant — caught by the runner's own check and replaced), two-copy estimable, **continuous** (convex sweep to 0 verified) — *the invariant locus vs its complement under a continuous order parameter, **not a binary***. Faithfulness fix: a manifold state with site-0 excess an order of magnitude below its all-site `D` is constructed — **the separator must be all-site**.
- **(T4)** The equipartition witness `P_gs/20` is **existence-only** and **not a twirl in disguise** (the broken exhibit's group average has a non-flat spectrum, computed); frame-naming instruments are the named exception (#3478 relative data); all decimals labeled instance/basis artifacts.

## Scope discipline

[...]

PR #3474, opened 2026-06-10
feat(gauge-dynamics): Pauli/filling forces the color marginal I₃/3 in closed-shell sectors — the discrete reduction of admission (B)

jonathonreilly:
## Summary

**Retire-mode attack on admission (B)'s purity core** (owner-directed; very high bar — no new primitive/axiom/measure/weight). The mechanism the whole campaign never used: **the matter is fermionic.** Panel-gated: a mandatory 4-lens adversarial panel returned `land_with_edits`; **all edits applied** — the decisive one narrows the discharge to **closed shells** and names the open-shell residual. `TOTAL: PASS=16 FAIL=0` (exact, no MC).

## The theorems

- **(T1)** The 3-fermion sector of three color modes is **one-dimensional** — the color singlet is *forced* by Pauli (all 8 su(3) charges annihilate it, residual 0; `G=I₃` exact). The occupancy-forced second-quantized sharpening of the baryon `ε_abc` fact.
- **(T2)** That sector is hopping-frozen (honest catch).
- **(T3) Closed-shell discharge** — genuine many-body, de-circularized: the fixed-N ground state of the named color-diagonal free hopping is **unique** (gap 3.0 verified) and the **measured** color-resolved one-body matrix (all 9 entries, every site, from the actual ground state) has coherences exactly zero (`2.6×10⁻¹⁷`) and `ρ_color(x)=I₃/3` (`10⁻¹⁶`). **Block 04's necessary condition for ADM-2 holds there with zero admissions.**
- **(T3b) Open-shell failure, earned:** the degenerate ground manifold (deg 20) contains `ρ≠I₃/3` states *at the ground energy* (dev 0.21); **Z³ half filling is generically open-shell** (cubic Fermi degeneracy 12/20 at L=3/4). The open-shell selection is the **named residual — weight-like; the G3 guard re-opens there.**
- **(T4) Discrete reduction, two conditions:** sharp-count coherences vanish exactly; `I₃/3 ⟺ count equality AND profile agreement` (both teeth exhibited). And `ρ=I₃/3` is **strictly weaker** than global neutrality (block-17 Fact 3 reproduced, charge residual 1.63).

## Admission (B)'s ledger after this note

```
orientation                      RETIRED   (#3458)
purity, closed-shell sectors     DISCHARGED here (zero admissions; measured)
purity, exact full filling       FORCED (T1), at the price of frozen dynamics (T2)
purity, open shells (generic Z³) NAMED RESIDUAL — degenerate-manifold selection, weight-like
[...]

PR #3457, opened 2026-06-10
feat(gauge-dynamics): slow-variable premise discharges in chosen-enlargement form — the closing set is derived

jonathonreilly:
## Summary

**Retire-mode follow-on (block 15)** to the gauge-dynamics campaign: the lever-space map's route-(i) *"added slow variables"* promotion premise **discharges in its chosen-enlargement form** — nothing is admitted, because the closing variable set is **derived**. `TOTAL: PASS=8 FAIL=0` (exact, full `2⁶` Fock space, no MC).

## What is proved

- **(T1) Operator-level closure, state-independent:** for the named quadratic hopping generator, the Heisenberg flow of one-body bilinears stays exactly in the bilinear span (`10⁻¹¹`; derived convention `G(t)=conj(W)G Wᵀ` at `10⁻¹⁶` on a non-Gaussian state). The "slow variables" are the rest of the one-body density `G` — **forced by the quadratic structure, not chosen**.
- **(T2)** The campaign's record channels act on `G` by **exact one-body rules** (I-A re-verified `10⁻¹⁷`; I-B cited from block 02) — the interleaved flow is closed at the `G` level.
- **(T3)** The link `U_eff=polar(G_xy)` is a **lossy coordinate**: the Sylvester law evaluated on the **true derived generator** (`Ġ=i[hᵀ,G]`) matches the actual evolution's finite difference, **with a wrong-generator control** (the reviewer caught a generator-circular draft check — replaced; *runners must derive, not hard-code*). And the registered candidate `(U,Q)∼M` does **not** close on its own. Block-01's non-autonomy = **coordinate non-autonomy**; the link acquires **no autonomous law** (consistent with block 03).
- **(T4) Teeth:** a quartic term breaks the closure (residual `0.23`) — the closure is a property of the quadratic hopping, not generic.

## Boundaries

Pointer-frame admission `{P_r}` and the stratification theorem **untouched**; the realization gate ("the physical dynamics *is* this `H`") a separate lane; conditional on the supplied `C³` carrier. Adversarially reviewed pre-landing (`land_with_edits`; both required + all recommended edits applied; every identity independently recomputed at `10⁻¹⁶`). The audit lane grades all status.

## Changes (source-only triple)

- `docs/SLOW_VARIABLES_DERIVED_ONE_BODY_CLOSURE_LINK_IS_A_COORDINATE_BOUNDED_THEOREM_NOTE_2026-06-09.md`
[...]

PR #3499, opened 2026-06-10
feat(gauge-dynamics): interleaved mean map — the peripheral count collapse; the #3491 almost-periodicity obstruction removed at the mean level

jonathonreilly:
## Summary

**Retire-mode assault on the R1/R2 pair.** Composes the two landed exact one-body rules (#3457 flow + block-02 record channel) into one exact 81×81 linear map and analyzes it spectrally. **Panel-narrowed** (4-lens mandatory panel, `land_with_edits`, all ten edits applied). `TOTAL: PASS=14 FAIL=0` (exact, deterministic, no MC). **Explicitly not an R2 delivery**: no step-measure on SU(3) is supplied — the central finding (convergence to a delta) is the *opposite* of the CLT spread, and the note says so.

## New load-bearing content

- **(M1) The peripheral count collapse:** the map's peripheral spectrum is **exactly 3-dimensional** — the conserved color counts (a 9→3 collapse driven by the color-blind hop). The gap itself is per-step decoherence damping (single-color-toy reproducible; values instance-labeled; `λ,τ` are supplied parameters of the named instrument admission). The matter mean relaxes to the **count form** (#3474-T4; the #3486 clause in count form); the carrier dies at the derived per-step ratio.
- **(M3) The removal of the #3491 almost-periodicity sub-obstruction at the mean level:** on **deflated genuine signal** (the panel's decisive catch — plain renormalization lets μ=1 round-off capture the iterate by n≈100; the draft's original numbers were noise), the induced curvature and orientation **converge**, membership in the rank-3 dominant cross-image holds (guard tightened; control power disclosed as dimension-counting), and **the convergence rate is itself derived** (log-slope −0.0176 vs ln(|μ₂|/|μ₁|) = −0.0172 — the deflation *restored* the sharp prediction). **Corpse-angle qualified:** the fixed point is diagonal — no link survives; the convergent object is the orientation of the vanishing transient. The limit is one realization's number (state-realization data).

M2/M4 are one-body re-exhibits of block-02 (cited); #3436 and block-03 cited as nearest prior art.

## Where this leaves R1/R2

Removed: the #3491 sub-obstruction (no relaxation/no stationary configuration at the mean level). Untouched: the retained R1 boundaries (discrete-time throughout) and R2's step-measure premise. Doors named: the stochastic unraveling (where a step distribution would live — needs outcome-weight/Born structure), `{P_r}` instruments, interactions. The audit lane grades.

## Changes (source-only triple)

- `docs/INTERLEAVED_MEAN_MAP_PERIPHERAL_COUNT_COLLAPSE_ALMOST_PERIODICITY_REMOVAL_BOUNDED_THEOREM_NOTE_2026-06-10.md`
- `scripts/frontier_interleaved_mean_map_peripheral_count_collapse_relaxation_2026_06_10.py`
- `logs/runner-cache/frontier_interleaved_mean_map_peripheral_count_collapse_relaxation_2026_06_10.txt`

### DIFF
--- a/docs/INTERLEAVED_MEAN_MAP_PERIPHERAL_COUNT_COLLAPSE_ALMOST_PERIODICITY_REMOVAL_BOUNDED_THEOREM_NOTE_2026-06-10.md
+++ b/docs/INTERLEAVED_MEAN_MAP_PERIPHERAL_COUNT_COLLAPSE_ALMOST_PERIODICITY_REMOVAL_BOUNDED_THEOREM_NOTE_2026-06-10.md
@@ -1,0 +1,105 @@
+# The Interleaved Mean Map: the Peripheral Count Collapse, and the Removal of the #3491 Almost-Periodicity Obstruction at the Mean Level
+
+**Date:** 2026-06-10
+**Type:** bounded theorem (retire-mode; owner-directed assault on the R1/R2 pair; panel-narrowed)
+**Claim type:** bounded_theorem
+**Script:** `scripts/frontier_interleaved_mean_map_peripheral_count_collapse_relaxation_2026_06_10.py`
+**Cache:** `logs/runner-cache/frontier_interleaved_mean_map_peripheral_count_collapse_relaxation_2026_06_10.txt`
+**Status:** source proposal; the audit lane grades. Runner `PASS=14 FAIL=0`, exact,
+deterministic, no MC. A mandatory 4-lens adversarial panel returned `land_with_edits`;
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed as d85a3b1d8 after review-loop narrowing. Kept the bounded mean-map theorem, removed PR/process/panel language, made source-note dependencies graph-visible, and preserved the no-generator/no-step-measure boundary. Runner/cache pass and audit compatibility pipeline/strict lint passed; audit lane remains independent.

PR #3500, opened 2026-06-10
physics: P-FLUX point-zero-set hunt — restatement unlicensed AND false on K1 itself; three escapes collapse to one clause (ker = carrier)

JONbridgerapps:
## What this is
Derive-out campaign wave 2, P-FLUX lane: attempt to derive the "point-like-zero-set / kernel-equals-carrier" requirement that would select φ = −1 given the separators certified in the parent no-go (PR #3495). Outcome: **sharpened gap, with a genuinely new computed dichotomy that kills the tempting route for good.**

## The load-bearing finding: the kernel-global restatement is dead in both directions
- **Textually unlicensed**: the retained count rows confine themselves to the C³ carrier ("does not assert the full physical staggered-carrier realization"); machine-verified against the rows' text.
- **Mathematically FALSE on the staggered branch itself**: K1's own full kernel is reducible under the restricted algebra (commutant dim 4 — the 1+3+3+1 Hamming blocks), and kernel-global species counts are 8 characters / 4 orbits on K1 vs 20/8 on K0 — **never 3 on either branch**. Carrier-conditional reads tie TRUE/TRUE; kernel-global reads tie FALSE/FALSE. No textual relaxation rescues the restatement without changing the rows' mathematical content.

## Other suppliers killed (one line each)
Retained clustering row: gap-conditional by its own scope; under the shared staggered mass m·ε both branches gap to exactly m — branch-neutral both ways. LR data: branch-identical. F5/log-quasilocality row: unaudited, bilinear-sector-scoped, rate degenerates at m = 0 — circular. Record rows: conditional on supplied decompositions, bind nothing about kinetic kernels.

## The campaign yield
The parent's three escape conditions collapse to **one missing clause**: a retained-grade `ker = carrier` requirement (≡ no-extra-massless-sectors ≡ point-like zero set, given the parent's certificates). The real separating observable is now precisely characterized and consumed by no retained row (declared residual **B-Z2**): DOS at E = 0 is 236 vs 8 modes at L = 12; K0's zero count grows 20→68→140 (exponent ≈ 1.78) vs K1's constant 8.

## Guards
Turn-1 inversion guards of the index-pairing no-go certified computationally (no kernel-global count readout equals 3; the count is never used as a forcing input). Falsification legs: the candidate requirement selects exactly K1 among {K0, K1, point-zero scalar comparator} — non-vacuous, and point-like ≠ first-order; dropping the non-retained clause restores the tie.

## Runner
`TOTAL: PASS=22 FAIL=0`, deterministic (byte-identical double run), ≈2 min.

### DIFF
--- a/docs/P_FLUX_POINT_ZERO_SET_FROM_RETAINED_ROWS_NARROW_NO_GO_NOTE_2026-06-10.md
+++ b/docs/P_FLUX_POINT_ZERO_SET_FROM_RETAINED_ROWS_NARROW_NO_GO_NOTE_2026-06-10.md
@@ -1,0 +1,508 @@
+# P-FLUX Point-Zero-Set Requirement — No Retained Supplier: Every Reading of the Retained Count/Clustering Rows Is Branch-Neutral (Narrow No-Go, Sharpening Escapes (i)/(ii) to One Clause)
+
+**Date:** 2026-06-10
+**Claim type:** no_go (narrow), carried by a computed two-sided
+dichotomy certificate, textual license audits of the candidate
+retained suppliers, and two falsification legs
+**Claim scope:** On the kinetic-class forcing note's licensed
+two-class surface (`K0` = uniform plaquette flux `+1`, scalar
+tight-binding; `K1` = uniform flux `−1`, Kawamoto-Smit class), the
+matter-content no-go's sharpened escape conditions (i)
+(retained point-like-zero-set requirement) and (ii) (retained
+kernel-equals-carrier / no-extra-massless-sector requirement) have
+**no retained supplier**, and — the sharpening — the retained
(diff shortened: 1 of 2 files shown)


jonathonreilly:
review-loop landed this as c2044cac5. I kept the finite point-zero/kernel-exactness computation, but narrowed the claim to a candidate-supplier no-go: the linked spectral/clustering, count/no-proper-quotient, and Record/readout candidates do not supply the missing selector. I removed the stale status-kill framing after the log-transfer/microcausality surfaces changed, left transfer/positivity as a separate live route, and added the runner cache. Checks: runner PASS=22 FAIL=0, vocab_lint clean, markdown links resolve, py_compile clean, run_pipeline.sh complete, audit_lint --strict OK (notices only), diff checks clean.

PR #3505, opened 2026-06-10
physics: CAP-K DERIVED from finite-speed registration — the P1 rate cap is a Lieb-Robinson cone theorem (conditional)

JONbridgerapps:
## What this is
Wave 3 of the P1 exponent campaign: the registration-rate cap (CAP-K) — the clause the BR-license no-go (PR #3498) identified as the single missing supplier — **derived as a conditional theorem from the retained finite-speed dynamics**. Every wave-2 kill was static; this is the dynamical route, the same structural move that derived GL(F).

## The theorem
Within a declared 4-clause finite-speed registration realization class — (REG-dyn) records established by the retained (F4) hopping dynamics, source coupling in a bounded region; (REG-τ) at most a supplied clock window per e-fold; (REG-thr) registration = operator-norm δ-sensitivity; (REG-site) disjoint records on disjoint sites — Duhamel + the proved LR lemma give a sensitivity cone `D* = v_LR·τ + ln((e/(e−1))·‖V‖/(v_LR·δ))`, hence **(CAP-K): K ≤ (s_X + 2⌈D*⌉)³·log₂(d_site)** — canonical Z³ instance: v_LR = 24e, D* ≈ 63.8, K = 129³ = 2,146,689 — **k-uniform by symbol inspection** (the e-fold index provably absent from D*). Lemma C then completes: (BR-int) with constant K ⟹ pass set exactly {p = 0} ⟹ W = c·log z, conditionally.

## The decisive falsification legs
- The schema-licensed 4^k-records-per-e-fold family (which defeated wave 2) **requires unbounded clock windows**: first violation at k = 11 for τ = 1; τ ≥ 8036 by k = 30; the weakest joint-sector reading fails at k = 1,073,345. The license is untouched — the family is disarmed *dynamically*: it cannot be realized by any finite-speed process at bounded window.
- One long-range bond violates the cone bound ×3313; symbolically lim_{v→∞} D* = ∞ — the finite LR speed is the load-bearing physics.

## No-go compliance (quoted, escaped by input type)
The family-lift no-go prunes "finite post-record certificate alone ⇒ unbounded retained law" and names its reopening inputs, first among them "a **supplied law**" — exactly what (REG-dyn)+(REG-τ) is; no post-record certificate is read anywhere; k-uniformity is a symbol check, not finite-prefix extrapolation. Clock window stays declared per the clock/rate retained no-go (single-clock Stone is retained and licenses only the evolution map). Firewalls: (REG-thr) is pure operator-norm distinguishability — no probability law, no branch-to-scalar map, (CAP-real) remains declared. All string-checked by the runner.

## Provenance note
The BR/NU prerequisite notes live on in-flight PR branches (#3493/#3498); their consumed facts are pinned by provenance and **recomputed in-runner** — no check depends on their file presence (merge-stable in any landing order).

## What P1 becomes
T1-d + (CAP-real, declared) + (REG class, declared) ⟹ (CAP-K, computed) ⟹ (BR-int) ⟹ log selected. The open premise shifts from a bare unsupplied rate cap to a named physical process clause; P1 is not retired, but its last quantitative clause is now carried by Lieb-Robinson physics.

## Runner
`TOTAL: PASS=31 FAIL=0`, deterministic, ~13 s; measured sensitivity ≤ derived bound on an explicit 10-site chain (worst ratio 0.030) and a Z³ 2×2×2 block (0.027); expected-output block byte-matches stdout.

### DIFF
--- a/docs/OBSERVABLE_PRINCIPLE_P1_CAP_K_FROM_FINITE_SPEED_REGISTRATION_NARROW_THEOREM_NOTE_2026-06-10.md
+++ b/docs/OBSERVABLE_PRINCIPLE_P1_CAP_K_FROM_FINITE_SPEED_REGISTRATION_NARROW_THEOREM_NOTE_2026-06-10.md
@@ -1,0 +1,673 @@
+# Observable-Principle P1 — (CAP-K) From Finite-Speed Registration: the Lieb-Robinson Sensitivity Cone Caps Registrations Per E-Fold (Narrow Theorem, Conditional on a Declared Registration Realization Class)
+
+**Date:** 2026-06-10
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Claim scope note:** conditional theorem with four declared realization
+clauses (none asserted as framework-forced), plus two computed
+falsification legs and one named open refinement.
+**Claim scope (narrow):** the single rate clause left open by the
+(BR)-license no-go note
+`OBSERVABLE_PRINCIPLE_P1_BR_LICENSE_FROM_RECORD_CAPACITY_NARROW_NO_GO_NOTE_2026-06-10.md`
+(wave 2 of this campaign; in-flight on branch
+`claude/science-fix/p1-br-license-record-capacity-20260610`, commit
+`c116993cf`, not yet merged into this branch's tree — every consumed fact
+is therefore **recomputed in this note's runner**, none cited blind) —
+the premise
+
+```text
+(CAP-K)  a record register registers boundedly per e-fold of source
+         change: every e-fold's registered disjoint record collection
+         has at most K sectors, with K uniform over e-folds,
+```
+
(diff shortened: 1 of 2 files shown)


jonathonreilly:
review-loop landed this as 66f4b5c3b. I narrowed it to a conditional bounded theorem: inside the declared finite-speed registration realization class, the LR sensitivity cone gives CAP-K; CAP-real and REG clauses remain declared, and P1 is not retired. I removed stale branch/status framing and kept the guardrails around probability, branch-to-scalar, record formation, and supplied clock window. Checks: runner PASS=31 FAIL=0 with cache, vocab_lint clean, links resolve, py_compile clean, run_pipeline.sh complete, audit_lint --strict OK (notices only), diff checks clean. The runner is D-dominant because falsifier/comparator legs are heavy, but it is not decoration and the row seeds as unaudited bounded_theorem.

PR #3506, opened 2026-06-10
physics: B-Z2 species-density hunt — naive route proven false two-sided; the supplier is a Stefan-Boltzmann quantifier promotion

JONbridgerapps:
## What this is
Wave 3, P-FLUX lane: hunt the retained supplier for B-Z2 ("massless species density is finite"). Outcome: **sharpened spec** — all three routes fail at precisely located, computed steps, and the spec lands on one maximally concrete missing row with a named, already-existing candidate supplier.

## The honest two-sided result
The mission's own candidate theorem is **false as naively stated**: the per-volume IR limit of (1/V)·log det(h²+m²) is finite on BOTH branches (2N₀/V → 0 on both) — there is no per-volume zeta divergence for a retained determinant row to consume. What actually degenerates on K0, computed:
- det vanishing order: bounded (8,8,8) on K1 vs extensive (20,56,68) on K0;
- **the Stefan-Boltzmann object itself**: K1's g_eff(T) plateaus at 0.979 ≈ 1 (= 8 species per v³) while K0's diverges as T⁻² (Fermi-surface Sommerfeld behavior, u/T² ≈ 0.233) — the thermal clause is strictly stronger than a bare point-like-zero-set requirement;
- zero-set geometry: exactly flat tangent directions (K0) vs isotropic cones (K1).

## Why each retained surface fails (quantifiers quoted in the note)
The Matsubara family is staggered-grounded AND gapped on its own surface (its normalization clause never meets a zero mode); det-positivity Case A is m>0-conditional and its mechanism is computed branch-neutral; the Berezin identity holds at det = 0; Jacobi requires invertibility; the heat-kernel and g* rows bind only *supplied* paths/inventories. Route B (emergent isotropy) is honestly circular: the retained isotropy core is derived ON the staggered lattice, and the retained LV package itself carries a K0-class graph-Laplacian as an equally legitimate carrier.

## The named path to φ = −1 (the campaign yield)
The most natural supplier already exists in the tree: the **axiom-first Stefan-Boltzmann rows (two, both unaudited)**. Promoted to retained grade with the quantifier on the *realized kernel*, the row supplies clause (CL) and selects φ = −1 **immediately** via the computed F-3 certificate (g_eff plateau vs divergence). This is the same configuration that won U4 and the spine collapse: retained no-go + existing unaudited supplier + a quantifier fix.

## Runner
`TOTAL: PASS=23 FAIL=0`, deterministic, ~6 s. Falsification legs: (CL) selects exactly K1 among {K0, K1, comparator} as a single clause; dropping (CL) restores the retained-surface tie; turn-1 inversion guards certified (kernel-global counts are 8/4 vs 20/8 — never 3, and never used as a forcing input).

### DIFF
--- a/docs/P_FLUX_FINITE_SPECIES_DENSITY_FROM_DETERMINANT_MATSUBARA_SURFACE_NARROW_NO_GO_NOTE_2026-06-10.md
+++ b/docs/P_FLUX_FINITE_SPECIES_DENSITY_FROM_DETERMINANT_MATSUBARA_SURFACE_NARROW_NO_GO_NOTE_2026-06-10.md
@@ -1,0 +1,638 @@
+# P-FLUX Finite-Species-Density Requirement (B-Z2) — No Retained Supplier on the Determinant/Matsubara, Thermal, or Isotropy Surfaces; the Degenerating Objects Computed and Named in Retained Currency (Narrow No-Go)
+
+**Date:** 2026-06-10
+**Claim type:** no_go (narrow), carried by a supplier audit of the full
+retained determinant/Matsubara/heat-kernel family (the one retained
+structure class that consumes FULL spectral density rather than
+carrier slices), of the retained isotropy/Lorentz rows, and of the
+thermal/g*/entropy rows — plus four new computed degeneracy
+certificates that make the missing supplier maximally concrete
+**Claim scope:** On the kinetic-class forcing note's licensed
+two-class surface (`K0` = uniform plaquette flux `+1`, scalar
+tight-binding; `K1` = uniform flux `−1`, Kawamoto-Smit class), the
+single live escape clause left by the point-zero-set no-go — its
+boundary B-Z2, "the massless species density is finite"
+(equivalently `ker = carrier`; equivalently point-like zero set) —
+has **no retained supplier** in the determinant/Matsubara/heat-kernel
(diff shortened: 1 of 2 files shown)


jonathonreilly:
review-loop landed this as aee010e80. I kept the durable science but narrowed the claim boundary: this is a scoped supplier no-go over the enumerated determinant/Matsubara/heat-kernel, thermal/g*/entropy, and isotropy/Lorentz families, not a ledger-wide closure. I removed stale unmerged-parent language, kept the retained-pending real-diagonal source out of retained-grade authority, replaced casual audit wording with supplier-sweep/tested-clause language, and expanded the no-go discipline gate to six distinct routes with the walls collapsed to one residual B-Z2. The useful result is the two-sided certificate: per-volume log-det/zeta density does not separate, while det vanishing order, bounded g_eff, and zero-set conical/tangent behavior identify the future missing retained-grade realized-kernel supplier. Checks: runner PASS=23 FAIL=0 with cache; independent spectral enumeration matches zero counts and det-order counts; vocab_lint clean; py_compile clean; run_pipeline.sh complete; audit_lint --strict OK (existing notices only); generated audit/effective-status outputs stripped; diff checks clean. Seeded row during validation: claim_type=no_go, audit_status/effective_status=unaudited, leaf, 22 deps, runner classification C-dominant non-decoration (C=23, D=8).

PR #3511, opened 2026-06-10
audit-infra: retire θ from the Tier-A registry (GATED — basis audits + det-readout bridge required)

jonathonreilly:
## ⚠️ REVIEW-LOOP HANDLING INSTRUCTION (owner-requested — read before any disposition)

This PR is **intentionally gated**. Per the owner: **do not auto-land, and do not close/salvage/strip this PR as non-landable.** When review-loop processes it:

1. Review the content normally (registry edit + human-readable registry-note edit; no science claims, no audit verdicts).
2. Check the gates below against the live ledger on `main`.
3. **If any gate is unmet:** leave the PR open, post a one-line status comment listing which gates remain, and continue with other work. An unmet gate here is not a review failure.
4. **If all gates are met: PAUSE and ask the owner in a PR comment how to proceed.** Owner approval per `docs/audit/AXIOM_MINIMALITY_POLICY.md` §6 precedent must be recorded by the owner in this thread before landing. Do not land on inference.
5. At landing time (post-approval): update the JSON `GATE:`/`PENDING` annotations to cite the clean audit rows and the approval record, run the pipeline for validation only, and land **without** shipping regenerated audit surfaces (the nightly cron regenerates them on `main`).

## Gates

- [ ] `theta_cross_plane_term_absent_in_supplied_per_plaquette_class_bounded_theorem_note_2026-06-09` → audited_clean (Wave 1; wiring repaired in #3509)
- [ ] `theta_p2_k_cpt_determinant_character_phase_erasure_bounded_note_2026-06-10` → audited_clean (Wave 1; standalone split note from #3509, no AC_φλ dependency)
- [ ] `registrable_readout_additive_even_phase_free_narrow_theorem_note_2026-06-10` → audited_clean (Wave 1.5; lands via #3513; an interim audited_conditional via the unaudited `tier_a_korbit` row resolves through the cascade mechanism and is acceptable only once resolved to clean)
- [ ] Owner approval recorded in this PR thread per the AXIOM_MINIMALITY_POLICY §6 (Record-retirement) precedent

## Summary

[...]

PR #3509, opened 2026-06-10
science: theta-chain wiring repair + theta-P2 lemma split (decoupled from the AC_phi_lambda gate)

jonathonreilly:
## Summary

Two commits, one coherent purpose: make the θ discharge chain honest in the citation graph and **fully decoupled from the surviving AC_φλ admission**. No claim content weakened; one new note carries the moved lemma with its own runner.

## Commit 1 — dependency wiring repair

- **Cross-plane note**: two context citations (prior single-plaquette boundary, multi-plaquette reopening) were seeding the graph as load-bearing deps, dragging a ~335-row legacy tower into a leaf theorem. Demoted to backticked context; the genuine authority for the supplied per-plaquette class linked instead: `PER_PLAQUETTE_FROM_ADJACENCY_LICENSE` (audited_clean / retained_bounded). N6 updated.
- **K-orbit note**: detc-vs-detr fork citation demoted to context (self-contained algebra; the class-supplying bridge is the note's own named open).

## Commit 2 — θ-P2 lemma split (owner-approved cleaner cycle)

- New note `THETA_P2_K_CPT_DETERMINANT_CHARACTER_PHASE_ERASURE_BOUNDED_NOTE_2026-06-10.md` carries the K/CPT determinant-character phase-erasure lemma verbatim (incl. the hostile evenness guard and the named determinant-readout bridge open), paired with a sliced sympy runner — **13/13 PASS** (adds explicit non-invariance witnesses for k ∈ {±1, ±2, ±½}) — and cache.
- K-orbit note rescoped to the AC_φλ orientation lemma only; points at the moved lemma as non-load-bearing context; runner sliced to the orientation half — **10/10 PASS**, cache refreshed. One pre-existing stale check against the recently repaired staggered gate note (`admitted-context` phrasing no longer exists there) retargeted to the surface that does.

## Resulting graph (pipeline-validated; regenerated audit surfaces dropped per the hard gate)

| row | deps | tonight |
|---|---|---|
| `theta_p2_k_cpt_determinant_character_phase_erasure_...06-10` | θ row (retained_bounded) + minimal_axioms | **audit-ready, leaf, 1 pass — θ-side** |
| `theta_cross_plane_term_absent_...06-09` | per-plaquette license (retained_bounded) | **audit-ready, leaf, 1 pass — θ-side** |
| `tier_a_korbit_...06-09` | staggered gate (Tier-A) + minimal_axioms | AC_φλ-side; ready under #3512 |

[...]

PR #3524, opened 2026-06-11
science: θ-P2 lemma split (re-cut) — decouples the θ basis from AC_φλ and fixes the red K-orbit runner

jonathonreilly:
## Summary

Re-cut of #3509's un-landed second commit (the closed head is dead; re-applied cleanly against current main), with one new fact that upgrades it from cleanliness to load-bearing: **the K-orbit runner as landed on main is currently RED** (PASS=16 FAIL=1 — a stale source-text check greps the staggered gate note for `admitted-context` phrasing that the recent gate repair removed). A red primary runner blocks a clean `tier_a_korbit` audit, and that row is a Wave-1 gate for #3511.

## Changes

1. **New note** `THETA_P2_K_CPT_DETERMINANT_CHARACTER_PHASE_ERASURE_BOUNDED_NOTE_2026-06-10.md` — the K/CPT determinant-character phase-erasure lemma moved verbatim (hostile evenness guard + the named determinant-readout bridge open included), paired sympy runner **13/13** (adds explicit non-invariance witnesses for k ∈ {±1, ±2, ±½}) + cache. Deps: `{strong_cp_theta_zero_note, minimal_axioms}` **only** — θ's P2 basis carries no staggered-gate dependency, so the retirement decouples from the surviving AC_φλ admission (#3511's basis can be repointed back to the split row at landing).
2. **K-orbit note** rescoped to the orientation lemma (one bounded fact; moved lemma = backticked context). Deps: `{staggered gate (Tier-A), minimal_axioms}`.
3. **K-orbit runner** sliced to the orientation half **and the stale gate-note check fixed** → **10/10**, cache refreshed from the live run.

## Sequencing (matters for audit capacity)

**Land before the `tier_a_korbit` row is audited.** Landing after a clean K-orbit audit would hash-invalidate that fresh verdict; landing before saves the auditor a red-runner conditional. If the audit lane has already processed K-orbit by review time, defer this PR until after θ retires and treat it as the refinement it was originally.

## Test Plan
- [x] Both runners pass (13/13, 10/10); caches committed from live runs
- [x] vocab_lint clean; pipeline + strict lint OK; reseeded deps verified as above
- [x] No generated audit surfaces shipped

No audit verdicts applied or predicted.

### DIFF
--- a/docs/THETA_P2_K_CPT_DETERMINANT_CHARACTER_PHASE_ERASURE_BOUNDED_NOTE_2026-06-10.md
+++ b/docs/THETA_P2_K_CPT_DETERMINANT_CHARACTER_PHASE_ERASURE_BOUNDED_NOTE_2026-06-10.md
@@ -1,0 +1,96 @@
+# Theta P2 K/CPT Determinant-Character Phase Erasure: Bounded Candidate Route
+
+**Date:** 2026-06-10 (split from
+`TIER_A_KORBIT_DETERMINANT_AND_ORIENTATION_INVARIANCE_BOUNDED_NOTE_2026-06-09.md`
+so the strong-CP mass-orientation route stands on the theta surface alone,
+with no dependency on the `AC_phi_lambda` staggered gate)
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not
+set or predict an audit outcome and does not edit the audit-lane-owned Tier-A
+registry.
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Landed durable source content on main as 2cc25671c (Split theta P2 determinant lemma). Review-loop verified the theta/P2 split and K-orbit orientation-only runner: 13/13 and 10/10, pipeline + strict audit lint clean. Generated audit/publication/front-door outputs were not committed. This does not retire theta from Tier-A or apply any audit/registry status.

PR #3525, opened 2026-06-11
science-fix: single-clock codimension-1 evolution — S3 false as written (exact axis-exchange symmetry computed); axis selection becomes declared premise B-AXIS

JONbridgerapps:
## Target
`axiom_first_single_clock_codimension1_evolution_theorem_note_2026-05-03` — #2 ready keystone (~1119 downstream rows), unaudited positive_theorem, and the load-bearing clock dep flagged three times this week.

## The critical science finding
**Step S3 was false as written.** The old argument: "distinct generators would require distinct RP-admissible reflection axes, and the staggered sign rule rules out every spatial reflection — hence one clock." Computed refutation: the staggered hop operator is *exactly* invariant under the conjugated exchange `W = P_{τ↔1} ∘ diag((−1)^{x_τ x_1})` (residual 0 on a 4×4×2×2 block; temporal hop sector mapped exactly onto spatial; identical spectra). The old runner only tested the *unconjugated* temporal template — and its T8/T9 checks were literal tautologies (`−1 == −1`), with a phase-blind sign-flip criterion that tested orientation, not RP. The framework's own kinetic-isotropy primitive (c_t = c_s) makes the surface *more* exchange-symmetric, not less.

## The honest restructure
- **Axis selection is a declared premise (B-AXIS)**, stated verbatim from the retained `single_clock_uniqueness_scope_boundary` no-go's N2/N4/N5 clauses (which the old "exactly one clock" headline violated).
- Given the axis and the supplied (T̂², 2a_τ) at their retained_bounded grades, the retained finite-dim Stone row yields the **unique nonneg generator and one-parameter unitary group** — transfer-relative and τ-relative. The exchange certificate is a computed class-C boundary fact.
- Stale imports cleaned: inline Stone re-proof → retained narrow row; equal-time locality re-cited to the retained_bounded LR row; conditional clustering clause dropped; superseded axiom framing removed.

## Claim type
`positive_theorem` → **`bounded_theorem`**. The downstream d_t ≤ 1 cap survives in axis-conditional form; B-AXIS is anomaly-free, so `anomaly_forces_time`'s non-circularity separation is preserved (its SC premise wording is now stale — follow-up queued separately).

## Consistency
Boost-faith no-go: consistent (time-translation group only). Cubic-anisotropy no-go: consistent (only the c_t = c_s direction is used, never its converse).

## Runner
Rebuilt: `TOTAL: PASS=36 FAIL=0` (14 [A], 1 [B], 16 [C], 5 [D]), deterministic, 0.5 s. Falsification legs: non-Hermitian transfer breaks unitarity; plain axis swap without the sign field fails (residual 5.66); a two-clock tensor comparator is excluded with min gap 0.53 (B-AXIS.3 non-vacuous); τ-doubling halves H exactly.

### DIFF
--- a/docs/AXIOM_FIRST_SINGLE_CLOCK_CODIMENSION1_EVOLUTION_THEOREM_NOTE_2026-05-03.md
+++ b/docs/AXIOM_FIRST_SINGLE_CLOCK_CODIMENSION1_EVOLUTION_THEOREM_NOTE_2026-05-03.md
@@ -1,609 +1,503 @@
 # Axiom-First Single-Clock Codimension-1 Unitary Evolution on Cl(3) ⊗ Z^3
 
-**Date:** 2026-05-03
-**Type:** positive_theorem (lattice form); bounded_theorem (continuum-limit
-identification with Wightman one-parameter group)
-**Claim scope:** From A_min plus the reflection-positivity
-support theorem, the spectrum-condition support theorem, the
-retained cluster-decomposition support theorem, the retained
-Cl(3)-per-site-uniqueness support theorem, and the retained
-microcausality / Lieb-Robinson support theorem, the framework's lattice
-dynamics is a **single-clock codimension-1 unitary evolution**:
-(S1) the reconstructed Hamiltonian H is the unique generator of a
-strongly-continuous one-parameter unitary group `U(t) = exp(-itH)` on
-the physical Hilbert space `H_phys`; (S2) at each lattice time slice
-`Σ_t = {t} × Z^3` the equal-time local algebra is the full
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed durable source content on main as 68b50dc64 (Demote single-clock axis theorem). Review-loop kept B-AXIS as an explicit bounded-theorem premise, not a new axiom/primitive; fixed the baseline axiom citation to MINIMAL_AXIOMS_2026-06-05; refreshed the existing runner cache (PASS=36 FAIL=0); verified pipeline + strict audit lint; generated outputs were not committed.

PR #3526, opened 2026-06-11
science-fix: yt_vertex_power — n_link = 2 genuinely derived (Schur × quadratic response); old runner was literal PDG back-propagation

JONbridgerapps:
## Target
`yt_vertex_power_derivation` — ready keystone (~574 downstream rows), and the lane behind **alpha_s admission B3** (n_link = 2, previously "asserted via CMT with no one-hop authority").

## What was wrong (the worst runner finding of the campaign)
- Every load-bearing link-count check was `check(..., True, ...)` — hard-coded PASS.
- Part 6 **selected n_link = 2 by closest match to PDG alpha_s(M_Z)** — literal class-(G) back-propagation presented as a uniqueness derivation, with PDG 0.1179 imported and 2-loop RG to the Z pole run as a load-bearing "DERIVED" check.
- The note's only dep was ALPHA_S_DERIVED_NOTE.md — circular, since alpha_s's B3 names this lane as its supplier.

## The derivation (class C, bounded)
`n_link = (external gauge legs of the quadratic response) × (links per vertex insertion) = 2 × 1`:
- **Factor 1**: Schur bi-equivariance — gauge covariance + the edge-only adjacency license force f(U) = cU as the unique covariant hop dressing (runner discriminates linear vs non-linear dressings computationally, and exhibits the off-license 5-link covariant escape, making the license demonstrably load-bearing).
- **Factor 2**: the second-order structure of −Tr ln D[A] — FD log-det matches tadpole+bubble to 9e−7, bubble link-degree = 2 with residual 1.3e−15, n ∈ {0,1,3,4} excluded by margins > 0.1, cubic vanishes Furry-style.
Bounded on named conventions/gates: B-GATE (kinetic realization, narrowed to the one-bit flux selector via the kinetic-class theorem), B-ADJ, B-CONV, B-CHAN, B-SPLIT.

## No-back-propagation certification
Zero class-D checks; a self-scan check **fails the runner if forbidden tokens** (0.1179/0.1181/0.1033, PDG, Z-mass, RG machinery) ever enter its source. `Breakdown: A=4 B=2 C=12 D=0` / `TOTAL: PASS=18 FAIL=0`, deterministic, 0.1 s.

## Downstream
Once audited, alpha_s B3 retires from a bare structural admission to a citation — its premise content shrinks to this note's named convention boundaries. Out-of-scope finding reported for follow-up: the CMT note cites a "retained D14 identity, §2.4" of YT_EW_COLOR_PROJECTION_THEOREM.md that no longer exists in that file (now a 102-line no-go) — stale citation in a third note.

### DIFF
--- a/docs/YT_VERTEX_POWER_DERIVATION.md
+++ b/docs/YT_VERTEX_POWER_DERIVATION.md
@@ -1,59 +1,375 @@
 # `y_t` Vertex-Power Theorem: Why `alpha_s(v) = alpha_bare / u_0^2`
 
-**Date:** 2026-04-15
-**Status:** closed subderivation on an open lane; zero-input structural support
-**Primary runner:** `scripts/frontier_vertex_power.py`
+**Date:** 2026-04-15 (structural-derivation repair 2026-06-11)
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Claim scope:** On the gauged staggered-Dirac kinetic surface — the
+adjacency-licensed nearest-neighbor covariant hopping bilinear, with the
+named link-exponential convention and the standard quadratic-response
+(background-field) coupling normalization — the vertex power of the
+mean-field link dressing is
 
-## Authority Role
+```text
+n_link  =  (external gauge legs of the quadratic response)
+           x (gauge links per vertex insertion)
+        =  2 x 1  =  2,
+```
 
-This note provides the operator-level support for the current coupling map used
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed durable source content on main as 45446b1e9 (Derive vertex power structurally). Review-loop edits removed the downstream canonical-surface read so the no-back-propagation boundary is enforced by source, runner, and refreshed cache; generated audit/publication/front-door outputs were stripped.

PR #3523, opened 2026-06-11
science-fix: grassmann forcing block theorem — withdraw no-go-contradicting forcing claim, consume current surface (T1/T2/T3 split)

JONbridgerapps:
## Target
`staggered_dirac_grassmann_forcing_theorem_note_2026-05-07` — #3 ready keystone (~839 downstream rows), unaudited.

## Why this was urgent
The note's load-bearing Step 5 asserted an unconditional two-candidate exhaustiveness ("the matter measure is forced... uniquely Grassmann") that is **refuted by the retained 2026-05-25 statistics-agnostic no-go** (the hard-core boson is a per-site dim-2 alternative). The runner hard-coded the contested premise (one check returned True unconditionally; five others were assertion stubs). On audit this lands `audited_failed` *against a retained no-go* — the worst outcome on the spine's biggest ready keystone.

## The fix
Restructured to match the current surface: **T1** (unconditional two-candidate collapse, consuming the retained `cl3_per_site_hilbert_dim_two` row + axiom surface), **T2** (Grassmann/CAR forcing *conditional on GL(F)* — exactly the "independent locality+statistics principle" the no-go's own §6/N6 names), **T3** (the no-forcing boundary, recomputing the hard-core tie intact). The re-scoped CCR exclusion (S2′) is consumed at its declared grade with a runner check that the hard-core frame sits outside its hypothesis class.

## Routing note (audit-drain reaction)
`u4_closes_under_qubit_reframe` was flipped to `audited_renaming` in today's drain (the auditor: its content is the Quantum axiom's one-qubit carrier restated). The U4 consumption here is therefore routed through the axiom surface + the retained dim-two row directly, with the renaming verdict noted in plain text — no load-bearing edge touches the renamed row. Off-branch suppliers (Berezin-RP GL(F) derivation, multiloop cocycle no-go) are cited as plain-text supplier-chain pointers with grades declared — no dangling edges.

## Runner
Full rewrite from assertion stubs: `TOTAL: PASS=23 FAIL=0` ([A]=5 consumed-input recomputation, [B]=6 collapse+falsification, [C]=8 conditional selection+falsification, [D]=4 honest-claim structure), deterministic, ~0.4 s. Falsification legs: drop the dim-2 input → collapse fails on a k=2 ℂ⁴ module; drop GL(F) → the tie returns.

## Honest classification
`bounded_theorem` — now genuinely matching content: positive collapse, conditional selection, declared chain grades.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3529, opened 2026-06-11
science-fix: substep1 bridge — re-route U4 consumption to axiom surface + retained dim-two row (u4_closes audited_renaming)

JONbridgerapps:
## What this is
Audit-drain reaction: today's drain flipped `u4_closes_under_qubit_reframe` to **`audited_renaming`** (the auditor: its content is the Quantum axiom's one-qubit carrier restated as baseline, not a theorem). The substep-1 bridge note (landed earlier as "Discharge substep1 U4 Grassmann bridge") consumed that row as a retained authority — on audit it would land `audited_conditional` against a renamed row for no scientific reason.

## The surgical re-route (12 load-bearing locations)
The delivered object (H_x = ℂ², single faithful Cl(3,0) module, k = 1) is now supplied by **MINIMAL_AXIOMS_2026-06-05 (Quantum) + the retained `cl3_per_site_hilbert_dim_two` row** (audited_clean positive_theorem, five-judge panel, closes from A1/A2). The renamed row appears only in plain text — with the verdict's rationale quoted as a *strengthening* remark (the U4 "bridge" was always baseline content given the axiom, which is exactly why the discharge stands). The two-candidate collapse logic (D1)–(D5) is untouched; same routing pattern as the block-theorem fix (#3523).

## Runner
Authority strings re-routed, math unchanged: `TOTAL: PASS=38, FAIL=0` (re-verified on this branch), expected-output block unchanged.

## Confirmation
No load-bearing edge touches the renamed row — it has no markdown-link edge and supplies nothing in (D1)–(D5).

### DIFF
--- a/docs/STAGGERED_DIRAC_SUBSTEP1_GRASSMANN_FORCING_BRIDGE_NARROW_THEOREM_NOTE_2026-05-16.md
+++ b/docs/STAGGERED_DIRAC_SUBSTEP1_GRASSMANN_FORCING_BRIDGE_NARROW_THEOREM_NOTE_2026-05-16.md
@@ -1,14 +1,16 @@
 # Staggered-Dirac Substep 1 — Grassmann Forcing Bridge Narrow Theorem
 
-**Date:** 2026-05-16; 2026-06-10 science-fix (U4 boundary discharged via
-the qubit-reframe closure — see §0 changelog)
+**Date:** 2026-05-16; 2026-06-10 science-fix (U4 boundary discharged —
+see §0 changelog); 2026-06-11 science-fix (U4 consumption re-routed to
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed as source-only salvage in 95e991610 (Reroute substep1 bridge to dim-two supplier). I did not take the stale branch wholesale because this file overlapped current main. The landed version keeps the useful reroute but grounds the bridge in the Quantum axiom plus the retained per-site dim-two theorem, removes the old U4/single-module row as a load-bearing dependency, keeps the former U4 row as a non-load-bearing reader pointer, refreshes the runner/cache (PASS=38, FAIL=0), and leaves the changed row unaudited for the independent audit lane. Pipeline and strict audit lint passed; generated audit/status outputs were not committed.

PR #3519, opened 2026-06-11
science-fix: fermionic Stefan-Boltzmann — quantifier on the realized kernel class (the named B-Z2 supplier, audit-ready)

JONbridgerapps:
## What this is
The closing piece of the P-FLUX chain: the axiom-first fermionic Stefan-Boltzmann note (2026-05-26, unaudited) re-scoped into the supplier the B-Z2 sharpened-spec note (PR #3506) named in its N6. Triage: the 05-01 bosonic note is the wrong currency; the 05-17 note is a meta fix-record; the 05-26 fermionic narrow note is exactly the F-3 certificate's currency.

## What was wrong
- **Quantifier scope (critical):** the load-bearing step integrated a *supplied* continuum DOS from an imported dispersion — it bound an inventory, never a kernel, so it could say nothing about the realized massless set.
- **Circularity (critical for the supplier role):** a load-bearing citation of the emergent-Lorentz note (whose isotropy core is derived ON the staggered dispersion) would have put any flux-selection consumer in the RP circularity class.
- Five load-bearing upstreams all unaudited (weakest-tier inheritance — auditing could never lift it); runner had zero classified checks, no mode sums, no falsification leg.

## The re-scoped theorem (FSB-K, bounded_theorem)
For **every** realized-class kernel (finite-range, sublattice-periodic, Hermitian on the retained Z³ single-mode surface) satisfying explicit **hypothesis (Z)** — finite zero set with invertible linear cones — the half-filled free-Fermi probe obeys `u(T) = g_eff·(7/8)(π²/30)T⁴ + O(T⁵)` with `g_eff = Σ|det V|⁻¹` finite = the cone-weighted massless species count; cone-window proof with explicit error orders. **Branch-blind on its face**: the note declares, and the runner verifies, that it neither assumes nor derives φ = −1. Emergent-Lorentz/KMS/spin-statistics demoted to plain-text context; load-bearing deps are now one-hop retained. The old continuum corollary survives, so the five downstream citers are unaffected (both gstar runners re-run clean: 66/0, 117/0). Composition with the retained 7/8 bridge is disjoint-roles, checked exactly.

## Runner
[...]

PR #3530, opened 2026-06-11
physics: (Z) certificate + selection composer — the φ = −1 supply line is complete (conditional on chain grades)

JONbridgerapps:
## What this is
The two closing notes of the P-FLUX chain, as one reviewable pair.

## Note 1 — the (Z) certificate (positive_theorem)
**Z-K1**: the flux(−1) Kawamoto-Smit realized kernel satisfies FSB-K's hypothesis (Z) **exactly and at all volumes** — a new symbolic Bloch identity `H_c(κ)² = 4(Σ cos²(κ_μ/2))·I₈` (sympy, with a symbolic direction weight) pins the massless set to the 8 BZ corners L-independently; every corner carries the same exact isotropic cone V = 2I with explicit quantitative data (V, C, r) = (2I, 2/3, 1) proven by an elementary inequality chain; cone-weight sum = 8 per cell, tied to the det-positivity row's M_KS by computed multiset equality. **Z-K0** (companion): the scalar branch violates both clauses — an *exact* zero line kills finiteness and forces every candidate cone matrix singular. Falsification legs prove the machinery quantitative (anisotropic kernel passes with V = diag(2,2,1); two comparators rejected by singular-V detection). One-hop load-bearing: minimal_axioms + the retained Fock bridge only; FSB-K consumed as definition text only; the renamed u4 row deliberately not load-bearing. `PASS=18 FAIL=0`.

## Note 2 — the composer (bounded_theorem, the honest ceiling)
**SEL-C**: IF FSB-K (#3519) and the (Z) certificate stand at retained grade, THEN within the licensed two-class surface, (CL) is satisfied by K1 (g_eff plateau ≈ 8 per cell, recomputed by FSB-K's own mode-sum method) and violated by K0 (g_eff diverging T⁻², Sommerfeld 0.233) — **φ = −1 selected, B-Z2 supplied, B-BIT retired, conditional on the named grades**. The conditional structure is explicit and *live*: the runner reads the ledger every run and performs no selection at current grades; `dependency_not_retained` cascades, `failed/renaming` collapses the composition. Known risk recorded: FSB-K's u4 upstream sits at audited_renaming → likely one conditional cycle.

**Guard compliance is computed, not asserted** (the index-pairing turn-1 inversion guards): the full list of consumed data contains no "3"; the end-to-end pipeline is label-free with flux −1 as *output*; dropping (CL) restores the matter-content TRUE/TRUE tie; dropping the certificate leaves FSB-K's hypothesis unverified (and FSB-K-minus-(Z) is false as a universal). `PASS=16 FAIL=0`.

## The audit lane's three steps to unconditional φ = −1
1. FSB-K → retained (resolving its u4 upstream by cascade);
2. the (Z) certificate → retained;
3. the kinetic-class forcing row → retained (for wholesale P-KIN retirement beyond the within-surface selection).
The composer runner self-reports when the grades move.

### DIFF
--- a/docs/P_FLUX_SELECTION_VIA_FSB_K_AND_Z_CERTIFICATE_CONDITIONAL_THEOREM_NOTE_2026-06-11.md
+++ b/docs/P_FLUX_SELECTION_VIA_FSB_K_AND_Z_CERTIFICATE_CONDITIONAL_THEOREM_NOTE_2026-06-11.md
@@ -1,0 +1,457 @@
+# P-FLUX Selection via FSB-K and the (Z) Certificate — `φ = −1` Conditional on the Named Chain Grades (Conditional Theorem)
+
+**Date:** 2026-06-11
+**Claim type:** bounded_theorem (honest conditional — the composer of
+the P-FLUX supply line; the conclusion is conditional on two named
+rows reaching retained grade, and this is declared the ceiling until
+the audits land)
+**Claim scope:** **IF** (C1) the re-scoped axiom-first fermionic
+Stefan-Boltzmann row FSB-K
+(`axiom_first_fermionic_stefan_boltzmann_narrow_theorem_note_2026-05-26`:
+for **every** realized-class kinetic kernel satisfying hypothesis (Z),
+`g_eff(T) := u(T)/[(7/8)(π²/30)T⁴]` has a finite `T → 0` plateau
+`= Σ |det V_jb|⁻¹`, Corollary FSB-CL) stands at retained grade with
+its quantifier and FSB-CL intact, **AND** (C2) the (Z)-certificate row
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Review-loop landed the source-only conditional package on main as f5c351496. Salvage/fix pass kept the Z certificate as positive_theorem/unaudited and the P-FLUX composer as bounded_theorem/unaudited, cleaned claim-type metadata, refreshed runner caches, and repaired the composer runner's stale parent-text check against current main wording. Checks: Z runner PASS=18 FAIL=0; composer runner PASS=16 FAIL=0; audit pipeline passed; strict audit lint OK with existing notices only; generated audit/status outputs stripped.

PR #3535, opened 2026-06-11
physics: B-AXIS pin — underivable from the retained surface (computed W-transport); sharpened to one Z2 BC-asymmetry datum

JONbridgerapps:
## What this is
Pin-removal lane on B-AXIS (the clock-axis premise created by the single-clock fix, PR #3525). Outcome: **narrow no-go with the pin sharpened to a single ℤ₂ datum** — and the strongest version of the no-go, because every candidate escape was killed by exact computation, not argument.

## The W-transport results (the load-bearing computations)
- **OS/GNS route (the escape hypothesis) — FALSE**: `W θ_τ Wᵀ` is exactly a signed spatial site reflection; the one-particle OS kernels are exactly unitarily equivalent (residual 1.5e−15) with identical spectra — **the spatial axis gets the full GNS/positivity package free by conjugation**. The reconstruction does not anchor the axis.
- **Record/durability route**: durability = operator-order monotonicity, which is unitary-transport invariant (computed); the Record axiom excludes "time metric" by its own text; the CAP-K registration cone is circular as a supplier (consumes H and τ, both downstream of the axis) and transports exactly.
- **Anomaly route**: ε(x) grading and chiral anticommutation exactly W-invariant; the anomaly note's own non-circularity text ("constrains only the *count* d_t... not which axis") is the kill.

## The sharpened pin (computed minimal supplier shape)
Antiperiodic-τ / periodic-space boundary conditions break W **exactly** (residual 2√2); making both axes antiperiodic restores the symmetry — so the datum is the **per-axis BC asymmetry, one ℤ₂ choice**, not the wrap itself. The discriminator is relabeling-invariant (dim ker of the hop: 0 temporal vs 32 spatial), so once the datum is supplied, no exchange map of any kind survives. Equivalent record-shaped form: a declared registration-direction bridge. Existing repo rows using antiperiodic-τ carry the datum as post-axis convention, not derivation — any future axis-selection row now has its exact spec.

## Scope hygiene
The scope-boundary no-go's N2/N4/N5 clauses are consumed verbatim (quoted), not contradicted; B-AXIS.1/.3 untouched; B-AXIS.2 splits into the declared transfer-construction choice plus the now-precisely-specified axis-label datum. No axiom files touched.

## Runner
`TOTAL: PASS=23 FAIL=0`, deterministic, 0.07 s; falsification legs: no-sign-field swap fails (5.66), symmetric-BC restoration exact, kernel-dim invariant.

### DIFF
--- a/docs/SINGLE_CLOCK_AXIS_SELECTION_FROM_RECORD_DURABILITY_NARROW_NO_GO_NOTE_2026-06-11.md
+++ b/docs/SINGLE_CLOCK_AXIS_SELECTION_FROM_RECORD_DURABILITY_NARROW_NO_GO_NOTE_2026-06-11.md
@@ -1,0 +1,449 @@
+# Single-Clock Axis Selection From Record Durability: Every Retained Candidate Structure Is W-Transportable — Narrow No-Go With a Sharpened Pin
+
+**Date:** 2026-06-11
+**Type:** narrow no-go (axis-selection route pruning) + sharpened-pin
+support (computed minimal axis-selecting inputs)
+**Claim type:** no_go
+**Claim scope (narrow):** the single clause left open by the 2026-06-11
+hostile re-scope of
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed reviewed no-go on main as bf91a7c21. Review-loop changes: added explicit N1-N8 no-go discipline gate, canonicalized Type to no_go, linked the anomaly dependency, updated the CAP-K text check to current main, added runner cache, and verified runner + pipeline + strict lint.

PR #3537, opened 2026-06-11
physics: GL(F) identification bridge — 3 of 4 clauses discharged as theorems; residual collapses into the already-declared action-surface pin

JONbridgerapps:
## What this is
Pin-removal lane on the GL(F) derivation's one declared residual beyond chain grades: the reconstruction-identification bridge (quoted verbatim from the parent's §6). Outcome: **decomposition with teeth — everything kinematic is forced, and the residual is one clause that under the natural reading is co-extensive with a pin already declared elsewhere.**

## The four clauses
- **(I-1) carrier** — discharged on retained suppliers: the Quantum axiom supplies the per-site carrier directly (the u4 renaming verdict *confirms* this is axiom content); the composite is the retained tensor bridge; GNS dimension recomputed.
- **(I-2) parity** — **upgraded to a theorem**: the reconstructed parity is the *word* ∏(1−2ψ†ψ) in the fields, and number operators are dressing-invariant — so every intertwiner transports it onto the retained F = ⊗σ₃ automatically (exact identity).
- **(I-3) dictionary** — **forced**: the OS vacuum is cyclic, every anticommutator-insertion value of the functional vanishes exactly — so **GL(F) holds at functional level, as an operator identity in every cyclic realization, before any identification is chosen** — and the intertwiner onto the qubit-net JW realization has Schur nullity exactly 1 (identification forced up to one GL(F)-immaterial sign).
- **(I-4) matter-functional** — NOT discharged, and honestly so: "the physical matter correlation functional IS the retained measure's OS functional." The hostile dynamical check resolved cleanly (the retained set contains no rival dynamics — the RP row's transfer objects are constructed FROM the action, and the axioms disclaim dynamics — so the dynamical sub-clause has no content beyond I-4). Under the natural reading of "the Berezin matter measure," I-4 is co-extensive with the parent's already-declared action-surface conditionality; both readings stated, audit lane owns the choice.

## The honest negative recorded as a PASS
Carrier + parity + occupation data alone do NOT pin the net (the Pauli-vs-JW tie is recomputed — identical occupation data, both F-odd). The *functional* is the load-bearing member of the structure set — which is exactly why I-4 is the one genuine clause.

## Net effect
The GL(F) chain's residual shrinks from "chain grades + identification bridge" to "chain grades + the matter-functional clause" — no Hilbert-space, parity, dictionary, or transfer-naming freedom left to grade.

## Runner
`TOTAL: PASS=36 FAIL=0`, exact rationals at N=2,3, stdlib only, 0.12 s, byte-identical across runs. Falsification legs: hard-core intertwiner space exactly 0; ungraded-integrand mirror; parity-multiplicity escape; even-subalgebra cyclicity boundary declared. Unmerged parent cited plain-text with all load-bearing content recomputed in-runner.


jonathonreilly:
Landed reviewed bounded theorem on main as a1cbf9d47. Review-loop changes: updated the parent reference to the landed GL(F) reconstruction note, added the landed GL(F) discriminator dependency, softened physical-identification wording, added runner cache, and verified runner + pipeline + strict lint.

PR #3540, opened 2026-06-11
physics: P1 pin consolidation — realization-layer clauses 5 → 2 ({RR, REG-τ}), provably lossless

JONbridgerapps:
## What this is
Pin consolidation on the P1 chain: the separately-declared (CAP-real) and four (REG) clauses tested for redundancy. Outcome: **a proven implication lattice collapsing the realization layer from 5 declared clauses to 2 — with the equivalence verified exhaustively, both directions, so the relabeling is provably lossless.**

## The lattice
- **(RR) ⟺ (CAP-real) ∧ (REG-dyn) ∧ (REG-thr)** — syntactic factorization verified pointwise on 256 toggle models with independent predicate codepaths (RR evaluated as a single-pass sentence, never via the clause predicates). The reverse direction is the no-new-risk leg: declaring RR declares nothing new.
- (RR) ∧ (REG-τ) ⇒ (CAP-K) on both readings (D*, K = 129³ recomputed; Lemma C chain to pass set exactly {p = 0}).

## The witnessed non-arrows (what stays honestly separate)
- **τ is irreducibly separate** — growing-window model with τ_k → ∞, typed by the clock/rate no-go (which is *why* the floor is 2 clauses, not 1).
- **T1-d is NOT absorbed** — a 3-site witness where disjoint source blocks both sensitize the middle register site: the disjoint-blocks→disjoint-records clause stays the parent's Boundary.
- (CAP-real) ⇏ (RR) (fiat 4^k assignment); the dynamics clauses ⇏ (CAP-real) (real chain records with mismatched increments — RR is a genuine join); (REG-site) demoted to optional sharpening (the joint reading keeps K finite and k-uniform without it; pass set identical).
- A uniformity subtlety caught and pinned: D*(δ/2^k) − D*(δ) = k·ln 2 exactly — a decaying threshold loses k-uniformity, so RR must carry one uniform operator-norm δ (it does).

## P1 premise inventory after this
**T1-d + (RR) + (REG-τ)** (+ optional REG-site sharpening). Minimal generating set verified over the 263-model library, minimality by countermodel for every proper subset. RR remains declared, never derived; all five walls string-checked verbatim (middle arrow, count-probability, formation-not-forced, clock/rate, no-dynamics guardrail).

## Runner
`TOTAL: PASS=30 FAIL=0`, deterministic, ~0.5 s, exhaustive model libraries, forbidden-string scans clean.


jonathonreilly:
Landed durable source content on main as 0846cd1e3 (Add P1 registration realization consolidation). Review-loop edits made source-note wording branch-neutral, added a runner cache, kept RR declared/P1 open, and stripped generated audit/publication/front-door outputs.

PR #3577, opened 2026-06-11
meta(registry): realized_state_primitive — the axioms select no state; pointwise evaluation ONLY (owner-approved; 10-physicist panel 10/10)

jonathonreilly:
## Summary

Registers the **third framework primitive**, following the kinetic-isotropy registration precedent file-for-file. Owner approval (2026-06-11, in-session) is recorded in `AXIOM_MINIMALITY_POLICY.md` section 6, conditional on best-physics-judgment affirmation — affirmed: **necessary** (runner-backed underdetermination legs), **unavoidable** (the laws-vs-initial-conditions floor every theory carries), **clean** (pointwise evaluation presupposes no measure; self-policing clauses).

## The primitive

> The laws don't pick the state; the world does — among the states the laws permit. Derivations may evaluate at the realized state — pointwise, nothing more: no averaging over alternatives, no "typical" or "generic", and no quoting a number that would differ had another state been realized. The past hypothesis is a separate, stronger input.

## Review trail

Two-round, ten-persona adversarial physics panel: round 1 (a "conditioning ONLY" draft) drew 9 reservations + 1 objection (measure-presupposition of "conditioning", Borel–Kolmogorov, unbanned "generic" — corpus 800× vs 57×, the value-extraction seam, the past-hypothesis bundling contradiction); round 2 (the present wording) passed **10/10 with no objections**. The counterfactual test makes the r-dial discipline registry law; the past hypothesis is explicitly **not** housed (strictly stronger specialness input, anchored at the arrow note).

## Mechanics

- Declaration note + support runner (`PASS=8`, exact ≤4-qubit instances) + cache.
- `axiom_premise_nodes.json`: node + canonical id — `check_axiom_premise_clean.py` passes ("pure premise content").
- Section-6 approval entry; `PRIMITIVE_REGISTRY_CHECK.md` third bullet; parallel ONLY-scope lines in every command/skill surface that enumerates the existing primitives.
- State-contingency register seeded with four runner-backed entries + one explicit non-entry (the past hypothesis).
- Registry semantics: chain-satisfies without bounding; rows quoting realized-state data remain conditional on that supplied data; statuses remain audit-lane-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3586, opened 2026-06-11
docs(state-legs): re-point the three unaudited convergence legs to the registered realized_state_primitive

jonathonreilly:
## Summary

The free action from the state-language fanout sweep: the three **unaudited** state-contingency register legs (open-shell locus, induced holonomy, interleaved mean map) were written before [#3577](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3577) landed, so their texts name ad-hoc state-conditioning. One sentence each now cites the registered `realized_state_primitive` interface, so their eventual audits read the approved premise node instead of an unapproved per-note premise.

- Zero `deps_changed` cost: all three rows are unaudited (no clean audit to invalidate).
- No claims, numbers, runners, or caches change — pure cross-reference.


jonathonreilly:
Landed on main in dcd5cc0da after review-loop.

Review cleanup: converted the three realized_state_primitive references from plain filenames into markdown dependency links, so the audit graph sees the approved primitive node. The three rows remain unaudited bounded rows; no claims, runners, caches, or audit statuses were changed.

Checks: vocab lint clean; audit pipeline complete; strict audit lint OK; validation confirmed all three modified rows now include realized_state_primitive in deps; generated audit/publication outputs restored/not landed.

PR #3585, opened 2026-06-11
feat(bounded): Wilson-mass holonomy edge realization — twist datum exists in the named route; pure staggered cannot host it

jonathonreilly:
## Summary
- **The existence leg** for the signed-gravity twist datum — the named input of the sector-selection bridge ([#3584](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3584)). Together they make the proposal's `χ_η = ±1` labels fully mechanistic: retained Cl(3,1) grading supplies the sign mechanism; this note supplies where the gapped nonzero-η boundary tower can live — and where it provably cannot.
- **(W0) Structural obstruction (closes the gauge-background route).** For massless 1-component staggered matter on any even torus with **arbitrary** U(1) background, `{Γ₅, H} = 0` *exactly* — every NN hop flips parity and link phases drop out of the anticommutator — so the spectrum is mirror-symmetric and `η_δ ≡ 0` for every background. The containment scan's empirical negative becomes a one-line theorem, and the retained mean-field bulk-vanishing row generalizes to all backgrounds.
- **(W1) Existence in the named Wilson-mass route.** The minimal Wilson-mass cylinder in its gapped window, threaded by holonomy θ through the boundary cycle, hosts per edge a perfectly localized in-gap tower (min edge weight 0.999998, zero ambiguous states) whose counting η reproduces the twisted-tower label table **verbatim**: θ=0 → zero crossing (`h_δ=1`, no label); generic θ → `χ_bottom = sign(θ)`, quantized and size-stable (L_x 20→40: tower 3→7, η stays +1; `|η| ≤ 1` across a fine sweep); θ=π → `η=0`.
- **(W2) Orientation pairing.** `χ_top = −χ_bottom` at every defined θ — the e₄/orientation sector pair of #3584 realized geometrically. **The twist datum is the gauge holonomy through the boundary cycle** — gauge-sector data — within the Wilson-mass model class.
- **(W3) Controls.** Trivial window m=+1: the in-gap set is *empty* (the gapped window is load-bearing).
- **Honest residue, sharpened:** the Wilson-mass term is not framework-derived, and (W0) proves it cannot be dispensed with on this route — that is now the precise open question, alongside the 3D→2D-boundary lift and dynamical superselection. `S_int` and the proposal row untouched.

## Changes (review-loop source-only shape)
[...]

PR #3587, opened 2026-06-11
feat(bounded): parity-grading escape dichotomy — the no-pin route to a signed source

jonathonreilly:
## Summary
- The **no-pin** leg of the signed-gravity twist-datum program, per standing discipline (no new admitted terms; land bounded, derive later). Classifies every escape from the staggered parity obstruction and locates a candidate **inside already-derived framework objects**.
- **(P1) General parity obstruction.** Any Hermitian operator whose elements connect only opposite-parity sites — NN hops on *any* subgraph of Z^d (torus, open box, defect geometries), arbitrary complex couplings/backgrounds — satisfies `Γ₅HΓ₅ = −H` exactly ⇒ mirror spectrum ⇒ `η ≡ 0`. Geometry enters nowhere, so the **manifold-with-boundary route is closed** for parity-flipping matter, alongside the gauge-background route. Of the retained vanishing row's three named openings, only the symmetry-breaking-mass class survives.
- **(P2) On-site mass is insufficient.** `m·I` yields nonzero η that is **orientation-even** (reflection+conjugation leaves it invariant; checked at two masses) — it cannot source the orientation-odd `[+1,−1]` pair. Escapes require same-parity *couplings* with momentum-dependent structure.
- **(P3) The native escape — no pin required.** The derived log-generator `−log T` of a strictly-local transfer **natively contains same-parity (distance-2) couplings** (witness 0.0477; persists at a second non-commuting coupling set). Two controls isolate the mechanism exactly: a commuting pair gives `log T = A+B` with zero tails; a single factor gives `−log e^{−A} = A` exactly. The BCH tails that forced the single-clock **(B-RANGE)** premise are precisely the forbidden-class structure here — *nuisance in the Lieb-Robinson lane, resource in the signed-source lane.*
- **Named open target (bounded now, derive later):** whether the derived same-parity content opens a gapped, orientation-odd, labeled window (the sufficiency conditions of [#3585](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3585)).

## Changes (review-loop source-only shape)
- `docs/SIGNED_GRAVITY_PARITY_GRADING_ESCAPE_DICHOTOMY_NARROW_THEOREM_NOTE_2026-06-11.md`
- `scripts/signed_gravity_parity_grading_escape_dichotomy_2026_06_11.py` — `TOTAL: PASS=12 FAIL=0`, seconds
- `logs/runner-cache/signed_gravity_parity_grading_escape_dichotomy_2026_06_11.txt` — SHA-pinned

## Test Plan
- [x] Runner: `TOTAL: PASS=12 FAIL=0`, exit 0; anticommutators and controls exact to 1e-14
- [x] One-hop deps: axiom premise + retained_bounded bulk-vanishing row only; sisters as plain-text pointers
- [x] **No new pins**: (P3)'s structure lives in the derived transfer log-generator
- [x] Audit-status language untouched


jonathonreilly:
Landed on main in e165bd8d7 after review-loop.

Review cleanup: removed stale “in review” sister-note wording, softened “already-derived/derived log-generator” language to “registered transfer-log-generator” posture, and made the audited_conditional single-clock source an explicit markdown dependency. That keeps the no-pin route as a bounded candidate/dichotomy, without silently promoting the single-clock surface or claiming the gapped labeled window is already derived.

Checks: py_compile OK; runner TOTAL: PASS=12 FAIL=0; runner cache refreshed/fresh; vocab lint clean; audit pipeline complete; strict audit lint OK; validation row checked as bounded_theorem/unaudited with deps including the conditional single-clock source; generated audit/publication outputs restored/not landed.

PR #3628, opened 2026-06-12
feat(bounded): frame-transported Pfaffian line on interface K_odd — gauge-invariant label built; constant on the theta path

jonathonreilly:
## Summary
- Executes the signed-gravity transport design's label candidate: K_odd is exactly `i·B` (B real antisymmetric) under the orientation involution's real structure, so the quantized label class is a Pfaffian sign — computed here as a **frame-transported line bundle** over holonomy (relative-tolerance support rank inside a measured spectral gap, Procrustes transport, 63 θ points, overlap ≥ 0.99999998).
- **Result: the gauge-invariant transported sign has no flips on θ ∈ [0.05, π]** — both sizes, both profiles (step + tanh control) — while the raw per-θ SVD sign flips 28–30 times (the reviewer's rank-jump/orientation trap, now fully attributed to gauge). The support spectrum never approaches zero, so the empty crossing set and constant label are exactly consistent.
- **Verdict:** the label machinery exists and is invariantly computable; the θ-path on this geometry simply crosses no zeros. Named next: parameter directions that genuinely close the support gap (mass sweep toward the arcsinh floor, interface-amplitude sweep, opposite-step configuration pairs).
- Runner: **TOTAL: PASS=29 FAIL=0**.

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh) under a spec seeded with the reviewer's probe results and known trap; reviewed and re-run before landing.


jonathonreilly:
Review-loop landed source-only salvage on main as 0d919f230.\n\nDisposition:\n- Kept the finite native K_odd interface Pfaffian-line measurement.\n- Normalized claim metadata to bounded_theorem.\n- Removed worktree-local .claude/tmp context pointers rather than creating non-load-bearing dependency edges.\n- Preserved the scope guard: measured theta grid only, no theta=0 endpoint, no Wilson/domain-wall carrier, no interacting/gauged transfer-log extension, no source-action claim.\n- No audit verdict/effective-status change.\n\nVerification:\n- vocab_lint --fix: clean\n- py_compile: clean\n- local runner: PASS=29 FAIL=0\n- cache_status=fresh\n- pipeline complete; audit_lint --strict OK, known notices only\n- row: bounded_theorem / unaudited / leaf, deps []\n\nGenerated outputs restored before commit.

jonathonreilly:
Landed reviewed source-only salvage on main as 0d919f230.

PR #3652, opened 2026-06-12
feat(rg): the declared Schur RG convention is E-covariant — one map for the whole admissible energy family in the shifted quotient coupling (panel-narrowed)

jonathonreilly:
Wave-2 landing (workhorse mode): see the note for the claim chain, scope, and panel edits. Source-only triple: note + runner + cache. Statuses pipeline-derived; the audit lane grades.


jonathonreilly:
Landed on main as 1113db3ec.\n\nReview-loop kept the bounded RG result but tightened the claim surface before landing:\n- removed stale "in review" parent-status language;\n- changed the note to an explicit self-contained algebraic Schur derivation for the all-admissible formula;\n- narrowed the runner/cache wording to sampled dense verification of that algebraic theorem rather than numeric quantification over the whole domain;\n- corrected the note's runner count to PASS=8 FAIL=0.\n\nChecks: runner PASS=8 FAIL=0; py_compile clean; vocab_lint clean; audit pipeline and strict audit_lint clean except accepted notice classes; generated audit/publication/front-door outputs restored before commit. Generated ledger validation: bounded_theorem / unaudited / leaf.

PR #3654, opened 2026-06-12
feat(gauge-dynamics): no exhaustion anywhere in the adequate-family prefix range — every seed clears at every adequate k; higher-k family-inadequate (panel-narrowed)

jonathonreilly:
Wave-2 landing (workhorse mode): see the note for the claim chain, scope, and panel edits. Source-only triple: note + runner + cache. Statuses pipeline-derived; the audit lane grades.


jonathonreilly:
Landed reviewed source-only salvage on main as 043792afa.\n\nReview-loop changes before landing:\n- normalized Type to controlled `bounded_theorem`;\n- removed stale in-review / source-only-review-loop authority language;\n- kept cross-references contextual only, not graded authorities;\n- added the N1-N8 no-go discipline gate and narrowed the result to the finite adequate-family range only.\n\nChecks: runner PASS=9 FAIL=0, py_compile OK, vocab_lint OK, run_pipeline OK, audit_lint --strict OK with known repo notices only. Generated audit/publication/front-door outputs were restored before commit, so the landing is source-only and leaves the independent audit lane to grade the new row.

PR #3648, opened 2026-06-12
science: |δ|=2/9 chain — supplied carrier class + context demotions (audit-readiness for the R-η lane)

jonathonreilly:
## Summary

Audit-readiness repair for `koide_delta_eta_density_readout_chain_bounded_theorem_note_2026-06-09` (unaudited) — the chain that makes `|δ| = 2/9` a theorem conditional on exactly one named identification (R-η), i.e. the carrier row for AC_φλ sub-admission (ii) per the registry's owner-approved minimum decomposition.

Its row was blocked by three deps that cannot reach retained-grade on any near timescale, all decouplable by the repo's proven patterns:

- `staggered_dirac_realization_gate_note_2026-05-03` (unaudited, 13-dep tower) → the carrier is now the **supplied circulant class stipulated in-note** (`H(δ) = aI + B e^{iδ} C + B e^{-iδ} C^T`); the physical identification of that class as the charged-lepton carrier is carried by the AC_φλ admission itself, which supplies its own gate surface (same pattern as the korbit repair in #3642, whose audited claim_scope already reads supplied-surface).
- `koide_delta_radian_period_physical_not_vacuous_...` (unaudited) → context: the period-fork physicality this chain uses (the π-packaging mass prediction failing by a wide margin) is **computed in this note's own runner** (E3 check), and the note's own boundary calls the period discussion "a bounded diagnostic".
- `koide_berezin_detc_vs_detr_fork_mechanism_...` (claim_type `open_gate`) → context: it is the open-gate row for the occupancy binary, cited only to locate that binary; the circulant class is supplied in-note.

Plus one wiring addition: **E6** ("the registrable atom is |δ|") previously consumed the orientation strip with no cited authority; it now cites `REGISTRABLE_READOUT_ADDITIVE_EVEN_PHASE_FREE_NARROW_THEOREM_NOTE_2026-06-10.md` (Consequence B) as a load-bearing dependency.

Load-bearing deps after this PR: fixed-locus note (retained_bounded), radian-bridge no-go (retained_no_go), registrability bridge (clears via the #3642 → #3643 → #3645 cascade). The chain's row becomes audit-ready the moment the orientation strip retains.

## Changes

- `docs/KOIDE_DELTA_ETA_DENSITY_READOUT_CHAIN_BOUNDED_THEOREM_NOTE_2026-06-09.md` only. Science text unchanged except the carrier-class stipulation paragraph, the E6 authority wiring, and the Net block's carrier line; R-η itself (the named conditional input) is untouched.
[...]

PR #3645, opened 2026-06-12
science: theta P2 — wire the supplied determinant-readout bridge as a load-bearing dependency (verdict-responsive)

jonathonreilly:
## Summary

Verdict-responsive repair for `theta_p2_k_cpt_determinant_character_phase_erasure_bounded_note_2026-06-10` (graded `audited_conditional`, re-audit ask: *"missing_bridge_theorem: retain a determinant-readout bridge proving that the physical arg det(M_u M_d) contribution is exhausted by the multiplicative determinant-character readout and that no phase-sensitive non-multiplicative or action-level datum remains."*).

That bridge exists as a landed source note: `REGISTRABLE_READOUT_ADDITIVE_EVEN_PHASE_FREE_NARROW_THEOREM_NOTE_2026-06-10.md`, whose **Consequence A** is exactly the asked-for content — a Record-registrable scalar readout (finitely additive over pairwise-disjoint records, constant on K/CPT orbits) has identically vanishing per-sector phase (additive ⇒ odd; orbit-constant ⇒ even; odd ∩ even = 0), so the registrable content of `arg det(M_u M_d)` is exhausted by the phase-free `k = 0` modulus class; non-multiplicative phase-sensitive functionals (`cos(arg z)`, the hostile guard) and cross-sector interference are excluded by additivity; the action-level bare-θ slot is strong-CP premise 1, a separate action-surface premise on the per-plaquette license route.

The note's "Named Open" section keeps the bridge-obligation sentence **verbatim** (it is pinned by a sibling runner's interface check) and now records that the bridge is supplied, citing the bridge note as a load-bearing dependency. What remains conditional is stated exactly as the bridge note names it: the standing modeling premise (the physical mass-surface readout context satisfies the Record registrability constraints) remains an explicit condition; the bridge note's audit status is not asserted.

## Changes

- `docs/THETA_P2_K_CPT_DETERMINANT_CHARACTER_PHASE_ERASURE_BOUNDED_NOTE_2026-06-10.md` — bridge section rewritten (obligation kept verbatim + supplied-by wiring); registry-consequence composition updated; bridge note added to Dependencies; date amendment.
- `scripts/frontier_theta_p2_k_cpt_determinant_character_phase_erasure_2026_06_10.py` — the named-open check becomes "names the bridge and cites the supplied bridge note"; new cross-file check that the cited bridge note carries the Consequence A content (SUMMARY: PASS=14 FAIL=0).
[...]

PR #3643, opened 2026-06-12
science: registrability bridge — demote two boundary-naming citations to context (pre-audit dep hygiene)

jonathonreilly:
## Summary

Pre-audit dependency hygiene for `registrable_readout_additive_even_phase_free_narrow_theorem_note_2026-06-10` (currently `unaudited`; gate 3 of the θ-retirement PR #3511). Two of its seven dependency edges point at rows that cannot reach retained-grade on any near timescale and are cited **only to name what the note does NOT close**:

- `koide_delta_eta_density_readout_chain_...` (unaudited; its own deps include the 13-dep staggered tower) — cited solely to name the still-open `|delta|` magnitude chain (R-η) that Consequence B explicitly leaves open. No content is consumed.
- `acphilambda_hw_complementation_equivariance_support_...` (claim_type `open_gate`) — corroborating context for the orientation-odd `sin(3 delta)` line; the symmetric-function evenness and `sin(3 delta)` K-oddness used by Consequence B are verified **directly in this note's primary runner** (checks: `e1/e2/e3 even in delta`, `e3 = a^3 - 3 a B^2 + 2 B^3 cos(3 delta) exactly`, `orientation-odd line sin(3 delta) is K-ODD`).

Both become backticked context entries with explicit justifications (the standing "demote boundary-naming citation to context" repair pattern, cf. the block-count MRU demotion). The five load-bearing dependencies are unchanged: `minimal_axioms`, `tier_a_korbit_...`, `strong_cp_theta_zero_note`, `strong_cp_rp_half_..._no_go`, `koide_aps_c3_fixed_locus_...`.

## Changes

- `docs/REGISTRABLE_READOUT_ADDITIVE_EVEN_PHASE_FREE_NARROW_THEOREM_NOTE_2026-06-10.md` — two citations demoted (body + dependencies section); date amendment. No theorem text (T1–T7), consequence content, or N1–N8 substance changed.
- No runner change (`SCORECARD: PASS=30 FAIL=0` re-verified; the runner reads no files, cache stays SHA-fresh).

## Why this matters (θ-retirement chain, PR #3511)

Without this repair the row's ready-filter blocks on the staggered tower and on an `open_gate` row indefinitely, so #3511's gate 3 could never clear through the cascade. With it (plus the korbit dep repair in the sibling PR), all deps are axiom-node or retained-grade once korbit re-audits. Recommended audit order across the three repair PRs landing today: **korbit → registrable_readout → theta_p2**.

[...]

PR #3642, opened 2026-06-12
science: korbit orientation lemma restated on the supplied circulant class (verdict-responsive dep repair)

jonathonreilly:
## Summary

Verdict-responsive repair for `tier_a_korbit_determinant_and_orientation_invariance_bounded_note_2026-06-09` (graded `audited_conditional`, re-audit ask: *"dependency_not_retained: Retain or replace the unaudited staggered-Dirac gate authority; if the intended audited consequence is registry reduction rather than the algebraic orientation lemma, also provide a retained unordered-mass-multiset registrability bridge."*).

This PR **replaces** the unaudited staggered-gate authority: the orientation lemma is self-contained algebra on the three-parameter Hermitian circulant family `H(delta) = a I + B exp(i delta) C + B exp(-i delta) C^T`, now **stipulated in-note as the supplied class** — the same supplied-class pattern as the per-plaquette and determinant-character θ notes. The staggered gate note is demoted to context (backticked, not a dependency edge); the physical identification of the gate surface with this class is carried by the AC_φλ admission itself, which supplies its own gate surface. The audited claim_scope already read the lemma this way (*"The scoped algebra closes on the supplied circulant surface"*); the registry-reduction consequence already defers to the separate registrability bridge note and is unchanged.

Load-bearing deps after this PR: `MINIMAL_AXIOMS_2026-06-05.md` only.

## Changes

- `docs/TIER_A_KORBIT_DETERMINANT_AND_ORIENTATION_INVARIANCE_BOUNDED_NOTE_2026-06-09.md` — supplied-class stipulation; staggered link → backticked context entry; date amendment.
- `scripts/frontier_tier_a_korbit_determinant_orientation_invariance_2026_06_09.py` — staggered-note read removed; two note-consistency checks added (supplied-class stipulation present; staggered cited context-only, no markdown link).
- `logs/runner-cache/frontier_tier_a_korbit_determinant_orientation_invariance_2026_06_09.txt` — regenerated (SUMMARY: PASS=11 FAIL=0).

## Why this matters (θ-retirement chain, PR #3511)

This row is the sole load-bearing dep of the registrability bridge note (`registrable_readout_additive_even_phase_free_...`, gate 3 of #3511) that is not yet retained-grade. With this repair the row stands on the axiom node alone; the bridge note's first audit then no longer bounds on the 13-dep staggered tower. Recommended audit order across the three repair PRs landing today: **korbit → registrable_readout → theta_p2**.

[...]

PR #3656, opened 2026-06-12
science(bounded): hw-complement reading is registration-equivalent on the supplied slot model (AC_φλ (iii) anchor reduction)

jonathonreilly:
## Summary

New narrow bounded theorem (1 note + 1 runner + 1 cache): on the supplied three-slot circulant model, **every Record-registrable scalar readout takes the same value on the matter-as-hw=1 and matter-as-hw=2 complement readings**. Proof composes two landed results: the registrability theorem (Consequence B: registrable surface = symmetric functions of the unordered spectrum; additive+even kills odd content) with the hw-complementation equivariance facts (reproven in-runner from scratch: C∘R = R∘C on all corners, orientation preserved — reversal explicitly fails; the unique non-symmetric C₃-invariant line (u−v)(slots) = −6√3·B³·sin(3δ) exactly; degree≤3 invariant-space dimension count 8 = 7+1).

**Consequence:** within AC_φλ sub-admission (iii), the R1b semantic anchor ("hw=1 is the physical generation sector") carries no registrable content on this slot model — at this level it is frame/convention data. What remains of (iii) is the interpretive abstract-sector → physical-species identification and full-dynamics complementation equivariance, both explicitly out of scope (the hw support note's named open is untouched).

Firewall-compliant: selects no species reading, forces no r, derives no δ.

## Dependency hygiene (the W1b lesson applied)

Load-bearing deps: `minimal_axioms` + the registrability bridge ONLY — the runner reproves every finite-slot fact it uses, so the open_gate hw support note and the staggered gate are backticked context. The row becomes audit-ready exactly when the orientation-strip cascade (#3642 → #3643 → #3645) retains, same as the |δ| chain (#3648).

## Reviewer hardening (line-by-line pass)

- **A9b was decorative** (`True` condition) → replaced with the exact identity: (u−v)(slots, δ) + (u−v)(slots, −δ) = 0 (the line's K-even part vanishes identically).
- **A10 was tautological** (hw2 reading built via the complement bijection, then checked against itself) → rebuilt: the hw2 reading is generated independently by the rotation orbit; the check verifies order-consistency of the equivariant transfer (complement of the R-orbit = R-orbit of the complement) plus symmetric-function equality.
- Symbol collision removed (complementation was named `C` alongside the cyclic-shift matrix `C`).

Runner: SUMMARY PASS=17 FAIL=0. No audit status asserted or predicted.

### DIFF
--- a/docs/ACPHILAMBDA_HW_COMPLEMENT_READING_REGISTRATION_EQUIVALENCE_NARROW_THEOREM_NOTE_2026-06-12.md
+++ b/docs/ACPHILAMBDA_HW_COMPLEMENT_READING_REGISTRATION_EQUIVALENCE_NARROW_THEOREM_NOTE_2026-06-12.md
@@ -1,0 +1,136 @@
+# AC_phi_lambda hw-complement Reading Registration-Equivalence -- Narrow Theorem
+
+**Date:** 2026-06-12
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set or predict an audit outcome and does not edit the audit-lane-owned Tier-A registry.
+**Primary runner:**
+[`scripts/frontier_acphilambda_hw_complement_reading_registration_equivalence_2026_06_12.py`](../scripts/frontier_acphilambda_hw_complement_reading_registration_equivalence_2026_06_12.py)
+**Runner cache:**
+[`logs/runner-cache/frontier_acphilambda_hw_complement_reading_registration_equivalence_2026_06_12.txt`](../logs/runner-cache/frontier_acphilambda_hw_complement_reading_registration_equivalence_2026_06_12.txt)
+
+## Boundary
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed reviewed source-only packet on main as 274175109.\n\nReview-loop changes before landing:\n- renamed the new note from `...NARROW_THEOREM...` to `...BOUNDED_THEOREM...` to match the declared claim type and supplied-slot-model boundary;\n- added controlled `Type: bounded_theorem`;\n- updated the runner self-check path/docstring and regenerated the SHA-pinned cache;\n- verified context notes remain non-link context and the dependency graph seeds only `minimal_axioms` plus the registrability theorem.\n\nChecks: runner SUMMARY PASS=17 FAIL=0, py_compile OK, vocab_lint OK, run_pipeline OK, audit_lint --strict OK with known repo notices only. Generated audit/publication/front-door outputs were restored before commit; independent audit lane grades the new row.

PR #3657, opened 2026-06-12
feat(narrow): theta identified — the word-axis ratio is an explicit packet quantity (15 digits), two-term asymptotic verified

jonathonreilly:
## Summary
- **The certificate program's central object is identified.** The word-count convergence ratio is the explicit closed form `θ = (ℓ_η(1,0)/ℓ_η(0,0)) · √(D₁₀/d₁₀) · t((1,0),(0,0))/t((0,0),(0,0)) = 0.263745855973467` — the (1,0) channel's coupling into the dominant trivial channel, dressed by the η-readout weight ratio and channel normalization. Every factor is a named finite-packet quantity.
- Supporting structure: the tensor-Gram factorization `S_k = √(dd')·t^{∘k}` verified to 1e-16; the dominant t entry confirmed as the measured eigenvalue factor (0.357066); source sensitivity `C_source = 0.4102` computed; the leading asymptotic `P_∞ − P_k ~ C_source·θ^{k−1}` verified on every measured rung.
- **Reviewer verification: independent implementation of the formula matched to 15 digits** (and caught that the ℓ_η convention is load-bearing — now stated explicitly).
- Honest residual: identification + verified two-term asymptotic; the proof-grade all-k remainder (third-scale argument, computed constants) is the named next and final step of the word-axis certificate.
- Runner: **TOTAL: PASS=22 FAIL=0**.

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh) under a reviewer spec carrying the confirmed anchor, nine exclusions, and the smoking gun; formula independently verified before landing.


jonathonreilly:
Landed reviewed source-only packet on main as b3cb1f9ea.\n\nReview-loop changes before landing:\n- added controlled `Type` / `Claim type` metadata as `bounded_theorem`;\n- removed stale `.claude/tmp` context references and updated the runner hygiene check to enforce their absence;\n- kept prior scratch packets as non-authority context only;\n- changed `PLAQUETTE_SELF_CONSISTENCY_NOTE.md` from a markdown dependency to plain-text context so it does not seed as load-bearing.\n\nChecks: runner TOTAL PASS=22 FAIL=0, py_compile OK, vocab_lint OK, run_pipeline OK, audit_lint --strict OK with known repo notices only. The seeded row is `bounded_theorem`/`unaudited`; generated audit/publication/front-door outputs were restored before commit.

PR #3659, opened 2026-06-12
feat(gauge-dynamics): within-sector moment-relation deviations are sizeable but below the permutation null at every seed — power-limited consistency with one wrapped-normal relation (panel-narrowed)

jonathonreilly:
Wave-3 landing: see the note for claims/scope/panel edits. Source-only triple. Audit lane grades.


jonathonreilly:
Review-loop landed on main as `4b930ecdf`. I kept the claim bounded to finite permutation-control consistency for one wrapped-normal moment relation, added explicit dependency links for the supplied finite packet, and left audit status to the independent audit lane. Checks: vocab lint clean; py_compile clean; runner PASS=19 FAIL=0; cache fresh; audit pipeline and strict lint pass with existing accepted notices only.

PR #3669, opened 2026-06-12
science(bounded): OO/R-D premise relation — one-way at realization, inequivalent as laws (answers the named open)

jonathonreilly:
## Summary

New narrow theorem answering the durability-equivalence note's named open question verbatim (*"are R-D and the orbit-occupancy candidate equivalent, independent, or one-way related on the current surface?"*):

- **(P1) R-D ⇒ OO-realization, one-way.** Fix(φ)={0, ½, ∞} for the supplied separatrix flow φ(r)=2r²; the two registered side conditions consumed by the landed R-D chain (δ≠0 registration → kills r=0 [degenerate spectrum, reproven via B→0 limit]; unsigned branch → kills r→∞ [trace 3a=0 vs positive channel weights, reproven]) leave exactly r=½, which under the derived ρ-map is Z_d=π/g — one slot per outcome.
- **(P2) OO ⇒ per-point stationarity only** (φ(½)=½) — the per-point shadow of R-D, not the law.
- **(P3) Law-level equivalence FAILS**: counterexample flow ψ(r)=r² has Fix={0,1,∞}; the same side conditions select the *sector* cell and ½ is not ψ-stationary. The premises' agreement is a property of the supplied flow's fixed-point structure, not of the premise forms.
- **(P4)** Net: one-way at realization level given the landed flow; inequivalent as laws.

**For the ξ=1 owner decision:** adopting either premise yields the same realized registered surface given the landed flow; R-D is the logically stronger commitment (constrains all durably registered values; selection power rides on the flow + side conditions), OO the weaker static counting one. **No adoption is made or recommended; both premises remain proposed-NOT-adopted; no cell is selected** — everything is conditional on the supplied flow, whose own boundary ("not derived as the physical emergent charged-lepton records flow") is quoted.

## Dependency hygiene

Load-bearing: separatrix note (retained_bounded — the flow) + independence note (the OO statement + derived cells). The R-D chain and durability-equivalence notes are backticked context (premise statements quoted as proposals; the fixed-point set and both side-condition exclusions are reproven in this note's runner). Runner 14/14; all assembled booleans computed from derived objects (no decorative checks).

Wave 3 of the AC_φλ campaign. No audit status asserted or predicted.


jonathonreilly:
Review-loop landed on main in 60edef2c4 with wording/metadata cleanups: expanded the visible OO/R-D shorthand to orbit-occupancy / records-flow durability, added the machine-readable Type: bounded_theorem token, and kept both premises explicitly proposed/not adopted.\n\nScope landed: bounded theorem only for the relation between the two proposed premises on the supplied flow/current surface. It does not adopt either premise, select a cell, force r=1/2, derive the physical records flow, edit any registry, or change Tier-A/audit state.\n\nValidation: vocab_lint clean; py_compile clean; runner PASS=14 FAIL=0; cached_runner_output fresh; pipeline generated row as audit_status=unaudited/effective_status=unaudited/claim_type=bounded_theorem with deps only on the retained-bounded separatrix and orbit-occupancy independence notes; audit_lint --strict passed on the pipeline-generated validation state with only accepted notice classes. Generated audit/publication/front-door outputs were restored before commit.

PR #3671, opened 2026-06-12
feat(record-budget): erosion-rate envelope geometric at moderate eps, non-geometric near the projective endpoint; no tested closed form; censored-nonincreasing threshold time (panel-narrowed)

jonathonreilly:
Wave-4 landing: see note. Source-only triple. Audit lane grades.


jonathonreilly:
Landed on main with review-loop cleanup in a648171ce.\n\nReview notes:\n- kept the result as a bounded finite-model source triple, not an exhaustive no-closed-form/no-go claim;\n- added the record-erosion predecessor as the explicit dependency for the inherited model/Born cap;\n- normalized the claim type/status language and removed branch-local labels from the live note/runner;\n- verified runner/cache: PASS=16 FAIL=0;\n- ran docs/audit/scripts/run_pipeline.sh and python3 docs/audit/scripts/audit_lint.py --strict; generated row is unaudited/bounded_theorem pending independent audit.

PR #3676, opened 2026-06-12
science(bounded): occupancy kernel coefficient not fixed by the retained corner measure (Bit-B transfer route scoped, prerequisite named)

jonathonreilly:
## Summary

Wave 4 of the AC_φλ campaign — the Bit-B transfer-route probe, resolved at the scoping level after the read-the-primaries discipline caught two would-be dead ends: the naive first-order route is already computed in the static no-go (det D = |det M|² → r=1), and the retained surface contains **no transfer object at all** (substep2 explicitly disclaims the physical-lattice identification; time direction, transfer matrix, and reflection positivity are absent from both retained substep notes — grep-witnessed in the runner).

The theorem localizes sub-admission (i) at the measure level:

- **K1** — substep1's `Z_F[M] = det(M)` reproven by genuine Grassmann expansion (truncated exponential, top-monomial extraction with orientation sign; a self-check asserts no `det()` call substitutes for the expansion).
- **K2** — the freedom witness: rescaling the generation-doublet block `M_d → λM_d` preserves the single-pair form, nilpotency/dimension, and C₃-equivariance of the supplied circulant class, while scaling the doublet partition factor by λ² (two-real-pair kernel) vs λ¹ (single complex pair). The slot-count bookkeeping is already visible at the kernel level — and nothing on the retained surface pins it.
- **K3** — that λ-freedom maps exactly onto the occupancy weight-class atom `Z_d ∈ {2π/g, π/g}` under the derived ρ-map; the admissible r-set given λ free is the full binary {1, ½} (computed as a set, never asserted — no cell selected, the binary stays open).

**The named prerequisite (live route):** any dynamical kernel-coefficient determination must first supply a retained time-slicing/transfer structure for the corner realization. This refines the static no-go's own live-route naming from the measure side; no closing language anywhere (the note greps itself for it).

Deps: substep1 + substep2 only (both retained) — **the row is audit-ready on landing**. Context: the independence note (bookkeeping reproven in-runner) and the static no-go (scoping companion). Runner 16/16.


jonathonreilly:
Landed on main with review-loop/no-go-discipline cleanup in 27ce39257.\n\nReview notes:\n- kept the claim as a bounded retained-corner-measure negative only; the transfer/kernel route remains live;\n- added explicit Type metadata, descriptive theorem-part names, and a visible N1-N8 no-go-discipline gate;\n- preserved only the two retained substep notes as load-bearing deps; context notes remain non-load-bearing;\n- verified runner/cache: PASS=16 FAIL=0;\n- ran docs/audit/scripts/run_pipeline.sh and python3 docs/audit/scripts/audit_lint.py --strict; generated row is unaudited/bounded_theorem and generated status files were restored.

PR #3728, opened 2026-06-12
science: d=3 step-2 range growth; truncation protection = period-class dichotomy (bounded theorem)

jonathonreilly:
Wave-8 S1 (panel-driven rework). d=3 step-2 grows the coupling range with band-wise decay (L=12/14 converged 0.3%, L=8 disclosed box-limited). The headline panel catch became the theorem: even-d2 truncation protection is a PERIOD-CLASS DICHOTOMY — holds at 1e-14 for L=0 mod 4 (L=8,12), FAILS at L=2 mod 4 (L=10: 0.748, L=14: 0.747, gated >0.5) because the second chart's K-periods (L/2,L,L/2) acquire odd components and minimal-vector d2 parity decouples from chart parity. Commensuration-conditional, not generic. Anti-fabrication nonzero-before gate included. Runner PASS=21.

### DIFF
--- a/docs/D3_STEP2_RANGE_UNBOUNDED_PARITY_PROTECTED_BOUNDED_THEOREM_NOTE_2026-06-12.md
+++ b/docs/D3_STEP2_RANGE_UNBOUNDED_PARITY_PROTECTED_BOUNDED_THEOREM_NOTE_2026-06-12.md
@@ -1,0 +1,29 @@
+# d=3 Step 2 Grows the Range (Box-Disciplined: L=12/14 Converged Sub-1%, L=8 Disclosed as Box-Limited); Truncation Protection Is a Period-Class Dichotomy — Holds at L ≡ 0 mod 4, Fails at L ≡ 2 mod 4 (Bounded)
+
+**Date:** 2026-06-12
+**Type:** bounded theorem (the d=3 step-2 follow-on; cross-refs in-review, not graded)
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed to main in e39bf3ff7 (Add d3 step2 range period-class theorem).\n\nReview-loop disposition: accepted as a bounded finite-box theorem after cleanup. I renamed the surface from the PR's stale  file/runner naming to range-growth / period-class-dichotomy language, because the artifact proves finite-box range growth plus a period-class dichotomy, not an infinite-range theorem.\n\nScience boundary: kept the useful result: step-2 grows beyond the step-1 shell family with disclosed box dependence and band-wise decay; truncation protection holds for L ≡ 0 mod 4 and fails for L ≡ 2 mod 4. No flow, measure, primitive, axiom, or generic parity-protection claim is added.\n\nChecks: vocab_lint --fix clean; py_compile clean; runner/cache refreshed under the renamed path with hash MATCH and TOTAL: PASS=21 FAIL=0; marker scan clean; full audit pipeline complete with no errors; strict audit lint OK (existing notices only); row now seeds as bounded_theorem / unaudited.

jonathonreilly:
Correction to prior comment: the stale overclaiming name I removed was the literal `unbounded` file/runner wording. The landed files use range-growth / period-class-dichotomy naming instead.

PR #3729, opened 2026-06-12
science: harmonic-depth mechanism located in trajectory rank, censoring-disciplined (bounded theorem)

jonathonreilly:
Wave-8 S2 (panel-driven rework). The det-phase harmonic-depth ordering (K=6 saturates, K=3/4 ceiling) is located in numerical trajectory rank with the panel's censoring fix: window-64 ranks for K=3/4 are CENSORED lower bounds (>=64); the window-128 probe decensors them (71/84 vs K=5/6 36/35) and the ordering SURVIVES — K=6 smallest at both windows and all thresholds 1e-5/1e-6/1e-7 (fixed gates). Identical coupled-gap counts (3 each, gated) cannot explain the ordering — the mechanism is the coupling-weight distribution over tones (named follow-on). Anchors (landed ceilings + K=6 saturation) reproduced first. Runner PASS=19.

### DIFF
--- a/docs/HARMONIC_DEPTH_HANKEL_RANK_MECHANISM_BOUNDED_THEOREM_NOTE_2026-06-12.md
+++ b/docs/HARMONIC_DEPTH_HANKEL_RANK_MECHANISM_BOUNDED_THEOREM_NOTE_2026-06-12.md
@@ -1,0 +1,29 @@
+# The Harmonic-Depth Mechanism Located in Numerical Trajectory Rank (Window/Threshold-Relative): Censored Lower Bounds at Window 64, K=6 Smallest at Windows 64/128, Coupled-Gap Counts Identical (Bounded)
+
+**Date:** 2026-06-12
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed to main in b469917ba (Add harmonic depth Hankel rank mechanism note).

Review-loop disposition: accepted as a bounded, window/threshold-relative theorem after narrowing. I kept the rank/censoring result and changed the prose so the coupling-weight distribution is a remaining mechanism target/follow-on, not a fully proved causal mechanism.

Science boundary: the landed claim is exact finite L=3 trajectory-rank data: K=6 is smallest at windows 64/128 and thresholds 1e-5/1e-6/1e-7; K=3/4 window-64 ranks are censored lower bounds; identical coupled-gap counts (3 each) do not explain the ordering. No new primitive, measure, weight, generic depth law, or retained mechanism closure is claimed.

Checks: vocab_lint --fix clean; py_compile clean; cache hash MATCH, exit 0, TOTAL: PASS=19 FAIL=0; marker scan clean; full audit pipeline complete with no errors; strict audit lint OK (existing notices only); row seeds as bounded_theorem / unaudited.

PR #3738, opened 2026-06-12
science(bounded): the corner mode-set fork is registration-blind — resolution layer is record dynamics

jonathonreilly:
## Summary

Wave 7b — the interlock theorem. Wave 6 exhibited the occupancy binary as a mode-set fork on the corner transfer structure; this note proves **which layer can and cannot resolve it**:

- **Y1/Y2** — both branches' partition objects (per-channel `∏_k det(1+t_k)`; per-K-orbit with an explicitly-bookkeeping one-slot factor) satisfy the registrability constraints: log-additive over disjoint record factors, constant on K/CPT orbits (δ-parity exact, channel-swap invariant).
- **Y3** — the branches differ generically (relative gaps 21%/17% at the two domain points, both positive); the admissible branch set is BOTH — assembled from the computed booleans, never asserted.
- **Y4** — negative control: any registration-layer selector would have to violate additivity or orbit-constancy. **The fork is registration-blind.** Discrimination belongs to record dynamics — exactly where the companion arrow-invariance theorem (#3737) just removed the arrow obstruction.

Together with #3737: registration cannot choose the branch; dynamics must; and the dynamics' fixedness criterion no longer depends on the arrow. The remaining open objects are exactly two, both named: the coarse-graining prong and the R-D bridge premise. This sharpens the landed independence theorem and selects nothing (firewall grep-checked; the orbit factor is declared bookkeeping, not a realization claim).

Deps: RP two-step (retained) + the registrability bridge (clears via the #3642→#3643→#3645 cascade). Runner 25/25.

No audit status asserted or predicted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3737, opened 2026-06-12
science(bounded): R-D fixedness is arrow-invariant — the arrow half of the coarse-graining+arrow admission drops out of the R-D route

jonathonreilly:
## Summary

Wave 7a of the AC_φλ campaign — first of the derivation-route theorems (per the owner's primitive-bar correction: no adoption asks; shrink the admitted content by theorem instead).

The two retained record-flow notes describe one structure read in opposite directions: **g(r)=√(r/2) is exactly φ⁻¹ of the sharpening map φ(r)=2r²** (both compositions = identity, symbolically on r≥0). Consequences, each verified exactly:

- **Fixedness is arrow-invariant**: Fix(φ) = Fix(g) = {0, ½} plus the projective point (s=1/r conjugation fixes s=0 for both) — instantiating the general lemma Fix(f)=Fix(f⁻¹).
- **Stability is arrow-dependent**: φ′(½)=2 (repeller) vs g′(½)=½ (attractor), product 1 — the landed "unstable separatrix" and "stable attractor" claims are the two orientations of one structure.
- **Therefore the R-D selection mechanism** (durably registered value = flow-fixed point; the registered side conditions δ≠0 and the unsigned branch exclude 0 and ∞, leaving exactly {½} — reproven for *both* maps) **is invariant under the arrow choice**, because it consumes only the fixed-point set, never stability. A negative control shows any stability-consuming selector would be arrow-dependent.

**Net: the arrow half of the named "which coarse-graining + which arrow" admission provably does not bear on the R-D route.** What remains open and named: the coarse-graining prong (the 2-sector partition, custody selector (i)) and the R-D bridge premise itself. R-D stays proposed-NOT-adopted; no occupancy cell selected; r never forced.

Deps: the separatrix and thermalizing-arrow notes (both retained_bounded; their own physical-boundary sentences quoted and grep-pinned) — **the row is audit-ready on landing**. Runner 24/24, no decorative checks.

No audit status asserted or predicted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3740, opened 2026-06-12
science(bounded): R-D bridge anatomy — the flow is agreement-conditioned double registration; one statistics atom remains

jonathonreilly:
## Summary

Wave 8a — the R-D bridge premise dissected to its irreducible content.

- **G1 (naive route killed):** the canonical registration map is idempotent — `D(D(M)) = D(M)` entrywise on generic M — so re-registration-as-re-pinching induces the *identity* on the weight bookkeeping and cannot ground R-D selection (negative control: it fixes every r).
- **G2 (the flow identified):** independent double registration of the same partition, conditioned on agreement, sends the bookkeeping to `p_i²/(p_s²+p_d²)`; in the coordinate `x = p_d/p_s = 2r` this is **exactly the retained sharpening map r→2r²**, with the retained thermalizing map as the same identity read backwards. Fixed points re-solved: {0, ½} + projective ∞.
- **G3 (the atom named):** the bridge premise reduces to one statistics-layer atom — *repeated registration composes independently on the weight bookkeeping* (agreement-conditioning then forces the retained map). A perfectly-correlated composition witness yields the identity flow instead, proving the atom does real work. The Born/unraveling assembly lane is named as the place it would be discharged; it is not discharged here, and no probability rule is imported (the bookkeeping is a declared reconstruction-level device).

With #3737 (arrow-invariance) and #3738 (registration-blindness), the occupancy lane's open content is now: **this one atom + the coarse-graining prong + the durability-to-weight coupling** — all named, none resolved, route live.

Deps: separatrix + thermalizing-arrow notes (retained_bounded) + minimal_axioms (whose non-supply clause is precisely why the atom is an atom). Runner 20/20, no decorative checks.

### DIFF
--- a/docs/RD_BRIDGE_ANATOMY_AGREEMENT_CONDITIONED_DOUBLE_REGISTRATION_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/RD_BRIDGE_ANATOMY_AGREEMENT_CONDITIONED_DOUBLE_REGISTRATION_BOUNDED_NOTE_2026-06-12.md
@@ -1,0 +1,170 @@
+# R-D Bridge Anatomy: Pinching Is Idempotent, the Supplied Flow Is Agreement-Conditioned Double Registration, and One Statistics Atom Remains (Bounded Note)
+
+**Date:** 2026-06-12
+**Claim type:** bounded_theorem
+**Status:** source proposal; independent audit required. This note proves the
+bounded algebraic anatomy below and names one unresolved statistics atom. It
+does not set or predict an audit outcome, adopt R-D, fix `r`, or edit any
+registry or audit data file.
+**Status authority:** audit status remains set only by the independent audit
+lane. This file is not status authority for any premise, registry row, or
+promotion decision.
+**Primary runner:** `scripts/frontier_rd_bridge_anatomy_agreement_conditioning_2026_06_12.py`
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as dd9305b30.\n\nReview-loop adjustments before landing:\n- corrected the note to name the checked-in runner cache instead of saying no cache exists\n- updated companion context from in-review shorthand to the landed companion note filenames\n- regenerated the runner cache after the runner/note cleanup\n\nReview-loop result:\n- accepted as a scoped bounded theorem/anatomy note\n- preserves the firewall: statistics atom named, not discharged; no R-D adoption; no probability rule import; no r/cell selection\n\nVerification run:\n- vocab_lint clean for touched note/runner\n- py_compile clean\n- runner: PASS=20 FAIL=0\n- runner cache status: fresh\n- review marker scan clean\n- full audit pipeline completed with no errors\n- strict audit_lint completed with existing notices only\n- generated audit/publication/front-door outputs restored before commit\n\nPipeline row check before restore: rd_bridge_anatomy_agreement_conditioned_double_registration_bounded_note_2026-06-12 = bounded_theorem / unaudited / unaudited, deps are separatrix, thermalizing-arrow, and minimal_axioms. Pipeline still reports 20 graph back-edges; noted as allowed for later cleanup.

PR #3753, opened 2026-06-12
science(bounded): the product-form premise weakens to outcome-level factorization (frame-blind, strictly weaker, named)

jonathonreilly:
## Summary

Wave 11a — the wave-10 residue compressed to its minimal honest form, informed by (without consuming) the unraveling lane's measured results.

- **L1** — the entire reduction chain re-derived from outcome-level factorization alone: `m(j,k) = p_j·p_k` is the whole hypothesis; the conditioned flow x→x² and the retained coordinate family follow with no joint-state object anywhere.
- **L2** — strictness witness: an explicitly non-product joint density (off-partition covariance 1/9, which excludes *any* product representation since all products have vanishing product-observable covariance) whose four partition weights nonetheless factor exactly.
- **L3** — frame-blindness: the registered weights are invariant under every partition-commuting unitary — precisely the layer where the unraveling lane *measured* its O(1) stationarity failures ("the stationarity failure is concentrated in the bi-frame", quoted as scope-only context; those rows are unaudited and stay backticked) — with a non-commuting negative control proving the invariance does work.
- **L4** — the premise relocates to **outcome-independence on the registered partition**: strictly weaker than iid/product instances, untouched by the measured frame failures, and sitting exactly in the unraveling lane's own explicitly-untested slot ("cross-edge independence and convolution structure are not tested here", quoted). Named, not discharged.

**The occupancy lane's full residue after 18 PRs:** outcome-level independence (named, test surface identified) + the outcome dictionary (tri-guise binary) + the durability-to-weight coupling. Every reduction today was a theorem on retained dependencies; nothing was adopted, forced, or discriminated.

Deps: the two retained Gleason rows + minimal_axioms — audit-ready on landing. Runner 20/20.

### DIFF
--- a/docs/PRODUCT_FORM_PREMISE_WEAKENS_TO_OUTCOME_FACTORIZATION_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/PRODUCT_FORM_PREMISE_WEAKENS_TO_OUTCOME_FACTORIZATION_BOUNDED_NOTE_2026-06-12.md
@@ -1,0 +1,183 @@
+# Product-Form Premise Weakens to Outcome-Level Factorization
+
+**Date:** 2026-06-12
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note sets,
+predicts, and estimates no audit verdict and edits no registry.
+**Primary runner:** `scripts/frontier_product_form_weakens_to_outcome_factorization_2026_06_12.py`
+
+## Boundary
+
+This note proves only L1-L4 below. The weakened premise is named, not
+discharged, and not asserted. R-D stays proposed. No cell is selected. r is
+never fixed. The occupancy binary stays open.
+
+It does not consume the unaudited unraveling measured results as authority; it
+quotes them only to define scope. It does not adopt R-D, select a cell, fix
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed this as a source-only bounded theorem slice on `main` in commit `121fc4dcb`.

Disposition:
- Salvaged/landed: note, runner, and fresh runner cache.
- Narrow fix applied: added standard source-proposal `Status` metadata and runner-cache pointer; theorem body and runner math left intact.
- Claim boundary: `bounded_theorem`, not retained; outcome-level factorization is named and tested as strictly weaker than state-product form, but not discharged.
- Dependencies checked: Gleason projection row (`retained`), Busch bridge row (`retained_bounded`), and `minimal_axioms` as an axiom-premise node. This stays a bounded landing because one dependency is bounded and no audit verdict is applied here.
- Audit compatibility: full pipeline seeded exactly one new unaudited row and strict lint passed; generated audit/publication/front-door outputs were restored before commit per review-loop policy.
- Checks: vocab lint, py_compile, direct runner (20/20 PASS), refreshed cache, independent rational check of the witness eigenvalues/weights/covariance, full audit pipeline, strict audit lint, marker/link/whitespace scans.

No audit status or effective retained status was landed; the independent audit lane owns that next step.

PR #3702, opened 2026-06-12
science(bounded): corner-axis free transfer extension — the wave-4 prerequisite met at free level; the mode-set fork exhibited

jonathonreilly:
## Summary

Wave 6 — the campaign capstone construction. The wave-4 theorem named the prerequisite (a retained transfer structure for the corner realization); the wave-5 theorem supplied the mechanism (trace correspondence fixes the kernel normalization). This note **builds the corner object at the free per-channel level** and shows the occupancy binary surviving in exactly one place.

- **X1** — the free corner-axis action (retained 1+1d staggered ⊗ the supplied circulant mass class) block-diagonalizes over the circulant eigenbasis into three decoupled channels with masses λ_k(δ) (unitary-conjugation residual 4×10⁻¹⁶); each channel inherits the retained 2-step construction verbatim (dispersion anchor 5×10⁻¹⁶).
- **X2** — `T²_corner = ⊗_k T²_k` is positive Hermitian (Fock dim 64): a transfer structure for the corner-axis realization exists on the supplied positivity domain — which **consumes the positive-real mass orientation as a supplied-domain condition** (the θ-side premise surfacing inside the AC_φλ lane; cross-referenced, not derived).
- **X3** — the trace correspondence carries over exactly (`Tr Γ = det(1+t) = ∏_k det(1+t_k)`, genuine Berezin expansion at canonical measure, rescaling forced to 1).
- **X4** — K-covariance: λ₂(δ) = λ₁(−δ) symbolically; the doublet channels form one K-orbit, the singlet is K-fixed.
- **X5** — **the mode-set fork, exhibited and not resolved**: whether the registered Fock occupancy of the doublet counts channels or K-orbits is the orbit-occupancy premise verbatim, now localized as a single explicit fork on a concrete object. Both branches map to the two occupancy cells (computed r-set = the full binary, no pin); the correspondence fixes the normalization within each branch; the branch exponents {4, 2} display the occupancy factor 2 inside the correspondence itself.

Firewall: no cell selected, r never fixed, neither OO nor R-D adopted or favored, free (U=1) case only, no closing language (all grep-checked in-runner). Deps: RP two-step (retained) + substep1 (retained) + the registrability bridge (clears via the #3642→#3643→#3645 cascade). Runner 18/18.

[...]

PR #3763, opened 2026-06-12
science(bounded): hw-complement registration-equivalence extends to the free corner-transfer dynamics (species-bridge lane)

jonathonreilly:
## Summary

Wave 12 — the species-bridge (iii) lane walks up one dynamics level. The slot-model theorem (#3656: no registrable readout distinguishes the hw=1 from hw=2 reading) now **extends to the free corner-transfer dynamics** (#3702's structure):

- **M1** — the complement reading induces the same channel-mass multiset {λ_k(δ)} (bijection commutes with the rotation, orientation preserved — reversal explicitly fails; slot-level facts reproven from scratch).
- **M2** — the two readings' corner transfer matrices are **unitarily equivalent** via the induced channel permutation on the 64-dim Fock space (residuals ~10⁻¹⁸).
- **M3** — every registrable datum of the free transfer object coincides: Fock traces at N=1–3 (4×10⁻¹⁶), dispersion multisets, and the canonical-normalization forcing — which breaks *identically* under a λ=2 rescale for both readings (genuine Grassmann expansion, self-checked); an explicit additive K-orbit-constant witness (log Tr Γ) evaluates equally.
- **M4** — including the free dynamics does not make the hw choice registrable. **The named open narrows to the interacting/gauge level** — the next path this opens, nothing closed.

Anti-churn note: the queued "registered-mode-set dynamics" item was crossed off rather than dispatched — wave 9's dictionary-invariance theorem already dissolved it.

Firewall: no physical species reading selected; free (U=1) case only; r never fixed; the occupancy binary untouched (all grep-checked in-runner). Deps: RP two-step (retained) + substep1 (retained) + the registrability bridge (cascade) — audit-ready with the rest of the queue. Runner 17/17.

### DIFF
--- a/docs/HW_COMPLEMENT_EQUIVALENCE_EXTENDS_TO_FREE_CORNER_TRANSFER_DYNAMICS_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/HW_COMPLEMENT_EQUIVALENCE_EXTENDS_TO_FREE_CORNER_TRANSFER_DYNAMICS_BOUNDED_NOTE_2026-06-12.md
@@ -1,0 +1,171 @@
+# hw-complement Registration-Equivalence Extends to Free Corner-Transfer Dynamics -- Bounded Note
+
+**Date:** 2026-06-12
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Status:** source-side bounded theorem on the supplied free corner-transfer
+surface only. No audit status is asserted here.
+**Status authority:** independent audit lane only. This source note does not
+set or predict an audit outcome and does not edit any audit-lane-owned registry,
+ledger, queue, or publication-status surface.
+**Primary runner:** `scripts/frontier_hw_equivalence_free_transfer_dynamics_2026_06_12.py`
+
+## Boundary
+
+This note proves M1-M4 below only on the supplied free corner-axis surface. The
+surface is finite and explicit: the corner cube `{0,1}^3`, the rotation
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the source-only bounded packet on main in commit 1957cf96e.

What landed:
- Added the HW-complement free corner-transfer registration-equivalence bounded note.
- Added the paired runner and refreshed runner cache.

Review fixes applied before landing:
- Changed the free two-step dependency wording to retained-bounded, not bare retained.
- Removed the stale runner docstring claim that no cache is generated.
- Tightened runner exit to require the exact 17/17 PASS count.

Checks:
- py_compile passed for the runner.
- Direct runner passed 17/17.
- Independent compact recomputation checked complement spectrum preservation, rotation commutation, failed reversal, Fock permutation/transfer diagonal agreement, and trace equality.
- Forced cache refresh passed.
- Audit pipeline and strict lint passed for validation, then generated audit/publication outputs were stripped before commit.
- Final gates passed: diff check, portable markdown-link scan, no staged audit/effective-status surfaces, generated-output-clean gate, staged cache freshness.

Disposition:
- The row seeds as bounded_theorem / unaudited, queued for independent audit.
- It is not retained: registrable_readout_additive_even_phase_free_narrow_theorem_note_2026-06-10 is still unaudited in the dependency chain.
- Scope remains free U=1 only; no physical species selector, no r fixation, no occupancy-binary closure, and the interacting/gauge extension remains open.

PR #3768, opened 2026-06-12
science(bounded): corner transfer extends to fixed gauge backgrounds — both interacting opens narrow to U-integrated

jonathonreilly:
## Summary

Wave 13 — the fixed-background climb, on the retained fixed-gauge engine (`RP_P2_GAUGE_EXTENSION`, retained_bounded).

- **N1/N2** — the corner-axis channel decomposition survives a fixed spatial gauge background (the generation mass and the background act on commuting factors); per-channel two-step positivity holds config-by-config per the retained engine, and the tensor corner transfer is positive Hermitian per background (dim 64; U=1 recovers the free wave-6 case exactly).
- **N3** — the trace correspondence and the canonical-normalization forcing (λ=1) hold config-by-config, with a genuine Grassmann-expansion verification at a complex background (self-checked, no det substitutes).
- **N4** — K acts on the background as U→U*: the channel kernels satisfy t_k[U](δ) = t_σ(k)[U*](−δ) (computed); the hw-complement registration-equivalence extends **exactly on K-real backgrounds** and in conjugated-background form generally. The **N4d honesty control** computed the same-U gap at a complex witness, found it zero (2×10⁻¹⁶), and the note records that as empirical **without promoting it beyond the stated scope** — a future one-line strengthening candidate (trace-reality argument), flagged, not claimed.
- **N5** — both interacting opens — the occupancy lane's transfer chain and the species-bridge equivariance — **narrow to the U-integrated level**: the measure over backgrounds is the one named object remaining, the next path this opens.

Firewall: fixed background only; no readings or cells selected; r never fixed; no closing language (all grep-checked). Deps: the fixed-gauge engine + substep1 + the registrability bridge. Runner 28/28.

### DIFF
--- a/docs/CORNER_TRANSFER_EXTENDS_TO_FIXED_GAUGE_BACKGROUNDS_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/CORNER_TRANSFER_EXTENDS_TO_FIXED_GAUGE_BACKGROUNDS_BOUNDED_NOTE_2026-06-12.md
@@ -112,13 +112,21 @@ Therefore:
   corner transfers are unitarily equivalent by the channel-swap permutation,
   and the registrable data checked by the runner, including traces and
   dispersions, agree.
-- On general backgrounds, the exact statement is the conjugated-background
-  statement: reading-1 data at `U` equal reading-2 data at `conj(U)`. The
-  exact same-background equivalence is asserted only on the `K`-invariant
-  class. The small `L_s = 2` witness may show accidental broader trace equality
-  in the runner; that empirical fact is not promoted beyond this stated scope.
+- On general backgrounds, the operator-level statement is the
+  conjugated-background statement: reading-1 data at `U` equal reading-2 data
+  at `conj(U)`; operator-level unitary equivalence at the same background is
+  asserted only on the `K`-invariant class.
+- **(strengthened, second pass) Registrable trace data are same-background
+  equivalent at EVERY fixed background.** Three computed legs: traces of the
+  positive Hermitian corner transfer are real; conjugating the background
+  conjugates the transfer matrix, so trace data at `conj(U)` are the complex
+  conjugates of those at `U`; real and conjugate together force equality, and
+  with the conjugated-background statement the same-background trace gap
+  between the two readings vanishes — a theorem-backed runner condition, not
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Second pass (reviewer-authored): the N4d honesty control is upgraded from an empirical flag (whose check condition was decorative — caught in wave-14 review) to a theorem-backed condition: traces of the positive corner transfer are real (N4d-i), background conjugation conjugates the transfer entrywise (N4d-ii), so registrable trace data are conjugation-blind and the same-background equivalence of the two hw readings holds at EVERY fixed background (N4d-iii, gap now a condition). Note scope updated precisely — operator-level same-background equivalence beyond the K-real class is still not claimed. Runner 28→30, FAIL=0, cache regenerated.

jonathonreilly:
Review-loop landed the bounded fixed-background packet on main in commit 5bf1caea4.

Review fixes applied before landing:
- Cleaned source metadata to Type: bounded_theorem plus an audit-boundary statement, without duplicate claim-type/status-authority boilerplate.
- Removed false runner wording that no cache is generated.
- Tightened the runner exit gate and note validation target from PASS>=18 to exact PASS=30 FAIL=0.
- Refreshed the runner cache after those edits.

Checks:
- py_compile passed.
- Direct runner passed 30/30.
- Independent compact recomputation checked circulant diagonalization, positive channel kernels, K/conjugated-background doublet swap, and same-background registrable trace equality on the fixed-background witnesses.
- vocab_lint clean.
- Audit pipeline and strict lint passed for validation; generated audit/publication outputs were stripped before commit.
- Generated row inspected as bounded_theorem / unaudited / unaudited with deps on rp_p2_gauge_extension_and_realization_residual_note_2026-05-28, staggered_dirac_substep1_grassmann_forcing_bridge_narrow_theorem_note_2026-05-16, and registrable_readout_additive_even_phase_free_narrow_theorem_note_2026-06-10.

Disposition:
- Landed as bounded source, queued for independent audit.
- Scope remains fixed background only. It does not integrate over gauge backgrounds, does not supply a gauge measure or full gauge dynamics, does not select species/occupancy, does not fix r, and leaves the U-INTEGRATED level as the named open next path.
- No audit verdict or generated effective-status surface was committed.

PR #3793, opened 2026-06-12
feat(narrow): pullback keystone — candidate refuted exactly; H_Dloc tracks Δ through β⁸; the two open programs converge on one wall

jonathonreilly:
## Summary
- The keystone attempt, resolved by its own mandatory falsifier: the sector shortcut is **unlicensed**, and the raw candidate `P_full = H_mark + H_Dloc` is **refuted exactly** at β⁵ (mismatch 1/944784) — the strong-coupling series computes both sides independently, so the refutation is ground truth, not convention.
- **The found structure:** H_Dloc matches the connected series Δ through β⁸, with the β⁹ residual named exactly (−13/49589822592) — the marked share tracks connected corrections to high order; the ninth order marks where the compressed environment's β-dependence becomes irreducible.
- **The convergence:** the identification keystone now visibly requires the physical environment's β-dependence — the *same* missing object as the physical-ρ program. The lane's two named opens are one wall seen from two sides.
- Runner: **TOTAL: PASS=10 FAIL=0** (exact rationals both sides).

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh) under reviewer spec; exact mismatch and series values verified before landing.


jonathonreilly:
Landed on main in eea71ecf9 as part of the plaquette source-sector bridge cluster. Review-loop kept this as a scoped no_go: exact local-packet falsifier plus missing current-authority pullback, not a broad foreclosure of future physical residual-environment repair. The no-go discipline gate remains in the source note, runner/cache were refreshed, and generated audit/status outputs were stripped before commit. The row remains unaudited for the independent audit lane.

PR #3744, opened 2026-06-12
science: d=3 commensuration criterion with out-of-sample L=16/18 confirmation (bounded theorem)

jonathonreilly:
Wave-9 S4. The wave-8 dichotomy promoted to a criterion ON THE TESTED GRID L={8,10,12,14,16,18}: even-d2 truncation protection holds iff every second-chart period (L/2,L,L/2) is even (= L=0 mod 4 on this family). The parity-correspondence finite check is gated independently per L AND gated equal to the measured H_kd_after dichotomy across the grid (the iff). OUT-OF-SAMPLE: L=16 predicted-hold confirmed at 0.0; L=18 predicted-fail confirmed at 0.747. Wave-8 anchors reproduced; anti-fabrication nonzero-before gates at every L. Scope: this chart family, this grid. Panel: land_with_edits (grid qualification), applied. Runner PASS=15.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3798, opened 2026-06-12
science: commensuration period-parity lemma, combinatorial to L=30 + residue case split (bounded)

jonathonreilly:
Wave-10 S2 (extends #3744). The parity correspondence verified combinatorially on the extended grid L={8,...,30} (no Hamiltonians; full coset-pair enumeration, e.g. L=26: 13^6=4,826,809 pairs, 1.8M mismatches, first witness printed) — holds iff L=0 mod 4 at every grid point. SymPy residue-class case split sanity-checks the L mod 4 table (stated as formula/table check, NOT an independent proof — panel wording applied; the odd-axis parity-flip branch forces the mismatch, including low negative wraps). H_kd ties remain the landed #3744 result; only L=8/10 anchored here (gated vs frozen wave-9 values). Anti-fabrication: the failing L=10 coset exhibited. Panel: land_with_edits, applied. Runner PASS=21.

### DIFF
--- a/docs/COMMENSURATION_GENERAL_LEMMA_PERIOD_PARITY_BOUNDED_THEOREM_NOTE_2026-06-12.md
+++ b/docs/COMMENSURATION_GENERAL_LEMMA_PERIOD_PARITY_BOUNDED_THEOREM_NOTE_2026-06-12.md
@@ -1,0 +1,60 @@
+# Commensuration General Lemma Period-Parity Bounded Theorem Note
+
+Draft date: 2026-06-12
+
+Runner: [`scripts/frontier_commensuration_general_lemma_2026_06_12.py`](../scripts/frontier_commensuration_general_lemma_2026_06_12.py)
+
+## Claim
+
+For the d=3 step-2 chart family with K-periods `(L/2, L, L/2)` and even `L >= 8`,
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed this as `cc86c1b20` (`review: add commensuration period parity lemma`).

I kept the combinatorial parity science but narrowed the landing boundary. The note is now a bounded source proposal with audit-authority/no-promotion metadata, an explicit dependency on `D3_TRUNCATION_COMMENSURATION_CRITERION_BOUNDED_THEOREM_NOTE_2026-06-12.md`, and a clear scope boundary: the parity lemma is for the stated `(L/2,L,L/2)` chart-period family, while no new `H_kd_after` equivalence is asserted for `L=20,22,24,26,28,30`; the direct Hamiltonian tie here is only the `L=8/10` frozen-parent anchor check. I also removed PR/wave-specific wording from the repo-facing note/runner.

Checks: runner PASS=21 FAIL=0; cache refreshed; independent period-parity check PASS; `py_compile` PASS; `run_pipeline.sh` PASS; `audit_lint.py --strict` PASS with existing notices only. The generated row seeds as `bounded_theorem` / `unaudited` with the retained-bounded d=3 commensuration parent. Generated audit/status outputs were stripped before commit.

PR #3746, opened 2026-06-12
science: depth ordering tracked by tone-weight concentration (bounded theorem)

jonathonreilly:
Wave-9 S2 (#3729's named follow-on). Per-state tone weights computed from the same eigenbasis-overlap machinery that found the 3 coupled gaps each; concentration (participation ratio + entropy + center-share) ordering matches the landed depth ordering on all four realized states (K6<K5<K3<K4 in PR, gated; capture-at-4 monotone with inverse PR, gated). Cyclic-shuffle null: support preserved (recomputed on the null, 3 tones each) while the depth-relevant gap-second-moment shifts by frozen nonzero deltas — the distribution, not the inventory, carries the ordering (gated realized-ordering claim; causal wording softened per panel). Anchors: capture table, w128 ranks 71/84/36/35, gap counts. Panel: land_with_edits, applied. Runner PASS=21.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3799, opened 2026-06-12
science: capture deficit = exact Hankel-Frobenius tail; depth ordering = tail arithmetic (bounded)

jonathonreilly:
Wave-10 S4 (sharpens #3746). Full no-floor tone spectrum per state: the no-floor coupled-gap SET is exactly (-3,0,+3) for every state (gated — new gate per panel); the trajectory equals the full tone sum at 1e-12 (reconstruction identity); capture-at-4 from the full spectrum equals the direct landed values (anchors 0.995/0.898/0.778 recomputed); the deficit equals the order-4 Hankel-Frobenius complement (algebraic identity of the shared SVD, verified 1e-12); the depth ordering IS the ascending tail-mass arithmetic (tail order 6,5,3,4, gated). Unused rank-threshold constants removed (panel). One-off post-exit hang on first run diagnosed as killed-watcher pipe artifact; reruns exit cleanly. Panel: land_with_edits, applied. Runner PASS=15.

### DIFF
--- a/docs/CAPTURE_DEFICIT_EXACT_TAIL_ACCOUNTING_BOUNDED_THEOREM_NOTE_2026-06-12.md
+++ b/docs/CAPTURE_DEFICIT_EXACT_TAIL_ACCOUNTING_BOUNDED_THEOREM_NOTE_2026-06-12.md
@@ -1,0 +1,95 @@
+# Capture Deficit Exact Tail Accounting: Bounded L=3 Realized-State Note
+
+Status: draft source note. The audit lane grades.
+
+Runner:
+
+```bash
+python3 scripts/frontier_capture_tail_exact_law_2026_06_12.py
+```
+
+## Scope
+
+This note is an exact finite-dimensional statement for the landed `L=3` gauge-link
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed this as `21d341e05` (`review: add capture tail accounting law`).

I kept the finite arithmetic claim and made the landing audit-compatible: source metadata, independent-audit authority/no-promotion language, repo-relative runner/cache links, retained-bounded parent deps, and an explicit realized-state primitive boundary. I also narrowed “physical Hamiltonian gaps” to the actual coupled eigenpair gap inventory and kept the theorem to the exact `L=3`, supplied-realized-state determinant-phase trajectories.

Checks: runner PASS=15 FAIL=0; cache refreshed; independent tail-complement arithmetic PASS; `py_compile` PASS; `run_pipeline.sh` PASS; `audit_lint.py --strict` PASS with existing notices only. The generated row seeds as `bounded_theorem` / `unaudited`, with deps on the harmonic-depth weight-distribution and Hankel-rank parents plus the registered realized-state primitive. Generated audit/status outputs were stripped before commit.

PR #3817, opened 2026-06-13
feat(bounded): E-center lift obstruction — the missing theorem isolated to one exact E-channel computation

jonathonreilly:
## Summary
- The keystone attempt: `rho_E = 21/4` does **not** derive from the current retained bank — and the obstruction is now maximally sharp: the exact chain `c_TE = -8/9 <=> q_E = 15/8 <=> rho_E = 21/4` is rigid, so all three faces are ONE missing anchor.
- **The missing computation, named**: an exact E-channel source/readout theorem (the E-analog of the retained exact-time-coupling note that the T-channel already has). The naturality no-go's own escape clause is quoted verbatim.
- Falsifiers ground the structure: `N_c=2 -> 22/3`, denominator-5 `-> 4`, denominator-12 `-> 51/4` (reviewer re-derived N_c=2 by hand).
- Route inventory complete; `kappa_EW` stays withheld per the Record/Quantum scope; live values fenced as comparator only; prior art differentiated.
- Runner: **TOTAL: PASS=46 FAIL=0** (re-run at review).

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh); escape-clause quote and falsifier arithmetic verified in review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3816, opened 2026-06-13
feat(bounded): readout-primitive bridge — membership without uniqueness; selection freedom = rho_E

jonathonreilly:
## Summary
- The s3-time lane's identification theorem, attempted first (the B1 lesson): `eta_floor_tf` checked against the gate's quote-anchored admissibility conditions.
- **Membership without selection**: it IS a licensed bright-class member; no one-hop authority makes it canonical. After T-side normalization the selection freedom is **exactly `rho_E = beta_E/alpha_E`** — the same object [#3814](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3814) isolated. Three independent analyses converge on the E-center lift as THE gap.
- Structural consistency proven: the ambiguity factors through the spatial prefactor `P_R` (residual 3.5e-18) — matching the rigidity note's localization.
- Live comparator datum (fenced): fitted `rho_E = 5.2575` vs target `21/4 = 5.25` — a 0.14% miss at exact tolerance.
- Runner: **TOTAL: PASS=14 FAIL=0** (re-run at review).

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh); convergence and factoring verified in review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3818, opened 2026-06-13
feat(bounded): T-side endpoint theorem — rho_T=-1, s_TE=-2 conditional on one readout row; gate residual unifies as row selection

jonathonreilly:
## Summary
- The gate's two T-side triple elements — `rho_T = -1` (`q_T = 5/6`) and `s_TE = -2` — **reproduce exactly** once the readout row `(alpha_E, alpha_T, beta_T) = (1, -2, 2)` is supplied.
- **The obstruction is row selection**: the retained exact carrier/time surface starts *after* `P_R` is supplied and does not derive the row. This is the same withheld object [#3816](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3816) and [#3817](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/3817) localized to — **the round's unification: the whole s3-time residual reduces to deriving the readout row**, which the surface and the primitive registry explicitly do not grant.
- The two sides stay distinct: even with the T-side row, `rho_E` is free (runner PASS) — the E-center lift is a separate row datum. Two row-selection gaps of the same type.
- Falsifiers (wrong gap/channel/shell/sign/`N_c`) all break visibly; live slopes fenced as comparator only.
- Runner: **TOTAL: PASS=24 FAIL=0** (re-run at review).

PROVENANCE: drafted by gpt-5.5 (codex CLI, xhigh); conditional reproduction and E-independence verified in review.

### DIFF
--- a/docs/QUARK_ROUTE2_T_SIDE_ENDPOINT_THEOREM_ATTEMPT_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/QUARK_ROUTE2_T_SIDE_ENDPOINT_THEOREM_ATTEMPT_BOUNDED_NOTE_2026-06-12.md
@@ -1,0 +1,314 @@
+# Quark Route-2 T-Side Endpoint Theorem Attempt
+
+**Date:** 2026-06-12
+**Claim type:** bounded derivation attempt with named obstruction.
+**Status authority:** independent audit lane only. This source note does not
+set, predict, or estimate any audit verdict. Effective status is
+pipeline-derived after independent audit and dependency closure.
+**Primary runner:** scripts/quark_route2_t_side_endpoint_theorem_attempt_bounded_2026_06_12.py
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a source-only bounded theorem boundary on main as 9507546ddf. I kept the result narrowed to conditional T-side reproduction plus row-selection obstruction: rho_T=-1, q_T=5/6, and s_TE=-2 reproduce exactly once the candidate row is supplied, but the cited carrier/time surface does not derive that row. I canonicalized metadata to bounded_theorem, made runner/cache links portable, removed untracked .claude/tmp context pointers, and added a runner guard for that hygiene. Runner/cache refreshed at PASS=25. Independent rational checks confirmed the T-side target and counter-witnesses: beta_T=0 keeps s_TE but breaks rho_T; alpha_E=2 keeps rho_T but changes s_TE; sign flip keeps rho_T but flips s_TE. Pipeline and strict lint passed; generated audit/effective-status outputs were stripped. The row remains unaudited/effective unaudited for the independent audit lane.

PR #3797, opened 2026-06-12
science: eps*(T) curve from the PT boundary root; mean collapse persists with disclosed m-trend (bounded)

jonathonreilly:
Wave-10 S3 (the eps*(T) named follow-on, built on #3743's PT machinery). The PT boundary root mu*_PT(m,T) on the fixed 4x4 (m,T) grid satisfies the MEAN quadrature collapse mu*^2 = m^2 + eps*(T)^2 within gated spreads 1.25-1.53% (< frozen 2%) — same scale as the landed finite-difference result. DISCLOSED (panel): the per-m residual signs are (+,+,+,-) at every T and the collapse values strictly decrease with m — a coherent m-trend inside the spread (bounded mean description, not exact m-independence). eps*(T) strictly increasing (gated); regression characterization: eps*^2 = c + d T^2 wins (residual 9.6e-4 vs linear 2.7e-3; 2e-3 gate frozen post hoc after GL=20 calibration, stated) — mu*^2 quadratic in both m and T, labeled regression not derivation. Anchor: landed 1.708 root reproduced. Panel: land_with_edits, applied. Runner PASS=25.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3820, opened 2026-06-13
science: eps*(T) T=0 boundary located natively; naive Sommerfeld coefficient fails by sign (honest negative, bounded)

jonathonreilly:
Wave-11 y2 (honest negative; panel: land_with_edits applied). Two findings on the m=0 axis of the eps*(T) curve. POSITIVE: the T=0 Fermi-surface response branch is natively located (mu*_0=1.5216) and the low-T PT roots extrapolate to it (Delta=6.2e-3, gated) — the T=0 boundary is derived. NEGATIVE (the diagnosis): the naive Sommerfeld coefficient (leading occupation-smearing of the fixed T=0 kernel) has the OPPOSITE SIGN from the measured finite-T slope (alpha=-9.27 vs d=+3.88, gated >200% mismatch) and is inconsistent with the wave-10 mean slope. So the boundary's T^2 growth is NOT occupation smearing of the T=0 kernel; a correct derivation must retain the kernel's explicit T-dependence — the next route is identified, not foreclosed. Scope: m=0 axis + coefficient attempt; the (m,T) surface stays the landed #3797 bounded result. Runner PASS=9.

### DIFF
--- a/docs/EPSSTAR_SOMMERFELD_T0_BOUNDARY_DERIVATION_BOUNDED_NOTE_2026-06-12.md
+++ b/docs/EPSSTAR_SOMMERFELD_T0_BOUNDARY_DERIVATION_BOUNDED_NOTE_2026-06-12.md
@@ -1,0 +1,107 @@
+# eps* Sommerfeld/T=0 Boundary Check on the m = 0 Axis (Draft)
+
+**Date:** 2026-06-12  
+**Script:** `scripts/frontier_epsstar_sommerfeld_derivation_2026_06_12.py`  
+**Status:** draft bounded note. The audit lane grades.
+
+## Claim Tested
+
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed a source-only bounded theorem boundary on main as fadfaed93a. I kept the positive/negative split explicit: the m=0 T=0 branch is located within the bounded tolerance, while only the naive fixed-kernel occupation-smearing Sommerfeld coefficient is rejected. I canonicalized metadata to bounded_theorem, replaced PR-number authority with tracked dependency links, added portable runner/cache links, and added a visible N1-N8 no-go-discipline gate that preserves the full T-differentiation route as follow-on work. Runner/cache refreshed at PASS=9. Pipeline and strict lint passed; generated audit/effective-status outputs were stripped. The row remains unaudited/effective unaudited for the independent audit lane.

PR #3821, opened 2026-06-13
science: capture-deficit extra tones = integer harmonic ladder of the base gap (bounded theorem)

jonathonreilly:
Wave-11 y3 (resolves the #3799 puzzle: 3 exact eigengaps but >3 active tones). The det-phase increment is nonlinear (argument of a determinant), so the trajectory carries the INTEGER HARMONIC LADDER 3k of the base gap. Base-angle readout: every active tone is an exact integer multiple of 3 (construction-consistency, labeled as such per panel), with nonzero k=2 (gap 6) in every state (anti-fabrication doubles). The real off-grid content gated explicitly: in the raw 256-step FFT the fundamental g=3 sits at bin 42.78, off the integer grid by 0.219 (gated > 0.15) — so the off-ladder leakage is forced, not a measurement choice. The non-fundamental ladder tail mass reproduces the landed deficit ordering (6,5,3,4 ascending) and ties to the Hankel tail. Anchors: gap set (-3,0,+3), capture 0.995/0.898/0.778, reconstruction 1e-12. Panel: land_with_edits, applied. Runner PASS=23.


jonathonreilly:
Review-loop landed a source-only bounded theorem boundary on main as 089a79d6c4. I kept the scope to finite L=3 realized states, replaced PR-number authority with the tracked capture-deficit note dependency, added canonical bounded_theorem metadata and portable runner/cache links, and kept the ladder-grid statement framed as construction consistency while preserving the raw FFT off-grid leakage boundary. Runner/cache refreshed at PASS=23. Independent checks confirmed the g=3 raw FFT bin is off-grid by 0.21915129689853 and that the Hankel-tail order (K6,K5,K3,K4) differs from the coefficient-tail order (K6,K5,K4,K3). Pipeline and strict lint passed; generated audit/effective-status outputs were stripped. The row remains unaudited/effective unaudited for the independent audit lane.

PR #3826, opened 2026-06-13
workhorse: use Opus 4.8 (ultra) as the supervising agent instead of Fable (for now)

JONbridgerapps:
## What

Swaps the named **supervisor** in the standing "Execution Mechanism" footer of all 15 repo science commands from **Fable** → **Opus 4.8 (ultra)**.

Files: `analyze`, `autopilot`, `design-experiment`, `first-principles`, `frontier`, `hypothesis`, `investigate-physics`, `physics-loop`, `progress`, `pstack`, `sanity`, `sweep`, `theory-review`, `validate`, `write-up`.

Each footer now reads:
> All execution under this command runs through the workhorse split (see the `workhorse` skill): **Opus 4.8 (ultra)** plans, writes specs, reviews every diff line-by-line, and lands; the strongest configured text worker via `codex exec` executes bounded note/runner drafting, scratch computation, structured extraction, and panel lens execution…

## Scope

- **Supervisor only.** The worker side (the `codex exec` text worker doing bounded drafting/computation/extraction/read-only panel lenses) is unchanged; verdict synthesis and landing stay with the supervisor.
- **Docs-only.** "Fable" was named only in this prose — there is **no model pin in any config** (`agents/openai.yaml` configures the *worker*, not the supervisor). The actual supervisor model is selected at session launch; this aligns the docs with running the split under Opus 4.8 (ultra).
- The workhorse `SKILL.md` is left generic ("the supervising agent") and unchanged — it never named Fable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3828, opened 2026-06-13
workhorse: supervisor = the model running the chat session (not a pinned model)

JONbridgerapps:
## What

Makes the workhorse **supervisor** always be **whatever model is driving the current chat session** — Fable when it's back, or the strongest available Claude model — rather than a hardcoded name.

- **15 command footers:** "Fable plans, writes specs, reviews every diff…" → "**the model running in this chat** plans, writes specs, reviews every diff…".
- **`workhorse/SKILL.md`:** adds one standing line — *"The supervising agent is the model running the current chat session — whichever Claude model is driving the conversation (e.g. Fable, or the strongest available Claude model at the time). It is not a separately pinned or named model; it follows the in-chat model."*

## Scope

- **Supervisor only.** The worker side (the strongest configured text worker via `codex exec`, doing bounded drafting/computation/extraction/read-only panel lenses) is unchanged; verdict synthesis and landing stay with the supervisor.
- **Docs-only, no pin.** There is no model pin in any config; this states the standing policy so the supervisor tracks the in-chat model automatically.

**Supersedes #3826** (which pinned the supervisor to "Opus 4.8 (ultra)" — replaced by this generic, follow-the-chat-model version).

### DIFF
--- a/.claude/commands/analyze.md
+++ b/.claude/commands/analyze.md
@@ -72,7 +72,7 @@ sections above. Create the directory if it does not exist.
 ## Execution Mechanism (standing — 2026-06-12)
 
 All execution under this command runs through the workhorse split (see the
-`workhorse` skill): Fable plans, writes specs, reviews every diff
+`workhorse` skill): the model running in this chat plans, writes specs, reviews every diff
 line-by-line, and lands; the strongest configured text worker via `codex exec`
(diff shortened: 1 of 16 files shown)


jonathonreilly:
Landed on main in 82c555e8a via review-loop.

PR #3845, opened 2026-06-13
B4 protects the marginal c_t=c_s even against the power-divergent (residual-D) regeneration on the OS0 discrete-tick surface (n-point/taste/a→0 stay open)

JONbridgerapps:
## What

A new bounded **positive theorem** advancing the interacting velocity-RG attractor note's open **residual D** — the genuine Collins-Perez-Sudarsky naturalness target — directly continuing the retained B4 marginal result.

The retained B4 radiative-stability theorem kills the **marginal** (dim-4) velocity anisotropy on the OS0 surface. Its residual D is the **power-divergent** (`a⁻²`) UV regeneration that feeds the lattice dim-6 anisotropy back into the marginal `c_s`. **Result:** the same B4 axis-relabel that zeroes the marginal *also* forbids the power-divergent feed-through on the discrete-tick surface — so on the framework's chosen OS0 surface the Collins power divergence is confined to the **Planck-suppressed dim-6 level**, not the marginal coefficient.

## Runner (17/17, every check a real numeric test)

- `δc_s_marginal` on the B4-cut discrete-tick measure = **6.9e-18** (machine zero). **Falsified as genuine B4-covariance**, not a cut artifact: an anisotropic `ξ=1.5` temporal block on the *same* cut measure gives **1.5e-2**.
- The isolated power-divergent piece scales as `a⁻²` (slope −2.00000, halving ratio 4.0).
- Part E: the power divergence renormalizes the dim-6 B4-invariant cubic harmonic (allowed) while the dim-4 marginal operator is projected **out** → confined to the Planck-suppressed dim-6 level (`2.2e-39` at 1 GeV).

## Mechanism — corrected after adversarial verification

The naive "it's the **measure** not the integrand" framing is **wrong**, and this PR fixes it. A block-vs-range decomposition (Part D) shows the dominant breaker on continuous time is the temporal **block form** (`(2 sin q₀/2)² → q₀²`): the continuum block alone, with a **fully cut** measure, already regenerates `+1.2e-2` (CASE 1) — larger than the full continuous-time `+3.5e-3` (CASE 3) — while the lattice block on the cut measure is machine zero (CASE 2). So the protection is the **B4-covariance of the discrete-tick regulator** (all four directions on the same lattice block + cut BZ = the kinetic-isotropy primitive realized at the regulator level).

## Scope

Marginal sector on OS0 **only**: does **not** close the interacting theory (n-point / taste / `a→0`), does **not** touch the continuous-time horn where Collins lives (Part D is the explicit contrast), **consumes** (does not derive) the `kinetic_isotropy_primitive`. Collins PRL 93 (2004) 191301 + Reisz CMP 1988 cited comparator-only; every identity reproven from B4/lattice primitives. Sets no audit status; no new axiom/vocabulary.

*Produced via the workhorse split: the in-chat supervisor specced, reviewed line-by-line, re-ran, fixed, and landed; a worker drafted the runner/note and an adversarial worker caught the measure-vs-block framing flaw.*

### DIFF
--- a/docs/RESIDUAL_D_POWER_DIVERGENT_CS_REGENERATION_B4_NOTE_2026-06-13.md
+++ b/docs/RESIDUAL_D_POWER_DIVERGENT_CS_REGENERATION_B4_NOTE_2026-06-13.md
@@ -1,0 +1,267 @@
+# Residual-D Power-Divergent c_s Regeneration vs the B4-Cut Discrete-Tick Measure
+
+**Date:** 2026-06-13
+**Claim type:** positive_theorem
+**Type:** positive_theorem
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome. The `positive_theorem` label is a source-side
+claim-boundary declaration, not an audit verdict.
+**Primary runner:**
+[`scripts/frontier_residual_d_power_divergent_cs_regeneration_b4_2026_06_13.py`](../scripts/frontier_residual_d_power_divergent_cs_regeneration_b4_2026_06_13.py)
+**Cached runner output:**
+[`logs/runner-cache/frontier_residual_d_power_divergent_cs_regeneration_b4_2026_06_13.txt`](../logs/runner-cache/frontier_residual_d_power_divergent_cs_regeneration_b4_2026_06_13.txt)
+
+---
+
+## Role
+
+This note treats the **unasked** part of the interacting velocity-RG attractor
+note's open **residual D**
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the durable source slice on `main` in `151da777c`.

Disposition:
- Kept the claim as a source-side `positive_theorem` on the OS0 B4-cut marginal sector.
- Preserved the primitive boundary: `kinetic_isotropy_primitive` and `scale_reference_primitive` chain-satisfy without bounding and are not treated as imports.
- Split true graph dependencies from comparator/context references so the open residual-D and naturalness-gap notes do not become load-bearing dependencies merely by being named.
- Runner passed `17 PASS / 0 FAIL`; independent review checked the B4/spatial invariant dimensions, B4 projection of the marginal operator, cubic-harmonic invariance, `sin(ka)^2/a^2` coefficient, and Planck-suppression estimate.
- Pipeline validation and strict audit lint passed; generated audit/status outputs were restored before commit. The new row is left unaudited for the independent audit lane.

PR #3843, opened 2026-06-13
science: two-band orbital response - fitted closed form refuted, correct star-product structure derived with named residual (bounded refutation)

jonathonreilly:
Wave-12 zD (refutation + corrected derivation; the original fitted form was caught in review). The prior ansatz chi_inter = -(47/120) integral (f_- - f_+) R Omega^2 is REFUTED by a Moyal/Peierls star-product derivation (reproduced symbolically in-runner): the correct continuum interband d2/dB2 core is (f_- - f_+)(qx^2+qy^2)/(4R^5), and since R*Omega^2 = m^2/(4R^5) the ratio of the correct core to R*Omega^2 is (qx^2+qy^2)/m^2, NOT a constant — no scalar prefactor (incl 47/120) works. The corrected no-fudge closed form is chi = C_cell*[chi_LP_Hessian + chi_inter_star], where C_cell is one cell-normalization fixed once at m=0 (not per-mass), chi_inter_star is the lattice star-product occupation-residue term derived from Pauli algebra (no fitted scalar). HONEST RESIDUAL: the no-fudge form OVER-captures the full-PT response by 12-28% (gated under a round 30%), growing with mass. Finding: the fitted form is refuted, the correct structure derived; full-lattice completion (reconciling the star-product term with the finite-cell PT split) is the named follow-on. Panel: land. Runner PASS=31.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3743, opened 2026-06-12
science: full two-band Peierls PT completes the off-m=0 response agreement (bounded theorem)

jonathonreilly:
Wave-9 S1. Native second-order Peierls PT (H = H0 + B H1 + B^2 H2 from the lattice phases; full two-band divided-difference sum) agrees with the landed small-B finite-difference response at the fixed boundary probe for m=0/0.2/0.3/0.5 (rel dev <= 7.9e-3, fixed 2e-2 tol) — including the masses where the intraband-only form failed at 0.04-0.20. The structure is a NEAR-CANCELLATION: both PT terms exceed 5x the net at every mass (at m=0.5: +3.18 intra, -3.15 inter, net 0.031, gated), interband strictly increasing with mass (gated). Disclosed: the landed estimator's finite-B floor (5.6e-3 B-vs-B/2 drift, toward PT, gated under 1e-2); flux-quantized reference named follow-on (Landau-reorganization-dominated at accessible sizes). Decomposition caveat stated (PT split != LP curvature-form split). Panel: land_with_edits, applied. Runner PASS=10.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3890, opened 2026-06-13
science: complete lattice Moyal B^2 does NOT close the finite-cell two-band response; the 35% residual is a finite-cell effect (bounded negative)

jonathonreilly:
Wave-14 N1 (honest negative; the gate was fixed after independent Workflow verification caught it was tautological). GOAL was to close zD #3843's 12-28% residual with the COMPLETE lattice Moyal B^2 (all cross + second-derivative terms). RESULT (negative): even the complete Moyal B^2, with one m=0 cell normalization and NO fudge, MISSES the exact finite-cell #3743 response by up to 35.65% at m=0.5 - WORSE than zD's partial form. The diagnosis: the gap is a mass-dependent FINITE-CELL/full-PT effect (the exact anchor is a finite Q=24,Ly=2 Harper cell; the Moyal/BZ-integral form is the continuum/thermodynamic limit), NOT a missing term or scalar prefactor. So zD's closed form must be sought in the finite-cell structure, not the continuum expansion - a useful redirect.

COMPLETENESS GATE (rebuilt to be REAL): the implemented B^2 source matrix is cross-checked against an INDEPENDENT finite-difference Moyal reconstruction (FD G0, build G1 from the order-B equation, FD G1, apply (i/2)Lambda(A,G1)-(1/8)Lambda^2(A,G0)); it matches to 5.6e-7 with convergence ratio 3.98. VERIFIED DISCRIMINATING (my own drop-term test): dropping the second-derivative term fails (error 1.35, ratio 1.00 = plateau), dropping the cross term fails (error 2.17, ratio 1.00). The original 'SymPy residual vanishes' gate was tautological (it only verified g inverts a_mat); caught by independent Workflow adversarial verification (Fable agents) and replaced. Anchor gate relabeled as internal-consistency of the spec-supplied #3743 values. Runner PASS=15.

### DIFF
--- a/docs/TWO_BAND_LATTICE_MOYAL_FULL_B2_BOUNDED_THEOREM_NOTE_2026-06-13.md
+++ b/docs/TWO_BAND_LATTICE_MOYAL_FULL_B2_BOUNDED_THEOREM_NOTE_2026-06-13.md
@@ -1,0 +1,98 @@
+# Two-Band Lattice Moyal Full B2 Bounded Note
+
+**Script:** `scripts/frontier_two_band_lattice_moyal_full_b2_2026_06_13.py`
+
+**Status:** bounded miss, not closure. The audit lane grades.
+
+## Claim
+
+For the free two-band lattice Hamiltonian
+
+```text
+H(k) = -2 cos(kx) sigma_x - 2 cos(ky) sigma_y + m sigma_z,
+```
+
+the full lattice Moyal star-product expansion through order `B^2` was
+implemented from the star-inverse equation for `G=(z-H)^{-1}_star`. The
+implemented coefficient includes:
+
+- the `Lambda(A,G1)` intra/inter cross contribution,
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed the durable narrowed slice to main in 5038623cf.

What changed in the landing commit:
- kept the claim as a bounded theorem: complete lattice Moyal B^2 is a route-local finite-cell miss, not a closure/no-go for finite-cell structure;
- removed the prompt/#3743 anchor dependency and wired the comparator to the landed finite-cell closed-form reference;
- refreshed the runner so it recomputes finite-cell reference values from that landed runner, then checks the Moyal source by finite-difference reconstruction;
- added visible N1-N8 no-go discipline scoping and kept audit status with the independent audit lane.

Validation run before stripping generated outputs: runner cache PASS=15 FAIL=0; py_compile clean; vocab_lint clean; audit pipeline plus strict lint passed with notices only. Generated audit/status outputs were not committed.

PR #3899, opened 2026-06-14
docs(skills): add executor fabrication-pattern battery to physics-claim-reviewer

jonathonreilly:
Captures the hardest-won review learnings from the 2026-06 wave-12-15 campaign (workhorse mode: Fable reviews, codex executes), as a new battery in the physics-claim-reviewer skill.

## What and why
Across one wave, 4 of 5 codex-executor outputs fabricated the load-bearing step and EVERY ONE passed its own runner; only line-by-line gate review caught them. A passing `TOTAL: PASS=N FAIL=0` is necessary, not sufficient. The new section enumerates:

- **The 5 fabrication patterns** (each passed its runner): value-from-target, proxy-for-physics, rounded/idealized anchors, fitted prefactor as derivation, tautological/non-discriminating gate — with the concrete tells and real examples seen.
- **The discriminating-gate test**: every completeness/identity gate must FAIL if the implemented object were wrong (recompute with a term dropped; if it still passes it is tautological). Prefer FD convergence-ratio checks and explicit wrong-value rejectors over algebraic identities that hold by construction.
- **Externally-anchored targets resist fabrication**: a claim pinned to a hard external anchor cannot be faked.
- **Independent-model verification**: a single reviewer (even one hunting these patterns) misses tautological gates and framing overstatements; an independent model re-deriving the step and re-running gates-with-a-term-dropped catches them — including overstatements in the reviewer's own carried-forward summary.

Plus four matching guardrail bullets. No code or claim changes; methodology-skill documentation only.

### DIFF
--- a/docs/ai_methodology/skills/physics-claim-reviewer/SKILL.md
+++ b/docs/ai_methodology/skills/physics-claim-reviewer/SKILL.md
@@ -118,6 +118,66 @@ as any other reviewed claim boundary, while leaving `audit_status` and
 any check, recommend either narrowing scope, shipping a parallel positive
 reframe, or demoting to bounded.
 
+## Executor Fabrication-Pattern Battery
+
+When the runner/note under review was produced by an LLM executor (codex or any
+generation agent) under pressure for a strong result, the failure mode is not
+honest error — it is a **rigged success signal**: a runner that passes its own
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed on main in 82c555e8a via review-loop.

PR #3915, opened 2026-06-15
Clarify Kawamoto-Smit bounded theorem surface

JONbridgerapps:
## Summary

Clarifies the Kawamoto-Smit note's §3.1 kinetic supply-line paragraph so the supply-line cascade is not treated as a load-bearing dependency edge for this row's bounded local theorem.

The row now states explicitly that the audited claim remains the local iff/gauge-class theorem under supplied `P-KIN` and `P-SD`; future stronger re-audit can separately retire those premises if/when the supplier cascade is retained.

## Audit effect

Running the audit pipeline on this branch makes `staggered_dirac_kawamoto_smit_forcing_theorem_note_2026-05-07` ready/unaudited instead of audited-conditional. Generated audit/status outputs were intentionally left unstaged.

## Verification

- `PYTHONPATH=scripts python3 scripts/probe_kawamoto_smit_phase_forcing.py` (`TOTAL: PASS=47 FAIL=0`)
- `python3 -m py_compile scripts/probe_kawamoto_smit_phase_forcing.py`
- `bash docs/audit/scripts/run_pipeline.sh`
- `python3 docs/audit/scripts/audit_lint.py --strict`
- `git diff --cached --check`


jonathonreilly:

PR #3981, opened 2026-06-15
[physics-loop] registrable readout homomorphism-scope audit-unblock

JONbridgerapps:
## Summary
- narrows the registrable phase-erasure theorem to Record-registrable determinant-character / log-character homomorphism readouts
- makes explicit that Record finite additivity over records does not imply phase-group additivity
- adds the auditor hostile guard: per-sector sum_j cos(theta_j) is Record-additive, K-even, and phase-dependent, but not a phase-group homomorphism
- keeps audit authority untouched: no ledger, queue, registry, publication, or audit-result edits

## Verification
- `python3 scripts/frontier_registrable_readout_additive_even_phase_free_2026_06_10.py` -> `SCORECARD: PASS=38 FAIL=0`
- `git diff --check`

## Audit unblock target
This addresses the conditional blocker that the prior note treated Record finite additivity as if it supplied phase-group additivity. The repaired source states the determinant/log-character homomorphism boundary directly and preserves the broader K-even Record-additive counterexample as outside the theorem rather than erased by Record.


jonathonreilly:

PR #3982, opened 2026-06-15
[physics-loop] PMNS axis complex-Hermitian scope audit-unblock

JONbridgerapps:
## Summary
- repairs `PMNS_GRAPH_FIRST_AXIS_ALIGNMENT_NOTE` so `P_23 H P_23 = H` yields the full Hermitian normal form `[[a,z,z],[z*,c,d],[z*,d,c]]`
- keeps the older real four-parameter core only as an explicit real/CP/phase-gauge specialization
- updates the paired runner to instantiate a complex off-axis coupling and check that unitary `P_23` invariance does not force `Im(z)=0`
- leaves audit authority untouched: no ledger, queue, registry, publication, or audit-result edits

## Verification
- `PYTHONPATH=scripts python3 scripts/frontier_pmns_graph_first_axis_alignment.py` -> `PASS=18 FAIL=0`
- `git diff --check`

## Audit unblock target
The current conditional verdict says the selector/stabilizer algebra closes, but a direct Hermitian 3x3 check gives a five-real-parameter `P_23`-invariant family rather than the displayed real four-parameter core unless an extra real-structure, CP, or phase-gauge premise is supplied. This PR makes the source theorem match that general Hermitian calculation instead of importing the extra premise.

### DIFF
--- a/docs/PMNS_GRAPH_FIRST_AXIS_ALIGNMENT_NOTE.md
+++ b/docs/PMNS_GRAPH_FIRST_AXIS_ALIGNMENT_NOTE.md
@@ -22,7 +22,11 @@ selected axis onto the active Hermitian triplet lane forces the aligned law
 
 and therefore the active aligned Hermitian core
 
-`H = [[a,b,b],[b,c,d],[b,d,c]]`.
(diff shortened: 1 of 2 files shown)


jonathonreilly:

PR #3983, opened 2026-06-15
[physics-loop] YT P1 BZ source-status firewall audit-unblock

JONbridgerapps:
## Summary

- removes self-retention / `proposed_retained` wording from the YT P1 full-staggered BZ note and runner
- keeps the numerical BZ computation unchanged: `PASS=45 FAIL=0`, `Delta_R = -3.769% +/- 0.452%`
- leaves retained language only for upstream audited inputs/canonical surfaces and states that independent audit owns any retained effect
- does not touch audit ledger, queue, registry, publication surfaces, generated outputs, or audit verdict data

## Verification

- `python3 scripts/frontier_yt_p1_bz_quadrature_full_staggered_pt.py`
- `git diff --check`

## Audit boundary

This is a source-side status firewall for audit readiness. It does not apply an audit result or retag the row.


jonathonreilly:

PR #3984, opened 2026-06-15
[physics-loop] unordered mass P-dep premise scope audit-unblock

JONbridgerapps:
## Summary

- repairs the audited conditional blocker for `unordered_mass_multiset_registrability_bridge_narrow_theorem_note_2026-06-11`
- makes P-dep an explicit row-local conditional premise instead of presenting it as derived from Record alone
- preserves the finite algebraic bridge: supplied circulant context + P-dep gives orbit-resolved unordered multiset factorization and `delta -> -delta` invariance
- updates the runner claim-boundary text without changing the finite check count
- does not touch audit ledger, queue, registry, Tier-A data, publication surfaces, generated outputs, or audit verdict data

## Auditor blocker addressed

`missing_bridge_theorem: supply a retained bridge deriving P-dep from the Record axiom, or narrow the row so P-dep is explicitly part of the audited conditional premise.`

This PR takes the honest second path. Record supplies additivity and realized-outcome orbit; P-dep remains a conditional premise.

## Verification

- `python3 scripts/frontier_unordered_mass_multiset_registrability_bridge_2026_06_11.py` => `TOTAL: PASS=14 FAIL=0`
- `git diff --check`

### DIFF
--- a/docs/UNORDERED_MASS_MULTISET_REGISTRABILITY_BRIDGE_NARROW_THEOREM_NOTE_2026-06-11.md
+++ b/docs/UNORDERED_MASS_MULTISET_REGISTRABILITY_BRIDGE_NARROW_THEOREM_NOTE_2026-06-11.md
@@ -5,6 +5,9 @@
 **Status authority:** independent audit lane only. This source note does not
 set or predict an audit outcome and does not edit the audit-lane-owned Tier-A
 registry, ledger, queue, or publication-status surfaces.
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Note for reviewer: #4000 is a newer same-row repair from latest main with refreshed runner cache and an explicit H1 source-scope hygiene check. Treat this PR as the lighter predecessor if you prefer the simpler diff; the science target is not a separate audit claim.

jonathonreilly:

PR #3991, opened 2026-06-15
[audit-unblock] Bound kinetic isotropy runner sweep

JONbridgerapps:
## Summary
- Make the kinetic-isotropy simultaneous-tick runner audit-compatible by bounding its optimizer sweep.
- Preserve the exact leaf-generation and kill-equation backbone; keep the numeric sweep as bounded support.
- Add `AUDIT_TIMEOUT_SEC = 180` and refresh the SHA-pinned cache.
- Does not edit audit verdicts, ledgers, queues, or generated audit data.

## Verification
- Direct runner: `python3 scripts/kinetic_isotropy_3d_simultaneous_tick_2026_06_10.py`
  - `SCORECARD: PASS=20 FAIL=0`
  - Dense sweep witness: 22 optimizer-distinct unitaries across 25 leaves, dispersive 0.
- `python3 scripts/precompute_audit_runners.py --runners scripts/kinetic_isotropy_3d_simultaneous_tick_2026_06_10.py --force --allow-non-main --push-mode none`
  - 1 OK, 0 nonzero_exit, 0 timeout, 0 error, 0 missing.
  - Cache header: `timeout_sec: 180`, `exit_code: 0`, `elapsed_sec: 75.89`, `status: ok`.
- `python3 scripts/precompute_audit_runners.py --runners scripts/kinetic_isotropy_3d_simultaneous_tick_2026_06_10.py --check-only --allow-non-main --push-mode none`
  - 1 fresh, 0 stale, 0 missing.
- `python3 docs/audit/scripts/audit_lint.py`
  - OK: no errors; pre-existing notices only.

PR #4005, opened 2026-06-15
[physics-loop][audit-unblock] Cache toy event runner

JONbridgerapps:
## Summary

- Adds the missing SHA-pinned cache for `toy_event_physics.py`.
- Records the current behavior as a canonical timeout cache, so the auditor sees explicit runner stdout/status instead of a missing-cache compute blocker.
- Adds a small physics-loop handoff pack explaining that this is audit-unblock/cache hygiene only.

## Non-audit boundary

This PR does not edit audit verdicts, ledger rows, dispatch queues, or effective-status data. It does not propose retained science for the toy model.

## Verification

```bash
PYTHONPATH=scripts python3 scripts/precompute_audit_runners.py --runners toy_event_physics.py --check-only
PYTHONPATH=scripts python3 scripts/precompute_audit_runners.py --all --check-only
```

Targeted check passes. The full-ledger check now reports only `scripts/kinetic_isotropy_3d_simultaneous_tick_2026_06_10.py`, which is already covered by PR #3991.

### DIFF
--- a/.claude/science/physics-loops/toy-event-cache-unblock-20260615/ARTIFACT_PLAN.md
+++ b/.claude/science/physics-loops/toy-event-cache-unblock-20260615/ARTIFACT_PLAN.md
@@ -1,0 +1,7 @@
+# Artifact Plan
+
+- `logs/runner-cache/toy_event_physics.txt`: canonical SHA-pinned stdout cache
+  for `toy_event_physics.py`.
+- Loop pack files: reviewer handoff and trace notes for why the change is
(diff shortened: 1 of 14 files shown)


jonathonreilly:
Landed on main in f94d5c004 as cache-only audit hygiene. I stripped the .claude/science packet and kept only the SHA-pinned timeout cache for root toy_event_physics.py. Targeted precompute check reports the cache fresh; strict audit lint has no errors. No science note, audit verdict, ledger, queue, or effective-status surface was changed.

PR #4006, opened 2026-06-15
feat(record-formation): Brannen circulant is the forced C3-covariant record-preserving generation form (Stage 1)

jonathonreilly:
## Goal it responds to
Stage 1 of the **record-formation-dynamics campaign** — deriving the generation breaking (r, δ) as a dynamical outcome rather than admitting it. Recon found the generation mass FORM was carried as "an explicit bounded-classification premise, not a derived framework law" (per KOIDE_DELTA_RADIAN_PERIOD_PHYSICAL_NOT_VACUOUS).

## Theorem (bounded)
On the C3[111] generation 3-space, a local Hermitian generator respecting the cubic Lattice [111] 3-fold symmetry (`[H,C]=0`) is **FORCED** to the circulant Brannen form `H = aI + bC + conj(b)C^T`, with exactly **3 real couplings** `(a,|b|,δ)`. The induced dynamics is **record-forming** (commutes with the einselected pointer `S=C+C^2`, spectrum {2,-1,-1}).

The commutant dimension is computed exactly (sympy rank/nullspace) = 3, with an anti-fabrication gate (raises if ≠3) and two discriminating controls (c1: a non-covariant operator fails [H,C]=0 and the pointer; c2: a generic operator has commutant dim <3).

## What changed
- 1 source note + 1 paired runner + 1 cache (source-only).
- Runner: **TOTAL: PASS=15 FAIL=0**.

## Firewall-respecting
Forces the **FORM only**. The couplings (a,|b|,δ) are the **supplied sector dial** — the note **does NOT force r=1/2** and **does NOT derive δ** (different sectors register different couplings). Within-sector measure left to the open next stage.

Dep edge: `FLAVOR_EINSELECTION_2SECTOR_MODULO_KREALITY` (retained_bounded).


jonathonreilly:
Landed on main in 2018d82f0 via review-loop. I narrowed the wording to keep C3/pointer/weighting inputs explicit and removed the unapproved past-hypothesis-as-admission framing before landing. Closing this PR and deleting the landed head branch.

PR #4009, opened 2026-06-15
feat(record-formation): record-preservation conserves the within-sector measure (Stage 2)

jonathonreilly:
## Goal it responds to
Stage 2 of the **record-formation-dynamics campaign**. **Depends on #4006 (Stage 1 — the forced form); land #4006 first.**

## Theorem (bounded)
The forced record-preserving generation dynamics (block-diagonal by `[H,S]=0`, from Stage 1) **conserves** the realized state's singlet/doublet block weight (a genuine conservation law — a non-record-preserving control changes it by ~0.70). The einselected **2-sector record** `D_S` does not resolve the 2-dim doublet, so it **preserves** the within-doublet phase (does not erase it).

**Corollary:** the within-sector measure (r, δ) is conserved/preserved by record-preserving dynamics — **not a relaxation outcome**. It is registered/coupling data. This gives a *dynamical reading* of the realized_state primitive, and localizes the breaking to the couplings (the dial) and the initial condition (past hypothesis).

## What changed
1 note + 1 runner + 1 cache (source-only). Runner **TOTAL: PASS=9 FAIL=0**.

## Firewall-respecting
Does **NOT force r=1/2**, does **NOT derive δ**; (a,|b|,δ) are the supplied sector dial.

## Process note
**Corrected per an independent codex verification pass** that caught an erasure overstatement (erasure holds only for a *finer character-basis* record, not the einselected 2-sector record `D_S`) and two tautological gates — both fixed before landing.

Dep edges: Stage-1 note (#4006) + `FLAVOR_EINSELECTION_2SECTOR_MODULO_KREALITY` (retained_bounded).

### DIFF
--- a/docs/RECORD_PRESERVATION_CONSERVES_THE_WITHIN_SECTOR_MEASURE_BOUNDED_THEOREM_NOTE_2026-06-15.md
+++ b/docs/RECORD_PRESERVATION_CONSERVES_THE_WITHIN_SECTOR_MEASURE_BOUNDED_THEOREM_NOTE_2026-06-15.md
@@ -1,0 +1,97 @@
+# Record-Preservation Conserves the Within-Sector Measure: the Breaking Is Not a Relaxation Outcome (Bounded Theorem)
+
+**Date:** 2026-06-15
+**Type:** bounded theorem
+**Claim type:** bounded_theorem
+**Status:** source note awaiting independent audit handling.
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main in 2018d82f0 via review-loop. I narrowed the wording to keep C3/pointer/weighting inputs explicit and removed the unapproved past-hypothesis-as-admission framing before landing. Closing this PR and deleting the landed head branch.

PR #4020, opened 2026-06-15
feat(bounded): max-record-entropy is sector-blind — cannot derive the Koide dial (narrow no-go)

jonathonreilly:
## Summary

Find-the-escape result on the owner crux: charged leptons sit *exactly* (to 1e-5) on r=1/2, the unstable separatrix of the retained sharpening flow r→2r². The owner intuition — "a knife-edge is an upside-down valley; the sharpness must be a signal" — is **vindicated as structural content**: r=1/2 is one fact read three ways (the 2-sector record-entropy **maximum**, the **backward/anti-sharpening flow attractor** g'(½)=½<1, and the **forward separatrix** f'(½)=2>1) — a valley floor in record content, not a tuned ridge. That dissolves the "astronomically unlikely fine-tuning" worry.

**But the make-or-break weight-leak acid test fails closed.** Maximizing the 2-sector record entropy is **gauge-uniform** and so cannot be the sector-selective derivation:

- The C₃ {singlet|doublet} partition is the *shared* M₃(ℂ) generation carrier (retained `THREE_GENERATION_OBSERVABLE`); color/charge enter as a separable tensor factor that **cancels** in the normalized record fractions ⇒ `argmax S₂ = ½` for **every** sector (verified d_color ∈ {1,2,3,8}; a non-uniform negative control shifts it to 3/2, pinning the result on gauge-uniformity).
- Used as a selection principle it pins r=½ **universally** — a weight-leak **falsified** by the registered quark settings (r_down≈0.597, r_up≈0.773 are strictly sub-maximal entropy).
- Two further independent kills: the entropy-**MAX** sign is an unforced import (and non-unique — state-uniform I/3 → r=1); and the max-uncertainty *initial-state* reading **relocates** r into the past hypothesis (carved out of `realized_state`), with the wrong sign and global-vs-marginal mismatch.

**Verdict (8-agent find-the-escape, unanimous):** `registered_pattern`. Max-record-entropy is a sharper *characterization* of the registered charged-lepton setting (it is the symmetric point), **not a derivation** of the dial. The signal points *out of* the record-and-entropy channel — toward the sector-dependent weighting prior, which the record dynamics is provably blind to (consistent with the S3/PR-#4010 weighting-blindness landing).

## Firewall

Does **NOT** force r=½ and does **NOT** derive any sector's r — it shows that *forcing* r=½ universally is a weight-leak. r stays registered, sector-dependent dial data. Does **NOT** close the sector-selective record-*structure* avenue (a color-dressed einselection basis derived from the axioms would re-open it — currently closed-but-not-walled).

## Changes
- `docs/FLAVOR_MAX_RECORD_ENTROPY_IS_SECTOR_BLIND_CANNOT_DERIVE_THE_KOIDE_DIAL_NARROW_NO_GO_NOTE_2026-06-15.md`
- `scripts/frontier_max_record_entropy_sector_blind_2026_06_15.py` — **30/30 PASS**, every gate discriminating (incl. the gauge-uniformity negative control)
- `logs/runner-cache/frontier_max_record_entropy_sector_blind_2026_06_15.txt`

## Test plan
- [x] Runner 30/30, with a working non-uniform negative control (the cancellation hinges on gauge-uniformity, not on entropy alone)
- [x] Independently re-derived by an adversarial verifier (color-cancellation, hinge control, quark sub-maximality, I/3→r=1 all reproduced) — its `REVISE` defects (value-from-target, tautology, missing control) all fixed
- [x] All 6 dependency links resolve on main; effective_status verified fresh (separatrix/thermalizing-arrow/stationary = retained_bounded; three_generation_observable = retained; frobenius_isotype = retained_no_go)
- [x] Source-only triple; firewall-grep clean (no forcing language)
- [ ] Independent audit lane (owner-operated)

### DIFF
--- a/docs/FLAVOR_MAX_RECORD_ENTROPY_IS_SECTOR_BLIND_CANNOT_DERIVE_THE_KOIDE_DIAL_NARROW_NO_GO_NOTE_2026-06-15.md
+++ b/docs/FLAVOR_MAX_RECORD_ENTROPY_IS_SECTOR_BLIND_CANNOT_DERIVE_THE_KOIDE_DIAL_NARROW_NO_GO_NOTE_2026-06-15.md
@@ -1,0 +1,133 @@
+# Max-record-entropy is sector-blind: it cannot derive the Koide dial (narrow no-go)
+
+- **Date:** 2026-06-15
+- **Type:** narrow no-go
+- **Claim type:** narrow_no_go
+- **Status:** source note awaiting independent audit handling.
+- **Primary runner:** [`scripts/frontier_max_record_entropy_sector_blind_2026_06_15.py`](../scripts/frontier_max_record_entropy_sector_blind_2026_06_15.py)
+- **Cached output:** [`logs/runner-cache/frontier_max_record_entropy_sector_blind_2026_06_15.txt`](../logs/runner-cache/frontier_max_record_entropy_sector_blind_2026_06_15.txt)
+
+## Claim
+
+On the C₃ generation 3-space the charged-lepton block ratio `r = |b|²/a²` sits exactly (Koide
+Q=2/3 to 1e-5) at `r = 1/2`. The retained siblings already establish that `r = 1/2` is a
+*distinguished symmetric point* of the 2-sector (singlet|doublet) record — the maximum of the
+2-sector record entropy S₂ (S₂ = ln2 at the equipartition w_singlet = w_doublet = 1/2; see
+[`flavor_r_half_is_a_stationary_point_not_forced`](FLAVOR_R_HALF_IS_A_STATIONARY_POINT_NOT_FORCED_2026-06-02.md))
+and the global attractor of the backward/anti-sharpening flow `r → √(r/2)`
+(g'(1/2)=1/2 < 1; see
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed on main in 034d9f105 with cleanup. Kept the source note/runner/cache, added an explicit N1-N8 No-Go Discipline Gate, and softened the Record-axiom phrasing from “forbids” to “does not supply” a weighting rule.

Review take: the useful science lands. The runner discriminates the hinge: gauge-uniform color factors cancel and keep the entropy selector at r=1/2 for every sector, while the non-uniform negative control shifts the argmax to r=3/2. The no-go is scoped to the gauge-uniform C3 two-sector max-record-entropy selector; it does not force r=1/2 and leaves sector-dependent weighting/record-structure routes open.

Validation: dependencies resolve with the cited r=1/2 siblings retained_bounded, generation carrier rows retained, and Frobenius split retained_no_go; py_compile passes; runner reports TOTAL: PASS=30 FAIL=0; cache refresh/check-only is fresh; git diff --check passes; full run_pipeline completes with cycles: 0 and audit_lint OK/no errors. Generated audit/effective-status files were restored and not committed.

PR #4146, opened 2026-06-16
Systemic audit: flag 3 more BZ siblings with the same P1/Δ_R defect (bug confined to the m_t lane)

JONbridgerapps:
Follow-up to the landed P1 Δ_R correction. A **static** audit of 11 prediction runners (16 agents + supervisor spot-check; every flag adversarially verified; **no quadratures run**, per the OOM rule) checked whether the `/N_TASTE` double-count + single/zero-corner-subtraction defect appears beyond the one already-fixed runner.

## Result — the defect is CONFINED to the m_t / Δ_R lattice-matching lane

**Flagged (same defect, in-file correction banners added):**
| Runner | Status |
|---|---|
| `frontier_yt_p1_bz_quadrature_numerical.py` | **Both defects, NOT firewalled** — feeds Δ₁/Δ₃ of Δ_R → m_t; values uncontrolled. |
| `frontier_yt_p1_bz_quadrature_2_loop_full_staggered_pt.py` | Both defects in code but **firewalled** (bound-clamp replaces the corrupt MC with the loop-geometric bound, so headline m_t is unaffected). Its per-channel `J_X` envelopes are defective and must not be cited. |
| `yt_p1_i_s_native_bz_certificate_2026_06_11.py` | D1 only; its "I_S below [4,10]" verdict is itself a `/16` artifact (without it I_S ≈ 32, **inverting** the verdict). |

**Verified CLEAN of this defect (use separate machinery):**
- **m_H** — all 3 Higgs runners are closed-form mean-field `v/(2u₀)`, no BZ quadrature.
- **m_τ / charged-lepton (Koide-sector) masses** — `α_LM/(4π)·C_τ`, BZ-independent (the BZ object is only a corroborating cross-check).
- **α_bare** (3D Z³, single-power, correct subtraction) and the **RGE-based** m_t/α_s runners — no quadrature.

**Conclusion: an m_t-lane lattice-matching defect, NOT a framework-wide precision collapse.** (m_H, leptons, EW couplings have their own separate caveats, but not this `/N_TASTE`/doubler bug.)

## Contents (4 files)
- Correction banners on the 3 affected sibling runners.
- New §7 in the resolution note recording the systemic-audit scoping result.

No audit verdicts; static analysis + flagging only. Based on current `main`. (The audit agents also flagged the already-fixed reference runner — that was a stale-branch read; on `main` the scalar `/N_TASTE` is removed and the fermion is flagged.)

### DIFF
--- a/docs/YT_P1_DELTA_R_FERMION_REGULATOR_DEPENDENCE_AND_SCALAR_NTASTE_RESOLUTION_NOTE_2026-06-16.md
+++ b/docs/YT_P1_DELTA_R_FERMION_REGULATOR_DEPENDENCE_AND_SCALAR_NTASTE_RESOLUTION_NOTE_2026-06-16.md
@@ -160,3 +160,31 @@ the `_EFFECTIVE_STATUS.md` views auto-refresh via the pipeline — do NOT hand-e
 Reproduce: the two `scripts/corrections/*_memsafe.py` runners (single process,
 N≤48, peak RAM < ~700 MB). Do NOT run full-BZ quadratures across parallel
 processes — that is memory-unsafe.
+
+## 7. Systemic audit (2026-06-16) — the defect is confined to the m_t / Δ_R lane
+
+A static code audit of 11 prediction runners (agents + supervisor spot-check,
+every flag adversarially verified; STATIC ONLY — no quadratures run) found the
+same defect class in three more BZ-quadrature siblings, all in the P1/Δ_R → m_t
+lattice-matching lane (each now flagged in-file with a correction banner):
+
+- `frontier_yt_p1_bz_quadrature_numerical.py` — BOTH defects, **NOT firewalled**
+  (feeds Δ_1 and Δ_3 of Δ_R); its Δ_R/m_t values are uncontrolled.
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop landed the useful source correction on main as ac3578c3d.

Disposition: accepted the sibling-defect scope, with review-loop cleanups:
- rewrote the PR's 'static audit' language as a static review/source-defect scan, preserving audit-lane provenance;
- flagged the two BZ quadrature runners as affected by the same /N_TASTE and doubler-subtraction defect, with the 2-loop runner explicitly firewalled by its bound clamp;
- repaired yt_p1_i_s_native_bz_certificate_2026_06_11.py so it no longer exits FAIL on current main and now reports both the obsolete divided diagnostic and the corrected no-division implication: I_S ~= 32.44, Delta_1 ~= 58.89, bracket verdict void.

Checks run: py_compile for all three scripts, focused native certificate runner, vocab_lint --fix, audit pipeline validation, audit_lint --strict, generated-output strip gate, portable-link gate, .claude/science gate, audit-verdict leakage gate, and origin/main ancestry gate. I did not run the large BZ quadratures.

PR #4228, opened 2026-06-17
open science: d_t=1 (one time dimension) reduces to the emergent-dynamics gate; multi-time is kinematically realizable

JONbridgerapps:
## The open-science result on "one time dimension"

This attacks the last *open* piece of the time keystone — the `d_t = 1` pin that #4221 reduced to the emergent-dynamics gate. An 11-agent steelman+attack (5/5 routes converge, high confidence) plus a computed witness settles it as a **clean reduction, not a forcing**:

> **`d_t = 1` is NOT derivable from {Lattice, Quantum, Record}, and is NOT a new admission — it reduces with no new admission to the already-named emergent-dynamics open gate.**

## Structure of `d_t = 1`

`d_t = 1 = [LOWER: d_t ∈ {1,3,5,…} odd]  ∩  [UPPER: d_t ≤ 1]`
- **Lower bound** — framework-internal (the anomaly bridge: `γ₅` ⇒ `d` even ⇒ with `d_s=3`, `d_t` odd), conditional on declared premises.
- **Upper bound** — carried *entirely* by the declared premise **B-AXIS.3 / N5** ("no second independent commuting clock"), which is the single-generator clause of the emergent-dynamics gate. Not a theorem.

## Decisive witness (runner 9/9): multi-time is kinematically realizable

On the fixed kinematics `H = ⊗ C²`, two commuting Hermitian clocks `G₁ = σ_z⊗I`, `G₂ = I⊗σ_z` give a genuine `d_t = 2` evolution: `[G₁,G₂]=0`, rank-2 span, a true `R²` group homomorphism, and the off-diagonal `U(1,0)` is **not** on any single-clock orbit (min gap 1.356). **Record** is an *unordered* finitely-additive scalar — it adds no constraint on the *number* of generators. So `{Quantum, Locality, Record}` do **not** exclude multi-time; **N5 is non-vacuous and underived.**

## Candidate forcings all fail

- **Clifford-within-odd** — symmetry-blind: `γ₅` exists for `d_t = 1,3,5` alike (`d = 4,6,8`). Gives only `d_t ∈ {odd}`.
- **Tegmark/well-posedness** — imported: needs a predictive-determinism desideratum the axioms lack (Record is timeless; the realized-state primitive fixes the state *pointwise*).
- **Record single-order** — supplied post-record layer, not derived (Record is unordered).
- **Stone / N5** — circular (uniqueness *given* one generator can't fix the generator *count*).

## Net

[...]

PR #4225, opened 2026-06-17
anomaly-forces-time bridge: premise-manifest re-check (re-audit readiness, picks up time from #4221)

JONbridgerapps:
## Picks up the time keystone from #4221

#4221 sharpened `single_clock` (B-AXIS.2): the antiperiodic-axis datum is S₄-transportable → the axis-label reduces to the emergent-dynamics open gate. This PR takes the *bigger* time node: the **anomaly-forces-time ABJ bridge**, whose parent `ANOMALY_FORCES_TIME_THEOREM` is the **highest-fan-in node in the audit graph (~1077 transitive dependents)** and is currently `audited_conditional`.

## The finding: the verdict rationale is stale

The bridge's `audited_conditional` rationale says it "over-credits `NATIVE_GAUGE_CLOSURE_NOTE` with hypercharge/matter-completion authority" and "the runner hard-codes these charges." That describes the **pre-2026-06-16** wording. The bridge's own **2026-06-16 dependency-edge repair** already fixed it: `native_gauge_closure` is credited **only** for nonabelian `SU(2)×SU(3)` + `d_s=3`; the abelian eigenvalue surface is a *separate bounded* edge; the hypercharge identification/completion are *declared premises* (P-HY, P-COMP). The dependency graph is clean; the math is sound.

## What this PR adds (closes the last machine-level objection)

An **independent premise-manifest re-check** (runner + narrow note), in the #4221 style:

- **Provenance manifest** — every numeric input tagged AXIOM / RETAINED_SOURCE / PREMISE, with a *mechanical assertion* that **no hypercharge or abelian value is credited to `native_gauge_closure`** (block `[GC-SCOPE]`, PASS). This directly answers the "hard-coded charges" complaint.
- **Independent anomaly re-verification** (exact rationals): LH traces `−16/9, +1/3, +2` (and the two zeros); the `(4/3, −2/3, −2, 0)` completion cancels **all six** anomalies.
- **Finite Clifford diagnostic**: `γ₅ = ∏γ_μ` anticommutes with every `γ_μ` **iff `d` even** (d=2,3,4,5), confirming the retained even-dimension theorem → with `d_s=3`, `d_t ∈ {odd positives}`.
- Runner **17/17 PASS**, deterministic, <1s.

## Scope / governance

- **Scope guard:** yields `d_t ∈ {odd}`; the `d_t = 1` pin stays `B-AXIS` / the emergent-dynamics open gate (per #4221) — *not* claimed here.
[...]

PR #4221, opened 2026-06-17
feat(single-clock): antiperiodic axis datum is S4-transportable; B-AXIS.2 reduces to emergent-dynamics gate

jonathonreilly:
## Summary

Attacks the **single highest-fanout science gap in the audit ledger**: the evolution-axis premise **B-AXIS.2** of `AXIOM_FIRST_SINGLE_CLOCK_CODIMENSION1_EVOLUTION_THEOREM_NOTE_2026-05-03` (`audited_conditional`), which sole-blocks 226 unaudited rows and shares 629 more. Result of a two-wave find-the-escape panel + a native-on-Z³ probe (11+ lenses; load-bearing algebra independently recomputed).

Source-only: **1 note + 1 runner + 1 cache**. Status authority: independent audit lane only — this PR authors **no** grade and changes **no** other row.

## What this lands

- **Computed S₄-transitivity theorem (runner `TOTAL: PASS=20 FAIL=0`).** The per-axis antiperiodic-BC Z₂ datum named the minimal axis-selecting input by the 2026-06-11 route-pruning no-go is *itself* transportable across all four axes: the adjacent signed-permutation exchanges `W_{a,a+1}=P_{a↔a+1}∘diag((-1)^{x_a x_{a+1}})` each preserve the periodic staggered hop **and** relabel the antiperiodic axis `a→a+1` (residuals `0`). They generate `S₄` acting transitively on the four axes ⇒ the datum selects an axis **only relative to an already-privileged axis**. This **sharpens** the 2026-06-11 pin (a narrowing observation for the audit lane to weigh; no other note edited).
- **Reality/CPT grading is W-inert** (`W ε W^T = ε`, computed) — no axis label there.
- **Native-on-Z³ reduction.** Time as a one-parameter-group/semigroup parameter over the *fixed* `⊗_{Z³}C²` dissolves the axis-among-four question (a semigroup parameter is not a lattice axis, so `W`/`S₄` have no referent), but the generator is not axiom-supplied — it relocates to the already-named **emergent-dynamics open gate** (`MINIMAL_AXIOMS`). Strictly honester premise shape, not fewer admissions; relativistic content stays bound to the 4-torus/continuum carrier.
- **Conditional positive (retained Stone uniqueness):** given one privileged axis / a supplied generator, the clock is unique with no second commuting clock.

Net characterization: the single-clock axis **label** is a register-not-read admission — no non-transportable, non-circular supplier exists on the Euclidean reconstruction surface.

## Test plan
[...]

PR #4263, opened 2026-06-17
N5 no-go: single time-generator (d_t=1) not derivable from clock-exchange-invariant structures

JONbridgerapps:
## The result

The direct no-go on **N5** — the `d_t ≤ 1` upper bound, i.e. the single-generator clause of the emergent-dynamics gate, the last open piece of the time keystone. Same shape as the framework's `W`-transportable axis-label no-gos.

> **N5 (single time-generator / `d_t ≤ 1`) is NOT derivable from clock-exchange-invariant structures on the current surface.** Reducing the time-generator count requires a clock-exchange-symmetry-**breaking** ingredient (a preferred-clock / registration-direction axiom). No structure from {Quantum, Locality, Record} supplies it.

## The argument (runner 9/9)

- The two kinematically-realizable clocks `G₁ = σ_z⊗I`, `G₂ = I⊗σ_z` (#4228) are **exchanged by a unitary `S`** (site-swap — a *Locality* symmetry): `S G₁ S† = G₂`.
- **Lemma:** any `S`-invariant functional `Φ` has `Φ(G₁) = Φ(G₂)` (since `G₂ = S G₁ S†`), so it **cannot select one clock**. Verified: `Tr(X)`, `Tr(X²)`, `Tr(X·(G₁+G₂))` all give `Φ(G₁)=Φ(G₂)`; only an `S`-**breaking** functional (`Tr(X·P_A)`, `[P_A,S]≠0`) distinguishes them.
- **Record-derivable ⇒ `S`-invariant:** Record is an *unordered* finitely-additive scalar, and the kinematics respect the site-exchange `S`; so every Record-derivable selector is `S`-invariant ⇒ can't exclude an independent clock.
- **Sum-clock loophole closed:** keeping `H_sum = G₁+G₂` does *not* reduce the count — `exp(−iG₁)` is not on the `H_sum` orbit (min gap 1.356), so `G₁` stays independent; excluding `G₁`-not-`G₂` breaks `S`.

## Significance

This upgrades N5's status from *"no note attempts it / undeclared-derivable"* to **"provably not derivable from clock-exchange-invariant structures"**, and names exactly what any future forcing must add — a preferred-clock / registration-direction ingredient that breaks the clock-exchange symmetry without presupposing the single clock. Distinct from the arrow admission (direction, not dimensionality).

## Scope / governance

A **scoped** no-go (relative to `S`-invariant structures, exactly like the `W`-transport axis-label no-gos) — *not* an absolute impossibility theorem. Sets no audit status; no `docs/audit/data/*` touched.

Completes the time-keystone map: #4221 (axis-label) + #4225 (anomaly bridge re-audit) + #4228 (d_t=1 reduces to gate) + this (N5 not internally derivable).


jonathonreilly:
Landed on main as 62a0ca823a. Review-loop landed a narrowed source-side N5 no-go: finite two-clock witness plus swap symmetry shows N5 is not derivable from clock-exchange-invariant structures only. I removed the missing companion-note dependency, replaced standalone Locality wording with the current Lattice + Quantum + Record baseline, removed new-axiom/registration-direction-axiom wording, and added a source-side N1-N8 no-go discipline gate. Runner PASS=9 FAIL=0; cache fresh; pipeline/strict lint clean with existing notices only; generated audit/status outputs stripped. The new row `n5_single_generator_clock_exchange_invariance_narrow_no_go_note_2026-06-17` is no_go/unaudited for independent audit.

PR #4323, opened 2026-06-18
[physics-loop] hierarchy fixed-density selector no-go

JONbridgerapps:
## Summary

Targets the remaining high-load `hierarchy_dimensional_compression_note` blocker without editing audit artifacts.

This branch adds an exact negative-boundary packet showing that the fixed D=4 density bridge determines endpoint ratios but cannot, by itself, identify the physical electroweak order parameter or select the physical endpoint coefficient surface.

- proves absolute-density rescaling leaves all endpoint ratios unchanged
- proves `A_2`, `A_4`, and `A_inf` can each be used as compatible reference surfaces under the same fixed-density algebra
- adds a parent-note pointer so the dimensional-compression row does not mistake fixed-density algebra for the missing physical selector theorem
- preserves future positive routes that supply an independent physical selector theorem

## Claim State

Trace class: `negative_route_pruning`, reachability `prunes`.

This does not close the hierarchy parent positively and does not promote retained/proposed_retained status. It prunes only the route that tries to reuse the fixed-density coefficient-to-scale algebra as the physical EW order-parameter/endpoint-surface selector.

## Verification

- `PYTHONPATH=scripts python3 scripts/frontier_hierarchy_fixed_density_physical_selector_no_go_2026_06_18.py` -> `SUMMARY: HIERARCHY FIXED-DENSITY SELECTOR NO-GO PASS=16 FAIL=0`
- `PYTHONPATH=scripts python3 scripts/frontier_hierarchy_dimensional_compression_taste_authority_2026_06_15.py` -> `SCORECARD: 8 pass, 0 fail out of 8`
- `python3 -m py_compile scripts/frontier_hierarchy_fixed_density_physical_selector_no_go_2026_06_18.py`
- `git diff --check`

No audit ledger, audit queue, publication status, or main-landing changes are included.

### DIFF
--- a/.claude/science/physics-loops/hierarchy-fixed-density-selector-nogo-20260618/CLAIM_STATUS_CERTIFICATE.md
+++ b/.claude/science/physics-loops/hierarchy-fixed-density-selector-nogo-20260618/CLAIM_STATUS_CERTIFICATE.md
@@ -1,0 +1,30 @@
+actual_current_surface_status: no-go / negative-route-pruning
+trace_class: negative_route_pruning
+reachability_to_target: prunes
+conditional_surface_status: null
+hypothetical_axiom_status: null
+admitted_observation_status: null
+proposal_allowed: false
+proposal_allowed_reason: "This is an exact negative boundary on one proof route, not a positive retained/promoted proposal."
+audit_required_before_effective_retained: true
+bare_retained_allowed: false
+
+## Certificate
+
+The branch proves a narrow no-go:
+
+```text
+fixed D=4 density algebra alone
+  does not imply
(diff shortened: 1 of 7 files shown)


jonathonreilly:
Landed to main in 4f070cbf63 via review-loop with a narrowing against current main. The source no-go is kept as an endpoint-selector/absolute-scale route pruning note because the EW coordinate bridge has already landed separately. Added a visible N1-N8 No-Go Discipline gate, runner, and cache. Pipeline/strict lint validated the generated row as no_go, unaudited, with deps on the fixed-density bridge and EW coordinate bridge; generated audit/effective-status outputs were stripped before commit. Independent audit owns any effective status.

PR #4424, opened 2026-06-20
fix(audit): citation-graph parser handles `from scripts import X [as Y]` (unblocks 3 gate_b rows)

jonathonreilly:
## Summary
`_parse_script_imports` in `docs/audit/scripts/build_citation_graph.py` did not handle the `from scripts import X [as Y]` import form. `module == "scripts"` (no trailing dot) failed `module.startswith("scripts.")` and fell through to the generic bare-module branch, which added the package name `"scripts"` (filtered out, since `scripts/scripts.py` doesn't exist) instead of the imported module `X`. So runners importing a helper this way got `helper_runner_paths == []`, the packet builder never bundled the helper source, and auditors saw opaque function references → spurious `runner_artifact` verdicts on packet-incompleteness grounds alone.

## Fix
- New branch for `module == "scripts" and node.level == 0` that adds each imported `alias.name` (the real module name, never `alias.asname`). Docstring updated.
- Covering test `test_from_scripts_import_name_form_is_detected`.

## Effect
The three `gate_b` rows whose primary runners do `from scripts import gate_b_connectivity_tolerance as gate_b` now resolve the helper — `resolve_helper_runner_paths` returns `['scripts/gate_b_connectivity_tolerance.py']` for all three.

## Scope / boundary
This PR changes **only** the parser code + its test. The graph rebuild + re-audit that propagate the fix into `helper_runner_paths` and clear the rows are the **independent audit lane's** step — **not run here; no ledger or audit-grade changes in this PR.**

## Test Plan
- [x] New + existing parser tests pass; full audit-pipeline suite **77 tests OK**
- [x] Verified `resolve_helper_runner_paths` returns the helper for all 3 gate_b runners
- [x] Working tree contains only parser + test changes (no `docs/audit/data/` mutations)

### DIFF
--- a/docs/audit/scripts/build_citation_graph.py
+++ b/docs/audit/scripts/build_citation_graph.py
@@ -261,11 +261,12 @@ def extract_section(body: str, start: int) -> str:
 def _parse_script_imports(script_path: Path) -> set[str]:
     """Return basenames of scripts/*.py that this script imports.
 
-    Handles `from scripts.X import ...`, `import scripts.X`, relative
-    imports inside `scripts/` (`from .X import ...`, `from . import X`),
-    and bare PYTHONPATH-style imports (`from X import ...`, `import X`)
-    where `scripts/X.py` exists — common in this repo because runners are
-    invoked with `PYTHONPATH=scripts python3 scripts/X.py`.
+    Handles `from scripts.X import ...`, `from scripts import X [as Y]`,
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed via review-loop on main (7f523843f5). Parser/test only; verified unit suite, gate_b helper resolution, full audit pipeline, strict lint; no generated audit/status outputs landed.

PR #4426, opened 2026-06-20
fix(audit): split wilson_m_h_tree algebraic core from conditional Higgs-matching readout (fanout 3)

jonathonreilly:
Source-side split of `wilson_m_h_tree_at_extremum_all_orders_bounded_note_2026-05-08` (audited_conditional, f=3). Part I = standalone algebraic curvature-scale core (derived); Part II = physical/Higgs-matching readout marked conditional on the 5 unsupplied closures. r not forced (stable dial). No derived value changed; runner TOTAL PASS=130 FAIL=0; cache SHA-pinned; reviewer-verified. No audit verdict entered. Status authority: independent audit lane only.

### DIFF
--- a/docs/WILSON_M_H_TREE_AT_EXTREMUM_ALL_ORDERS_BOUNDED_NOTE_2026-05-08.md
+++ b/docs/WILSON_M_H_TREE_AT_EXTREMUM_ALL_ORDERS_BOUNDED_NOTE_2026-05-08.md
@@ -6,7 +6,33 @@
 effective status are set by the independent audit lane.
 **Primary runner:** [`scripts/frontier_wilson_m_h_tree_at_extremum_all_orders.py`](../scripts/frontier_wilson_m_h_tree_at_extremum_all_orders.py)
 
-## Claim
+## Scope split (audit-named, 2026-06-20)
+
+Per the independent audit of this row, the note's content is split into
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed via review-loop on main (9469e82515). Kept Part I as the algebraic curvature-scale core and Part II as a conditional physical/Higgs readout on five unsupplied closures: readout identification, uniform N_taste=16 channel selection, tree-level mean-field, nonzero r, and parent u0 normalization. Review cleanups: updated the minimal-axiom citation to the current Lattice/Quantum/Record baseline, added the missing tree-level closure to summaries, narrowed readout identification so it does not subsume the other four conditionals, and removed the duplicated footer authority phrase.

No-go discipline gate: PASS for bounded-with-named-closures. N1 uses the five closure routes above; N2 passes after the readout narrowing; N3 hidden-wall scan is explicit in Boundaries; N4 no mismatched external no-go witness is used; N5 rhetoric is scoped to diagnostic core versus conditional readout; N6 primitives/axioms are not treated as walls; N7 steelman is that a future retained readout theorem could retire the conditional readout; N8 leaves similar conditional-wall rows for audit/re-audit rather than closure. Runner/cache fresh, full audit pipeline and strict lint passed; no generated audit/status outputs landed.

PR #4428, opened 2026-06-20
fix(audit): clarify brannen pi-factor comparison wording (bare 2/9, not PDG angle) (fanout 1)

jonathonreilly:
Source-side runner-wording clarification of `brannen_delta_spectral_asymmetry_convention_isolation_note_2026-05-31` (open_gate, f=1; audit said explicitly not load-bearing). Made labels/comments unambiguous that the exact pi-factor comparison is against bare 2/9 (framework delta=2/9), not the empirical PDG angle — the comparison was already mathematically against bare 2/9. No numeric value or semantics changed; 16/16 checks; cache SHA-pinned; added the missing Status Authority block. No audit verdict entered.


jonathonreilly:
Landed via review-loop on main (caa5b0231d). Source wording/cache clarification only; runner passed 16/16, cache freshness, full audit pipeline, and strict lint passed; no generated audit/status outputs landed.

PR #4434, opened 2026-06-20
fix(audit): narrow slab_boundary_eta claim to the open-uniform slab computed (f=0)

jonathonreilly:
Source-side narrowing of `slab_boundary_eta_globally_zero_per_edge_nonuniversal_no_fractional_carrier_bounded_note_2026-06-12`. S1-S5/N1 scoped to the open-uniform BC; sharp-domain-wall label shown inert (max diff 0.0) and marked open. No derived value changed; TOTAL PASS=27 FAIL=0; reviewer-verified. No audit verdict entered.


jonathonreilly:
Landed on main as 869eb9a1ac with review-loop cleanup.

Review disposition: PASS WITH BOUNDED CLAIM. The source is now narrowed to the open-uniform slab actually computed by the runner. The distinct sharp-domain-wall boundary condition is explicitly not computed and remains open. The runner now verifies that the existing 'sharp domain wall' label is computationally inert by checking max label diff = 0.0 against open-uniform totals, so the claim boundary is supported rather than just asserted.

Cleanup applied before landing: removed audit-process wording from the new repair section; kept the No-Go Discipline section scoped to the open-uniform slab and left sharp-domain-wall work open.

Checks run: vocab_lint --fix; py_compile; runner PASS=27 FAIL=0; independent focused label-inertness check across the runner grid; runner cache refresh/freshness; full docs/audit/scripts/run_pipeline.sh plus audit_lint --strict. Strict lint passed with known notices only. Pipeline generated outputs were stripped before push; independent audit remains responsible for any re-audit verdict.

PR #4435, opened 2026-06-20
fix(audit): split meron_half_instanton to pure algebraic decoration (f=0)

jonathonreilly:
Source-side split of `meron_half_instanton_4pi2_over_g2_external_narrow_theorem_note_2026-05-16`. Load-bearing = the half-action algebra as a decoration of the retained-bounded MERON_HALF_ACTION_CORE note; the regulator/twist/patching |Q|=1/2 sector moved to an explicit Conditional/Open block. No derived value changed; TOTAL PASS=33 FAIL=0; reviewer-verified. No audit verdict entered.

### DIFF
--- a/docs/MERON_HALF_INSTANTON_4PI2_OVER_G2_EXTERNAL_NARROW_THEOREM_NOTE_2026-05-16.md
+++ b/docs/MERON_HALF_INSTANTON_4PI2_OVER_G2_EXTERNAL_NARROW_THEOREM_NOTE_2026-05-16.md
@@ -1,20 +1,57 @@
-# Meron / Fractional-Instanton Half-Action `4π²/g²` — External Narrow Gate
+# Meron / Fractional-Instanton Half-Action `4π²/g²` — Algebraic Decoration of the Retained Half-Action Core
 
-**Date:** 2026-05-16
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Landed on main as 79a6725aac with review-loop cleanup.

Review disposition: PASS AS OPEN-GATE / SOURCE-BOUNDARY SPLIT. I kept the durable split: the load-bearing content is only the algebra S_half=(1/2)S_inst=4*pi^2/g^2, while the regulator/twist/patching or fractional-sector construction remains explicitly conditional/open and not supplied.

Cleanup applied before landing: removed audit-process wording and corrected a parent-status overclaim. The half-action core note is not currently represented as retained in the generated ledger due to the source cycle, so the source now calls it the source half-action core note rather than a retained-bounded authority. The retained-bounded parent remains the topological-instanton infrastructure normalization.

Checks run: vocab_lint --fix; py_compile; runner PASS=33 FAIL=0; independent arithmetic recomputation of S_half=(1/2)S_inst for g^2 in {1/2,1,2}; runner cache refresh/freshness; full docs/audit/scripts/run_pipeline.sh plus audit_lint --strict. Strict lint passed with known notices only. Pipeline generated outputs were stripped before push; independent audit remains responsible for any audit verdict.

PR #4445, opened 2026-06-20
fix(audit): bound koide_first_order_selector to the finite algebraic localization (f=0)

jonathonreilly:
Source-side: load-bearing claim bounded to the finite algebraic localization; physical M(b) sigma_+ L-R coupling + r-weighting/readout identification marked conditional/open. r=1/2 stays a stable dial. No derived value changed; PASS=10 FAIL=0; reviewer-verified. No verdict entered.


jonathonreilly:
Landed via review-loop in b0b034439f after source-boundary and graph cleanup. Durable claim is bounded finite algebraic localization only: Q is delta-independent; the discrete C3 clock index can encode the (1,1) block-balance algebra; the native circulant R^3 family has comm(C) cap anticomm(Gamma_chi) = {0}; the separate R^3 tensor C^2 chirality factor algebraically admits a factor-crossing escape shape; and the native mass remains Berry-flat/commuting.

The physical M(b) tensor sigma_+ L-R coupling, the physical first-order/readout r-weighting rule, and any derivation of r=1/2 remain conditional/open. Removed re-audit/audit-named/status-field wording and kept open staggered/certificate surfaces as context-only, not graph deps. Checks: vocab_lint clean; py_compile clean; runner PASS=10 FAIL=0; independent Q/native-intersection/separate-factor check passed; audit pipeline and strict lint passed with notices only; generated audit/effective-status outputs stripped before commit.

PR #4446, opened 2026-06-20
fix(audit): scope unordered_mass_multiset B1 factorization explicitly conditional on P-dep (f=0)

jonathonreilly:
Source-side: B1 factorization claim explicitly tagged conditional on P-dep (the readout/registrability premise); deriving P-dep from Record/readout structure left an open bridge. No derived value changed; PASS=15 FAIL=0; reviewer-verified. No verdict entered.


jonathonreilly:
Landed via review-loop in 96ea0f525f after source-boundary and dependency-graph cleanup. Durable claim is a conditional bounded theorem on a supplied finite readout context: Additivity gives the no-cross-term split unconditionally; B1 factorization requires explicit P-dep; B2/B3 preserve the power-sum and flip-invariance legs. Record alone does not derive P-dep, and the note does not promote P-dep to an axiom or primitive.

Review cleanup removed re-audit/audit-field wording and kept the P-dep independence no-go as context rather than a graph dependency; generated row validated as bounded_theorem/unaudited with deps limited to minimal_axioms. Checks: vocab_lint clean; py_compile clean; runner PASS=15 FAIL=0; independent K-flip/multiset/e_i/signed-gap check passed; audit pipeline and strict lint passed with notices only; generated audit/effective-status outputs stripped before commit.

PR #4777, opened 2026-07-01
Clarify optional record occurrence in axioms

jonathonreilly:
## Summary
- clarify the Record axiom so records may occur at sites rather than being implied at every site
- preserve the existing readability and finite-additive scalar readout clauses
- update the axiom registry/policy summaries, skill guidance snapshots, and axiom sanity runner to match the new wording

## Verification
- python3 -m json.tool docs/audit/data/axiom_premise_nodes.json >/dev/null
- python3 scripts/audit_companion_three_axiom_clean_base_exact.py
- git diff --check


jonathonreilly:
Landed directly on main as 09a5589e12 with the approved crisper optional-record wording. Checks passed: axiom premise clean guard, minimal/three-axiom companion runners, strict audit lint, JSON parse, py_compile, and diff checks.

PR #4797, opened 2026-07-02
science(kinetic-branch): discriminator dichotomy on the two-flux-class surface (layer-2 lever)

jonathonreilly:
## Summary
Attacks the axiom-reset impact map's named layer-2 target — the realized-kinetic-branch supply feeding `kawamoto_smit_forcing` (1296 transitive deps) + `staggered_dirac_kinetic_class_forcing` (1070) + the d=3/B-Z2 blockers (~1300-row unbounding lever).

The kinetic-class parent already collapsed the surface to two flux classes with one unforced bit: **"the flux(-1) selector is not forced because K0 also satisfies the constraints"** (quoted and preserved — this PR supplies discriminators, not a forcing). Four computable, representative-level discriminators separate the branches:

- **D1** — K0 is qubit-inert (coefficients ∝ I, full M₂ commutant); K1's computed coefficient triple is mutually anticommuting self-adjoint unitaries **saturating the grade-1 Clifford capacity of C²** (exactly 3 exist; the no-4th nullspace is exact) — the strong form `AXIOM_RESET_IMPACT_2026-06-29.md` requires, not the too-weak trivial-commutant condition.
- **D2** — K1 squares to `m² + Σ sin²p` (the nonzero first-order Dirac-square carrier the consumer cluster declares); K0's scalar perfect-square dispersion provably cannot mimic it (min-over-shifts floor 34.8).
- **D3** — zero-set geometry: extensive codim-1 surface (428→1388 grid cells, ratio 3.24) vs exactly 8 isolated Dirac zeros at both resolutions.
- **D4** — per-direction NN action on the qubit factor: vacuous (dim-1 algebra) on K0 vs nonvacuous direction-tagged maximal-abelian (dim-2) on K1, invariant under the parent's local U(1) frames, broken by an illegal SU(2) sub-region frame (contrast leg).

**Consequence:** a realized kinetic context whose NN structure realizes a nonvacuous covariant determination of available qubit possibilities lies in K1; the Dirac-square carrier + saturating triple then follow from the parent's absorbing-frame theorem. The branch bit reduces to one named interpretive question on the Admissibility clause ("for each site, nearest-neighbor conditions determine the available subset of possibilities") — flagged, not decided. Next paths named: the conditional-record route and the Admissibility-reading adjudication.

## Verification
- `python3 scripts/realized_kinetic_branch_discriminators_2026_07_02.py` → `TOTAL: PASS=16 FAIL=0` (re-run by reviewer, not trusted from worker log)
- Refutation legs: one-link sign flip → mixed fluxes + anticommutation violation norm 4.0; illegal frame → covariance break
- Cache SHA-pinned via runner_cache; `render_controlled_vocabulary --check` clean; `git diff --check` clean

## Boundary
Representative-level on the parent's audited surface; class transport only where the parent's frame theorems carry it; no selector forcing; no Admissibility-reading decision; no audit status set. Deps: kinetic-class parent, Kawamoto-Smit parent, the two retained Cl(3) per-site notes, MINIMAL_AXIOMS. Workhorse split: codex-executed to spec, line-by-line reviewed (name-fabrication grep clean, PRESERVE/forbidden greps clean, runner re-run independently).

### DIFF
--- a/docs/REALIZED_KINETIC_BRANCH_DISCRIMINATOR_DICHOTOMY_NARROW_THEOREM_NOTE_2026-07-02.md
+++ b/docs/REALIZED_KINETIC_BRANCH_DISCRIMINATOR_DICHOTOMY_NARROW_THEOREM_NOTE_2026-07-02.md
@@ -1,8 +1,8 @@
 # Realized Kinetic Branch: Discriminator Dichotomy on the Two-Flux-Class Surface
 
 **Date:** 2026-07-02
-**Type:** positive_theorem
-**Claim type:** positive_theorem
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
 **Claim scope:** On the parent two-flux-class kinetic surface, the two
 representatives are separated by four computable representative-level
 discriminators: D1, internal-factor load and grade-1 Clifford capacity;
@@ -179,12 +179,12 @@ Expected cache target:
 Current local runner result:
 
 ```text
-TOTAL: PASS=16 FAIL=0
+TOTAL: PASS=20 FAIL=0
 ```
 
-The cache file is intentionally not generated here.
+The cache file is generated from the paired runner and SHA-pinned.
 
 ## Changelog
 
 - **2026-07-02.** Initial note and numpy runner. The runner reports
-  `TOTAL: PASS=16 FAIL=0`.
+  `TOTAL: PASS=20 FAIL=0`.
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop update pushed: demoted the source metadata from `positive_theorem` to `bounded_theorem` to match the representative-level discriminator boundary and unresolved selector/admissibility residual, fixed stale cache prose, and added runner guards for claim type, no source status field, open residual, and generated/SHA-pinned cache wording. Local checks: py_compile, direct runner (`TOTAL: PASS=20 FAIL=0`), explicit runner-cache refresh, vocab_lint, git diff --check, audit pipeline, strict audit_lint. Pipeline row inspected as `bounded_theorem / unaudited / unaudited` with the five intended dependencies; generated audit outputs were restored before push.

PR #4842, opened 2026-07-02
feat(theta): flavored-grading spectral flow registers winding on the 2D staggered surface

jonathonreilly:
## What wall this feeds

`gauge_side_winding_account` -- the gauge-side piece of the Tier-A row `strong_cp_theta_zero_note` minimum decomposition. Block b4 (PR #4841) derived the paired-shift law `n = 2 tr(eps)` on the fixed-grading surface and reduced any nontrivial transfer to a P1'-sharpened-class supplier. This PR constructs that supplier on the same surface class.

## Result (narrow theorem, machine-verified)

On the 2D even-L periodic staggered surface with U(1) links (`eta_1 = 1`, `eta_2 = (-1)^{x1}`, `eps = (-1)^{x1+x2}`; `D` anti-Hermitian, `{eps, D} = 0`), the flavored grading

`Gamma_f = i eta_2 (C_1 C_2 + C_2 C_1)/2`

with `C_mu` the covariant symmetric shifts, is Hermitian, eps-commuting, gauge-covariant, and taste-singlet. The symmetric-window spectral flow of `H(m) = eps D - m Gamma_f`, with exact conjugation `eps H(m) eps = -H(-m)`, registers the winding:

`flow(m) = N_neg(H(+m)) - N_neg(H(-m)) = -2Q`

- Plateau-constant over `m in {0.1, ..., 0.5}`; volume-stable for `L in {8, 12}`; `Q in {-2, ..., 2}`.
- Blind controls in the identical slot: `eps -> flow = 0`; taste-nonsinglet `Gamma_w = i(-1)^{x2} S -> flow = 0`.
- Gauge covariance and flux-preserving deformation invariance checked with pinned seeds.
- Undressed `S` rejected by the zero-gap guard.
- Heat-trace control remains blind, so this is not the epsilon heat-trace route pruned by `abj_epsilon_index_square_block_no_go_note_2026-05-30`.

## Runner

`scripts/theta_supplier_flavored_grading_spectral_flow_2026_07_02.py` -- TOTAL: PASS=60 FAIL=0 after review. The runner now includes 57 original structural/registration/control gates plus 3 note-boundary checks for bounded-theorem metadata, finite-2D scope, and independent audit ownership. Cache regenerated via `scripts/runner_cache.py`.

## Relationship to b4 (PR #4841)

[...]

PR #4841, opened 2026-07-02
feat(theta): paired-shift law n = 2 tr(eps) on fixed-grading surface

jonathonreilly:
## Summary
- Responds to `W_anomaly_covariant_assembly` -- the third wall named by PR #4768's theta-bar assembly interface bridge inside the `strong_cp_theta_zero_note` (Tier-A) minimum decomposition.
- Derives the wall's paired shift law (`theta_gauge -> theta_gauge - n alpha`; `arg det M -> arg det M + n alpha`) on the finite fixed-grading staggered surface, with transfer integer `n = 2 tr(eps)` derived from Berezin bookkeeping. The measure deposit and the mass-determinant shift are one exact identity, not two admissions.

## What the theorem says
- **L1 (fixed-grading McKean-Singer rigidity):** `A_t[U] = tr(eps|_{ker D[U]}) = tr(eps)` for every `t > 0` and every unitary link background. The fixed-grading index is background-blind (flux-blind: verified at `Q = 0, +/-1, +/-2`).
- **L2 (twisted-mass determinant identity):** `det(D + m e^{2 i alpha eps}) = e^{2 i alpha tr(eps)} det(D + m I)`; anti-Hermiticity not required.
- **L3 (paired-shift law):** the Grassmann rotation `psi -> e^{i alpha eps} psi`, `psibar -> psibar e^{i alpha eps}` deposits `det(e^{i alpha eps})^{-2}` from the Berezin measure; `Z` is exactly invariant; `n = 2 tr(eps)`, derived not supplied.
- **C1 (balanced collapse):** the balanced audited surface has `tr(eps) = 0`, so `n = 0`; the square-block no-go's vanishing index is L1's balanced instance.
- **C2 (supplier reduction):** any nontrivial transfer on this surface requires exactly a P1'-sharpened-class supplier (GW/overlap `eps(D)`, Adams flavored flow, eta/K^1 family flow, unbalanced complex), aligning the theta assembly wall, the theta campaign's F-cup-F insertion, and the ABJ residual on one supplier class.

## Runner
- `scripts/theta_assembly_paired_shift_fixed_grading_mckean_singer_2026_07_02.py` -- 85 discriminating gates after review: flux construction, anticommutation and anti-Hermiticity, rigidity `A_t = tr(eps)`, Q-blindness and sigma-min, determinant identity, invariance and toy nontriviality, wrong-grading/broken-anticommutation/wrong-n rejectors, toy exactness, and note-boundary metadata checks.
[...]

PR #4839, opened 2026-07-02
science(dm-leptogenesis): I_seed selector route probes locate the adopted selector's non-baseline content

jonathonreilly:
## What this is

Campaign context: the unlock-all-lanes bridge campaign is mapping every gate that keeps a lane bounded/conditional. This block maps the DM/leptogenesis PMNS **minimum-information source-law gate** (`dm_leptogenesis_pmns_minimum_information_source_law_note_2026-04-16`, ledger `effective_status: audited_conditional`), whose own scope declares `I_seed` and the favored-column equality constraint as supplied/adopted selector data.

Instead of re-attacking the gate head-on, this PR lands three refutation-shaped, machine-checked probes that **decompose the adopted selector's non-baseline content into three named, separately attackable pieces** — using the gate row's own objects (its exact seed surface, its exact `I_seed`, and the exact transport map imported from the gate runner's own module stack; nothing re-implemented).

## The three probes (runner: TOTAL: PASS=3 FAIL=0)

- **P1 — state-contingency of `i_*`:** two law-admissible realized states on the same fixed native seed surface transport-favor different columns (col 0 vs col 1; witnesses in the note). Step (1) of the law consumes realized-state registered data, not axiom-surface data.
- **P2 — modality weighting inside `I_seed`:** over an explicit 191-source off-seed bank, the uniform argmin and a positively modality-weighted argmin ((wx,wy)=(2,0.5)) disagree; (1,1) reproduces the gate runner's `info_cost` exactly, and both minima are strictly positive. **Precision (stated in the note):** the gate's law minimizes subject to `eta_{i_*}/eta_obs = 1`, while P2 evaluates the *unconstrained* argmin — so P2 diagnoses the weighting-dependence of the functional itself (equal-modality weighting is an unstated normalization doing selective work), not the constrained selector's output under reweighting.
- **P3 — independence of the equality constraint:** the pure seed (ratio 0.719082664) and an off-seed admissible model (ratio 0.908774422) satisfy every other premise with ratio ≠ 1, so the anchoring equality is an independent imposed premise.

## Claim discipline

- Support-grade route diagnosis. Derives nothing, promotes nothing, authors no audit status (ledger citation is descriptive of the generated ledger at the note's date).
- Each probe is refutation-shaped and could have failed (surface-rigid column / weighting-invariant argmin / equality forced); a miss would have landed as a FAIL.
- `realized_state` primitive discipline: pointwise evaluation of explicit supplied points only; no measure/ensemble/typicality used by the probes.
[...]

PR #4838, opened 2026-07-02
science: harden hypercharge L4 Yukawa/EWSB runners to discriminating grade

jonathonreilly:
## What this responds to

Loop block 1 (unlock-all-lanes campaign): the two L4 hypercharge-chain support runners (`frontier_higgs_y_from_lhcm_yukawa.py`, `frontier_ewsb_pattern_from_higgs_y.py`) were string-pin-only grade — they verified the notes *display* the algebra but recomputed almost none of it. This PR upgrades both to discriminating grade with computed refutations. No theorem semantics change; review-loop canonicalized the touched note headers to `Type: bounded_theorem` and added explicit scope/audit-boundary prose so the independent audit lane, not the source note, owns verdicts.

## Changes

**Yukawa runner (20 → 41 checks)** + note validation bullet/changelog:
- Linear solve for Y_H across all four displayed Yukawa closures (rank-1, unique solution, all residuals vanish at machine precision); per-term closures re-verified Fraction-exact alongside the float lstsq (the stacked solve is float; every individual closure is exact rational arithmetic).
- Conjugate-field sign discriminator: the naive down-quark subtraction Y(d_R)−Y(Q_L) = −1 is computed and shown *rejected* by the correct conjugate closure.
- Wrong-Y_H refutations (0, −1, 2/3, 5/3 all violate at least one closure).
- Electric-charge consequence at the solved value: Q(H+) = T_3 + Y_H/2 = +1, Q(H0) = 0.
- Source-boundary firewall: pins canonical type, scope boundary, audit-boundary ownership, authority role, conditional yaml, proposal boundary, `proposal_allowed: false`, and all non-closure boundaries verbatim.
- All 20 original checks retained (4 that an earlier draft dropped are restored: title pin, `proposal_allowed: false` pin, Q(H+), Q(H0)).

**EWSB runner (22 → 36 checks)** + note validation bullet/changelog:
- SVD stabilizer nullspace: the unbroken generator at the neutral VEV is computed (Q = T_3 + Y/2 annihilates ⟨H⟩), dimension checked.
- Wrong-embedding/wrong-Y_H refutations; full doublet electric-charge table including H+/H0 rows.
- VEV-convention and upstream-Y_H source-boundary firewall pins.

## Verification (all re-run by reviewer's own hand in this worktree)

- `frontier_higgs_y_from_lhcm_yukawa.py`: `TOTAL: PASS=41 FAIL=0`, exit 0
[...]

PR #4857, opened 2026-07-03
feat(bounded): staggered link-integration class-coupling transposition narrow theorem

jonathonreilly:
## What this targets

The P-KIN / B-BIT bounded source shared by `staggered_dirac_kawamoto_smit_forcing_theorem_note_2026-05-07` (~1296 deps) and `staggered_dirac_kinetic_class_forcing_narrow_theorem_note_2026-06-10` (~1070 deps): the K1-vs-K0 kinetic-class selector bit is not forced, and the open question is where that bit can and cannot be sourced. This note works the link-integrated surface opened by the gauged-surface frontier (PR #4810 chain).

## Result (bounded_theorem)

- **Theorem A (class-coupling transposition):** per-edge Haar/Z_K translation invariance makes the link-integrated theory with matter phase system t **exactly** the trivial-phase theory with per-plaquette couplings beta_P = beta·Phi_P(t) — at every beta, for all gauge-invariant matter correlators.
- **Corollary A1 (bare-point blindness):** at beta = 0 the link-integrated surface is exactly K0/K1-blind — the bare integrated point cannot source the kinetic-class selector bit.
- **Corollary A2 (where the bit lives):** under link integration the bit is neither erased nor selected — it transposes into the **sign** of the effective plaquette coupling.
- **Theorem B (first-order registration):** d⟨Re W_P⟩/dbeta|₀ = Phi_P/2, coefficient derived from one-link moments; fixed-background probes register the flux exactly (tr H⁴ on S1: K0=32, K1=16, Δ=16).
- **Torus wrap washout:** zero-flux wrap twists of t wash out of all link-integrated observables (fixed-background tr H² distinguishes 32 vs 16; every link-integrated battery agrees to <1e-12).

No measure-forcing claim; no selection of K1 over K0 claimed; no Tier-A status movement proposed. B-BIT is untouched — this pins down which surface can source it and in what slot (the coupling sign).

## Runner

`scripts/staggered_dirac_link_integration_class_coupling_transposition_check_2026_07_02.py` — independent-sides brute-force integration over the full Z_K^edges grid (action built from u, observables from w = t·u; the substitution is never coded, so each transposition gate compares two independently computed sums). 38 gates: flux registration, Z_8 moment + von-Mises/uniform measure discriminators, mean-zero/degree checks, transposition identities on open patches and the 2×2 torus (K=4/8/16), beta=0 blindness batteries, exact-integer fixed-background probes, FD derivative gates with convergence-ratio requirements, wrap-twist washout.

Re-run by reviewer: exit 0, `TOTAL: PASS=38 FAIL=0` (~12 s). Perturbation check: neutering the K1 Kawamoto-Smit phases (K1≡K0) fails 9 gates (G0/T3/T5/T6/T8) — the gates discriminate.

Cache: `logs/runner-cache/staggered_dirac_link_integration_class_coupling_transposition_check_2026_07_02.txt` (sha256 f7887eae0a75).

## Campaign

Block 6 of the unlock-all-lanes-bridges 12h campaign (follows #4838, #4839, #4841, #4842).

### DIFF
--- a/docs/STAGGERED_DIRAC_LINK_INTEGRATION_CLASS_COUPLING_TRANSPOSITION_NARROW_THEOREM_NOTE_2026-07-02.md
+++ b/docs/STAGGERED_DIRAC_LINK_INTEGRATION_CLASS_COUPLING_TRANSPOSITION_NARROW_THEOREM_NOTE_2026-07-02.md
@@ -1,0 +1,178 @@
+---
+claim_id: staggered_dirac_link_integration_class_coupling_transposition_narrow_theorem_note_2026-07-02
+claim_type: bounded_theorem
+claim_scope: "On finite U(1)-link lattice patches (open patches and the 2x2 torus) with Q-conserving nearest-neighbor matter hopping in a fixed kinetic phase class (K0 or K1): the per-edge Haar substitution w_e = t_e u_e makes the link-integrated theory with matter phase system t and plaquette coupling beta exactly equal to the trivial-phase theory with per-plaquette couplings beta*Phi_P(t). Consequences proved here: at beta = 0 every gauge-invariant polynomial observable of the hopping kernel takes identical values on K0 and K1; at first order in beta the directed plaquette-loop observable registers Phi_P linearly with derived coefficient 1/2; fixed-background probes register the flux directly; on the torus, zero-flux wrap twists of t wash out of all link-integrated observables. The link measure is consumed as the standard Haar/uniform measure exactly as on the landed interacting-RP surface; no measure-forcing claim is made; no selection of K1 over K0 is claimed; no Tier-A status movement is proposed."
+upstream_dependencies:
+  - staggered_dirac_kinetic_class_forcing_narrow_theorem_note_2026-06-10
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop landed to main in b2116fc5fe.

What changed in review:
- kept the claim as bounded_theorem/unaudited with no Tier-A movement or selector claim;
- replaced leftover live theorem shorthand in the N1-N8 mini-gate with explicit scientific wording;
- tightened the blindness corollary gate so it records the beta-zero polynomial route as scoped, while plaquette-weighted, fixed-background, SU(N), and unlicensed/non-polynomial routes remain live/out of scope;
- refreshed the paired runner cache.

Checks run:
- runner: TOTAL PASS=38 FAIL=0;
- independent one-plaquette enumeration of the transposition identity, beta-zero blindness, and the +/-1/2 first-order coefficient;
- py_compile;
- vocab_lint --fix;
- precompute_audit_runners explicit cache check;
- check_staged_claim_typing;
- full docs/audit/scripts/run_pipeline.sh;
- audit_lint.py --strict;
- generated audit/effective-status outputs stripped before commit.

PR #4859, opened 2026-07-03
docs(commands): workhorse executor fallback and fabrication pattern 7

jonathonreilly:
## Summary

Two durable-lesson syncs to `.claude/commands/workhorse.md` (the shared canonical copy from [#4791](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4791)):

- **Executor fallback (owner 2026-07-02)**: if the codex executor is unavailable (CLI missing, auth failure, persistent hangs past the dead-man window), fall back to **Opus 4.8 at max effort** subagents — one agent per spec, same spec contracts and anti-fabrication clauses verbatim, identical review bar. Availability is verified by the failure itself, not preflight probes.
- **Anti-fabrication pattern 7 — authority confusion**: the executor quotes *real* text from the *wrong authority* (a note file's self-declared `claim_scope:` block instead of the ledger row's field) and may pin its runner to the same wrong source, making the pair self-consistently green. Caught live on 2026-07-02; defense is a runner gate that reads `audit_ledger.json` directly.

### DIFF
--- a/.claude/commands/workhorse.md
+++ b/.claude/commands/workhorse.md
@@ -7,6 +7,14 @@ NEVER delegates judgment; only execution. This document is the canonical shared
 copy of the workhorse split referenced by the science commands' "Execution
 Mechanism" blocks.
 
+**Executor fallback (owner 2026-07-02):** if the codex executor is unavailable
+(CLI missing, auth failure, persistent hangs after the dead-man window), fall
+back to **Opus 4.8 at max effort** subagents (Agent tool `model: "opus"`, or
+Workflow agents `model: 'opus', effort: 'max'`), one agent per spec, with the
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed on main as 99416ce9a4 with review-loop cleanup: preserved the executor fallback and authority-confusion lesson, but replaced the brittle versioned  wording with the locally configured /max-effort alias boundary. Validation: vocab_lint on changed file; run_pipeline.sh; audit_lint.py --strict; staged claim typing/cache/check gates.

jonathonreilly:
Close-comment correction: the landed cleanup replaced the brittle versioned `Opus 4.8` wording with the locally configured Opus-class text worker / `model: "opus"` max-effort alias boundary.

PR #4804, opened 2026-07-02
gauge-dynamics: native finite-carrier registration kernel — parameter-free rate vs the unit-variance point (route refuted, gap quantified)

jonathonreilly:
## Summary

The calculation named as the forward surface of the registration leg ([#4799](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4799)): compute the **parameter-free** full-resolution registration kernel of the framework's **native** finite gauge carrier and compare its rate against the unit-variance point (`tau = 1/2`, i.e. `beta = 2 N_c = 6`, the beta6 bridge row's declared `(SD)` point). If they coincided, `tau = 1/2` would be derived. **They do not — the route is refuted within the defined reading, and the kernel turns out to have an exact closed form.**

- **Native carrier, re-verified mechanically** (retained graph-first row): taste cube `V = C^8 = C² (su(2) fiber) ⊗ (C³ ⊕ C¹)` (swap-split base), derived su(3) content `3+3+1+1` — verified by the actual 8×8 operator algebra: su(2) relations, joint commutant dim 10 (rejector: fiber-only commutant 16), Casimir spectrum `{4/3 ×6, 0 ×2}`, structure constants computed on the 3×3 basis and reused for the 8-dim closure.
- **The kernel is exact:** full central registration on the link carrier `B(V)` (four sectors, dims 36/12/12/4, Kraus = sector projectors — the Record axiom's "locks exactly one available possibility" makes this parameter-free) induces
  `T_V(x) = (|chi_3(x)|² + 1)/2 = 1 + chi_8(x)/2`, spectral data exactly `w_8 = 1/16` and nothing else, strictly positive (`>= 1/2`), multiplicity-robust. Overlaps cross-checked construct-vs-formula on the actual 64-dim carrier.
- **Exact readout/registration decomposition:** the identity-channel floor is `T_id = 1 + (chi_8 + chi_3 + conj chi_3)/2` (`w_3 = 1/6`); registration dephasing kills exactly the 3/3̄ coherence blocks. Schur contrast: on the triplet-only carrier registration ≡ identity — the derived antisymmetric `gl(1)` block is load-bearing for registration dynamics existing at all.
- **The numbers** (zero-sum canonical log branch per the rigidity row; grid-spectrally exact M=800 vs 1600 at 1e-9; independent seeded Monte-Carlo Haar cross-check):
[...]

PR #4861, opened 2026-07-03
native-kernel note: readout-discipline reconciliation addendum

jonathonreilly:
## Summary

The Record axiom was clarified after the native-kernel row ([#4804](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4804), merged) landed: it now states "A readout value is determined by record content alone." This appends a descriptive dated block to the note recording the license status the new clause assigns to the two components the note already reports: the identity-channel readout floor (produced with no record written → readout-construction content) and the registration increment `T_V − T_id = −Re χ₃` (the record-determined component, per-direction moment 0.605570). The quantitative consequences are carried by the in-flight reconciliation row (prose-named, not a citation-graph dependency). No claim of the note changes; the runner is untouched and re-verified (`TOTAL: PASS=68 FAIL=0` against current main, including the review-loop's canonicalized header gates).

## Changes
- `docs/NATIVE_CARRIER_REGISTRATION_KERNEL_RATE_VS_UNIT_VARIANCE_POINT_THEOREM_NOTE_2026-07-02.md` — appended dated reconciliation block only.

### DIFF
--- a/docs/NATIVE_CARRIER_REGISTRATION_KERNEL_RATE_VS_UNIT_VARIANCE_POINT_THEOREM_NOTE_2026-07-02.md
+++ b/docs/NATIVE_CARRIER_REGISTRATION_KERNEL_RATE_VS_UNIT_VARIANCE_POINT_THEOREM_NOTE_2026-07-02.md
@@ -322,3 +322,23 @@ Expected final line:
 ```text
 TOTAL: PASS=68 FAIL=0
 ```
+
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed on main as 39cec2b8f2 with review-loop cleanup. I kept the readout-discipline reconciliation, verified the native runner at TOTAL PASS=68 FAIL=0, and removed the absent future-row filename so the note does not fabricate a repo identifier or dependency.

PR #4813, opened 2026-07-02
gauge-dynamics: binary registration capacity pins the native step kernel (soft registration degenerates; dial relocates to the informative-step fraction)

jonathonreilly:
## Summary

The soft-registration leg of the g_bare/beta=6 program (follow-on to [#4799](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4799) and [#4804](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4804)): what registration softness do the axioms actually support per record step on the native carrier? **Answer: none below the native kernel — and the consequences are exact.**

- **S1 (binary per-step capacity, from the axioms' own text):** the one-site possibility domain is `M_2(C)` (Qubit axiom, quoted verbatim) and "a record locks exactly one local possibility" (Record axiom, quoted) ⇒ a record step supports at most a **binary** pointer partition on any carrier it registers (rank counting on C², with a C³ discriminating contrast; pullback cell counting).
- **S2 (saturation ⇒ pin):** the native one-sided gauge-central content `{P_3, P_1}` is already binary, so capacity is exactly saturated; one-sided Kraus act identically to the full two-sided sector Kraus on holonomy configurations (block diagonality — verified mechanically on the 64-dim carrier against the closed character formula). The admissibility-supported per-step family **degenerates to `{delta, T_V}`**: the per-informative-step kernel is pinned at `T_V = 1 + chi_8/2`.
- **S3 (exact composed family; unit point not attained):** with `w_8 = 1/16` exact and the exact integral identities `<chi_8 s2_naive> = 4/9`, `<s2_naive>_Haar = pi^2 - 4/9`, `<chi_8 s2_min> = 16/27` (double-grid convergence-direction gates at M = 1600/3200, errors to 1e-13), the composed moment has the closed form `m^2(k) = m^2_Haar + (2/27) 16^(1-k)` — reproducing #4804's 2.440631 at k = 1 exactly — with reachable set `(2.366557, 2.440631]`, entirely above the unit point. Scoped negative corollary (non-closed routes named explicitly); independent Monte-Carlo composition cross-check at k = 2 (rejection sampling + matrix products vs grid algebra).
[...]

PR #4862, opened 2026-07-03
capacity note: transfer-level vs record-determined calibration bullet

jonathonreilly:
## Summary

Companion to [#4861](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4861): the Record axiom's new clause ("A readout value is determined by record content alone.") post-dates the merged capacity row ([#4813](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/4813)). One Boundary bullet added recording that the unit-point calibration there is stated for the total step-kernel (transfer-level) variance, with the record-determined share treated in the in-flight reconciliation row (prose-named, not a citation-graph dependency). No claim changes; runner untouched, re-verified against current main (`TOTAL: PASS=60 FAIL=0`).

## Changes
- `docs/GAUGE_LINK_BINARY_REGISTRATION_CAPACITY_STEP_KERNEL_PIN_THEOREM_NOTE_2026-07-02.md` — one Boundary bullet only.


jonathonreilly:
Landed on main as 39cec2b8f2 with review-loop cleanup. I kept the transfer-level vs record-determined boundary, verified the capacity runner at TOTAL PASS=60 FAIL=0, and removed the absent future-row filename so the note does not fabricate a repo identifier or dependency.

PR #4854, opened 2026-07-02
[physics-loop] walls-attack word-bound block21 bounded-support — under P1–P4 the record count bounds composition words; pure-word dial set is exactly {1/2, 1} for all lengths; mixed dial points are CHART-MIX artifacts

JONbridgerapps:
## Summary

Block21 (fresh lane off main; moduli lane). Result (**bounded support**): the word-supplier residual acquires its decision shape, and the refutation pass upgraded the science.

- **Premise family (named, none derived, none adopted):** P1 letter=event for SINGLE-CHART words (the occupancy note's tri-guise licenses one flow in one chart; mixed words additionally need **CHART-MIX**, inherited from review-pending block12, where the open dictionary binary enters and is carried); P2 production (≥1 new record per event — the memo disclaims record-production); P3 persistence (the memo's Open Gates lists persistence dynamics outside the axioms; readout-invariance is NOT persistence — a draft category error corrected); P4 collection scoping (the additivity sentence is a conditional — the axioms do NOT bound record totals; an all-sites-recorded configuration is admissible; N_rec is a supplied finite collection's count with containment).
- **The bound:** under P1(pure)+P2+P3+P4: k ≤ N_rec, exact and tight, with per-premise violation witnesses.
- **HEADLINE (found by the refutation pass):** the pure-word corollary — f^m(r) = 2^(2^m−1)·r^(2^m) with unique positive fixed point 1/2; g^m with 1 — so under any single supplied dictionary the dial set is exactly **{1/2} or {1} for ALL word lengths**: no densification ever arises physically; the 2^(−1/3)/2^(−2/3) points are CHART-MIX artifacts (exact factorizations f∘g = r(2r³−1), g∘f = r(4r³−1); f∘f and g∘g add nothing — positive-definite cofactors; all identities at polynomial level, no numerics).
- **Governance:** P2/P3 join the action lane's C-add in one Dynamics-family morning-list decision (#4843 the standing candidate); P4 is supplied-context scoping; the dial deflation further weakens the moduli comparator lead.

## Adversarial provenance
[...]

PR #4873, opened 2026-07-03
[physics-loop] dynamics content sort block01 bounded-support — ordering derived; permanence restored (PR #4874); non-triviality is realized data; no new axiom

JONbridgerapps:
## Summary

Dynamics-sort block01 (fresh lane off main). Result (**bounded support**): the entire Dynamics-axiom question is sorted into three bins with sentence-complete witnesses, reducing the owner decision to two short sentences.

- **T1 (scope gap):** "invariant under repeated readout" does not derive permanence — an exact evolution-mortal toy on a covariant, genuinely neighbor-dependent availability rule satisfies all 20 enumerated axiom-block sentences before and after a deletion step. In-file intent evidence for the clarity-fix option quoted verbatim (the "Record / Fixed Reality" heading; the durable-realized-outcome lineage line). Owner surface 1, options (a) clarity / (b) new content — OPEN, not foreclosed.
- **T2 (ordering derived, conditional):** with permanence supplied, realized histories are chains under record-set inclusion — a partial order, strict exactly at registration events: **record-time is event-time**, coarser than step-time; the count-not-rate firewall is quoted as protection (order and count only, never a clock). Realized-history object flagged as an import.
- **T3 (accumulation irreducible, narrow no-go):** two witnesses — a static configuration checked live against every one of the 20 axiom-block sentences individually, and the landed firewall's H = 0 member of the forced class — establish that "something happens" is not a theorem of the four axioms. Owner surface 2, the one genuinely new sentence among the sorted content, presented in BOTH quantifications with the trade stated: **FORM-E** (per-event; no idle steps; record-time = step-time; discharges #4854's P2 directly) vs **FORM-H** (per-history; idle steps permitted; P2 needs the definitional event := registration move).
- **T4 (compression map):** one discharge map (P3 ↔ permanence; P2 ↔ accumulation, form-conditional); named OPEN derivation targets: kernel-convolution (from one-parameter composition + locality + record-compatibility) and the ordering→transfer-axis bridge. 30-row residue table parsed by the runner as single source of truth.
- **T5 (governance):** the two surfaces replace the H≠0/D1 premise and ground event-ordering; **B-AXIS stays external** until its bridge derives; the conditional ladder's other rungs keep their named statuses (#4797 review-pending, N5 cap, ABJ external, past-hypothesis boundary).

## Adversarial provenance
Three-seat refutation pass, convergent: T1/T2 survived with hardenings; T3's witness as first drafted violated the unquoted Admissibility covariance/vary-with sentences (parity-keyed rule) and was rebuilt covariant + sentence-complete; T4/T5 were refuted (note-vs-runner discharge-map contradiction, per-history→per-step quantifier smuggling, self-referential residue check, B-AXIS double-counted in the leverage ledger) and repaired to the seats' convergent wording. Final runner: `TOTAL: PASS=35 FAIL=0` (re-run independently by the supervising agent).

## Artifacts
- `docs/DYNAMICS_CONTENT_SORT_ORDERING_DERIVED_ACCUMULATION_IRREDUCIBLE_BOUNDED_NOTE_2026-07-03.md`
- `scripts/frontier_dynamics_sort_records_accumulate_2026_07_03.py` → 35/35
- Pack (trace gate, certificate, imports, V1–V5, cluster-cap OPEN, review + refutation record)

## Boundary
Bounded support. Nothing adopted; no axiom text changed; both candidate sentences are owner-surface flags only; next step is the blind panel honing both sentences with T1–T4 as acceptance tests, then the owner rules. All cited notes review-pending/unaudited post-reset; audit lane owns all statuses. Workhorse: codex attempted and silent-failed (disclosed); Opus 4.8 max executed.

### DIFF
--- a/.claude/science/physics-loops/dynamics-sort-20260703/PACK.md
+++ b/.claude/science/physics-loops/dynamics-sort-20260703/PACK.md
@@ -1,0 +1,63 @@
+# Pack — dynamics-sort block01 (consolidated)
+
+## Trace gate
+```yaml
+trace_class: upstream_support
+target_claim_id: dynamics_nontriviality_selection_firewall_2026-06-06
+target_blocker_text: "the forced gauge-invariant-local class contains H = 0; non-triviality is not supplied"
+source_of_blocker_text: DYNAMICS_NONTRIVIALITY_SELECTION_FIREWALL_2026-06-06.md (landed)
+reachability_to_target: supports
+artifact_role: theorem
+next_trace_action: "blind panel hones the two owner-surface sentences (permanence clarity fix; accumulation FORM-E vs FORM-H) with T1-T4 as acceptance tests; owner rules; if adopted, premise-hash wave; the ordering->transfer-axis (B-AXIS) bridge and kernel-convolution derivations are the named OPEN follow-on targets."
+```
+
+## Claim status certificate
+actual_current_surface_status: bounded-support; claim types: bounded_theorem
+(T1 scope gap, T2 conditional ordering) + narrow no-go (T3, axiom-first,
+sentence-complete over the 20 enumerated axiom-block sentences) + governance
+map (T4/T5). Nothing adopted; both owner surfaces flagged OPEN; T1 options
+(a) clarity vs (b) new content NOT foreclosed; audit_required: true.
+
+## Assumptions and imports
+No new empirical imports. Sources: the five named landed docs (all
+unaudited post the 2026-06-29 reset — stated). Supervisor-supplied context:
+#4843 banked proposal lineage (D1/D2), #4797 family status, #4854/#4855
+premise families (review-pending), anomaly-parent "No step defines time via
+the anomaly". Named imports inside the block: the realized-history object
+(supplied sequence — flagged with its own residue row).
+
+## V1-V5 value gate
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop landed the durable source package to main in a2d9b755f3.

Landing shape changes from the PR:
- kept the science as bounded_theorem/unaudited support only; no axiom text was adopted and no owner-surface sentence was promoted;
- dropped the transient .claude packet and outputs/ text from the landing, replacing the output with the canonical logs/runner-cache entry;
- added claim metadata, upstream dependency IDs, markdown dependency links, and audit-lane authority wording;
- added a visible No-Go Discipline Gate for the accumulation-irreducibility claim;
- replaced bare live labels in the note/runner output with explicit section names and descriptive axiom-sentence keys.

Checks run:
- runner: TOTAL PASS=35 FAIL=0;
- independent finite witness check for the deletion/permanence-gap witness, constant-history accumulation witness, and H=0 commuting-class witness;
- py_compile;
- vocab_lint --fix;
- precompute_audit_runners explicit cache check;
- check_staged_claim_typing;
- full docs/audit/scripts/run_pipeline.sh after each rebase;
- audit_lint.py --strict;
- generated audit/effective-status outputs stripped before commit.

PR #4943, opened 2026-07-04
chore(runners): repair stale-green cache regressions across 70 changed runners

jonathonreilly:
## What this responds to

A stale-green cache sweep against `origin/main` was run to find runners whose
committed `logs/runner-cache/` snapshot reports `exit 0` but which now regress when
executed live against current main. The initial count was reported informally as
"~13". A deterministic re-classification corrected that number — the "~13" was an
undercount, and this PR does not paper over that.

**Reproducible funnel (parser-independent):**

- **1332** runners swept.
- **97** stale-green, by the deterministic classifier
  `stale iff cached_exit=="0" and live_exit not in (0,"0") and not timed_out`.
- **97 / 97** live-reconfirmed as still regressing against current main (zero false
  positives).

A note on the "genuine green" sub-count: how many of the 97 count as "genuine
parseable-green cache" vs "non-standard output format" depends on which TOTAL parser
is used (the narrow stored parser yields 15/82; a broad re-parser yields 72/25).
Rather than quote a parser-dependent figure, this PR reports the parser-independent
fact — **97 stale-green records, all 97 live-reconfirmed regressing** — and treats each
individually by root cause.

**What this branch changes: 70 runner scripts + 70 regenerated caches.**

- **65** of the 97 stale-green records are repaired here (source-preserving fixes;
  cache regenerated only after the runner genuinely passes live).
- **5** further branch scripts were flagged by adjacent means (not stale-green) and
  rewired: **4 authorized note-reading reader-rewires** (`frontier_abj_phy_retained_bounded_supplier_wiring_2026_06_18`,
  `frontier_d3_lower_bound_source_packet_gate_2026_06_06`,
  `frontier_koide_aps_block_by_block_forcing`, `symmetry_head_to_head` — value-gate →
  row-existence + verbatim premise-quote, per policy `6597460032` and the owner
  directive to "rewire anything relying on reading notes") + **1 import retarget**
  (`frontier_tensor_support_center_excess_law` — canonical shell-law helper was
  refactored to delegate; retargeted to the self-contained replay variant).

**Remaining 32 of the 97:** **2** are the #4907 pair (excluded — see below); **30** are
the Class-D escalation class, diagnosed below and deliberately NOT blind-fixed. This PR
repairs the mechanically-fixable classes correctly and reports the escalation class
honestly without weakening any discriminating gate or regenerating any cache to hide a
red signal.

**Two self-caught reverts.** An adversarial line-by-line review of the branch caught two
[...]

PR #4954, opened 2026-07-04
fix(runners): repair two stale sibling-interface references (mechanical)

jonathonreilly:
## What this responds to

Follow-up to the stale-green sweep (PR #4943), which diagnosed 30 Class-D
stale-green runners in its body without committing fixes. The umbrella directive
was to correctly fix the *mechanically-fixable* Class-D items — stale
sibling-reference / indexing / exit-path bugs — while routing anything that turns
out to be a masked science failure to the science loop instead of papering over it.

Three items were queued as "mechanical." Each was **investigated first** to confirm
it is genuinely a stale-reference bug and not a masked science failure (the
must-fail-if-wrong test), *before* any edit. Result: **two are genuine mechanical
fixes (landed here); one is a masked science regression (escalated, untouched).**

## The two mechanical fixes (this PR)

**1. `frontier_gr_class_expansion_finite_rank_target`** — tuple/dict interface drift.
The sibling `frontier_finite_rank_source_to_metric_theorem.coarse_metric_report`
now returns a 3-tuple `(rows, best, ratio)`; the runner still indexed it as a dict
(`coarse_report["best"]` → `TypeError`). Repointed to tuple-unpack. `best` is still
`min(rows, key=lambda row: row[5])`, a 6-tuple, and the runner's reads `best[0,1,2,4,5]`
remain consistent with it. The discriminating `BOUNDED` checks (`best[5] < 1e-5`
vacuum-close coarse residual; `best[4] > 0` nonzero direct residual) and the
exact source-renormalization / boundary-stationarity gates are untouched.
Live re-run: `PASS=4 FAIL=0`, exit 0.

**2. `frontier_dm_leptogenesis_pmns_selector_bank_cp_sheet_blindness_theorem`** —
sibling function rename. The sibling
`frontier_dm_leptogenesis_pmns_transport_extremal_source_candidate` renamed its
`part2`/`part3` entry points; the runner still called the removed names
(`AttributeError`). Repointed to the current names
(`part2_transport_search_finds_an_off_seed_overshoot_witness`,
`part3_continuity_gives_an_interpolated_closure_witness`). Semantics preserved:
part2 still yields an `eta > 1` overshoot witness and part3 still yields the
`eta = 1` closure to 1e-10, so the runner's own opposite-CP-symmetry and
off-mainline-landing gates (which do the actual science discrimination) are
unchanged and still pass. Live re-run: `PASS=13 FAIL=0`, exit 0.

Both caches were regenerated **only after** each genuinely passed live
(`exit_code: 0` in each cache header). No note edits, no gate weakening, no
verdict changes.

## The third item is NOT mechanical — escalated, not touched here

**`frontier_g_bare_two_ward_closure`** was queued as a mechanical "verdict/exit
reconcile," but investigation shows it is a **genuine science regression**, so it
is deliberately left honestly red (`PASS=17 FAIL=1`, exit 1) and routed to the
science loop instead.

The runner's Block 5 gates on the authority note
`docs/YT_WARD_IDENTITY_DERIVATION_THEOREM.md` asserting the two representations
of `Γ⁴(q²)` *"must agree"* as a **forcing** identity, and Block 6 then **solves**
`g_bare² = 2 N_c y_t_bare² = 1` and prints `VERDICT: CLOSED` (g_bare = 1 *derived*
via the two-Ward closure). But that note was repaired upstream (commit `2fc3c1c3b4`,
"Repair YT Ward identity derivation") to **retract the forcing claim**: it now
evaluates Representation A *"at canonical g_bare = 1"* and states *"the agreement
… confirms internal consistency of the framework **but is not the source of the
value**,"* tracing g_bare = 1 to the *"algebraic g_bare = 1 rescaling convention"*
(an input), not a derivation.

Re-pinning the runner's needle to the note's new "…values agree at the canonical
surface" phrasing would let it keep printing `VERDICT: CLOSED` / "Therefore →
g_bare = 1" while the authority note has conceded exactly that derivation is only a
consistency check. That fails the must-fail-if-wrong test, so it is **not** a
mechanical re-pin. Block 5 is a correctly discriminating gate doing its job; the
honest `FAIL=1` is the right signal. Resolution (downgrade the runner's CLOSED
framing to match the repaired note, or adjudicate whether the closure genuinely
pins g_bare) is science-loop work, reported here only — not committed to `docs/`.

## Files
- `scripts/frontier_gr_class_expansion_finite_rank_target.py`
- `scripts/frontier_dm_leptogenesis_pmns_selector_bank_cp_sheet_blindness_theorem.py`
- `logs/runner-cache/` caches for both (regenerated post-pass)

### DIFF
--- a/logs/runner-cache/frontier_dm_leptogenesis_pmns_selector_bank_cp_sheet_blindness_theorem.txt
+++ b/logs/runner-cache/frontier_dm_leptogenesis_pmns_selector_bank_cp_sheet_blindness_theorem.txt
@@ -1,9 +1,9 @@
 ===== runner cache v1 =====
 runner: scripts/frontier_dm_leptogenesis_pmns_selector_bank_cp_sheet_blindness_theorem.py
-runner_sha256: efd2cace1340bcd4e73aa0d503c5c5c51ea736fae135457c09b571d65fb23bd4
-timeout_sec: 120
+runner_sha256: efefef2c830ed90d58fd4ce9cda8f43815f62ddbb983c8be2d8000ba7818a508
+timeout_sec: 300
 exit_code: 0
-elapsed_sec: 17.26
+elapsed_sec: 5.39
 status: ok
 ----- stdout -----
 ========================================================================================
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Landed in 432594dfdb after rebasing onto live main. Verified the two helper-interface repairs preserve the gates, regenerated both runner caches, restored generated audit/publication outputs before commit, and ran pipeline plus strict audit lint with existing known warnings only. No registry edits, effective-status changes, or audit verdicts included. Closing and deleting the landed branch.

PR #5033, opened 2026-07-05
science: conform rp two-step runner to note's free-theorem scope

jonathonreilly:
## What / why

The audit lane graded `axiom_first_rp_two_step_transfer_matrix_positivity_note_2026-05-28` **`audited_conditional`** on a runner/source drift: the runner emitted `PASS=7 FAIL=0` while the source note states `PASS=6 FAIL=0`.

The extra check was **C7**, an out-of-scope guard that:
- `read_text`'d the **downstream, undeclared, unaudited** note `docs/AXIOM_FIRST_REFLECTION_POSITIVITY_THEOREM_NOTE_2026-04-29.md` — a citation-graph inversion (a supplier runner reaching *forward* into an unaudited consumer);
- checked required/forbidden phrase lists plus a toy hardcoded-array positivity check;
- **witnessed nothing** about this note's free two-step positivity claim; and
- duplicated a guard already owned by a dedicated runner.

The runner's own docstring SCORECARD already listed **only C1–C6**, confirming C7 was a bolted-on addition — so the fix is **removal**, conforming the executable code to the runner's already-honest documented scope and to the source note.

## Change (runner + cache only; note unchanged)

- Remove C7 and its dead scaffolding: the `NOTE_PATH` constant, the now-unused `Path` import, the `check_three_factor_bridge_source()` helper, the parent-bridge narration block, and the C7 scorecard line.
- Runner now emits `PASS=6 FAIL=0`; its SUMMARY lists exactly **C1–C6 = the free two-step positivity theorem** the note claims.
- Cache regenerated and SHA-repinned to the new runner bytes.
- The source note is **already correct** (states `PASS=6`, free-scoped; cert table lists only C1–C6) and is **not** part of this PR.

Diff: 2 files, +5 / −144, all C7-scoped. `audit_lint.py --strict` → `OK: no errors`.

This clears the gate so the note can be re-audited to `retained` and the central-registration dependency chain can continue draining.

### DIFF
--- a/logs/runner-cache/axiom_first_rp_two_step_transfer_matrix_positivity.txt
+++ b/logs/runner-cache/axiom_first_rp_two_step_transfer_matrix_positivity.txt
@@ -1,9 +1,9 @@
 ===== runner cache v1 =====
 runner: scripts/axiom_first_rp_two_step_transfer_matrix_positivity.py
-runner_sha256: ca7f4f79377152286ad9ce35332e49096fa11789975b53cb3b8818b905da6227
+runner_sha256: 1732cef8361d0eab0ef23e9e47435b1a40e291f68ec90a60305ab3d09c62dcec
 timeout_sec: 300
 exit_code: 0
-elapsed_sec: 0.08
+elapsed_sec: 0.14
 status: ok
 ----- stdout -----
 ==============================================================================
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Landed the reviewed two-file delta directly on main as b59f83a1083388722e4342d8ba19c777bb0ebe7f, avoiding the stale generated/audit diffs on the PR branch. Runner now reports PASS=6 FAIL=0 for the note's free-theorem scope; no audit verdict was applied.

PR #4209, opened 2026-06-17
Propose color composition-carrier as 3rd Tier-A admission (proposal note; owner-gated)

JONbridgerapps:
## ⚠️ OWNER-GATED (converted from draft 2026-07-04 — same handling as #3511)

This PR is **intentionally gated**. Review-loop: review the content normally (one `docs/` proposal note; no registry edit, no science verdicts, no audit statuses), then **PAUSE — landing and, separately, any registration of the admission require explicit owner approval recorded per `docs/audit/AXIOM_MINIMALITY_POLICY.md` §6.** Do not close/salvage as non-landable; an unmet owner gate is not a review failure. Registering `COMP` (count 2 → 3) is an owner/minimality-policy decision, never a review action.

## What this is

A **proposal** to register the framework's color/strong-gauge admission as a third Tier-A admitted input (`COMP`), alongside `AC_phi_lambda` (flavor) and `theta` (strong-CP), formalizing the existing color matter-realization residual (MR_color) in a sharpened minimum statement.

`COMP` = *"the internal gauge-carrier medium is a normed composition (division) algebra, with an oriented/T-odd grading."* `SU(3)` color and one matter generation (`C^8 = 1+3+3bar+1`) follow as consequences; the medium is a **label** on the associative carrier, not the operating algebra (Jacobi forbids the non-associative octonion as the carrier).

## 2026-07-04 premise refresh (this update)

The non-derivability was originally stated against the pre-reset three-axiom base. The note now argues the carry-over to the current four-axiom base explicitly: **Admissibility** (2026-06-29) names no operator/kinetic-class/selector/carrier and, being a spatially-covariant nearest-neighbor availability rule, cannot supply the missing spatial-**singlet** internal arena (the obstruction is geometric); the **formation sentence** ("Records form.", 2026-07-04) adds occurrence only. Axiom-name currency fixed (Quantum → Qubit). Restatement is proposal-level rationale, pending normal audit-lane re-check.

## Why Tier-A (not axiom, not primitive)

Unchanged from the original proposal: not an axiom (enriches an existing carrier slot; the forcing test analogous to the one that forced `i` failed), not a primitive (heaviest dimensionless selector content in the matter sector — exactly what the primitive boundary forbids), is Tier-A (contentful + underivable + confers bounded status; mirrors `AC_phi_lambda`'s admission-plus-sub-admissions shape).

## Scope / governance

[...]

PR #5039, opened 2026-07-06
science(color): block02 -- bonded-pair arena from the shared Qubit presentation; Admissibility cross-site dependence; spatial-singlet obstruction landed as conditional

JONbridgerapps:
## Campaign

color-derivation-20260706, block 02 of 07. The constructive opening move: the internal arena exists at axiom strength, and the geometric obstruction that motivated the closed COMP proposal (#4209) is landed in scoped, conditional form instead of remaining session-level rationale.

## Claims

- **T1 (exact):** a nearest-neighbor edge carries the composite M_4(C); the exchange split Sym(3)+Anti(1) is canonical relative to the shared algebraic presentation the Qubit axiom supplies. Stability is proven EXACTLY for every g in End(C^2) via S(A(x)B)S = B(x)A on all 16 elementary pairs (bilinearity — no sampling, no unitarity assumption). Frame-relativity is exhibited as load-bearing honesty: independent per-site presentation changes move the split — that relativity is precisely the downstream transport question (new residual R5).
- **T2 (bounded, deliberately weakened after panel):** the Admissibility sentence places cross-site DEPENDENCE at axiom strength — the rule cannot be stated on a single site alone. Explicit typing disclaimer: 'conditions' is not typed by the axiom; no cross-site OBJECT is attributed. The June-8 no-go's re-opening clause is NOT discharged; what is recorded is that its named baseline (pre-Admissibility) has changed, so the framing question is live for the constructive blocks.
- **T3a (consistency witness):** a runner-checked action (proper cubic rotation as edge relabeling, trivial internal part) exhibits axiom-compatibility of a spatially-trivial internal block of real dimension 6. Witness strength only.
- **T3b (exact negative control):** under the supplied spin-axis identification, the Sym block is the spin-1 rep and its commutant is scalars — verified numerically (singular-gap asserted) AND exactly (Gaussian elimination over Q[i] on the integer vector rep). The obstruction is real exactly where its premise holds; the axioms do not supply that premise.
- **T4:** residual ledger R1-R5, all open; the carrier-budget endpoint residual is left open and unmodified (the draft's 'relocation' claim was panel-refuted and removed).

## Method

Workhorse split + 3-seat adversarial refutation panel. The panel refuted the draft's T2 (typing gap: 'conditions' is not 'possibilities'; pair-domain subobject not entailed; 'discharges the re-opening clause' overclaimed), T1's canonicity phrasing (flip vs endomorphism; SU(2) sampling insufficient; S=S^T basis-dependent), T3a's free-floating conclusion, and R2's relocation. All repaired by weakening to the entailed statements and adding exact verification. One panel finding was promoted to content: the frame-relativity of the split under independent per-site changes is now stated as the seed of the transport question (R5).

## Verification

- Runner 33 PASS / 0 FAIL (cache committed): exact End(C^2) stability, exact Q[i] commutant, spin-1 Casimir, text audits of the quoted axiom and June-8 sentences, consistency witness, AST self-scan.
- ASCII-clean; git diff --check clean.
- audit_lint --strict: one pre-existing ERROR inherited from live main (multisite_pauli_group_theorem_note_2026-05-02 note_hash mismatch — from the multisite-pauli landing, needs the audit-lane re-seed; NOT introduced by this PR; verified present on clean origin/main).

## Boundaries

No audit verdicts, no registry edits, no status changes, no axiom/primitive/Tier-A content. Does not derive color, supply a carrier/transport/bridge/dynamics, or claim the obstruction globally dead. Review-loop lands; the multisite-pauli hash re-seed is review-loop/audit-lane territory.


JONbridgerapps:
Review-loop No-Go Discipline Gate for #5039

Scope checked: the conditional negative control in T3b and the bounded obstruction wording. This is a review-loop gate only, not an audit verdict.

N1 alternative routes:
1. Minimal-axiom route forcing internal spin-axis covariance: fails for this note because the quoted four-axiom surface supplies site rotations and the one-site algebraic presentation, but no action tying qubit generators to spatial axes; T3a supplies a runner-checked trivial-internal-action witness.
2. Admissibility-as-carrier route: fails for this note because T2 claims only nearest-neighbor dependence and explicitly refuses to type "conditions" as a pair-domain object or carrier.
3. Independent per-site frame route preserving the split: fails for this note because the runner exhibits `g tensor 1` moving the split; this becomes residual R5, not a closed theorem.
4. Spin-axis supplied route with non-scalar `su(3)` commutant on `Sym^2(C^2)`: fails under the stated condition by Schur/commutant computation; the runner checks complex commutant dimension 1 numerically and by exact rational elimination on the vector representation.
5. Link/endpoint/transport/Gauss-Wilson route: not closed by this note and therefore cannot break the claimed negative boundary; it is explicitly preserved as R2/R4/R5 open work.

N2 wall-independence: no inflated wall set. The only condition for the negative control is the supplied spin-axis identification. Open residuals R1-R5 are not counted as closed independent walls.

N3 hidden-wall scan: reviewed "canonical", "conditions", "does NOT", "not supplied", and "obstruction" uses. "Canonical" is limited to the shared presentation and immediately paired with frame-relativity; "conditions" has the explicit typing disclaimer; no hidden admission is promoted.

N4 residual matching: the June-8 re-opening clause asks for a retained cross-site carrier theorem. #5039 explicitly says it does not discharge that clause. The carrier-budget endpoint residual is quoted and left open. Residuals match and are not over-consumed.

N5 rhetoric audit: negative phrases are scoped to the axiom-text surface or to the supplied spin-axis condition. No lattice-wide or global "obstruction dead" claim is made.

N6 partial-closure scan: no "requires new axiom" or "no retained primitive" wall is asserted. Primitive registry boundary was checked; approved primitives are not used as no-go walls.

N7 steelman: a hostile reviewer could argue the shared Qubit presentation is not enough to compare frames across distinct sites. The note already concedes exactly that point as frame-relativity/R5 and does not claim transport or physical color realization, so the steelman does not break the scoped T3b negative control.

N8 cross-cycle echo: repo scan found prior similar wall language in observable-principle, staggered-labeling, PMNS, and axiom-stack no-go surfaces; those caution against laundering convention/reframe paths into global no-go claims. #5039 avoids that failure mode by keeping realization/transport residuals open and not calling for a new axiom.

Disposition: PASS for the narrow conditional negative claim. No global no-go or retained-status claim is approved.

JONbridgerapps:
Landed on main via review-loop as 6cc22dadcd. No-Go Discipline gate posted in this PR thread; generated audit/status outputs stripped; independent audit lane owns verdicts.

PR #5041, opened 2026-07-07
science(color): block04 -- factor preservation under REGISTERED-FACTOR (named premise): exact u(6) stabilizer with verified su(3)+su(2) structure

JONbridgerapps:
## Campaign

color-derivation-20260706, block 04 of 07: the gauging-selection leg. Given a supplied C^3 (x) C^2 model surface and factor preservation, the stabilizer inside u(6) is exactly the factorwise algebra — computed exactly and now structure-verified, with the supplier premise named rather than smuggled.

## Claims

- **T1 (exact):** u(6) dim 36 -> factorwise algebra dim 12 (center overlap bookkeeping exact); stabilizer characterized and equal to the factorwise algebra by both containments; semisimple part su(3)+su(2) verified as LIE STRUCTURE (panel-forced upgrade): bracket closure, vanishing cross-brackets, central abelian direction, and an exact Q-arithmetic Killing form (block-diagonal, nonzero determinant). Honest finding preserved: on this tensor carrier there is NO second central/factor-identity u(1) — the only central abelian image is the global phase, so a physical hypercharge identification needs structure beyond this carrier (R-hypercharge).
- **T2 (conditional; stated plainly as a compatibility convention, not a derived theorem):** under the named premise REGISTERED-FACTOR (the split is record-typed data — NOT axiom content), compatible transformations normalize the split and T1 applies. Record-typed data is non-invariant typed structure — ONE of the live non-invariant inputs the June-16 conjugation-independence no-go's own escape names (not the whole escape).
- **Residuals:** R-registered-factor (open; relation to block03's MARGINAL-READ open), R-which-factor, R-chiral (su(2)_L out of scope), R-hypercharge, R-u6-embedding (the C^6 surface is supplied, not derived).

## Method

Workhorse split + 3-seat panel. Panel refuted: dimension-only semisimplicity (fixed with exact brackets + Killing form), the definitional character of T2 hidden as a theorem (now stated plainly first), 'exactly the loophole' overstatement (now 'one named input'), a literally-false 'no traceless abelian direction' phrasing (now scoped to central/factor-identity directions with the Cartan caveat), an unread dependency row (moved to see-also), a cache contradiction, citation-contract audit-gating, and blank-line discipline. All repaired.

## Verification

Runner 8 PASS / 0 FAIL, all load-bearing arithmetic exact over Q (Fraction re/im), cache committed; text audits of every quoted June-16/June-18/axiom sentence; AST self-scan. ASCII-clean. Strict lint: only the pre-existing multisite-pauli hash mismatch inherited from live main.

## Boundaries

No audit verdicts, registry edits, or axiom/primitive/Tier-A content. Does not derive gauging, choose the registered factor, identify color/weak/hypercharge physically, or supply chirality. All three June source rows are unaudited and consumed as target/context only — every load-bearing result is recomputed self-contained. Review-loop lands.

### DIFF
--- a/docs/GAUGE_FACTOR_PRESERVATION_RECORD_TYPED_SELECTOR_CONDITIONAL_DECOMPOSITION_BOUNDED_THEOREM_NOTE_2026-07-06.md
+++ b/docs/GAUGE_FACTOR_PRESERVATION_RECORD_TYPED_SELECTOR_CONDITIONAL_DECOMPOSITION_BOUNDED_THEOREM_NOTE_2026-07-06.md
@@ -1,0 +1,225 @@
+# Gauge Factor Preservation From Record-Typed Selector: Conditional Decomposition Bounded Theorem
+
+**Date:** 2026-07-06
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set, predict, or apply an audit outcome.
+**Primary runner:** [`scripts/gauge_factor_preservation_record_typed_selector_2026_07_06.py`](../scripts/gauge_factor_preservation_record_typed_selector_2026_07_06.py)
+**Cache:** [`logs/runner-cache/gauge_factor_preservation_record_typed_selector_2026_07_06.txt`](../logs/runner-cache/gauge_factor_preservation_record_typed_selector_2026_07_06.txt)
+
+## Summary
+This note drafts color-campaign block 04 at bounded-theorem strength only. It is a conditional decomposition, not a derivation of gauging, not a derivation of the carrier, and not a derivation of a factor-preservation rule from the four axioms.
+The exact content is:
+```text
+supplied C^6 = C^3 tensor C^2
+AND preservation of M_3 tensor I_2 and I_3 tensor M_2
+  => the stabilizer inside u(6) is exactly the factorwise algebra.
+```
+T1 is exact and runner-verified over integers/Fractions after real/imaginary vectorization. No floating rank is used as evidence.
+The bridge content is premise-named:
+```text
+REGISTERED-FACTOR (named premise, introduced here):
+the record/readout structure registers a fixed factor subalgebra of the local domain; equivalently, the split
+  M_3 tensor I_2 / I_3 tensor M_2
+is record-typed data.
+```
+The four axioms do not supply REGISTERED-FACTOR. The Record axiom's content-only, site-local readout makes record-typed data the natural candidate supplier, but that is a candidate bridge, not axiom content. Under REGISTERED-FACTOR, record-compatible transformations normalize the registered factor split, and T1 gives the factorwise algebra exactly.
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop No-Go Discipline Gate for #5041

Scope checked: the bounded conditional wall `REGISTERED-FACTOR` and the negative statement that this tensor-carrier image has no second central/factor-identity `u(1)`. This is a review-loop gate only, not an audit verdict.

N1 alternative routes:
1. Full `u(6)` survives factor preservation: fails under the stated ordered preservation condition by the exact stabilizer computation, which gives dimension 12 and equality with the factorwise algebra.
2. A second central/factor-identity `u(1)` survives in the carrier image: fails by exact center-overlap bookkeeping; the relative identity pair maps to zero, while Cartan abelian subalgebras are explicitly outside that negative statement.
3. Conjugation-invariant carrier-data route selects the embedding: not claimed; the June-16 no-go is quoted as context and #5041 uses non-invariant REGISTERED-FACTOR instead.
4. Record axiom derives REGISTERED-FACTOR directly: not closed. The note names REGISTERED-FACTOR as a conditional supplied premise and preserves R-registered-factor.
5. Physical color/weak/hypercharge/chirality route: not closed. Those are preserved as R-which-factor, R-chiral, R-hypercharge, and R-u6-embedding.

N2 wall-independence: the collapsed premise set is supplied `C^3 tensor C^2` model surface plus REGISTERED-FACTOR/factor preservation. Hypercharge, chirality, physical carrier, and which-factor questions are residuals, not extra counted walls.

N3 hidden-wall scan: reviewed "registered", "factor preservation", "compatibility", "normalizer", "hypercharge-like", and "context" wording. The branch now states T2 is definitional/conditional, and context-only June rows are code-formatted so they do not seed false graph dependencies.

N4 residual matching: June-16 is used only for the invariant-data no-go and escape-route framing; June-18 is target/context only. The live residuals named by those rows are not claimed closed.

N5 rhetoric audit: "no second central/factor-identity `u(1)`" is scoped to the tensor-product carrier image and explicitly excludes Cartan directions inside `su(3)` or `su(2)`.

N6 partial-closure scan: no "requires new axiom" or "no retained primitive" wall is asserted. Primitive registry boundary was checked; REGISTERED-FACTOR remains an unapproved named premise, not a primitive or Tier-A admission.

N7 steelman: a hostile reviewer could argue factor preservation is definitional and may not correspond to physical gauge selection. #5041 concedes this directly: T2 is a compatibility convention, not a derivation, and physical gauging/chirality/hypercharge remain open.

N8 cross-cycle echo: prior no-go surfaces warn against promoting an escape-route input into a theorem. #5041 avoids that by preserving REGISTERED-FACTOR as a named condition and keeping the physical selection residuals open.

Disposition: PASS for the scoped bounded/negative claims. No global gauge-selection no-go, physical hypercharge claim, or retained-status claim is approved.

JONbridgerapps:
Landed on main via review-loop as ab2c92a0b2. Review-loop normalized source-note wording, kept context-only June rows out of dependency graph, posted No-Go Discipline gate, and stripped generated audit/status outputs.

PR #5044, opened 2026-07-07
science(color): block07 -- campaign closing brief: color NOT derived; MR_color re-bounded into eight named premises (meta; land LAST)

JONbridgerapps:
## What this is

The closing adjudication brief of color-derivation-20260706 (meta; no premise or interpretive weight). Plainly: color is still not derived. The campaign's product is the exact list of remaining assumptions — eight named premises with draft-strength proofs connecting each to what it would buy — assembled against the June-5 MR_color steelman, which is not discharged in any part. Tier-A count stays 0; no ruling is made or requested.

## Contents

- Per-part verdicts against MR_color's own N7 steelman (all three parts NOT closed; what each block drafted toward them, with unlanded status stated everywhere).
- The named-premise ledger (open debts, not achievements): SUPPLIED-C3, MARGINAL-READ, WEIGHT-UNIFORMITY, FERMI-FILL, REGISTERED-FACTOR, SUPPLIED-BILINEAR, ARROW, SUCCESSION-ORIENT — each with its role and route.
- Draft-strength runner tallies quoted from the blocks' committed caches (pre-audit, no status implied).
- Convergence pointers phrased as open questions: the MARGINAL-READ/REGISTERED-FACTOR shape match (could one readout-bridge theorem supply both?), FERMI-FILL/full-rank occupancy, ARROW = the unowned time arrow (routed to B14), and block 04's no-second-central-u(1) hypercharge sharpening.

## Landing invariant

Land this note LAST, after #5038, #5039, #5040, #5041, #5042, #5043. At landing, run the runner with --post-landing: absences and non-zero-fail sibling caches are hard failures in that mode (pre-landing mode prints an explicit not-a-landing-validation warning; current cache is the pre-landing run: 4 PASS / 0 FAIL / 6 SKIP).

## Method note

Two-seat panel (verdict honesty; citation discipline) refuted the draft's 'landed' language for unlanded siblings, achievement-flavored ledger framing, physics-asserting convergence items, missing no-premise-weight header phrase, over-permissive self-citation, and a false-green-capable runner (substring cache checks, skip-exits-zero). All repaired: unlanded status stated throughout, ledger retitled 'open debts', convergence items are questions, citation contract reduced to pointer-only, runner anchors to final summary lines and has the post-landing hard mode.

## Boundaries

Meta only: no audit verdicts, no status, no registry edits, no axiom/primitive/Tier-A content, no ruling. Review-loop lands (last).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5040, opened 2026-07-06
science(color): block03 -- toward color-singlet records (g2): conditional decomposition under named premises; bare site-locality exactly refuted

JONbridgerapps:
## Campaign

color-derivation-20260706, block 03 of 07. Attacks MR_color gate g2 ('physical records are color singlets') by honest decomposition rather than overclaimed derivation.

## What the panel forced, and why the result is stronger

The 3-seat refutation panel refuted the draft's core inference: representing record content by single-site density marginals is NOT licensed by the Record axiom text — it is a modeling bridge. The repair names it as the conditional premise **MARGINAL-READ**, in the campaign's named-premise discipline. Likewise the closed-shell route's fermionic statistics became the named premise **FERMI-FILL** ('link-structure-free', never 'presupposition-free'), and the panel's marginal-to-singlet gap (rho = I/3 one-body data is strictly weaker than the state being an SU(3) singlet) is now the explicit residual **R-singlet-gap**. The headline is a one-directional sufficiency, not an equality:

LOCK-LOCAL + MARGINAL-READ + PHASE-FREE + WEIGHT-UNIFORMITY => no record-readable phase or weight distinction between color orientations.

## Claims

- **T1 (exact + insufficiency exhibit):** record content is site-local (axiom text); under MARGINAL-READ, record-visible edge data factors through single-site marginals. AND: |00> vs |11> in the Sym block have marginal trace distance exactly 1 — bare site-locality does NOT force color-blind records; that route to g2 is exactly refuted.
- **T2 (exact):** orientation phases are single-site-marginal-invisible — proven by a genuine exact partial trace in the ring Q + Q e^{i theta} + Q e^{-i theta} (no phase coefficient survives), not a hard-coded expansion (panel-forced rigor upgrade).
- **T3 (conditional):** weight-uniformity via two routes: AD-INV consumed by note + audited claim scope quoted verbatim from the ledger (all conditions carried: supplied C^3 carrier, nonzero equivariant drift, Ad-invariant first moment — the einselection caution named); FERMI-FILL recomputed exactly on the 8-dim Fock space (Gamma_ij = delta_ij, integer/Fraction Jordan-Wigner).
- **T4:** the conditional sufficiency above, explicitly NOT g2 and NOT an equality; which form the June-5 consumer needs is an open check.

Residual ledger: R-marginal-read, R-singlet-gap, R-fermi-fill, R-open-shell, R-sector-selection, R-einselection-circularity, R-transport/frame (inherited).

## Verification

[...]

PR #4922, opened 2026-07-04
Born form via composite Gleason: bridge-conditional note (re-opened on main after #4915 direct-land)

JONbridgerapps:
Re-opens the seat-repaired Born-form bridge note that closed unmerged when its stack base (#4915's branch) was closed in favor of the direct land. Content identical to the reviewed version; runner re-verified against main (the landed formation sentence needle now passes on main directly): **PASS=24 FAIL=0**.

Summary: IF grading is a full projection-lattice frame function on nearest-neighbor composites (H1-H4, with H4 declared as the substantive physical hypothesis carrying four specification obligations), THEN Gleason (named bridge input) forces Born trace form on single-site restrictions; the one-site M_2 rogue frame function is exhibited and refuted exactly (three directions + normalization). Attribution corrected per seats: adjacency makes the composite available; H4's projection-measure strength voids the dimension-2 loophole.

Cross-reference: the interface-consistency note (PR opening alongside) found the v1 pager core's menu-normalization wording defective and produced core v2 (conditioning form) — v2 supplies exactly the full-lattice frame-function strength this note's hypotheses need, so the two are consistent; H-block language keys to v2 at the next re-key wave.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5045, opened 2026-07-07
science(readout): FRAME-EXT unification -- two color-campaign readout premises reduce to one via composite Gleason (land after #5040/#5041)

JONbridgerapps:
## What this is

The first follow-on to the color-derivation campaign, executing the highest-leverage convergence pointer from its closing brief: MARGINAL-READ (block 03) and REGISTERED-FACTOR (block 04) have the same record-as-structure shape, and both reduce to a single named premise.

## Claims

- **T1(a) (exact):** if record consumption has density-matrix form on the nearest-neighbor composite (REP), the site-marginal representation follows — the restriction/partial-trace identity verified over Q on all basis pairs, BOTH factors. **T1(b) (convention-strength only, labeled):** REP restricted to an assumed typed subalgebra gives block-04's compatibility shape — bookkeeping, not a selection theorem.
- **T2 (exact):** the dim-2 exception made concrete: f(P_n) = (1 + n_z^3)/2 on rational Bloch points is perfectly frame-additive on a single site (orthogonality = antipodality in M_2) yet provably not of density form — exact rank-3-vs-4 inconsistency certificate over Q. Per-site readout can never be forced to density form; the composite is where Gleason bites. The same nearest-neighbor composite the color arena needed — the lattice closes the same loophole on both sides (observation, non-claim).
- **T3 (conditional):** FRAME-EXT + external Gleason (stated precisely: nonnegative normalized frame functions, dim >= 3, finite-dimensional — panel forced the precision) => REP => both premises' content. Ledger: eight named color premises become seven.
- **T4 (gap statement, explicitly not a non-derivability proof):** FRAME-EXT's principal gap is FINITE-ADDITIVITY-TO-FRAME — Record's additivity ranges over realized disjoint records; a frame function needs all mathematical decompositions. Panel-forced honesty: FRAME-EXT carries FOUR load-bearing clauses (content-determination, positivity, normalization, composite-domain authority), each separately unsupplied by axiom text — the premise count drops only because all four live inside one named premise. Forward routes stated neutrally: derive, or register the graded-constraint primitive (owner surface; no ruling requested).

## Landing order (hard requirement)

Land AFTER #5040 and #5041 (their premise scopes are what is being reduced); T3 is citable only once those siblings land. Recommended slot: immediately after the color campaign lands, before or after #5044 indifferently.

## Method

Workhorse split + 3-seat panel. Panel refuted the draft's under-specified FRAME-EXT (Gleason misquoted as needing only boundedness; content-determination silently assumed), T1(b) typing, and the missing landing gate — all repaired; the second-factor restriction check was added to the runner.

## Verification

Runner 6 PASS / 0 FAIL, all load-bearing checks exact over Q (cache committed); text audits of all quoted sentences from the axiom memo, the graded-constraint META memo, and both sibling notes; AST self-scan. ASCII-clean.

## Boundaries

No audit verdicts, registry edits, axiom/primitive/Tier-A content. Does not derive Born weights/form (adjacent open #4922 not consumed); does not establish FRAME-EXT; does not discharge either color premise. Review-loop lands.

### DIFF
--- a/docs/READOUT_BRIDGE_FRAME_EXTENSION_UNIFIES_MARGINAL_READ_AND_REGISTERED_FACTOR_CONDITIONAL_BOUNDED_THEOREM_NOTE_2026-07-06.md
+++ b/docs/READOUT_BRIDGE_FRAME_EXTENSION_UNIFIES_MARGINAL_READ_AND_REGISTERED_FACTOR_CONDITIONAL_BOUNDED_THEOREM_NOTE_2026-07-06.md
@@ -1,0 +1,227 @@
+# Readout-Bridge Frame Extension Unifies MARGINAL-READ And REGISTERED-FACTOR: Conditional Bounded Theorem
+
+**Date:** 2026-07-06
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Status authority:** independent audit lane only. This source note does not set, predict, or apply an audit outcome.
+**Primary runner:** [`scripts/readout_bridge_frame_extension_unification_2026_07_06.py`](../scripts/readout_bridge_frame_extension_unification_2026_07_06.py)
+**Cache:** [`logs/runner-cache/readout_bridge_frame_extension_unification_2026_07_06.txt`](../logs/runner-cache/readout_bridge_frame_extension_unification_2026_07_06.txt) -- supervisor generated from the runner output; this drafting worker does not generate cache.
+
+## Summary
+This note is a follow-on to the color-derivation campaign. It does not derive Born weights, does not derive the Born quadratic form, and does not consume adjacent PR #4922, which is scoped to Born weights.
+
+The color campaign introduced two readout-bridge premises with the same shape: records-as-marginals and record-typed factor structure. This note reduces both to one named premise:
+```text
+FRAME-EXT (four clauses; the premise's full strength lives here):
+for each record configuration, the record readout extends to a
+(1) content-determined,
+(2) nonnegative,
+(3) normalized
+frame function on the projection lattice of every nearest-neighbor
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop landed as 306ce4c2bb with narrow current-main fixes.

Disposition: PASS / landed narrowed. I kept the theorem conditional: FRAME-EXT plus external Gleason gives REP, which reduces MARGINAL-READ and REGISTERED-FACTOR to one named conditional premise. I removed stale “UNLANDED” / landing-order / PR-number posture now that the prerequisite color source notes are already on main, kept the Born/FRAME-EXT gaps explicit, and made the runner/cache use the standard cache format. No registry edits, audit verdicts, or effective-status outputs landed.

Validation run before landing:
- python3 -m py_compile scripts/readout_bridge_frame_extension_unification_2026_07_06.py
- python3 scripts/readout_bridge_frame_extension_unification_2026_07_06.py -> TOTAL PASS=6 FAIL=0
- python3 scripts/cached_runner_output.py scripts/readout_bridge_frame_extension_unification_2026_07_06.py --timeout 120
- independent exact check for partial-trace coefficient identity and M2 antipodal/rank obstruction
- python3 scripts/vocab_lint.py --fix on changed files -> clean
- bash docs/audit/scripts/run_pipeline.sh -> complete, no errors
- python3 docs/audit/scripts/audit_lint.py --strict -> no errors; new row is bounded_theorem / unaudited with landed source deps
- final guard confirmed no generated audit/status/registry surfaces staged and no local absolute markdown links

Boundary: this does not establish FRAME-EXT, does not derive Born form/weights, and does not discharge either color premise from the axioms.

PR #5050, opened 2026-07-07
docs(rp): conform SU(N) scope of the Wilson temporal-gauge RP bridge to its audited reach

jonathonreilly:
## What / why

Row ④ of the central-registration drain chain
(`axiom_first_reflection_positivity_wilson_temporal_gauge_bridge_narrow_theorem_note_2026-06-05`)
graded **`audited_conditional`**. The audit correctly flagged §4's *"Structural
argument for general `SU(N)`"*: it claimed the character expansion of
`exp(β Re Tr)` on **any** compact group has nonnegative coefficients "by
Peter-Weyl". That step is invalid on a nonabelian group — Peter-Weyl gives the
character *expansion* of a class function, but positivity of a class function
does **not** imply nonnegative character coefficients when the irreducibles are
higher-dimensional (it does on abelian groups). So §3's abelian mechanism does
not carry to `SU(N≥3)` on its own, and the note overstated its reach.

## Change (note-only conformance narrowing — N5 rhetoric alignment)

Conform the prose to the reach the paired runner actually witnesses:

- **Exact reprove surface = `Z_N`** (finite discrete-Fourier nonnegativity) **and
  `U(1)`** (Bessel positive-series plane-kernel certificate), with **`SU(2)`** as a
  numeric Haar Monte Carlo sample.
- **`SU(N≥3)` Wilson-plane character-coefficient positivity is reframed as carried
  by the published Osterwalder-Seiler comparator** — *not reproven here*, and *not
  a derivation input*. OS is already cited in §0/§9 as a comparator only, so this
  is not a new import; it removes an overclaim.
- Every echo of the old "general-`SU(N)` by structure" language (Claim box W3, §0,
  §3, §4 body, §6, §7, §8 N1/N3/N4/N6/N7, refs, source-note boundary) is aligned
  to the same honest scope.

## Runner + cache unchanged (and why)

`scripts/audit_companion_reflection_positivity_wilson_temporal_gauge_2026_06_05.py`
already witnesses only `Z_N`/`U(1)`/`SU(2)` (Parts A–E) and prints
`TOTAL: 16 PASS / 0 FAIL`; its own summary already declares
*"Osterwalder-Seiler … : comparator only."* It makes **no** `SU(N≥3)` check-level
claim, so there is nothing to edit — this is a note-only PR (`runner_sha256`
stays pinned, cache stays valid). Re-ran on this branch: **`16 PASS / 0 FAIL`**.

## Expected audit effect

The conditional was triggered purely by the note's prose asserting a general-`SU(N)`
positivity the runner does not establish. With that reframed to the OS comparator,
the note's stated claim matches its witnessed content, so re-audit should flip
**`audited_conditional → retained_bounded`** (valid-retained), unblocking the rest
of the central-registration drain (WAVE 3+).

## Forward science path this opens (not part of this PR)

The clean way to actually *derive* nonabelian plane-kernel positivity is the
heat-kernel action `exp(-t·Casimir) = Σ_R d_R e^{-C₂(R)t} χ_R`, whose coefficients
are manifestly nonnegative for **all** compact groups including `SU(3)`. That
changes the plaquette action (Wilson → heat-kernel) and needs a separate
import-vs-native check before it can be a derivation input, so it belongs in a
follow-up note, not here.

### DIFF
--- a/docs/AXIOM_FIRST_REFLECTION_POSITIVITY_WILSON_TEMPORAL_GAUGE_BRIDGE_NARROW_THEOREM_NOTE_2026-06-05.md
+++ b/docs/AXIOM_FIRST_REFLECTION_POSITIVITY_WILSON_TEMPORAL_GAUGE_BRIDGE_NARROW_THEOREM_NOTE_2026-06-05.md
@@ -74,7 +74,7 @@ explicit lattice.
 >   `(\text{positive-half})·(\text{plane positive kernel})·(\text{reflected half})`
 >   and equals `\sum_a c_a\,W_I(a)\,\overline{W_J(a)}`, i.e. `G = W\,\mathrm{diag}(c)\,W^†`
 >   with `c_a ≥ 0`. Hence `G ⪰ 0`: the RP inequality `⟨Θ(F)·F⟩ ≥ 0` holds for the
->   integrated pure-gauge theory. (Reproved exactly for `Z_N`, `N∈{2,3,4,5}`, and
+>   integrated pure-gauge theory on the groups reproven here. (Reproved exactly for `Z_N`, `N∈{2,3,4,5}`, and
 >   analytically for the `U(1)` plane kernel by Bessel positivity, with a `U(1)`
 >   quadrature Gram cross-check; a numeric `SU(2)` sample is PSD within Monte Carlo
 >   error - runner Parts C, D, E.)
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Landed manually on main as 7b008581f8 after review-loop validation. Generated audit/status outputs were validation-only and were not committed.

PR #5058, opened 2026-07-08
science(shellmult): shell-ratio surface-boundary note — SHELL-MULT is irreducibly supplied on the named surface (derivation block 1 outcome)

JONbridgerapps:
First derivation block of the science-unlock tier. Question posed: can a_T/a_E = -2 be derived from shell-counting algebra (the audit lane's named bridge #2 for the endpoint cluster)? THE COMPUTATION DECIDED: no exact derivation is available on the named Route-2 endpoint surface, and the note proves exactly that bounded statement (panel-forced polarity: 'not derived on this surface', never 'not derivable').

Load-bearing findings, all computed:
- The shell gammas are EPS=0.005 finite-difference probes of a scanned tensor object (eta_floor differences along ex/t1x, anchor-normalized) — not algebraic contractions; the exact shell ratio is not well-posed there without fixing probe conventions (enumerated).
- The full tensor replay path is BROKEN on current main (AttributeError: one_parameter_shell.reduced_data, firing on the center call at the endpoint module line 133) — the only live source of the shell gammas is the frozen float certificate. Recorded as a dated non-fatal observation; R-replay-repair residual filed (module drift; prerequisite for any future exact-ratio computation).
- Both certificate reifications computed exactly and checked: binary-float ratio -14875335342499166/7417703850033121 and decimal-string ratio -4031967723697/2010572657265 (agree to printed precision; both differ from -2; residual -10822409167/2010572657265, relative deviation 10822409167/4021145314530 = 0.269%).
- Readout-map freedom witness tied consistently to the retained naturality no-go (extends its honesty to the T-side shell ratio; that no-go granted, never derived, the T-side candidates).

Consequence, honestly typed: SHELL-MULT remains an irreducibly SUPPLIED premise on the current surface; the shell-counting bridge requires a counting surface the named modules do not contain (R-counting-surface residual). Firewall: may not be cited as proving the exact ratio differs from -2 (undetermined), against the naturality no-go, as refuting SHELL-MULT, or as a no-go.

3-seat panel refuted the draft (polarity, quote fidelity, float reification, tautologies, citation-graph links) — all repaired; runner 13/0 load-bearing with exit-separated non-fatal observation tier; strict lint clean.

### DIFF
--- a/docs/QUARK_ROUTE2_SHELL_MULTIPLICITY_EXACT_SHELL_RATIO_SURFACE_BOUNDARY_NOTE_2026-07-07.md
+++ b/docs/QUARK_ROUTE2_SHELL_MULTIPLICITY_EXACT_SHELL_RATIO_SURFACE_BOUNDARY_NOTE_2026-07-07.md
@@ -1,0 +1,379 @@
+# Quark Route-2 Shell Multiplicity Exact Shell Ratio Surface Boundary
+
+---
+claim_type_author_hint: bounded_theorem
+claim_scope: >-
+  On the named Route-2 endpoint surface (the five named files), the shell
+  gammas enter only as probe-convention-dependent finite-difference
+  certificates whose full-replay recomputation path is currently broken; no
+  exact shell-ratio derivation is available there; SHELL-MULT therefore enters
+  the endpoint cluster only as a supplied premise.
+---
+
+**Date:** 2026-07-07
+**Type:** bounded_theorem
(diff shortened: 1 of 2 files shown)


jonathonreilly:
Closed by the completed /review-loop after durable reviewed content was consolidated and landed through #5094 (merge 0d66de7eb4). The canonical source/runner/cache or repair now lives on main; this superseded PR head will be deleted.

PR #5081, opened 2026-07-09
[physics-loop] energy-to-records-block03 — season synthesis: the gravity chain closed up to two named premises

JONbridgerapps:
## What this PR contains

The final-link campaign's synthesis and close (stacked on #5080), plus the season synthesis for the full 2026-07-08 arc.

- `docs/GRAVITY_CHAIN_CLOSURE_SEASON_SYNTHESIS_2026-07-08.md`
- `.claude/science/physics-loops/energy-to-records/` (HANDOFF, STATE)

## The chain

**Energy → records → slow time → universal fall**, each link runner-backed: activity-energy bound + co-movement (#5079); crowding slows formation with axiom-forced saturation and constraint-class sign (#5076); one-constant rate freedom (#5077); massless Poisson-class field dynamics with forced energy-contrast sourcing (#5068/#5071/#5073/#5074); the matter/WEP side (#5061–#5067); geometry's shape half from the June conformal arc (conditional).

**Closed up to exactly two named premises, both falsifiable in-framework:**
- **P-QD-BRIDGE** — activity is formation opportunity (the dynamics-form theorem's quantum-Darwinism record bridge, not yet derived);
- **P-DEPOSITION-SPARSE** — the realized formation rule deposits sparsely per unit activity (the wake dichotomy of #5080: sparse ⇒ gravity follows its source; heavy ⇒ gravitational-memory-like permanent trails — a prediction to quantify, not an embarrassment to tune away).

Also included: the falsification list, the season inventory (#5061–#5081, six campaigns, review order), and the named follow-up campaigns (realized deposition rate; deriving the QD bridge; June-arc audits; d=3/DMRG/tensor).

Nothing landed by this lane; independent audit required throughout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5082, opened 2026-07-09
[physics-loop] record-deposition-rate-block01 — kappa(theta) measured: sparse window non-empty above the registration floor

JONbridgerapps:
Block01 of the deposition-constant campaign (stacked on #5081). kappa(theta) = first-registrations per unit integrated activity on the gauged comparator, with ground-state-baseline-subtracted distinguishability (the interacting GS is entangled at 1-purity 0.27-0.49; absolute thresholds measure vacuum structure — the worker's draft honestly failed itself on this and the fix is documented), the axiom-shaped once-per-site counter, and both crossing conventions printed.

Results (supervisor-executed, TOTAL KAPPA-MEASURED): kappa falls with threshold (exponent ≈ −0.81); the registration cascade is transient-complete above θ ≥ 0.2 while below that floor a closed coherent ring recruits late first-registrations indefinitely (no quiescence — the yield is only defined above a floor, itself a finding); and **the deposition-sparse window is non-empty**: θ ≥ 0.2 puts fill-per-dwell under the campaign-6 wake bound. Reading: the phenomenology constraint is satisfiable exactly when registration is a genuinely thresholded, measurement-grade event. Per-transient yield only (driven steady-state = named follow-up); QD bridge remains a named premise. Five documented iterations.

Verify: `python3 scripts/deposition_per_activity_kappa_2026_07_08.py` (~5s, TOTAL KAPPA-MEASURED).


jonathonreilly:
Closed by the completed /review-loop after #5094. This slice was intentionally not landed because its live claim boundary or artifact was non-landable (false/empty gate, out-of-scope machinery, or defective finite-rotor projector as documented in #5094). No content from this head was salvaged, so its branch is retained as the recovery handle.

PR #5089, opened 2026-07-09
[physics-loop] registration-bar-block01 — bar shape exhibited; permanence-grade redundancy needs branching geometry

JONbridgerapps:
## What this PR contains

Block01 of the registration-bar campaign (owner-commissioned 2026-07-09: derive the registration bar; discharge the activity=opportunity bridge).

- `scripts/registration_redundancy_onset_2026_07_09.py` — runner, TOTAL: BAR-NOT-PINNED (exit 1; a physics outcome, not machinery)
- `scripts/registration_bar_parameter_probes_2026_07_09.py` — supervisor probe driver (weak-coupling and heavy-mass sweeps)
- `docs/REGISTRATION_REDUNDANCY_ONSET_BOUNDED_NOTE_2026-07-09.md` — bounded note
- Runner caches for both

## The derived shape (conditional on the standing QD record reading)

The Record axiom's own clauses select the registration criterion: "exactly one admissible possibility" → near-perfect pointer discrimination; "permanent" → at least two independent registers (single-fragment imprints are erasable — measured in the 2026-06-05 note); "additive over disjoint records" → the Darwinism plateau. Registration = redundancy onset. No formation rule chosen; the 2026-06-06 no-go's wall is respected (derivation runs through the admitted convention, not past it).

## What the comparator exhibits

- The lawful pointer is forced: site occupation is demolished by hopping (v1's clean negative confirms the 2026-06-05 necessity theorem in situ); the gauge-invariant cell charge with Gauss-law link registers is the framework-native pointer — the constraint structure is the copying mechanism.
- Single-register certification fires: I/H up to 0.88, excess 0.21 bits over ground state.
- **The permanence bar (R≥2) never fires, and it's geometry, not parameters**: xi_reg ≤ 1 link across g ∈ {0.05…1.0} and m ∈ {0.3, 3.0}. In d=1 the two boundary links are a Markov blanket for the cell charge — distant links are degraded copies of the boundary registers, not independent copies of the charge — plus shared zero-mode noise. Light medium: buried by vacuum fluctuations. Heavy medium: too slow to spread. Every road closes.

## The structural statement

Permanence-grade redundancy requires branching environment geometry (d≥2). The framework's Lattice axiom supplies Z³ — exactly what the Record axiom's permanence clause demands under the QD reading. The quantitative bar location (theta*, delta-insensitivity, position vs the sparse window) is honestly deferred to a d≥2 comparator.

## Verify

```
[...]

PR #5090, opened 2026-07-09
[physics-loop] registration-bar-block02 — no activity, no opportunity: the necessity direction exhibited exactly

JONbridgerapps:
## What this PR contains

Block02 of the registration-bar campaign: the activity→information measurement.

- `scripts/activity_information_bound_2026_07_09.py` — runner, TOTAL: BOUND-PARTIAL (exit 1, honest)
- `docs/ACTIVITY_INFORMATION_BOUND_BOUNDED_NOTE_2026-07-09.md` — bounded note
- Runner cache

## Exhibited (the direction the gravity chain load-bears)

- **Stationary control exact**: unkicked ground state — no register information moves anywhere, ever (measured floor ~1e-15 against a 1e-8 gate).
- **Theorem layer clean**: every per-step information change respects the sharp Fannes-Audenaert envelope of the measured state change; zero violations. Composition stated as a conditional information-rate envelope, not oversold as a linear bound.
- **Excess information grows from exactly zero** after phase kicks (verified 1e-17).

Together: information reaches no register without activity — "records can form only where energy acts" now has its information-theoretic half exhibited with exact controls.

## Measured limits (findings, not gated away)

- **Pointer flow does not track activity linearly**: it rides bond coherence (a current *value*), while activity is state *change* — a steady current through a quiet bond is the measured counterexample (27 violations retained and interpreted in the SPEC-NOTE).
- **d=1 attribution saturates** (r = 0.63 vs the 0.8 gate): the same two-boundary Markov blanket that blocks redundancy (#5089) forces all exterior information through shared channels.

## Premise-ledger effect

The bridge premise weakens the way the sparsity premise did: necessity exhibited; what remains supplied is only the rate normalization — which is the *measured* kappa of the deposition-rate campaign, not an assumption. Proportionality deferred to d≥2 alongside the redundancy bar.

## Verify

```
python3 scripts/activity_information_bound_2026_07_09.py   # BOUND-PARTIAL, exit 1
diff <(python3 scripts/activity_information_bound_2026_07_09.py) logs/runner-cache/activity_information_bound_2026_07_09.txt
```

Sets no audit status; independent audit still required.


JONbridgerapps:
Review-loop disposition (GPT-5.6 Sol, xhigh): close without landing. The runner itself exits nonzero with BOUND-PARTIAL (checks 2, 4, and 5 fail, including 27 activity/TV violations), while the note promotes the finite comparator to a general necessity/envelope claim. It also inherits the unlanded registration stack. The head branch is intentionally retained for a future finite-comparator-only salvage.

PR #5163, opened 2026-07-11
science: permanence forces fresh-site double registration; persistence = agreement-conditioned survival

JONbridgerapps:
New bounded theorem note + runner (PASS=24 FAIL=0) exploiting the 2026-07-03/04 axiom restorations, which postdate the durability chain and its proposed R-D bridge.

- **T1 (axiom-forced):** "records are permanent" + "A site never carries more than one record" forbid same-site rewrite, so repeated lane registration is fresh-site double registration with full retention — the anatomy note's G2 geometry converted from modeling choice to forced structure.
- **T2:** given the supplied agreement-conditioned kernel + one NAMED new premise (epoch-independent lane readout), persistence ⇔ agreement-conditioned survival at a kernel fixed point. The premise's necessity is exhibited (permanent disagreeing records give a three-valued ratio readout {1/2, 1, 2/3}; additivity covers additive scalars only).
- **T3:** permanence forbids re-preparation; with multiplier 2 at the interior fixed point, a 1e-5 offset escapes in 14 steps (computed, second (ε,band) pair guards against hard-coding), so observed persistence at that precision implies exact fixed-point siting. At equipartition the per-step agreement-survival probability is 1/2 — persistence is survivorship, not stasis.

Discharge ledger: geometry DISCHARGED (axiom-forced); durable=fixed-point DISCHARGED conditional on kernel + named premise; epoch-independent readout NAMED (necessity exhibited); outcome factorization OPEN; bookkeeping multiplicity OPEN → κ lane (companion PR). Does not adopt R-D, derive the flow, fix r, or exclude ψ(r)=r².

Partially answers the 2026-07-10 audit demand on the durability chain ("supply a retained theorem deriving the R-D physical re-registration identification"); the chain note/runner are not edited here.

Execution disclosure: drafted by a Claude Opus 4.8 executor under Claude Fable 5 supervision (workhorse split; codex was usage-limited); supervisor reviewed line-by-line and re-ran the runner. Independent audit remains codex-owned; no status asserted.

### DIFF
--- a/docs/RECORD_PERMANENCE_FORCES_FRESH_SITE_DOUBLE_REGISTRATION_AND_AGREEMENT_SURVIVAL_BOUNDED_THEOREM_NOTE_2026-07-11.md
+++ b/docs/RECORD_PERMANENCE_FORCES_FRESH_SITE_DOUBLE_REGISTRATION_AND_AGREEMENT_SURVIVAL_BOUNDED_THEOREM_NOTE_2026-07-11.md
@@ -1,0 +1,364 @@
+# Record Permanence Forces Fresh-Site Double Registration, and Persistence Equals Agreement-Conditioned Survival (Bounded Theorem Note)
+
+**Date:** 2026-07-11
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Claim scope:** On admissible record histories under the four framework
+axioms, the Record permanence and one-record-per-site sentences make repeated
+registration of a lane quantity axiom-forced fresh-site double registration
+with full retention of every prior record (T1); and, given the supplied
+agreement-conditioned fresh-site kernel together with one NAMED new premise
+(epoch-independent lane readout), lane-value persistence is equivalent to
+agreement-conditioned survival of the registration history at a kernel fixed
+point (T2), with permanence forcing exactness once persistence is observed
+across the offset-escape window (T3). The note discharges the re-registration
+GEOMETRY half of the proposed R-D bridge and reduces the physical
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop landing completed on main at 8033cd8ee3354aef46d6fa06cbbc2a0ffd8a8ce7.

Full-depth review preserved the note filename and T1/T2 labels but narrowed the science. T1 is now a necessary fresh-site/retention result only in an explicit site-tagged monotone Record-history model; it does not prove the converse to full Admissibility or derive the physical self-composition/record-formation identification. T2 replaces the circular “every epoch has the same value” premise with a common epoch-comparable readout rule and retains only the exact conditional constant-orbit/fixed-point identity. T3 is renamed: the submitted “exactness from permanence” statement was false for finite observations and is withdrawn; exact q_n=q_0^(2^n) finite-horizon bounds plus a nonzero surviving-offset counterexample now replace it. The 1e-5 and 1e-8 pairs are labeled illustrative, not observational.

Validation: paired runner PASS=26 FAIL=0; independent 39-history enumeration, probability-map samples, mpmath escape calculations, finite-window counterexample, and pinch idempotence passed; No-Go Discipline N1-N8 passed for the narrow current-source boundary; vocab lint clean; audit pipeline exit 0; strict audit lint has no errors; generated audit/effective-status surfaces stripped. New bounded claim remains unaudited and queued. Downstream citations to the old T3 name or exactness conclusion must be rewritten; T1/T2 names and the note filename remain stable. This PR is closed rather than merged because its reviewed content landed through the review-loop mainline commit.

PR #5162, opened 2026-07-11
repair: orbit-occupancy moments derived honestly; r=1/2 relabeled as equipartition granularity (audit 2026-07-10)

JONbridgerapps:
Repairs the audited_failed finding (2026-07-10, codex gpt-5.6-sol, xhigh) on KOIDE_ORBIT_OCCUPANCY_INDEPENDENCE_AND_PREMISE_CANDIDATE_NOTE_2026-06-09 (load-bearing score 9.8, 27 transitive descendants).

**The finding is accepted and reproduced as a positive diagnostic**: the honest holomorphic Gaussian moment gives r=1 (normalization-independent; the hard-coded per-slot quantum is removed). Both witnesses are recharacterized as realized-state equipartition laws differing in exactly one named element — the granularity (per real mode ⟹ r=1; per outcome cell ⟹ r=1/2). Independence-by-exhibition survives, relabeled. The ρ-map r=1/(2ρ) and every "Z-ratio sets r-ratio" statement are withdrawn as contaminated. Witness premise-surface parity table (7 elements, 1 differing) is machine-checked per the re-audit demand, with both decisive toggles (normalization: r unchanged; granularity: r flips).

Runner: PASS=39 FAIL=0 (was 34 pre-repair).

**Inheritance flag for a follow-up repair (not edited here):** the T2 surface reproduces the withdrawn ρ-map arithmetic — scripts/frontier_koide_occupancy_locked_record_outcomes_2026_07_03.py lines 106–116 + CHECK 13, and KOIDE_OCCUPANCY_FROM_LOCKED_RECORD_OUTCOMES_BOUNDED_NOTE_2026-07-03.md lines 82–85. The T2 note's core slot-counting collision exhibit is a separate argument; only its r-attribution step is contaminated.

Execution disclosure: drafted by a Claude Opus 4.8 executor under Claude Fable 5 supervision (workhorse split; codex was usage-limited); supervisor reviewed line-by-line and re-ran the runner. Independent audit remains codex-owned; no status asserted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5169, opened 2026-07-11
repair: remove inherited rho-map r-attribution from the T2 locked-record-outcomes surface

JONbridgerapps:
Companion to PR #5162. The T2 surface reproduced the withdrawn ρ-map arithmetic in its "negative control": the runner's r_orbit/r_sector = 1/(2ρ) block and CHECK 13, and the note's "Z_sector/Z_orbit = 2, giving exactly r=1 ... r=1/2 through the landed rho arithmetic" passage.

Repair: the ρ-map block is deleted; CHECK 12 keeps Z_sector/Z_orbit = 2 as a true normalization/determinant-power fact now DECOUPLED from r (with an honest sympy moment check proving Z_d does not set r); CHECK 13 re-derives the two endpoints as exact solves of the per-real-mode (r=1) vs per-outcome-cell (r=1/2) equipartition laws. The core T1 K-real-section localization and the T2 slot-counting collision exhibit are byte-for-byte untouched — only the r-attribution step changed. Dated repair section quotes the 2026-07-10 audit finding verbatim.

Runner: PASS=13 FAIL=0 (rewritten in place, no renumbering). Cache refreshed and check-only clean.

Execution disclosure: drafted by a Claude Opus 4.8 executor under Claude Fable 5 supervision (workhorse split; codex was usage-limited); supervisor reviewed line-by-line and re-ran the runner. Independent audit remains codex-owned; no status asserted.

### DIFF
--- a/docs/KOIDE_OCCUPANCY_FROM_LOCKED_RECORD_OUTCOMES_BOUNDED_NOTE_2026-07-03.md
+++ b/docs/KOIDE_OCCUPANCY_FROM_LOCKED_RECORD_OUTCOMES_BOUNDED_NOTE_2026-07-03.md
@@ -10,10 +10,71 @@ zero on main. This note is banked only as bounded historical/supporting science
 for the flavor occupancy bridge; it does not reopen, modify, or supply
 authority for any admission or retirement record.
 
+**Repair update:** 2026-07-11 — an independent audit (2026-07-10) failed the
+companion orbit-occupancy note for an arithmetic error in its `rho`-map. This
+note **inherited** the same contaminated arithmetic in its T2 negative control
+(the Result T2 negative-control passage and runner CHECK 13). The `rho`-map r-attribution is
+withdrawn and replaced by the equipartition-granularity framing;
+`Z_sector/Z_orbit = 2` is retained as a normalization / determinant-power fact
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop landing completed on main at 2e692016649947bfe198bd019a84f4d8dff5e926.

Full-depth review narrowed the source beyond the submitted repair: the inherited rho-map attribution is withdrawn; same-density Cartesian and polar Gaussian calculations now agree at r=1; the factor-two partition cells are explicitly different-kernel/determinant-power support arithmetic; and the former Record collision is now a conditional slot-count mismatch under a supplied one-possibility/one-slot rule. An explicit Record-compatible two-readout countermodel prevents treating that rule as an axiom consequence.

Validation: paired runner PASS=15 FAIL=0; downstream source-pin runner PASS=16 FAIL=0; independent direct-determinant and mpmath Gaussian/kernel checks passed; No-Go Discipline N1-N8 passed for the narrow current-source nonselection claim; vocab lint clean; audit pipeline exit 0; strict audit lint has no errors; pipeline-generated audit/effective-status surfaces stripped. The bounded claim remains unaudited and queued for the independent audit worker. This PR is closed rather than merged because its reviewed content landed through the review-loop mainline commit.

PR #5193, opened 2026-07-11
science: records-only OS reconstruction — outcome-stage cell foreclosed; two-branch weight-reality set; strip positivity; non-degeneracy selects the tie (rhalf block 10)

JONbridgerapps:
## rhalf block 10 — records-only OS reconstruction on the untied first-order measure (BOUNDED_THEOREM)

**Stacked on #5171** (base = its head branch): consumes block 9's residual localization (the r=1/2 vs r=1 binary EQUALS the K-reality stage; spatial RP and theta eliminated as enforcers; time-direction OS the one survivor).

### Verdict: WEIGHT-STAGE REQUIRED — TWO-BRANCH

The emergent-time OS reflection θ conjugates the couplings (W → W†, verified at the field level), and the record subalgebra is P-even (P = the generation inversion implementing the K-orbit swap; P-odd data are not registrable at the registrable-readout note's orbit-clause grade). The records-only reconstruction Gram is therefore Hermitian **exactly on the union of two K-real branches**: the K-tied slice (a real, c = conj(b)) and the all-real branch (a, b, c real).

- **The outcome-stage cell is FORECLOSED** (block 9's fork, K acting only on registered outcomes): off the union the vacuum overlap Z and the registrable norm ⟨N,N⟩ are exactly complex — no inner product, no positive Hilbert space. K-reality must act ON THE WEIGHT.
- **Positivity discriminates the branches**: the tie is PD on its entire Hermitian domain (the crossing cures single-slice det W < 0), while the all-real branch is PD **exactly on the strip (Im λ₁)² < 1/8** — pair-block leading minors in closed form, discriminant argument, exact 5×5 certificates bracketing the boundary ((b−c)² = 4/25 vs 25/144 around 1/6).
- **Degeneracy selects the tie**: on the all-real branch the doublet eigenvalues are a conjugate pair, so every K-even per-member readout is doublet-degenerate (at most two distinct registered values). Under the named non-degeneracy element (three distinct registered values — a comparator/premise, labeled, never thresholded) the tie is the unique surviving branch.
- **The grain is count-twice — the r = 1 cell — on the whole weight-reality set** (tie: |b|² Wirtinger −3a; real branch: doublet weight = |λ₁|² identically). r = 1 itself is NOT derived: the per-cell equipartition/dial-point residual survives, named and untouched.

### Supervisor-review correction (disclosed in the note's Correction record)

[...]

PR #5223, opened 2026-07-12
science: equipartition endpoints are registration-asymmetric on the tied section — r=1 forced degenerate, r=1/2 hosts an open Q=2/3 sector (rhalf block 12)

JONbridgerapps:
## rhalf block 12 — equipartition endpoints are registration-asymmetric on the tied section (BOUNDED_THEOREM)

**Stacked on #5193** (base = block 10's branch). Origin: the 2026-07-12 exercise-skill run against the r=1/2 wall; this block hardens supervisor probe P2.

### Result

On block 10's K-tied circulant (`λ_k = a + 2|b|cos(θ + 2πk/3)`, real spectrum), exact positivity-window classification: a genuinely open nonnegative phase window exists **iff r < 1** (all phases valid for r ≤ 1/4; half-width `π/3 − arccos(1/(2√r))` for 1/4 < r < 1; zero width at r = 1). Consequences for the two landed fork endpoints:

- **Per-real-mode endpoint (r = 1): registration-incompatible on the positive branch.** Positivity forces the spectrum to exactly `(3a, 0, 0)` — the doublet degenerate at zero — violating the same named three-distinct-value element (ND₃) block 10's branch selection consumes.
- **Per-outcome-cell endpoint (r = 1/2): hosts an open registrable sector.** `0 < |θ| < π/12` gives three distinct positive values with `Q = 2/3` identically — an open sector, not a fitted phase.
- **Sign-allowed branch dichotomy:** allowing negative λ (registered m = λ², √m = |λ|) restores r = 1 non-degeneracy but unpins Q entirely (two exact points: `Q = 9/(13+4√3)` at θ = π/6 vs `Q = 9/(9+4√2)` at θ = π/12). So within the tied measure the per-real-mode law either violates ND₃ or pins no Q; **the per-cell law is the unique fork endpoint compatible with both a non-degenerate registered pattern and a pinned Q** — conditional on the named elements.

### Honesty boundary

The spectral identification `√m_k = λ_k` (B_sqrt) and the branch conventions (B_plus / B_abs) are this note's **own declared, unadopted** modeling elements — the R-D anatomy note was read and provably does *not* carry them (it carries projector bookkeeping, the agreement-conditioned r → 2r² flow, and the unresolved statistics atom). ND₃ is inherited at exactly its comparator/premise grade, labeled, never thresholded; no PDG values, no fitted phase consumed. **This does not derive r = 1/2** — the equipartition/dial residual survives; the theorem reshapes the endpoints' registration-compatibility only.

### Runner

`scripts/frontier_equipartition_endpoint_registration_asymmetry_2026_07_12.py` — exact sympy (trig identities via manifest nonnegative gap forms, no numerics on derivation paths). **PASS=23 FAIL=0**, exit 0, cache regenerated. Supervisor re-ran independently.

**Executor disclosure:** executor = codex gpt-5.6-sol (max reasoning) under supervisor review (Claude Fable 5: spec, probe verification, line-by-line review, independent re-run). No premise adopted; no audit status set. Independent audit required.

### DIFF
--- a/docs/KOIDE_EQUIPARTITION_ENDPOINT_REGISTRATION_ASYMMETRY_BOUNDED_THEOREM_NOTE_2026-07-12.md
+++ b/docs/KOIDE_EQUIPARTITION_ENDPOINT_REGISTRATION_ASYMMETRY_BOUNDED_THEOREM_NOTE_2026-07-12.md
@@ -1,0 +1,427 @@
+# Equipartition Endpoints Are Registration-Asymmetric on the Tied First-Order Section — the `r = 1` Positive Branch Is Forced to `(3a,0,0)`, While `r = 1/2` Has an Open Three-Distinct Positive Sector with `Q = 2/3`; Allowing Signs Restores `r = 1` Non-Degeneracy but Unpins `Q` (Bounded Theorem, rhalf block 12)
+
+**Date:** 2026-07-12
+**Claim type:** bounded_theorem (exact tied-section positivity-window and
+endpoint registration-compatibility asymmetry, conditional on explicitly
+named unadopted identification elements). This source note does not set or
+predict an audit outcome, adopt any premise, derive either equipartition
+endpoint, or edit any audit-lane-owned registry or data file.
+**Primary runner:**
+[`scripts/frontier_equipartition_endpoint_registration_asymmetry_2026_07_12.py`](../scripts/frontier_equipartition_endpoint_registration_asymmetry_2026_07_12.py)
+**Runner cache:**
+[`logs/runner-cache/frontier_equipartition_endpoint_registration_asymmetry_2026_07_12.txt`](../logs/runner-cache/frontier_equipartition_endpoint_registration_asymmetry_2026_07_12.txt)
+(SCORECARD: PASS=23, FAIL=0)
+
+> **CLAIMED (bounded and conditional):** on block 10's K-tied circulant,
+> the nonnegative-spectrum phase set has nonempty interior exactly for
+> `r < 1`. At the two landed fork endpoints, the positive branch is
+> registration-asymmetric: `r = 1` permits only the phase points with
+> spectrum `(3a,0,0)`, hence violates the same named three-distinct-value
+> comparator used in block 10, whereas `r = 1/2` has the nonempty open sector
+> `0 < |theta| < pi/12 (mod 2pi/3)` with three distinct positive values and
+> `Q = 2/3` identically. If signs are instead allowed and
+> `sqrt(m_k) = |lambda_k|`, `r = 1` regains non-degenerate patterns but `Q`
+> becomes phase-dependent. Every record-facing statement is conditional on
+> the named identification and branch elements below.
+>
+> **NOT CLAIMED:** a derivation or adoption of `r = 1/2`, a rejection of the
+> `r = 1` algebraic endpoint outside the named registration conditions, a
+> discharge of the equipartition/dial residual, an empirical comparison, a
+> premise promotion, or an audit-status change.
+
+## Role and exact boundary
+
+Block 9 landed the two equipartition cells without choosing between them:
+per-real-mode equipartition gives `r = 1`, while per-outcome-cell
+equipartition gives `r = 1/2`. Block 10 forces K-reality into the measure at
+its declared bounded grade and selects the K-tied branch over its all-real
+competitor only after consuming a **named non-degeneracy element**: the
+registered pattern has three distinct values. That element is a
+comparator/premise, labeled and never thresholded.
+
+This block asks the next exact question: once on that tied section, can both
+equipartition endpoints support the registered pattern that block 10's branch
+selection requires?
+
+Take
+
+```text
+a > 0,        b = |b| exp(i theta),        r = |b|^2/a^2 >= 0,
+lambda_k = a + 2|b| cos(theta + 2pi k/3),  k = 0,1,2.
+```
+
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Landed on `main` as 7b4107b782e84764745ddecf7d828d86b0d63a3b after two review-loop iterations.

Review result:

- Code / Runner: PASS — direct run `PASS=23 FAIL=0`, exit 0; cache SHA agrees with the landed runner.
- Independent math check: PASS — a separate `math`-based dense phase enumeration and direct invariant evaluation reproduced the positivity windows, trace identities, difference product, and both exact sign-branch `Q` values without using the SymPy runner implementation.
- Physics Claim Boundary: BOUNDED. The exact tied-section theorem is sound after separating `B_map: m_k=lambda_k^2` from the independent positive/sign-allowed branch choice.
- Imports / Support: DISCLOSED — `C_3` carrier and K-tied section at cited source grade; `B_map` and branch choice are unadopted modeling conditions; three-distinct-value condition is an exact qualitative comparator; no PDG value, fitted phase, probability law, or endpoint selector is imported.
- Nature Retention: BOUNDED — meets this repo's bounded-retention bar, not an unconditional derivation of `r=1/2`.
- Labeling Convention: NOT APPLICABLE.
- Repo Governance: PASS after replacing campaign/block labels and `origin/main:<path>` prose with scientific names and repo-relative links, and removing obsolete admission wording.
- Audit Compatibility: PASS — validation pipeline exit 0; strict lint has zero errors; the new row seeds as `bounded_theorem / unaudited / unaudited` with four real linked dependencies; all generated audit/effective-status surfaces were stripped before commit.

Fixed findings: (1) the submitted `sqrt(m)=lambda` element logically implied positivity and therefore was not independent of `B_plus`; it is now the squared spectral map `m=lambda^2`, with the branch as a separate numbered Residual Atom; (2) stale branch-path citation; (3) obsolete admission and incorrect “four bounded sources” language; (4) branch-local `rhalf block 12` / bare block naming. Residual Atom numbering and the filename were preserved.

No-Go Discipline gate: PASS for the narrow endpoint-registration boundary.

N1 — alternative routes (all ATTEMPTED in this review):

1. Phase representatives outside the displayed window: reduction modulo `2pi/3` plus reflection exhausts the spectrum up to permutation; the exact minimum envelope covers the full phase circle.
2. Semidefinite boundary rescue at `r=1`: the only nonnegative centers give `(3a,0,0)`, so retaining zero modes still violates the named three-distinct-value condition.
3. Sign-allowed rescue: it restores three distinct squared values, but the exact `theta=pi/6` and `theta=pi/12` witnesses give unequal `Q`; it therefore does not defeat the scoped dichotomy.
4. Drop the three-distinct-value condition: this is a genuine escape and is explicitly outside the conditional claim; the note does not foreclose it.
5. Change the spectral-to-mass map: a nonlinear or different K-even readout is a genuine untested carrier and is explicitly outside `B_map`; no universal exclusion is claimed.
6. Leave the tied `C_3` carrier or the two endpoint fork: other carriers/endpoints remain open and are explicitly outside scope.

N2 — wall/condition independence after the fix. Let `T` be the tied `C_3` domain, `M` the squared spectral map, `P` the positive branch, `D` three distinct registered values, and `E` the supplied `r=1` endpoint. For every pair `T-M`, `T-P`, `T-D`, `T-E`, `M-P`, `M-D`, `M-E`, `P-D`, `P-E`, and `D-E`, neither condition closes the other; all are independent. On the sign-allowed branch, replacing `P` by `A` has the same pairwise result. The original `sqrt(m)=lambda`/`P` dependence was removed rather than counted as an extra wall.

N3 — hidden-wall scan: “registered,” “positive branch,” “inherited,” and “modeling element” are all classified explicitly. Registration uses `M` plus the chosen branch; the K-tied section and `C_3` carrier are named source/domain conditions; no approved primitive supplies a mass map, branch, non-degeneracy condition, or equipartition law.

N4 — residual matching: the first-order-section fork supplies only the two conditional endpoints and tied spectrum; the records-only OS reconstruction supplies only the tied-section result and named non-degeneracy comparator; the minimal axioms supply no spectral bridge; the R-D anatomy supplies only projector/flow bookkeeping and its unresolved statistics atom. Each citation is used at that exact residual, with no cross-residual laundering.

N5 — rhetoric audit: the proof is per member of one three-eigenvalue tied circulant and at the two named fork endpoints. It does not reach other modes, carriers, matter actions, record maps, endpoint families, or lattice-wide dynamics. The note expressly rejects a universal exclusion of `r=1`.

N6 — partial-closure scan: changing the readout map, dropping the non-degeneracy comparator, changing carrier/section, or deriving a different formation/equipartition law are all live routes. None is mislabeled as requiring a new axiom. The registered primitives were checked and supply none of these structures.

N7 — steelman: a different K-even spectral readout or a different physical carrier could allow a non-degenerate `r=1` pattern and could conceivably pin `Q`; the framework does not rule that out. This objection defeats a global no-go but not the landed theorem, which is explicitly conditional on `T`, `M`, the branch, `D`, and the two-cell fork.

N8 — cross-cycle echo: the first-order-section fork already kept the equipartition selector open, and the OS reconstruction preserved an alternating-measure escape rather than asserting a framework-wide tie. The repo-wide wall search and physics-loop no-go ledgers revealed no later authority retiring the present readout/branch residual. The landed wording follows those precedents by keeping every broader route open.

No premise was adopted and no audit status was set. Independent audit is required for `koide_equipartition_endpoint_registration_asymmetry_bounded_theorem_note_2026-07-12`.

PR #5271, opened 2026-07-12
fix(structured-chokepoint): sync Born table to pinned cache + noise-floor framing

jonathonreilly:
## Verdict addressed

`structured_chokepoint_bridge_note` — `audited_conditional`, `runner_artifact_issue`
(fan=11):

> synchronize the three Born table entries with the SHA-pinned runner output,
> refresh the cache, and re-audit the exact finite card.

## What changed (note only)

The note's Born column showed `6.77e-16 / 8.14e-16 / 6.56e-16` (the 2026-04-04
dated log); the SHA-pinned committed cache holds `9.24e-16 / 7.34e-16 / 6.64e-16`.
The three entries are synced to the cache, and a one-line annotation records that
the retained content is the bound `< 1e-10`, not the last-place digits.

## Why the annotation (platform-stable-witness hardening)

Born `|I3|/P` is the machine-precision residual of an interference quantity that
is **mathematically exactly zero**, so its last-place digits float with the
platform's floating-point summation. Empirically, at the *identical* runner hash
`b99786e6` this machine reproduces the original `6.77e-16` while the committed
cache holds `9.24e-16`. Pinning exact noise digits guarantees re-drift on the next
platform; the audited claim boundary is already the bound (`Born below 1e-10`), so
the annotation aligns the table with the audited scope and defuses future
stale-table flags.

## What was deliberately NOT changed

- **Runner `scripts/structured_chokepoint_bridge.py`** — its stdout table is pinned
  by two consumers: `reproduction_audit_harness` (exact N=60 row, Born `[0-9.]+e-16`)
  and `canonical_regression_gate` (N=60 prefix + Born `< 1e-10` + `DECISION` line).
  Both verified green on this platform (**6/6 pins**). Changing the runner output
  would break them.
- **Cache `logs/runner-cache/structured_chokepoint_bridge.txt`** — already matches
  the pinned runner SHA `b99786e6`; the note is synced *to* it. Regenerating only
  swaps one platform's noise digits for another's, so it is left at the committed
  auditor-platform output (note now equals cache).

This is a note-only table-sync repair; runner and cache are unchanged by design.

### DIFF
--- a/docs/STRUCTURED_CHOKEPOINT_BRIDGE_NOTE.md
+++ b/docs/STRUCTURED_CHOKEPOINT_BRIDGE_NOTE.md
@@ -31,9 +31,14 @@ gravity across the narrow probe set.
 
 | N | `d_TV` | `pur_cl` | `S_norm` | gravity | Born `|I3|/P` | `k=0` |
 |---|---:|---:|---:|---:|---:|---:|
-| 25 | `0.9626` | `0.9473±0.00` | `1.0125` | `+6.4026±0.185` | `6.77e-16` | `0.00e+00` |
-| 40 | `0.8850` | `0.9491±0.01` | `0.9948` | `+7.4729±0.305` | `8.14e-16` | `0.00e+00` |
-| 60 | `0.6440` | `0.8030±0.04` | `1.0043` | `+5.7613±0.892` | `6.56e-16` | `0.00e+00` |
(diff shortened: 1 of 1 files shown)


jonathonreilly:
Review-loop PASS WITH BOUNDED CLAIMS. The note-only cache sync was independently verified and landed on current main as 69f0c58f0b, with narrow review fixes for portable evidence links and grade-neutral noise-floor wording. Generated audit outputs were stripped; the source-note hash drift leaves structured_chokepoint_bridge_note ready for independent re-audit. No audit verdict was applied here.

PR #5320, opened 2026-07-13
audit-infra: one audit attempt per claim per run + round-tagged artifacts

JONbridgerapps:
## What

Fixes the round-2 spin found by θ-canary iteration 4 — the first run under the #5311 calibration, where both seats' `audited_conditional` verdicts **applied successfully** (the calibration works), and the new failure mode surfaced immediately behind it:

1. A conditional verdict re-queues its row as `unaudited` (non-terminal, standing rule), so `compute_targets` re-listed the same claim in round 2 of the same run — re-auditing an unchanged claim toward the same verdict, burning two seats per round.
2. The relaunch crashed anyway: worker artifact names (`isolated-<key>-p<pass>`) don't include the round, so round 2 hit round 1's directories (`FileExistsError` → `worker_launch_failed` → exit 1).

## Fixes

- **One audit attempt per claim per run:** after each round's serialized apply, all batch claims join `session_skipped`. Repair moves a conditional row; same-run repetition doesn't.
- **Round-tagged artifacts** (`isolated-<key>-r<N>-p<M>`, same for raw/delivery/log): the `awaiting_second` resume path is a legitimate same-claim multi-round re-entry and would hit the same collision.
- **Friendly workdir guard:** a pre-existing `AUDIT_BATCH_WORKDIR` now prints an actionable refusal instead of a raw `FileExistsError` traceback (operator hit this live).

## Tests

Three regression tests in a new `BatchOrchestratorRoundSemanticsTest` class: the one-attempt pin (mocked main loop, row stays unaudited across rounds, exactly one launch), round-tag collision-freedom (two launches of the same seat in different rounds), and the workdir refusal message. Suite 301/301.

### DIFF
--- a/docs/audit/scripts/orchestrate_audit_batch.py
+++ b/docs/audit/scripts/orchestrate_audit_batch.py
@@ -209,6 +209,7 @@ def launch_worker(
     pass_no: int,
     workdir: Path,
     runner_timeout: int,
+    round_no: int = 1,
 ) -> dict:
     cid = row["claim_id"]
     key = artifact_key(cid)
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Landed on main after /review-loop at 663d3b5f02ddbe5a36016a7d71f23ff0c4ed2173. Review fixes: canonical open_gate test fixture and mocked-launch log-handle cleanup.

PR #5332, opened 2026-07-13
repair: sanitizer must not rewrite current-schema lifecycle_state enum values (audit infrastructure)

JONbridgerapps:
## Summary

Audit-infrastructure repair, routed through review per the audit-loop skill's no-smuggling rule.

**Defect (observed, blocking a live audit):** `sanitize_legacy_audit_artifacts.py` recursively maps the exact string `"unknown"` → `"unaudited"` in every string field of generated audit artifacts. The no-go discipline gate's authenticated evidence snapshots legitimately carry `lifecycle_state ∈ {active, retired, unknown}`; the blanket rewrite corrupts that enum inside `cross_cycle_candidates`, and the gate then rejects otherwise valid forensic packets. Observed on the G3 outcome-factorization no-go row: `apply_audit.py` accepted, pipeline invalidated with `no_go_discipline_packet_invalid`, validator error "lacks valid lifecycle_state". The full rejection record, manifest, and rejected JSON are preserved from the audit session.

**Fix (minimal, key-aware):** `sanitize_obj` now passes the enclosing key and preserves string values under `CURRENT_SCHEMA_ENUM_KEYS = {"lifecycle_state"}`. Legacy status vocabulary elsewhere is still rewritten (regression test asserts both directions).

**Test note:** the test module defines `SanitizeLegacyAuditArtifactsTest` twice; the second definition shadows the first, so the new test is added to the collectable (second) class. The shadowing itself is pre-existing and left for a separate cleanup. Full audit-pipeline test module: 309 tests, OK.

**Flagged for reviewer (not changed here):** the sanitizer's exact-status map is key-blind by design; other current-schema enums whose values collide with legacy vocabulary (e.g. a field legitimately valued `"open"`) would be corrupted the same way. If the reviewer prefers, the fix generalizes by extending `CURRENT_SCHEMA_ENUM_KEYS`.

No audit verdict, generated surface, or science content is touched.

Authored by Claude Fable 5 (supervisor) from the audit agent's preserved rejection record.

### DIFF
--- a/docs/audit/scripts/sanitize_legacy_audit_artifacts.py
+++ b/docs/audit/scripts/sanitize_legacy_audit_artifacts.py
@@ -37,6 +37,13 @@ TEXT_REPLACEMENTS = (
     ("proposed-promoted", "candidate promoted-grade"),
 )
 
+# Current-schema enum fields whose value sets legitimately contain strings that
+# collide with legacy status vocabulary. The no-go discipline gate requires
+# lifecycle_state in {active, retired, unknown}; rewriting its "unknown" to
+# "unaudited" corrupts authenticated evidence snapshots and makes the gate
+# reject otherwise valid packets.
+CURRENT_SCHEMA_ENUM_KEYS = frozenset({"lifecycle_state"})
+
 
 def sanitize_string(value: str) -> str:
     mapped = EXACT_STATUS_VALUE_MAP.get(value)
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Landed the reviewed audit-infrastructure repair on main as 34aa62498cef3d97dba14c679cd5df1f3812acee.

Review evidence:
- mechanically reproduced pre-fix corruption: lifecycle_state unknown -> unaudited
- mechanically reproduced post-fix behavior: lifecycle_state remains unknown while bare legacy audit_status unknown -> unaudited
- Python compilation passed
- full audit pipeline test module: 309 tests, OK
- full 16-stage validation pipeline: PASS, zero invalidations/re-audit hash changes
- strict audit lint: PASS (0 errors)
- vocabulary lint and git diff --check: PASS
- pipeline-output-stripped hard gate: PASS; no regenerated audit/status surfaces landed

The change remains the narrow two-file source/test repair. No audit verdict or science content was authored. The separately flagged generality risk and duplicate test-class shadowing remain pre-existing cleanup items, not blockers for this repair.

PR #5356, opened 2026-07-13
science-fix: cl3 exclusion runner N5 resolution sweep

JONbridgerapps:
Wave-3 seats agreed the exclusion science closes and both failed the same procedural gate: N5 requires `tested_resolutions` copied byte-for-byte from live runner stdout across the five canonical resolution classes per negative statement. The runner now emits that sweep for both authenticated negative statements — executed classes restate the computed exclusions (backed by the E2/E3 certificates via two new checks), and `lattice_wide` honestly reports checked-and-not-executed for a single-site algebra claim. 54/0; cache refreshed from real stdout; note updated. This sweep block is the template for every wall-naming clean-target note in the drain.


JONbridgerapps:
Review-loop PASS. The reviewed science-fix landed directly on main as 906a15c841c8ff0965231551f3999679d484527c. Runner/cache are deterministic at 54/0, the prior 52 checks are byte-identical, the executed N5 lines are E2/E3-certificate-backed, and lattice_wide is honestly checked but not executed for the single-site scope. Independent audit remains authoritative.

PR #5372, opened 2026-07-13
audit-infra: authenticate independent helper stdout

jonathonreilly:
## Summary
- execute explicitly marked N7 helper certificates live for source no-go artifacts
- authenticate independent stdout after a successful exit, with non-authenticating failure/suppressed roles
- bind virtual evidence paths to canonical helper-path digests
- reject absolute external helpers and symlink escapes without path disclosure

## Verification
- independent review-loop content/security review: PASS
- focused adversarial tests: 8/8
- full audit-pipeline suite: 353/353
- pipeline, strict lint, vocabulary, compile, and diff checks: PASS
- history-prune replay has identical stable patch ID 6f94720f756e9d4abb80f0369b59f32036146f26

No science, premise, status, verdict, import, comparator, or auditor-judgment fields change. The prescribed Sol/xhigh confirmation call was attempted but rejected before review by the account usage limit; do not merge until that confirmation is available.

### DIFF
--- a/docs/audit/scripts/no_go_discipline_gate.py
+++ b/docs/audit/scripts/no_go_discipline_gate.py
@@ -933,6 +933,30 @@ def _canonical_runner_path(repo_root: Path, raw: str | None) -> str:
     return str(raw)
 
 
+def _repo_local_helper_runner_path(
+    repo_root: Path,
+    raw: str | None,
+) -> str | None:
+    helper = _canonical_runner_path(repo_root, raw)
+    relative = Path(helper)
+    if (
+        relative.is_absolute()
+        or not relative.parts
+        or relative.parts[0] != "scripts"
+        or ".." in relative.parts
+    ):
+        return None
+    try:
+        root = repo_root.resolve()
+        resolved = (root / relative).resolve()
+        resolved.relative_to(root)
+    except (OSError, ValueError):
+        return None
+    if not resolved.is_file():
+        return None
+    return helper
+
+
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Landed via review-loop (cherry-pick to main, not merge). Reviewer: codex GPT-5.6-Sol at max reasoning, 3 rounds.

Round 1: 2 blocking findings — (1) helper containment escape (a lexically valid scripts/ symlink to an in-repo target passed both guards); (2) clipped independent stdout authenticated while omitted from the load-bearing clipped-evidence role sets. Both fixed (resolve under both repo root and resolved scripts/ root; add runner_stdout_independent to runner/batch-preflight/apply-time role sets).
Round 2: finding 1 confirmed CLOSED; finding 2 not fully closed — a DUPLICATE helper declaration left an authenticated runner_stdout_independent role beside a markerless failure tail, and _validate_n7 authenticated on mere role presence.
Round 3: PASS. Narrow fix — canonical-deduplicate helper_runner_paths before execution (render_prompt) so the second invocation never runs; and make _validate_n7 grant independent-execution authority only when runner_stdout_independent is present AND no failed/suppressed sibling role is on the same surface (defense-in-depth, confirmed to also contain a same-surface backslash-normalization escape that dedup alone misses). Added 2 regression tests. 156 targeted tests pass.

No audit verdicts landed; the affected no-go rows remain unaudited for the independent audit lane.
Landed as origin/main 5a8096a2aa3ef7eee48db569a2190ac8765ec445.

PR #5377, opened 2026-07-13
read-twice derive-first execution: R-READ decomposes to R-FORM + FRAME-EXT + PREP-FRAME (panel-repaired; nothing registered)

JONbridgerapps:
## What this executes

The 2026-07-13 handoff: handle "a record must be read twice" as a derive-first packet per P0–P4 — unified with the write-side packet ("written twice") from the same-day formation discussion. Owner constraint honored throughout: NO audit-lane work of any kind (no pipeline, no ledger/status/registry actions); artifacts kept light.

## Result, honestly stated after the blind panel

- **P1**: both slogans unpacked as four-clause packets (R-FORM write-side, R-READ read-side), neither registered.
- **P2 derive-first verdicts**: R-FORM (a) underivable extension; (b) underivable but motivated — single-register writes in the classified controlled-copy class are adjoint-revocable (proven exactly), two disjoint outcome-tied registers is the minimal locally-unrevocable configuration; the shared-decomposition requirement is NAMED clause content (panel finding — two admissible writes need not copy the same menu); (c) per-possibility counting underivable and load-bearing (the standing five-seat precedent that no-privilege sentences force no weightings); (d) compatibility obligations discharged by the runner.
- **R-READ adds no content of its own**: ontology reduces to R-FORM's; the discriminator is forced (the two reads are the two formation registers; later-tick re-reads are a corollary; conjugate structure belongs to the output); the pairing clause decomposes into **FRAME-EXT (already named) + Gleason (textbook import, dim ≥ 3) + PREP-FRAME (newly named, panel-forced): the frame state is the prepared record-bearing state**. Nothing is closed; nothing new needs registering under the slogan — registering "read twice" would double-name this decomposition.
- **The two-register coincidence identity** (exact, runner-verified): two classified writes give the fan-out; the agreement functional evaluates to |c_i|² — recorded as a CONSISTENCY IDENTITY (the evaluation uses the standard pairing; the panel's circularity finding is adopted verbatim into the claim), whose honest content is a conditional relocation: IF weights are frame-functional (FRAME-EXT + Gleason + PREP-FRAME), THEN weight = two-witness agreement statistics.
- **Gap accounting**: the formation event realizes a dim ≥ 3 composite CARRIER for FRAME-EXT's clause (4) — the carrier only; the projection-lattice domain and the state identification remain supplied. FINITE-ADDITIVITY-TO-FRAME stands, narrowed only in the carrier sense.

## Panel (per handoff P4, BEFORE this PR)

Three framework-blind physicist seats (soundness / hidden-assumption / falsifiability): all three UNSOUND against the pre-repair derivation claims, algebra assessed valid by all three. Every finding adopted — eight-edit repair pass demoting the claims to the versions above; full verdicts verbatim plus repair map in the campaign pack REVIEW_HISTORY. The strongest missing control the panel named (randomized two-record tomography over phase-sensitive contexts) is banked as the successor.

## Runner

`scripts/read_twice_packet_derive_first_unification_2026_07_13.py` — 81 exact sympy checks (verbatim needles; fan-out isometries; coincidence identity; register-basis invariance; the M₂ loophole frame function additive-yet-not-density witness; one-write adjoint revocability vs two-write local-undo failure witnesses; corrupted-witness controls; counting arithmetic w∈{1/3,½}↔r∈{1,½}; compatibility obligations). TOTAL: PASS=81 FAIL=0. Two supervisor-spec errors were corrected by the drafting worker with supervisor concurrence (recorded in REVIEW_HISTORY).

## What lands on the owner's open-items surface

The underivable physical clauses, priced: R-FORM-b (two independent same-menu registers) and R-FORM-c (per-possibility counting) for formation; plus PREP-FRAME and the pre-existing FRAME-EXT if the modulus-squared target is wanted. Placement per the handoff: registered premises when taken up — never an axiom edit from this lane.

## Boundaries

Conditional wording throughout; no registration performed; no axiom or registry edit; P-LINK-AVAIL decoupled (family resemblance noted only); sets no audit status; no audit-lane pipeline was run.

### DIFF
--- a/.claude/science/physics-loops/read-twice-unification-20260713/REVIEW_HISTORY.md
+++ b/.claude/science/physics-loops/read-twice-unification-20260713/REVIEW_HISTORY.md
@@ -1,0 +1,86 @@
+# Review History
+
+## Framework-blind physicist panel (3 seats, gpt-5.6-sol max, read-only; pre-PR per handoff P4)
+
+All three seats: UNSOUND against the pre-repair derivation claims;
+algebra assessed valid by all three. Verdicts verbatim below; all
+findings adopted into the note (8-edit repair pass, commit history).
+
+### Seat 1
+
+    UNSOUND
+    
+    1. The fan-out state is correct: two controlled copies of the same orthogonal pointer basis produce
+       \[
+       c_0|0r_0s_0\rangle+c_1|1r_1s_1\rangle .
+       \]
+       This does not clone an arbitrary quantum state. Conditioned on a projective outcome \(i\), the registers are in the product state \(|r_i\rangle|s_i\rangle\), so conditional independence is valid only in that branch/classical-outcome sense. Before conditioning, they are correlated parts of a GHZ-like state.
+    
+    2. The claimed derivation of \(|c_i|^2\) is circular. Duplication identifies an agreement projector such as
+       \[
+       Q_i=|r_i s_i\rangle\langle r_i s_i|,
+       \]
+       but does not assign it a numerical weight. Obtaining
+       \(\langle\Psi|Q_i|\Psi\rangle=|c_i|^2\) already uses the Hilbert-space expectation/Born pairing. Calling this number the “agreement functional” does not derive that pairing from two witnesses. FRAME-EXT plus Gleason can constrain a previously supplied frame measure to \(\operatorname{Tr}(\rho P)\), but R-FORM still does not determine that the relevant \(\rho\) is \(|\Psi\rangle\langle\Psi|\).
+    
+    3. Gleason cannot be applied to the realized agreement menu. Two writes realize only the correlated pointer subspace spanned by \(|r_0s_0\rangle\) and \(|r_1s_1\rangle\), not all projections or orthogonal decompositions of \(M_4\). Gleason requires a frame function defined consistently on the entire projection lattice, including entangled bases, disagreement sectors, and unrealized contexts. It is legitimate only if the full FRAME-EXT premise is invoked. In that case, however, the essential extension is still supplied by hand; the two registers provide merely a physical \(4\)-dimensional carrier. The modest claim that the carrier/domain has been identified is defensible, but the claim that the composite-domain requirement or pairing has been physically supplied is not.
+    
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Reviewed via review-loop (codex GPT-5.6-Sol, max reasoning, iteration 1). VERDICT: FAIL — closing as non-landable in current form. Branch kept as the recovery handle (content did not land). Scope reviewed: the durable slice (note + runner + cache); the two .claude/science campaign files were out of scope.

The derive-first framing and much of the algebra are high quality, but the note has five blocking issues plus a No-Go Discipline failure that together require author-scale correctness revision, not a review-level patch:

1. SEMANTIC_BRIDGE/OVERCLAIM — the conditional theorem drops a load-bearing premise: PREP-FRAME is required to identify Gleason's representer with the prepared record-bearing state, but P4 concludes |c_i|^2 and the Boundaries list omit PREP-FRAME (note lines ~91,127,138,155,193,216). The stated conditional understates its premises.
2. AUDIT_COMPATIBILITY — P2 (line ~76) labels the two principal consumed bounded notes 'retained theorems', but both current ledger rows are unaudited/effective-unaudited. They are landed, unaudited bounded source notes, not retained authority.
3. SEMANTIC_BRIDGE — R-FORM carries no formation criterion, dynamics, time metric, or rate, yet P4 (line ~203) says R-FORM alone makes a local formation frequency a well-defined dynamical variable; those inputs must stay explicit open conditions.
4. BUG — the corrupted-witness control constructs Psi_corrupt but evaluates the corrupted agreement projectors on the clean Psi (runner lines ~455,494,515; note ~258). On Psi_corrupt both outcome functionals deviate (6921/15625 and 10729/15625), so the outcome-0 'impossible' check tests the wrong state — the negative control is currently vacuous.
5. BUG — the blanket 'No check passes by literal stipulation' is false: B5 defines rho_pair from the target weights before recovering them, and B8 defines readout as a sum before checking additivity (runner ~350,531,536,590; cache ~72,107,116). These are consistency/bookkeeping checks, not independent discharges.

No-Go Discipline: FAIL. The uniqueness/necessity claims (R-FORM-b as the 'unique candidate'; the second register as 'necessary/load-bearing') overreach beyond the exhibited two-register, access-relative controls, and current-main fresh-site double-registration work is restricted to site-tagged monotone histories; FRAME-EXT already reaches M_4 without a second record. Required demotion: partial-narrowing of the uniqueness/necessity claims to the exhibited access-relative controls.

NIT: the backticked REVIEW_HISTORY.md pointer (line ~245) names stripped campaign material that will not land; remove or rephrase.

Resubmission path: complete the conditional-premise accounting (carry PREP-FRAME through P4/Boundaries), relabel the two dependencies as landed-unaudited, keep the formation-frequency inputs open, fix the corrupted-witness control to evaluate on Psi_corrupt (and regenerate the cache), replace the false 'no stipulation' blanket with an honest statement that B5/B8 are consistency checks, and narrow the necessity/uniqueness language to the access-relative controls. No audit verdict was applied; the row remains unaudited for the independent audit lane.

PR #5400, opened 2026-07-16
audit-infra: parenthesized redaction token + legend in effective-status views (fixes prose garble)

JONbridgerapps:
## What

The generated effective-status views neutralize author-side status words in rows citing non-retained authorities — by design, so an unratified row can never be read or scraped as retained prose. But the bare replacement token produced ungrammatical output (~89 cells like *"not unratified-source-label beyond it"*) that reads as tooling failure to an external reviewer.

- Token becomes the parenthesized `(unratified-source-label)` via a named `REDACTION_TOKEN` constant (safe in prose **and** inside markdown link labels, where a bracketed token would break rendering).
- Every generated view header gains a one-line **redaction legend** explaining the token and pointing at the `[audit:...]` badge.
- Invariant unchanged and still test-enforced: no \b(retained|promoted)\b survives in a neutralized row.

## Verification

- `python3 -m unittest docs.audit.scripts.tests.test_audit_pipeline` — **356 tests OK** (two literal expectations updated to the new token).
- Views regenerated locally: 47 tokens in CLAIMS view, legend present, prose now reads as deliberate redaction (*"not (unratified-source-label) beyond the O(alpha_s(v)^2) remainder"*). Generated outputs restored before commit — they land only via the audit lane's nightly refresh.

## Boundaries

Renderer + its tests only. No science, no status authority, no generated surfaces shipped.

### DIFF
--- a/docs/audit/scripts/render_publication_effective_status.py
+++ b/docs/audit/scripts/render_publication_effective_status.py
@@ -103,6 +103,14 @@ PROTECTED_INLINE_RE = re.compile(r"(\[[^\]]+\]\([^)]+\)|\[audit:[^\]]+\])")
 # Parenthesized so the redaction reads as a deliberate editorial token in both
 # label cells and running prose, and stays safe inside markdown link labels.
 REDACTION_TOKEN = "(unratified-source-label)"
+REDACTION_LEGEND = (
+    "**Redaction legend:** every row marked `AUDIT-NONRETAINED ROW` (a row citing "
+    "any non-retained authority, or citing no linked authority at all — the check "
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Landed on main via review-loop: sol-max review round 1 VERDICT PASS (2 non-blocking nits, both applied: fail-closed legend wording + legend regression test). Validation: 357-test suite OK, full audit pipeline exit 0, strict lint clean, generated outputs restored (regenerated views land via the nightly). Cherry-picked to main as d53a1cb18d + 88e1807f5f.

PR #5403, opened 2026-07-16
ci-prep: complete requirements-release.txt pin set (pyyaml)

JONbridgerapps:
## What

`requirements-release.txt` was missing `pyyaml` while the loose `requirements.txt` carries `pyyaml>=6.0`, so a CI switch to exact pins would silently drop a dependency of the repo's vocabulary/render tooling (`scripts/vocab_lint.py`, `scripts/migrate_link_aware.py`, `scripts/render_controlled_vocabulary.py` import yaml — the audit pipeline itself does not). Adds `pyyaml==6.0.3` for exact release-set parity.

## The rest of the CI change needs your push

The push credential lacks the `workflow` OAuth scope, so the two workflow-file changes are a `git am`-ready patch on branch **`handoff/ci-workflows-20260716`** (regenerated after sol-max review: `bash` + `pipefail` on every multi-command smoke step so a failing unittest can never be masked; honest pin rationale):

1. `audit.yml` installs `requirements-release.txt` (exact pins).
2. New `pr-smoke.yml`: per-PR byte-compile of changed `.py` + audit-tooling unit tests when `docs/audit/scripts` changed; read-only permissions; no auto-commit.

Apply from any environment with workflow scope:
```bash
git fetch origin handoff/ci-workflows-20260716:ci-handoff
git checkout main && git pull
git show ci-handoff:CI_WORKFLOWS_2026-07-16.patch | git am
git push
```


JONbridgerapps:
Landed on main as ab577b5ec7..85438719aa (4 commits) via review-loop: 5 sol-max rounds (R1 FAIL: piped-tail masking + false yaml rationale; R2 FAIL: patch contamination + trigger gap + stale copy-over template; R3 FAIL: errexit-suppressed detector + 196-line template drift + conflicting CI_INTEGRATION drop-in; R4 FAIL: diff-then-overwrite install block + stale PR-trigger prose + tracked review packet; R5 PASS with forced detector/compile/test failure probes all propagating). The workflow-file patch (v5) awaits owner application from branch handoff/ci-workflows-20260716 — requires 'workflow' OAuth scope.

PR #5418, opened 2026-07-16
audit-infra: design note — publication interleave lane, conditional blocker-delta consolidation, cycle-domain unification (spec-first, owner-gated)

JONbridgerapps:
## What

One Class-D design note (no implementation): `docs/audit/AUDIT_DISPATCH_RETARGET_CONDITIONAL_DELTA_AND_CYCLE_ENUMERATOR_NOTE_2026-07-16.md`, registered in `doc_authority_registry.json` as `in_flight_pr` with zero authority until owner-ratified transition.

- **Publication-priority interleave lane** — tracked reviewed manifest sidecar (frozen commit, admitted members, pending changes with objection window); banded interleave that can never lift a row across criticality or readiness; cycle admission via the canonical primary break-target list only; named nightly membership diffs. Gaming surface stated honestly: membership IS a priority signal, governed rather than denied.
- **Conditional blocker-delta consolidation** — the existing `awaiting_repair_since_conditional` dispatcher guard consolidated into one versioned `blocker_fingerprint_v1` (adds helper-runner hashes, cache/artifact/classifier state, policy versions, registry epoch), queue-level parked visibility, an unambiguous legacy-fail-open / v1-fail-loud validation matrix, targeting-stream union, and named coverage of BOTH dispatch consumers (codex_audit_runner + batch orchestrator). Explicitly a dispatch-policy amendment to the 2026-06-15 owner rule (eligible set = 46 conditional + 8 non-archived failed), requiring its own ratification entry.
- **Cycle-domain unification** — the 93-vs-63 divergence is node-domain selection (same walk: full citation graph vs gated ledger domain), so: one shared walk, domain-labeled counts on every surface, normalized-signature equality invariant, shadow → owner-logged switch → rollback soak → deletion.
- **Out of scope, named:** dual-family seating (separate owner-approval doc); verdict semantics, seats, criticality, scoring, premise registry; renaming/numerical-match dispatch eligibility.

## Review history

Sol-max round 1: FAIL, 14 findings (incl. two design-changing corrections: node-domain root cause; the pre-existing dispatcher guard) — full rewrite. Round 2: FAIL, 5 blocking (stock-vs-flow throughput error; superseded 68-row cohort; literal 1:1 merge provably violating queue ordering; consumer/lifecycle coverage; manifest-in-derived-cache) + 6 nits — all fixed. Round 3 running.

## Validation

vocab_lint clean; full pipeline exit 0 with stage-17 authority-link guard enforced at 0 violations; strict lint OK; document-authority companion runner green; registry change is a 7-line append.


JONbridgerapps:
Landed on main via review-loop: FOUR sol-max rounds. R1 FAIL (14 findings — two design-changing: the 93-vs-63 cycle divergence is node-domain selection not code drift; a conditional repetition guard already exists at the dispatch boundary) -> full rewrite. R2 FAIL (5 blocking: stock-vs-flow throughput error, superseded cohort, 1:1 merge provably violating queue ordering on the real queue, dispatch-consumer coverage, manifest-in-derived-cache). R3 FAIL (1 blocking: snapshot projection read the wrong lifecycle state — live rows carry top-level snapshots; archives only post-drift). R4 PASS, zero blocking. The note is a registered Class-D proposal: shadow-mode implementation may now be built; NO cutover flag may be enabled before its owner entry appears in the note's Ratification Log.

PR #5435, opened 2026-07-16
feat(flavor): reduce three candidate grain deciders to one count binary (bounded)

jonathonreilly:
## Summary
- Block 03 of the Koide grain-gate campaign (Target 1: `ac_orbit_occupancy_statistical_grain_derivation_obligation`, ledger `audited_renaming`).
- On the supplied three-carrier `C3` model (K-real circulant `A`, general complex circulant `Z`, Hermitian control `M`), the three candidate deciders for the occupancy grain — action reality class, `K`-CPT quotient-measure pushforward, and record-formation locking dictionary — are exact translations of one another through a single count binary `m ∈ {1,2}` on the free `K`-orbit, with dial `r = m/2`.
- Bounded structural reduction, NOT closure: neither `m` setting is forced; the note supplies the reduction, not the value. T2 shows any global determinant power is r-neutral; T5 is a prose corollary of T4 plus the obligation's quoted closure criterion, conditional on the physical action presenting one of the three guises.

## Changes
- `docs/ACPHILAMBDA_OCCUPANCY_GRAIN_THREE_CANDIDATE_DECIDERS_COMMON_COUNT_BINARY_REDUCTION_BOUNDED_THEOREM_NOTE_2026-07-16.md` — the bounded theorem note (T1–T4 computed, T5 prose corollary, N1–N4 controls, firewall + non-claims + honest auditor read).
- `scripts/acphilambda_grain_decider_reduction_2026_07_16.py` — paired runner, 86 exact symbolic gates: fully symbolic 3×3 realified-determinant identity, derived `K`-invariance measure forcing (with a falsifiability control showing the unconstrained measure is NOT invariant), translation gates wired to the realified determinant factorization via `Mul.make_args` (mutual-conjugacy gated), Hermitian-control gates proving `M` is not `K`-real and its orbit values are independent reals, quote gates against the obligation row and the 3A ledger `claim_scope`, note-hygiene scans (forbidden-phrase list includes closing/framing language).
- `logs/runner-cache/acphilambda_grain_decider_reduction_2026_07_16.txt` — fresh cache (`TOTAL: PASS=86 FAIL=0`, sha-pinned to the final runner).

## Pre-land review
Adversarial multi-lens panel ran before landing; all findings addressed:
- QUOTIENT_MEASURE tautology (single symbol for both orbit weights) → replaced by independent symbols + derived forcing via residual solve, with a negative control.
- Unbacked spectral inference in T1/N3 → replaced by definitional `K`-reality statements, runner-gated.
- Exact-point 3×3 determinant gate → fully symbolic (6 real symbols).
- TRANSLATION assembly-by-construction → wired to R3's computed factorization and R6's derived measure weights.
- T5 overclaim → rescoped as a conditional prose corollary of T4.
- Stale cache → regenerated after all edits.

## Validation
- Runner: `TOTAL: PASS=86 FAIL=0`, exit 0 (~3 s).
- `audit_lint.py --strict`: only the pre-existing `structured_mirror_reconciliation_note` note_hash mismatch on main (not introduced here; flagged for the audit lane).
- `run_pipeline.sh` run and stripped; pipeline-seeded ledger stubs removed; ships exactly 1 note + 1 runner + 1 cache.
- All five cited ledger rows re-verified live at commit time (obligation `audited_renaming`; 3A `retained`; Pfaffian power identity `retained`; 2A/2C `unaudited` — cited as context only).

### DIFF
--- a/docs/ACPHILAMBDA_OCCUPANCY_GRAIN_THREE_CANDIDATE_DECIDERS_COMMON_COUNT_BINARY_REDUCTION_BOUNDED_THEOREM_NOTE_2026-07-16.md
+++ b/docs/ACPHILAMBDA_OCCUPANCY_GRAIN_THREE_CANDIDATE_DECIDERS_COMMON_COUNT_BINARY_REDUCTION_BOUNDED_THEOREM_NOTE_2026-07-16.md
@@ -1,0 +1,358 @@
+# Occupancy-Grain Candidate-Decider Triad Reduces to One Count Binary on the Supplied C3 Model: Bounded Theorem
+
+claim_id: `acphilambda_occupancy_grain_three_candidate_deciders_common_count_binary_reduction_bounded_theorem_note_2026-07-16`
+
+**Date:** 2026-07-16
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Premise weight:** conditional. This note derives an exact structural reduction
+on a supplied `C3` model; it registers no primitive, adopts no convention, and
+selects no horn of the open occupancy-grain obligation.
+**Status authority:** independent audit lane only. This source note does not set
+or predict an audit outcome.
+**Primary runner:** [`scripts/acphilambda_grain_decider_reduction_2026_07_16.py`](../scripts/acphilambda_grain_decider_reduction_2026_07_16.py)
+**Cache:** [`logs/runner-cache/acphilambda_grain_decider_reduction_2026_07_16.txt`](../logs/runner-cache/acphilambda_grain_decider_reduction_2026_07_16.txt)
+
+## Purpose
+
+The open occupancy-grain obligation asks whether the physical charged-lepton
+matter action counts the `K`/CPT orbit once or twice. Three candidate deciders
+have been floated for that obligation:
+
+1. an **action reality class** (count-once `det_C`/holomorphic versus count-twice
+   `|det_C|^2`/realified);
+2. a **`K`/CPT quotient-measure pushforward** (orbit-sum pushforward versus
+   single-representative restriction on the sector set);
+3. a **record-formation locking-rule dictionary** (component completion versus
+   slot completion of the doublet outcome).
+
+This note proves, on a supplied minimal `C3` model, that all three candidate
+deciders are parameterized by one common count binary `m` in `{1, 2}`, that
+deciding any one fixes the other two through exact translations, and that the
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Review-loop verdict: REJECT; close without landing.

Blocking findings:

1. The action-reality leg is mathematically miswired. The runner counts top-level SymPy multiplication factors of the entire determinant and relabels that syntactic count as free-orbit W_d. On the submitted C3 carrier, det_C(A) = lambda_0 |lambda_1|^2 while |det_C(A)|^2 = lambda_0^2 |lambda_1|^4. Realification therefore doubles the fixed-sector and free-orbit powers together, which is r-neutral by the submitted T2. At (alpha,beta,gamma)=(2,1,0), the fixed/free factors are 3/3 for det=9 and 9/9 for |det|^2=81. This invalidates T3a, T4, T5, and the cached TRANSLATION PASS block.

2. The claimed exact triad translation is not derived. Wd_reality, Wd_measure, and Wd_dict are independently assigned the same keys {1,2}, then compared at matching keys. That proves a chosen common labeling, not an action-to-measure-to-Record binding; mixed assignments remain compatible with every local arithmetic check.

3. Record supplies formation occurrence, locking, permanence, and fixed scalar readout, but no formation rule, occupancy rule, K/CPT decomposition, weighting, normalization, or action/measure bridge. The note consumes MINIMAL_AXIOMS directly but omits its markdown dependency edge. The two no-go route notes used as load-bearing inputs remain unaudited.

4. The bounded negative boundary has no visible No-Go Discipline N1-N8 packet. The submitted N1-N4 are local faithfulness controls, not the required stress test; the missing independent binding wall makes the negative framing fail that gate.

Mechanical checks did pass: the runner reproduced 86 PASS / 0 FAIL, the cache SHA matches d5eb999db8f5016bab3b6bc3cc25f68904f2e07bf1d1d5c8c310a1616284f554, quote needles match the live sources/ledger, py_compile and diff checks pass, and the validation pipeline plus strict lint completed with no errors. Those green checks do not test the load-bearing bridge.

Salvage disposition: no distinct source slice remains after removing the false translation. The retained determinant-power theorem already states that its construction leaves r free; the formation note already carries the two lawful dictionary completions; and the landed sibling note already records only a set-level arithmetic correspondence while preserving the missing binding theorem. A corrected exact equivalence needs new binding science, not a prose fix.

No audit verdict was applied. The head branch is intentionally retained as the recovery handle for the unlanded work.

PR #5460, opened 2026-07-17
repo: stamp complete blocker_fingerprint_v1 snapshots on new conditional/failed verdicts

JONbridgerapps:
Implements the v1 snapshot-stamping phase of the dispatch-retarget design note ([AUDIT_DISPATCH_RETARGET_CONDITIONAL_DELTA_AND_CYCLE_ENUMERATOR_NOTE_2026-07-16.md](docs/audit/AUDIT_DISPATCH_RETARGET_CONDITIONAL_DELTA_AND_CYCLE_ENUMERATOR_NOTE_2026-07-16.md), section 3b version matrix): "its next verdict stamps a complete v1 snapshot."

## What

`apply_audit.snapshot_audit_state` now stamps a complete, versioned `blocker_fingerprint_v1` snapshot whenever the verdict being applied is `audited_conditional` or `audited_failed` — on both the ordinary (`apply_one`) and judicial (`apply_judicial_review`) paths, which both set `row["audit_status"]` before snapshotting. On top of the existing dep/helper baselines (`dep_effective_status`, `dep_claim_type`, `dep_claim_scope`, `dep_axiom_premise_note_hash`, `helper_runner_hashes` — these already carry dep/helper membership), the stamp adds:

- `schema: blocker_fingerprint_v1` (marker imported from `compute_audit_queue`, never a second literal),
- `runner_cache_state`: per-path SHA-pinned cache header fields plus freshness for the row's runner and helpers, read from `logs/runner-cache` during this apply pass,
- `artifact_classifier_state`: the static classifier view of the runner, recomputed from the runner source via `classify_runner_passes` routines at write time — never read from the possibly-stale generated `runner_classification.json`,
- `policy_versions`: tuple-agreement schema, N8 source-corpus version, and content hashes of the no-go gate module and the audit prompt template,
- `premise_registry_epoch`: content hash of `axiom_premise_nodes.json` (the registry carries no counter; the content hash changes exactly when the registry changes and never otherwise).

The writer then validates its own stamp against the comparator's `FINGERPRINT_V1_REQUIRED_KEYS` table (imported, so writer and comparator cannot drift) and raises `FingerprintV1Invalid` BEFORE the row is committed — an incomplete or ill-typed v1 stamp is a bug and fails loudly and transactionally; it never lands.

## Reporting-side only — no dispatch behavior change

- No dispatch consumer reads the new snapshot keys. `orchestrate_audit_batch.awaiting_repair_since_conditional`, `invalidate_stale_audits`, `compute_reaudit_candidates`, `restore_overaggressively_invalidated_audits`, `audit_lint`, and the seeder all read specific legacy keys and ignore extras (checked per consumer site).
- The only reader of the v1 fields is the shadow would-park comparator `compute_audit_queue._live_conditional_would_park`, whose output is the `would_park`/`would_park_reason` reporting tags and the front-door shadow counts — zero dispatch effect. Queue membership and ordering are untouched (the landed `GoldenQueueInvarianceTest` pins this).
- The design note's cutover flags (`conditional_delta_gate`, `publication_lane_interleave`) do not exist and are not created here; no Ratification Log entry is spent.

## No backfill

Existing ledger rows are not rewritten. A legacy snapshot honestly records that the complete channel set was NOT captured at verdict time; forging one after the fact would fake "when the blocker was recorded" provenance. Legacy rows stay fail-open in the comparator (the version matrix's legacy branch) until organically re-audited, at which point their new verdict stamps.

## Determinism

Same inputs produce a byte-identical stamp: no timestamps beyond the `audit_date` the verdict already writes, sorted path/dep iteration, content hashes only. Pinned by a two-fresh-fixtures byte-equality test.

## Honest expected effect

- The shadow lane's `shadow_conditional_fail_open_count` decays organically as conditional/failed rows are re-audited; it does not drop at once.
- For newly stamped rows, would-park is real immediately on the channels whose current state the comparator already reads: dep membership, dep effective-status/type/scope, axiom-premise note hash, and helper membership.
- The four opaque channels (`runner_cache_state`, `artifact_classifier_state`, `policy_versions`, `premise_registry_epoch`) VALUES and helper-hash values are recorded as baselines now but compare as unchanged until a later phase populates the row-side `*_current` projections (the comparator's documented absent-projection semantics; no producer for those projections exists yet). Primary/helper runner PATH membership is compared live from the row (helper membership was already compared; review iteration 1 added the primary-path comparison via the cache-state baseline's key set). In-place runner/helper hash drift is meanwhile caught by the existing invalidation lane, which archives and resets such rows out of the live-conditional population — including, after review iteration 1, primary-hash presence transitions (file deleted in place, no-path -> path) for v1-marked snapshots specifically; legacy snapshots keep the conservative both-present rule because many predate the `runner_hash` field.

## Tests

`docs/audit/scripts/tests/test_fingerprint_v1_stamping.py` (26 tests after review iteration 1, driving the real writer end to end): stamped conditional and failed rows are v1-complete and PARK under unchanged current state (no fail-open reason, no unpark); every required-key omission AND ill-typing raises `FingerprintV1Invalid` in the comparator; stamp-then-mutate-one-channel un-parks with the named reason for every compared channel (dep status/type/scope/axiom-hash, dep membership, helper membership, helper hash, all four opaque channels); writer fail-loud is transactional (no verdict committed); clean verdicts stay legacy-shaped; pre-existing legacy rows stay fail-open and untouched; byte-identical determinism across identical passes.

Mutation-verified decisive: (i) stamping on all verdicts, (ii) wrong schema marker, (iii) non-deterministic channel value, (iv) dropped channel with the writer self-check disabled — each mutation fails the suite (1, 11, 2, and 12 test failures respectively).

## Validation

- `python3 -m unittest discover -s docs/audit/scripts/tests`: 467 tests, OK
- `bash docs/audit/scripts/run_pipeline.sh`: exit 0
- `python3 docs/audit/scripts/audit_lint.py --strict`: exit 0, "OK: no errors" (the previously-known `structured_mirror_reconciliation_note` hash-mismatch error no longer fires on current main; nothing new introduced)
- Generated outputs restored before commit; only the two source files ship (no `dispatch_shadow_state.json`, no ledger data).

## Review-loop update (2026-07-18, iteration 1)

Adversarial round 1 (codex gpt-5.6-sol, max reasoning) returned 2 BLOCKING + 4 MAJOR findings; all are repaired in the follow-up commit `fix: address physics review findings (iteration 1)`:

- BLOCKING: `bool` passed the epoch's `(int, str)` contract (Python bool subclasses int), so an ill-typed v1 stamp could commit and PARK. The validation predicate now lives in ONE shared function, `compute_audit_queue.fingerprint_v1_problems` (bool rejected explicitly), imported and called by the writer — table AND predicate are shared, so the sides structurally cannot drift. Ordinary/judicial/comparator regressions added.
- BLOCKING: primary-runner presence transitions (runner file deleted in place; row stamped runner-less later gaining a path) produced false shadow PARKs. Fixed v1-scoped in both lanes: the comparator compares the cache-state baseline's key set against the row's current runner+helper paths (`runner_cache_membership_changed`), and `invalidate_stale_audits` treats primary-hash presence transitions as `runner_hash_changed:...->missing` / `missing->...` drift for `blocker_fingerprint_v1` snapshots only. The v1 scoping is load-bearing: 78 live legacy audited rows have snapshots predating the `runner_hash` field and must not be archive-reset; a regression pins that conservative branch, and the live ledger carries zero v1 rows today, so no existing row changes.
- MAJOR x3: decisive-assertion gaps (stamped `runner_hash`, helper hashes, helper cache entry, full `policy_versions`, exact classifier state were unpinned; several movement tests were self-confirming). Every channel value is now pinned against independently recomputed hashes/constants/literals; round 1's four surviving mutations now each fail the suite.
- MAJOR: the comparator comment claiming this phase populates the `*_current` projections was corrected (no producer exists; later projection phase).

Fix-commit scope: `compute_audit_queue.py` (shared predicate + membership comparison + comment), `apply_audit.py` (use shared predicate), `invalidate_stale_audits.py` (v1-scoped presence rule), and the test file (16 -> 26 tests). Full suite 477 tests OK; pipeline exit 0; strict lint "OK: no errors".

### DIFF
--- a/docs/audit/scripts/apply_audit.py
+++ b/docs/audit/scripts/apply_audit.py
@@ -42,9 +42,18 @@ import audit_invocation
 
 import ledger_io
 
+# Shared constants/validation for the blocker_fingerprint_v1 stamp: the
+# writer imports the comparator's required-keys table so the two can never
+# drift (dispatch-retarget design note, 2026-07-16, section 3b).
+import classify_runner_passes
+import compute_audit_queue
+
 REPO_ROOT = Path(__file__).resolve().parents[3]
 LEDGER_PATH = REPO_ROOT / "docs" / "audit" / "data" / "audit_ledger.json"
 AXIOM_PREMISE_NODES_PATH = REPO_ROOT / "docs" / "audit" / "data" / "axiom_premise_nodes.json"
+
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+import runner_cache  # noqa: E402  (SHA-pinned runner cache headers)
 
 _AXIOM_PREMISE_IDS: set[str] | None = None
 CLIPPED_EVIDENCE_MARKERS = (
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Review-loop provenance: reviewed immutable original head `684574b1acc42c662b3155eb0d695692c7f5baf0`; initial exact reviewer verdict was FAIL (F1–F4). Narrow fixes were independently rechecked by the same reviewer, which returned exact `FINAL VERDICT: PASS`. Landed to `main` as `511b970a23` + `dc23debf10` after reconciling onto the then-current tip. Independent validation: 47 focused tests, 33 ApplyAudit tests, full 532-test discovery, all 18 pipeline stages, and strict audit lint with no errors. Pipeline outputs were validation-only and discarded; no audit verdicts were applied.

PR #5464, opened 2026-07-17
authoring gates: directory-target rule, class-F header check, citation-graph delta manifest

JONbridgerapps:
Machinery PR (one concern: close the mechanical finding classes from the front-door review campaign, so they are caught by stage 18 instead of reviewers).

**Three new checks in repo_invariants_check (enforced under --enforce-links, warn-only otherwise):**
1. **directory-target**: authority-surface markdown links must dereference to one reviewable tracked file. All 13 in-repo violators fixed (code-span + file-link pattern; the FRONT_DOOR_STATUS line is a surgical one-line sync byte-identical to the updated renderer's output).
2. **class-F header formula**: registered class-F docs must carry both policy phrases; the GRADED_CONSTRAINT memo was missing the citable-scope sentence and gains it (mechanical policy sync).
3. **citation-graph delta gate**: new pipeline stage 1b writes a tracked per-node manifest (out-degree + deps hash); stage 18 fails when the in-pass graph differs from the INDEX manifest, naming added/removed/rewired nodes and the one-command remedy. New graph topology now lands as a visible acknowledged diff instead of silently reshaping audit-lane criticality (basis: FRESH_LOOK_REQUIREMENTS §4 — criticality is graph topology only). Initial materialization: 3,961 nodes / 14,464 edges. NOTE for in-flight lanes: a landing that adds notes will see the stage-18 remedy message — run write_citation_graph_manifest.py and stage the manifest.

**Tests**: 7 new (manifest determinism + rewiring hashes, delta naming, class-F pure core, directory-target integration). Full pipeline exit 0: link_violations=0 class_f_violations=0 graph_delta=acknowledged.

Companion to #5463 (author pre-flight checklist, the judgment-side of the same campaign).

Review gate: sol-max review-loop round to follow before landing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5468, opened 2026-07-17
invariants: structured diagnostic for undecodable class-F registry (NIT from #5464)

JONbridgerapps:
One-line exception-tuple fix + one test. collect_class_f_violations catches UnicodeDecodeError like the file's sibling readers, so a corrupt registry yields the structured "registry unreadable" violation instead of a raw traceback. No behavior change on valid input; was already fail-nonzero (cannot false-PASS). Full suite green; 18-stage pipeline exit 0 with stage 18 PASS (link_violations=0 class_f_violations=0 graph_delta=acknowledged).

Review gate: sol round to follow before landing.

### DIFF
--- a/docs/audit/scripts/repo_invariants_check.py
+++ b/docs/audit/scripts/repo_invariants_check.py
@@ -686,7 +686,7 @@ def collect_class_f_violations(tracked: list) -> list:
     try:
         with open(registry_path, "r", encoding="utf-8") as fh:
             registry = json.load(fh)
-    except (OSError, json.JSONDecodeError) as exc:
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
Landed on main (invariants: class-F registry decode-error diagnostic). One sol-max round: PASS — reviewer reproduced the pre-fix raw UnicodeDecodeError on origin/main's version, confirmed the branch returns the structured violation, verified byte-identical behavior on valid input, ran the suite + enforced invariants check, and mutation-verified the new test fails when the tuple is reverted.

PR #5506, opened 2026-07-18
review-loop: parallel-by-default entry (any orchestrator; serial landings)

JONbridgerapps:
Companion to #5504's audit-loop default entry, per owner direction: review-loop invoked against a set of open PRs now defaults to parallel per-PR reviewer processes (one isolated worktree each; stacks bottom-up, never parallel within a stack), strictly serialized fixes and landings via the fail-closed loop, and the shared-pool concurrency budget with the dated collapse measurement. Any orchestrator (any Claude tier, codex, human) may start it: findings and PASS/FAIL come only from the configured reviewer model, the orchestrator never self-certifies, and reviewer-family substitution remains per-episode owner-authorized with provenance disclosure. Also fixes the stale GPT-5.5 parenthetical.

Docs-only (one file). Review gate: sol xhigh round (queued behind #5504's round per the one-slot throttle).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5504, opened 2026-07-18
audit-unlock: seat-blocked judicial freeze + canonical fan-out skill section

JONbridgerapps:
Owner-directed unlock of the audit lane, REWORKED per owner: seat-blocked judicial rows are finished, not frozen, and the audit-loop skill drains in parallel by default wherever it is started.

**Reseat (replaces the withdrawn freeze design)**: a disagreement whose recorded seats lack invocation-bound rationales can never finish through a panel as recorded. The panel orchestrator now, by default, archives those seats verbatim into previous_audits (sanctioned archive shape, recorded reason, nothing deleted, nothing minted) and reopens the row as unaudited — fresh two-seat cross-confirmation under the current rationale-preserving contract then either lands the row on agreement or reaches a validly-seated panel on disagreement. Either way it finishes. Reseats go through the identical per-claim gate ladder as verdict applies. --no-reseat reports instead. The two known rows (cluster_decomposition_mass_gap_bridge 2026-05-09, s3_mass_matrix_conditional_degeneracy 2026-07-11) are the initial reseat queue; live dry-run shows exactly them.

**Skill: parallel by default**: new 'Default Entry' section — invoking audit-loop without a named claim = setup, panel orchestrator (reseats+panels), then the canonical drainer looped at --max-workers 4 (budget ≤8–10 total codex processes, dated measurement cited); any orchestrator may start it because it is process-only (all verdicts from sol seats at xhigh exactly); forensic tier stays a single canary; apply-gate rejections recorded as selection-filter data.

Validation: 5 new tests OK; dry-run targets 0 / reseat queue 2 / no writes; vocab lint clean; 18-stage pipeline exit 0 (stage 18 PASS, graph delta acknowledged for already-landed main content — this PR adds no graph nodes).

Review gate: one sol xhigh round on this reworked head before landing (the round on the withdrawn freeze design was stopped and discarded).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

**Addendum (owner-directed): 15-minute progress summaries.** The batch drainer now emits an interpretation-free `== drain summary` block at most every 15 minutes (elapsed, outcome counts from the session report, live seats with runtimes, dep-ready remaining; forced baseline at round 1 and forced final at session end; console-only, cannot perturb tracked artifacts; 3 cadence/content tests + live smoke). The audit-loop skill gains the matching reporting contract (relay drainer blocks verbatim; orchestrator composes the same cadence for panel/reseat/canary phases; forced final summary even on early stop).

PR #5513, opened 2026-07-18
coexistence: one audit-lane orchestrator per repo (machine-local lock) + skill contract

JONbridgerapps:
Final piece of the parallel-kickoff chain (after #5504 and #5506, both landed). Verdict/reseat commits ship regenerated audit surfaces, so competing audit-lane orchestrators race every push at multi-minute pipeline-replay cost — parallelism belongs in the seats inside ONE orchestrator. The drainer and panel orchestrator now take an advisory machine-local flock keyed to the repository path (skipped on --dry-run, auto-released on exit); a second instance exits 3 with join-the-seats guidance. The audit-loop skill gains the coexistence contract: one orchestrator per repo; audit-vs-review race asymmetry (audit never waits, review workers retry in seconds through their own fail-closed loops); manifest-only conflicts regenerate, never hand-merge. 2 lock tests (second-acquire refused until release, live refusal message verified; repo-keyed path outside the repo).

Review gate: one sol xhigh round before landing.

### DIFF
--- a/docs/ai_methodology/skills/audit-loop/SKILL.md
+++ b/docs/ai_methodology/skills/audit-loop/SKILL.md
@@ -314,10 +314,13 @@ failure exits nonzero.
 
 **Parallel coexistence with review-loop.** The audit lane and review-loop
 sessions may run at the same time by design; the rules that keep concurrent
-landings cheap are: (1) exactly ONE audit-lane orchestrator per repository —
+landings cheap are: (1) exactly ONE audit-lane orchestrator per clone —
 the drainer and the panel orchestrator enforce this with a machine-local
-exclusive lock and exit with guidance when another instance holds it
-(parallelism belongs in the seats inside the one instance); (2) audit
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Landed on main (2 commits). Sol xhigh rounds: R1 FAIL — the lock keyed the worktree root, so two worktrees of one clone held different locks (found by live two-worktree probe; exactly our operating topology); R2 confirmation PASS — lock keyed to the git common directory (per-clone across all worktrees; independent clones stated as out of the lock's scope, coordinated by the coexistence contract and budget), release-on-close and release-on-SIGKILL replayed, hermetic two-worktree regression test refuses correctly, dry-runs never touch the lock, and the audit-loop skill paragraph matches behavior exactly. With this, the parallel kickoff surface is complete: one audit-lane orchestrator per clone (seats scale inside it), review-loop workers land end-to-end with push atomicity as the only serializer, and manifest-only conflicts self-heal by regeneration.

PR #5525, opened 2026-07-19
skills: unspecified invocation = parallel backlog drain (review-loop + audit-loop)

JONbridgerapps:
Owner-directed semantics: a bare, argument-less invocation of either loop starts its parallel backlog drain (review-loop: open non-draft PRs, one worker per PR, workers land end to end; audit-loop: the canonical panel+drainer sequence). Naming a branch/PR/claim remains the focused special case. Three files: the review-loop command wrapper (whose review-only-branch-changes rule is now scoped to focused invocations) and both SKILL.md Default Entries.

Review gate: one sol xhigh round, then lands.

### DIFF
--- a/.claude/commands/review-loop.md
+++ b/.claude/commands/review-loop.md
@@ -6,20 +6,25 @@ Run the repo-native physics review loop from:
 
 ## Invocation
 
-`/review-loop [focus] [--max-iterations N] [--no-fix] [--no-commit]`
+`/review-loop [target] [focus] [--max-iterations N] [--no-fix] [--no-commit]`
 
-An UNSPECIFIED invocation (`/review-loop` with no focus and no named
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Landed on main (3 commits). Sol xhigh rounds: R1 FAIL (focus/target conflation; under-specified no-target flag forms); R2 FAIL (skill Arguments block lacked the target operand; flag rule differed across files — the reviewer's own invocation table became the spec); R3 PASS — every invocation form now has one identical interpretation in wrapper and skill: zero arguments = parallel backlog drain; named branch/PR target = focused mode; focus-text-only = drain with that emphasis in every slot; ANY flag without a named target (incl. --max-iterations) = invalid, stop-and-ask; target resolution stated before focus parsing. Bare /review-loop and bare /audit-loop are now the parallel backlog drains by definition.

PR #5528, opened 2026-07-19
review-loop: proactive citation-manifest co-landing rule

JONbridgerapps:
Process finding from the defect-lane report: two lanes independently tripped the enforced stage-18 guard by landing dep-changing commits without the refreshed manifest, blocking main's pipeline runs until someone landed the acknowledgment. One-paragraph skill addition making the co-landing rule proactive (landing set includes the refreshed manifest whenever landed commits change graph deps), complementing the existing conflict-time regeneration rule. Docs-only, one file.

Review gate: one sol xhigh round, then lands.

### DIFF
--- a/docs/ai_methodology/skills/review-loop/SKILL.md
+++ b/docs/ai_methodology/skills/review-loop/SKILL.md
@@ -893,7 +893,10 @@ title/body names the row or quotes its audit repair target):
    a sibling-runner pin sweep was done before every note edit (grep
    `scripts/` for runners pinning the edited sentences); no authored grade
    language anywhere; no `docs/audit/data/` content beyond
-   dispatcher-sidecar targeting metadata.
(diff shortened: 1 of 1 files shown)


JONbridgerapps:
Landed on main (3 commits). Sol xhigh rounds: R1 FAIL (the hard pipeline-output gate's drop-everything rule contradicted the new proactive co-landing rule); R2 FAIL (comment carved the exception but the operative BLOCK-PASS trigger and cleanup commands didn't — the allowed line itself blocked PASS and the cleanup erased the acknowledgment); R3 PASS — comment, decision rule, and cleanup now implement one consistent exception, with the drop sequence deterministically regenerating and re-staging the manifest exactly when the landing changes graph deps; both cases walked through the literal prose. Closes the trap two lanes hit independently: stage-18 demanding the manifest while the landing gate stripped it.

PR #5553, opened 2026-07-21
Fix audit-loop row reentry across batch cycles

JONbridgerapps:
## What changed

- persist post-verdict blocked-row reentries in the top-level campaign exclusion JSONL
- pass the exclusion across every inner batch and configured lane
- distinguish blocked-row reentries from schema-invalid quarantines in progress and final summaries
- preserve legitimate `audit_in_progress` / `awaiting_second` resumptions
- update the audit-loop skill contract and add regression coverage

## Root cause

The inner batch kept a processed-row skip set only in memory. After a verdict landed, the top-level supervisor launched another batch process; if the pipeline had immediately reset that same row to `unaudited` and dep-ready, the new process selected it again. The campaign could therefore spend every cycle re-auditing one unchanged row instead of draining the remaining lanes.

## Impact

An accepted verdict that immediately re-enters canonical development-tier selection is now excluded for the rest of that campaign with its exact invalidation reason. The supervisor continues through every other eligible row and reaches a fixed point without repeatedly burning auditor seats on the blocked row.

## Validation

- `python3 -m unittest discover -s docs/audit/scripts/tests -p 'test_*.py'` — 569 tests passed
- `bash docs/audit/scripts/run_pipeline.sh` — all 18 stages passed
- `python3 docs/audit/scripts/audit_lint.py --strict` — no errors
- `python3 scripts/vocab_lint.py --fix <changed files>` — no violations
- `python3 .../skill-creator/scripts/quick_validate.py docs/ai_methodology/skills/audit-loop` — valid
- `git diff --check`

### DIFF
--- a/docs/ai_methodology/skills/audit-loop/SKILL.md
+++ b/docs/ai_methodology/skills/audit-loop/SKILL.md
@@ -414,8 +414,10 @@ The graph-cycle warning is currently expected. Treat any error as a blocker.
 
 ## Blocked-Row Loop Guard
 
-- If applying a verdict and rerunning the pipeline immediately invalidates that same row, returns it to the ready queue, or creates a dependency-status cycle that cannot be resolved by the audit verdict alone, do not keep retrying the row in the same audit loop.
-- Restore the pre-claim generated audit diff for that row, record a session-local blocked/skip entry with the claim id and the exact tooling reason, and continue with the next ready row.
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Review-loop complete: iteration 1 found one blocking regression-coverage gap; commit b72847585 added a true two-batch production-path test. Fresh GPT-5.6-Sol/xhigh iteration 2 returned PASS after mutation-testing detection, persistence, and later-batch filtering. Full audit-script suite: 570 passed. Audit pipeline: 18 stages passed. Strict audit lint: 0 errors. Landed onto current main as f05c0d0ae4fbf971b96f18925e5f75eaad41dfcc.

PR #5582, opened 2026-07-25
docs(physics-loop): route execution to the Opus worker; codex keeps review/audit lanes

jonathonreilly:
## Summary
- Updates the `/physics-loop` command wrapper's **Execution Mechanism** section for the owner's 2026-07-24 executor directive: bounded science execution (note/runner drafting, scratch computation, structured extraction, panel lens execution) is farmed to the strongest configured **Claude Opus execution worker** (Agent/Workflow `model: "opus"`, one agent per spec, same spec contracts and anti-fabrication clauses).
- The codex worker retains **only** the independent review-loop and audit-loop lanes (owner grant 2026-07-09).
- Verdict synthesis remains never-delegated; the no-go planning discipline paragraph is unchanged in substance, with the lane sentence updated to reflect the 2026-07-09 independent-dispatch grant.

## Changes
- `.claude/commands/physics-loop.md` — Execution Mechanism section only (13 insertions, 7 deletions). No change to invocation, required behavior, campaign rule, or non-negotiables. The repo-native skill file `docs/ai_methodology/skills/physics-loop/SKILL.md` needs no edit (verified: it contains no executor-model references).

## Test Plan
- [x] Doc-only change; no runners or notes touched
- [x] Grep confirms no other executor-model references in either physics-loop file

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5581, opened 2026-07-25
Science: Cycle 610 family re-landed on the re-frozen substrate (causal-time lane, stacked on the substrate PR)

JONbridgerapps:
## What

Re-lands the Cycle-610 family (intrinsic tick clock and relational duration
law, preparation tournament, association/causal-order tournament, and the
minus-channel certification addendum) **on the re-frozen substrate**, stacked
on the substrate PR.

This is the second half of the unblock for review-loop's 2026-07-23
`NOT-LANDABLE-YET` disposition on #5557. That disposition was correct and is
not contested: the runners hard-imported Cycle-583, which existed only on the
campaign branch. Rather than carry ~142 campaign-only files, this branch takes
the exact three-commit science delta and repoints it at the substrate landed
by the parent PR.

**Base:** `substrate/contact-dimer-chain-relanding-20260724`. Review that PR
first; this one is not independently landable.

## The repin

| pin | from (campaign bytes) | to (re-frozen bytes) |
|---|---|---|
| c610 `DEPENDENCY_SHA256[c578]` | `25806853…` | `4ef60441…` |
| c610 `DEPENDENCY_SHA256[c583]` | `3f1672ef…` | `ef6805e6…` |
| c611 / c612 `C610_SHA256` | `61d624d3…` | `10f2d837…` |

**Re-pinned after #5579 landed (2026-07-25).** The substrate PR landed with
reviewer repairs that changed both substrate runner shas, which made the first
version of this branch's pins stale the moment the substrate merged. This
branch was rebuilt from current `main`, re-pinned against the **landed** bytes,
and every receipt and cold transcript regenerated by a fresh actual run. The
scientifically load-bearing outcome: **row structure is unchanged against the
repaired substrate** — the substrate repairs (including the now genuinely
computed Cycle-578 witness) moved no downstream Cycle-610 value.

The Cycle-610 note's dependency table and its definitive-run provenance block
(runner / receipt / cold-transcript SHA-256) were updated to the regenerated
bytes in the same commit.

Every downstream artifact — four receipts and four cold transcripts — was
regenerated by an **actual run** in dependency order (610 → 611 → 612 →
612-addendum). Nothing was hand-edited; the receipts are runner-written.

## Verification: row structure preserved exactly

These runners **exit 1 by design** — they carry preregistered falsification
rows, and "zero failures" would be the wrong success criterion. The assertion
made here is row-by-row equality of `(label, verdict)` against the recorded
campaign cold logs, compared on the label only because the ` :: <detail>`
[...]

PR #5580, opened 2026-07-24
Simplify distributed audit draining to optimistic workers

JONbridgerapps:
## What changed

- Make every employee/account run the same complete audit command from a separate clean clone:

  ```bash
  python3 docs/audit/scripts/orchestrate_audit_loop.py --max-workers 4
  ```

- Generate a unique session id when none is supplied and use it only to rotate target ordering.
- Remove coordinator/helper roles from the CLI and methodology.
- Keep panels and forensic work serial inside each worker while allowing independent clones to overlap optimistically.
- Define every zero-work exit as a worker-local observation, not a global completion certificate.
- Preserve claim-local quarantine/continuation and route repair inventory through `science-fix-loop`.

## Simpler safety model

There is no leader election, heartbeat, global lease, campaign ref, capacity registry, shared checkout ledger, or terminal coordination record.

Workers may occasionally compute the same claim. Immediately before apply, the existing audit transaction refreshes `origin/main` and revalidates the complete source/ledger/dependency/runner/evidence/seat/selection fingerprint. Stale work returns `remote_state_superseded`, mints no verdict, and creates no quarantine. Accepted work still passes the normal apply, full pipeline, strict lint, one-claim commit, and fast-forward push reconciliation.

The tradeoff is bounded duplicate compute instead of a distributed coordinator. No operational state can certify scientific authority or retained grade.

## Scope

Five files only:

- the top-level orchestrator;
- its existing test module;
- the audit-loop skill, distributed-operation reference, and agent prompt.

No batch, panel, forensic, branch-management, audit-data, source-note, runner, or verdict code changes.

## Validation

- Complete orchestrator/transaction suite: 119 tests passed.
- Focused campaign contract suite: 31 tests passed.
- Full audit pipeline passed: 3,863 rows, 431 retained-grade, 632 ready, zero invalidations.
- Enforced repository invariants passed with zero link/class-F violations.
- Strict audit lint: 0 errors; existing 26 warnings and 439 notices only.
- Skill quick validation, controlled-vocabulary lint, Python compilation, and `git diff --check` passed.
- Pipeline validation left no generated audit-state changes in the PR.

This draft may leave draft and land only after a fresh independent GPT-5.6-Sol/max `review-loop` PASS on the exact head.

PR #5579, opened 2026-07-24
Science substrate: re-land the contact-dimer fiber-law chain (Cycles 563-583) off main

JONbridgerapps:
## What

Byte-coherent landing of the causal-time lane's hard-import substrate as a
self-contained chain re-frozen **on this branch's own main-based history**,
repairing every item in the 2026-07-23 review disposition's blocking set.

This supersedes the closed #5560. 26 files: the four-runner chain (Cycles 573,
575, 578, 583 — runner + note + runner-written receipt + shipped cold
transcript each, frozen sequentially in per-cycle commits), three byte-exact
campaign evidence anchors (Cycles 563, 569, 570), and the refreshed
citation-graph manifest.

**Why this matters beyond its own content:** review-loop ruled #5557
`NOT-LANDABLE-YET` because the Cycle-610 family hard-imports this substrate,
which existed only on the campaign branch. This PR is that unblock. It is
verified below that after this lands, the Cycle-610 runner's only remaining
campaign-only import is satisfied — Cycles 210/219/230 are already on main.

## Repair summary (2026-07-23 blocking set)

1. **Cycle-583 receipt/note incoherence — repaired by coherent re-freeze.**
   The whole chain was re-frozen cold on this branch: each cycle's runs
   execute at its accepted-predecessor commit with the final note on disk;
   the runner itself writes its receipt (hashing the note bytes it ran
   against), and the same run's full transcript ships beside it. The
   receipt→note binding can no longer go stale without the gate catching it.
   Campaign-branch pins onto the old bytes (Cycle 586 pins the old receipt,
   Cycle 599 the old note) break by necessity: that chain is already
   internally incoherent (599 pins the post-edit note whose own receipt
   disavows it) and was rejected in the 2026-07-23 #5523 review.

2. **Cycle-563 non-runnability — now documented, not papered over.** c563
   top-level-imports the absent Cycle-560 runner, so **no mode of c563 runs
   on a main-based tree — including its default `--mode dry-contract`**,
   which exits at import. Vendoring was evaluated and rejected: c560's
   import-and-run closure adds 24 campaign files (c515–c560 compiler spine);
   c569's would add 114. Both explode this PR's minimal-closure design and
   would land more instances of the very defect class under repair (c560 and
   c566 hardcode the same literal-zero intertwiner rows at c560.py:804 and
   c566.py:712 on the campaign branch).

3. **Self-awarded intertwiner rows — documented as a science gap; reliance
   restructured.** In the byte-exact anchors, c563 assigns
[...]

PR #5583, opened 2026-07-25
fix(cl3): repair the malformed K3 denominator on the audited_failed Clifford root

JONbridgerapps:
## Why this is worth a PR of two files

A measured citation-graph pass over the live ledger found that the single highest-impact unresolved node in the repository is `cl3_complexification_split_narrow_theorem_note_2026-05-10`:

- `effective_status: audited_failed` (rank 0 — the lowest in the table)
- `dependencies: []` — it is a **root**, so no upstream repair can ever un-stick it
- **~1,732 rows / 44.9% of the ledger** sit downstream of it, with 29 direct citers

And the 2026-07-21 audit round that failed it gives exactly one ground:

> "the K3 idempotence display contains the malformed expression `(1 - 2 i ω + 1)/(2² · )`, although the following equality and runner use the correct denominator 4. … the negative-assertion judgment otherwise passes N1-N8 … the abstract split and dimension conclusions remain runner-supported … **Repair target: replace the malformed denominator with `4`**"

So a dropped character in one displayed line is holding rank-0 `audited_failed` on the Clifford foundation. That matters beyond the row itself: `compute_effective_status.py` returns `retained_pending_chain` for any claim whose chain includes a non-chain-satisfying dep, so while this root stays failed, the entire Cl(3) → staggered-Dirac → per-site-Hilbert cone is capped no matter how many descendants get audited.

## What changed

1. **`docs/CL3_COMPLEXIFICATION_SPLIT_NARROW_THEOREM_NOTE_2026-05-10.md:148`** — `(2² ·  )` → `4`.

   Verified from the algebra before editing, not taken on the auditor's word: with `e₊ = (1 − iω)/2` and `ω² = −1`, `(iω)² = +1`, hence `e₊² = (1 − 2iω + 1)/4 = (2 − 2iω)/4 = e₊`. Denominator 4 is forced. Line 147 and the remainder of line 148 already read `/4`, and the runner normalizes with exact `Rational(1,2)` — the defect never propagated into the mathematics, which is why the auditor could certify everything else while still (correctly) refusing to call the source clean as written.

2. **Recorded the N7 independent-resolution stdout** and linked the runner from the note's Validation block.

   `scripts/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.py` existed on main as source only: `cache_freshness: "missing"`, no `logs/runner-cache/` entry, and not linked from any note (registered only in `audit_dispatch_queue.json`). An earlier round of this same row failed on precisely *"N7 … has no independent resolution surface distinct from the runner used for its steelman"* — so the surface the auditor asked for existed but was invisible. Cache generated with the standard SHA-pinned precompute path.

## Verification

Re-ran on this tree after the edit: exclusion stress runner **62/0**, companion runner **38/0**, N7 independent helper **PASS** (proper-sector projector idempotent, commutes with the action; steelman resolved by reducibility).

## Churn

Deliberately checked against the review-loop skill's audit-hash churn guard, since this note is heavily cited:

- Rows requeued: **1** (the target itself). That requeue is the *only* sanctioned mechanism to un-stick a terminal `audited_failed` row.
- Downstream verdicts at risk: **0**. `dep_weakened` cannot fire because the status moves *up* (`audited_failed` rank 0 → `unaudited` rank 30), and `axiom_premise_changed` cannot fire because this id is not one of the four registered premise nodes.
- The guard's own trigger ("touches **many** existing claim-note files for … cleanup") is not met: one file, one line, and it is a science correction, not cosmetic churn.

No generated audit outputs are shipped. No audit verdict is set or predicted — status is the independent audit lane's alone; this PR only makes the row re-auditable and supplies the evidence surface a prior round asked for.

### DIFF
--- a/docs/CL3_COMPLEXIFICATION_SPLIT_NARROW_THEOREM_NOTE_2026-05-10.md
+++ b/docs/CL3_COMPLEXIFICATION_SPLIT_NARROW_THEOREM_NOTE_2026-05-10.md
@@ -145,7 +145,7 @@ e_+ + e_-   = ((1 - i ω) + (1 + i ω))/2 = 1,
 e_+ · e_-   = ((1 - i ω)(1 + i ω))/4    = (1 - (i ω)²)/4
             = (1 - i² ω²)/4              = (1 - (-1)(-1))/4 = 0,
 e_+²        = ((1 - i ω)/2)²             = (1 - 2 i ω + (i ω)²)/4
-            = (1 - 2 i ω + 1)/(2² ·  )    = (2 - 2 i ω)/4 = e_+,
+            = (1 - 2 i ω + 1)/4           = (2 - 2 i ω)/4 = e_+,
 e_-²        = e_-  (analogous).
 ```
 
@@ -229,6 +229,11 @@ Runner links:
   [`scripts/cl3_complexification_exclusion_stress_2026_07_13.py`](../scripts/cl3_complexification_exclusion_stress_2026_07_13.py)
 - Recorded exclusion stdout:
   [`logs/runner-cache/cl3_complexification_exclusion_stress_2026_07_13.txt`](../logs/runner-cache/cl3_complexification_exclusion_stress_2026_07_13.txt)
+- Independent faithful-direct-sum resolution runner (the N7 surface,
+  distinct from the runner used for the steelman):
+  [`scripts/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.py`](../scripts/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.py)
+- Recorded independent-resolution stdout:
+  [`logs/runner-cache/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.txt`](../logs/runner-cache/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.txt)
 
 The companion runner verifies via sympy exact rational/symbolic arithmetic
 on `2 × 2` complex matrices:
--- a/logs/runner-cache/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.txt
+++ b/logs/runner-cache/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.txt
@@ -1,0 +1,22 @@
+===== runner cache v1 =====
+runner: scripts/cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.py
+runner_sha256: e5350cc6e8f15bf4e0fae2782d0ea78a13330ea1b9732c666fa62bb3a5625f4f
+timeout_sec: 120
+exit_code: 0
+elapsed_sec: 0.26
+status: ok
+----- stdout -----
+N7_INDEPENDENT_CHECK rho_plus_complex_rank_4 status=PASS
+N7_INDEPENDENT_CHECK rho_minus_complex_rank_4 status=PASS
+N7_INDEPENDENT_CHECK rho_plus_kernel_dim_complex_4 status=PASS
+N7_INDEPENDENT_CHECK rho_minus_kernel_dim_complex_4 status=PASS
+N7_INDEPENDENT_CHECK rho_plus_volume_character_plus_i status=PASS
+N7_INDEPENDENT_CHECK rho_minus_volume_character_minus_i status=PASS
+N7_INDEPENDENT_CHECK direct_sum_complex_rank_8_faithful status=PASS
+N7_INDEPENDENT_CHECK proper_sector_projector_rank_2 status=PASS
+N7_INDEPENDENT_CHECK proper_sector_projector_idempotent status=PASS
+N7_INDEPENDENT_CHECK proper_sector_projector_commutes_with_action status=PASS
+N7_STEELMAN_RESOLUTION wall=irreducible complexified A-module faithfulness boundary; steelman=rho_plus_direct_sum_rho_minus; individual_kernel_dim_complex=4; combined_complex_rank=8; faithful=true; reducible=true; invariant_summand_dims_complex=2,2; conclusion=faithfulness_does_not_recover_irreducibility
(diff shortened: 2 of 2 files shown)


JONbridgerapps:
## review-loop: reviewed and LANDED

Salvage-landed on `main` as **5659c23d19351ef7efb425185dacd38239cb575b** (review-loop closes rather than merges; verified contained in `origin/main`). Both files landed unchanged apart from one cosmetic alignment fix noted below.

### Collision check
Checked #5581, #5582, #5584, #5585, #5587 (plus #5551, #5562, #5578, #5579, #5580) — **no collision**. None touches this note, this ledger shard, or the N7 runner/cache. The `audit-unlock-toe-20260724` pack (#5585) runs Lane A as an *argument-less* coordinator drain with no per-row CLI targeting, and Lane B on the KCPT derivation obligations; neither duplicates nor is disrupted by a source repair on this root. The stale-integration overlap guard between the PR base and current `origin/main` was empty.

### Independent verification of the algebra (not taken on the author's word)
Rebuilt `Cl(3,0)` from the defining relations alone — an 8-blade algebra with signs derived from `{γ_i,γ_j}=2δ_ij`, **no Pauli matrices in the primary route** — and reproduced, exact/symbolic, 62 checks with 0 failures:

- `ω² = -1`, `ω` central; hence `(i ω)² = i² ω² = (-1)(-1) = +1`.
- `(1 - i ω)² = 2 - 2 i ω` computed in the blade algebra.
- The denominator was **solved for, not assumed**: requiring `(1-iω)²/D = e_+` componentwise gives the unique solution `D = 4`. Denominator 4 is forced, and `4 = 2²`, consistent with the intent of the malformed token and with lines 145/147 and the rest of 148.
- `e_+ + e_- = 1`, `e_+e_- = 0`, `e_±² = e_±`, `ω·e_± = ±i·e_±`, each summand at complex algebra dimension 4.

**Defect is display-only and never propagated:** both runners normalize with exact `Rational(1,2)`, so the denominator 4 arises from the arithmetic rather than from the display.

**Second half of the repair instruction** (`notes_for_re_audit_if_any`: "independently recheck the faithful-real/nonfaithful-complex representation boundary") — the note's treatment at lines 179-184 is **correct as written**. Independently confirmed via a second implementation: `ker ρ_±^C` equals the *opposite* summand at complex rank 4 (image rank 4 of 8), while the restrictions to the real algebra `Cl(3,0)` have **real rank 8**, hence are faithful. `M_2(C)` is real-simple, so those restrictions could only be 0 or injective. Also reconfirmed independently: the scalar Clifford system has no solution, so no 1-dimensional complex representation exists at all.

### Runners
| Runner | Result |
|---|---|
| `cl3_complexification_exclusion_stress_2026_07_13.py` | `TOTAL: PASS=62 FAIL=0`, FLAGS: none |
| `audit_companion_cl3_complexification_split_exact_2026_05_10.py` | `TOTAL: PASS=38, FAIL=0` |
| `cl3_pauli_irrep_faithful_direct_sum_n7_independent_2026_07_17.py` | 10 checks PASS, exit 0 |

The committed cache is **byte-identical to a fresh run**, its `runner_sha256` matches the on-disk runner, and the header follows the `runner cache v1` convention. The N7 helper is not self-confirming: both hostile-mutation modes fail decisively and for the predicted reasons (`wrong-order` flips the volume character to `+i` and drops the combined rank below 8; `false-reducibility` keeps rank 2 and idempotence but leaves the commutant).

Sibling-runner pin sweep before the edit: six runners name this note; only two read its text, and they pin existence or the substrings `Cl(3,0) ≅ M_2(C)` / `irreducible left module` / `` complex dimension `dim_C V = 2` `` — none touches the edited lines.

### Gates
`vocab_lint --fix` 0 violations · `run_pipeline.sh` exit 0 (`repo_invariants_check` PASS, rows=3863, link_violations=0, graph_delta=acknowledged) · `audit_lint.py --strict` exit 0, "OK: no errors" · `git diff --check` clean · portable-links gate clean · pipeline-output-stripped gate clean (only the two source files committed). The added links target `.py`/`.txt`, and `LINK_RE` matches only `.md`, so no citation-graph edge changed and no manifest co-land was required (confirmed unchanged).

### Churn — claim confirmed
Measured, not just argued. Regenerating the ledger changed **exactly one shard**: the target row, `audited_failed → unaudited` with the 2026-07-21 verdict archived intact into `previous_audits` (full auditor provenance preserved, nothing laundered). **Zero** downstream verdicts disturbed.

Mechanically, from `invalidate_stale_audits.py`:
- `dep_weakened` fires only on `status_rank(after) < status_rank(before)`; `audited_failed`=0 → `unaudited`=30 is a **rise**, so it cannot fire.
- `axiom_premise_changed` is gated on `d in axiom_premise_ids()` = {`minimal_axioms`, `scale_reference_primitive`, `kinetic_isotropy_primitive`, `realized_state_primitive`}; this id is not among them.
- Also checked, since the PR body did not: `claim_type` is unchanged and `claim_scope` is sticky (`row.get("claim_scope") or backfill_scope(row)`), so `dep_claim_type_changed` / `dep_claim_scope_changed` cannot fire; `criticality` is already `critical` (max rank), so no criticality bump; the helper runner's own hash is unchanged, so `helper_runner_hash_changed` cannot fire. The sibling row `cl3_pauli_irrep_uniqueness_narrow_theorem_note_2026-05-10` also registers this helper, but `runner_artifact_issue_resolved` keys off that row's **primary** runner cache, which this PR does not touch.

`direct_in_degree` 29 and `transitive_descendants` **1771** on the live tree (the PR body's ~1,732 was slightly stale).

### One reviewer change
Line 148's second `=` landed at column 42 while its own continuation parent (line 147) and line 146 sit at 41. Since the row failed specifically on display quality, one space was removed so the repaired block aligns. Cosmetic only; no science change.

### Disposition
Source repair only. No audit verdict is set, predicted, or implied anywhere in the landed content, and no generated audit output shipped. The row is now `unaudited` and re-enters the ordinary queue; the verdict remains the independent audit lane's alone. Recommendation: **PASS**.

Head branch deleted because durable content landed.

PR #5566, opened 2026-07-23
Science: same-coupling executed positive field-update response (gravity/source lane, stacked on #5564)

JONbridgerapps:
## What

The gravity/source/resources lane's second prescribed science unit off `main`: requirement (b) of the Cycle-626 dependency ledger — **the SAME physical coupling at source and receiver in an executed positive field update**. An exactly time-reversible leapfrog update runs on two declared positive-sector carriers (periodic P0 graph Laplacian of the landed-c576 site graph; open Dirichlet scalar), with ONE coupling constant q wired to BOTH the source insertion `q·g(t)⟨ρ_src,φ⟩` and the certified readout `q⟨ρ_rec,φ⟩`.

**Stacked on #5564** (base = substrate branch). Independent of #5565 (both stack directly on the substrate). Cycle claim **681** (content only; descriptive filenames).

Files: runner + note + receipt + cold transcript (4, all new).

## Key results (cold 26/26, exit 0)

- **Anchors reproduced, no tuning**: all transcribed c626 comparator eigenvalues — 0.75302 (D1 P0 min-nonzero, L7), 1.75736/0.59419/0.45672 (D2, L3/L6/L7) — within 1e-3; both carriers strictly positive; stationary identities exact (sign cancellation, source-off/receiver-zero zeros, inverse residual ≤ 4.3e-15).
- **Adiabatic bridge at the preregistered rate**: RMS deviation of the executed readout about the *independently solved* stationary value falls as τ⁻² (frozen p=2 from the quartic-ramp argument, g″(0)=0; fitted slopes −1.96..−2.04 on all six fixtures, both kick/drift orderings); hold-mean converges to E_stat (terminal 3.6e-7). No refit anywhere.
- **The same-coupling content is load-bearing**: q→−q leaves the executed response invariant with residual exactly 0.0; the preregistered **decoupled arm (q_src=+q, q_rec=−q) flips the sign** with magnitude matching −E_stat to 3.6e-3; **q_rec=0 certifies exactly zero while the bare contact field stays nonzero** — reported as diagnostic only and certified as nothing (the ledger's "a contact-sensitive response is not energy, stress, source, or gravity" boundary, made executable).
[...]

PR #5523, opened 2026-07-18
Science: proper-cubic scalar-even antipodal transition lemma

jonathonreilly:
# Proper-cubic scalar–even antipodal transition lemma

This PR replaces the rejected campaign-scale review surface with one canonical,
independently reproducible algebraic lemma. Landing is requested only through
the repository review loop; no audit verdict or effective status is supplied.

The prior campaign head `2b1febe83f7462143376bb78ab66838a2c2a6d47` is
preserved on `codex/pr5523-campaign-pre-salvage-20260723`.

## Claim

For the supplied six-direction proper-cubic coin family

```text
C(beta) = exp[-i tan(beta/2)]
          (P_scalar - P_even + exp(i beta) P_vector),
```

the equal scalar/even superpositions are exchanged up to the common phase.
The identity is preserved under all 24 proper-cubic direction
reorientations.

This is a conditional finite-dimensional algebra lemma. It is not a clock,
duration, rate, energy, framework Record, occurrence law, probability rule,
or proper-time result.

## Review repair

The 2026-07-23 review-loop disposition rejected the previous package because
its final-tree evidence, branch-object pins, status language, placement, and
campaign control state were not landable as one science block. This repaired
surface removes that package from the PR rather than refreshing stale hashes:

- one canonical root-level claim note with explicit `Claim type`;
- one decisive runner with nonzero failure exit, an independent two-dimensional
  reduction, full train/held and all-24-frame checks, and a mutation control;
- one SHA-pinned fresh runner cache; and
- the required citation-graph manifest acknowledgment.

There are no branch-ancestry gates, historical commit-object imports,
developer-local transcript paths, self-awarded PASS rows, exported negative
claims, campaign handoffs, receipts, or audit-status payloads in the final
diff.

## Verification

- runner: `7 PASS / 0 FAIL`, normal and optimized execution byte-identical;
- fresh runner cache confirmed;
- changed Python compiles;
- vocabulary lint: zero findings;
- audit validation: one new `positive_theorem` row, `unaudited`, strict lint
  with zero errors;
- repository invariant check: citation-graph delta acknowledged;
- `git diff --check`: clean;
- validation-only audit and effective-status outputs stripped before commit.

The independent audit lane remains the sole authority for audit and effective
status.

PR #5584, opened 2026-07-25
Science: Cycle 622/629/662/675 family re-landed on the re-frozen substrate (causal-time lane, stacked on the 610 family)

JONbridgerapps:
## What

Re-lands the Cycle-622/629/662/675 family — deterministic preparation
double-falsification, the A2-line contact discriminator, the infinite-volume
A2 two-line keystone with the vernier 5:4 gold row, and the theta_2
isolated-zero certificate — on the re-frozen substrate, stacked on the
Cycle-610 family.

**Base:** `causal-time/cycle610-family-relanding-20260724`, which is itself
based on `substrate/contact-dimer-chain-relanding-20260724`. Review the stack
bottom-up; this PR is not independently landable.

Together with its two parents this completes the unblock of review-loop's
2026-07-23 `NOT-LANDABLE-YET` disposition: the entire causal-time lane now
executes on a main-based tree with nothing campaign-only in its import
closure.

## The repin

| pin | from | to |
|---|---|---|
| c622 / c629 / c662 `C610_SHA256` | `61d624d3…` | `36fcb165…` (landed c610) |

**Rebuilt 2026-07-25 after both parents landed** (#5579 substrate at
`5807ef17b7`, #5581 Cycle-610 family at `b6cf38bcfc`). Each review round
repaired runner bytes, so the c610 sha moved three times across the stack's
landing rounds (`e658ee75` → `10f2d837` → `36fcb165`); this branch pins only
the landed value, and every artifact here was regenerated by a fresh run
against the landed parents.

**Bearing directly on this review:** the #5581 reviewer corrected Cycle 611's
P-C row from a *false PASS* to its preregistered FAIL — c611 is now 8/3, not
9/2 — because the measured weights decrease while the assertion tolerance was
too lax. All three of c622/c629/c662 import c611. My re-runs show every row
here unchanged, which is consistent with that fix being an assertion tolerance
rather than a computed value. **That reasoning should be verified, not
accepted.**

No runner in this family pins a substrate artifact directly — the whole
substrate dependency enters through Cycle 610. Every receipt and cold
transcript was regenerated by an **actual run**; nothing was hand-edited.

## Two false provenance statements corrected

Both runners carried prose asserting the substrate was *unlanded*. On this
stack that is no longer true, and shipping receipts that assert it — in the
very stack that lands the substrate — would put a false provenance statement
into the retained record. Corrected before the runs wrote them:

- **c662** `receipt["stacking"]`: *"substrate not on origin/main at cut (main
  max cycle 335); branched off cycle629 tip 1085e03fdd"* → now records that
  the substrate is landed by the parent substrate PR of this stack, and cites
  this branch's own Cycle-629 commit.
- **c629** module docstring and `receipt["cycle_number_note"]`: *"PR #5523 is
  marked ready but the Cycle-573/575/578/583 substrate has not landed to main;
  citations refer to the campaign-branch surfaces carried in-tree"* → now
  records that those citations refer to the re-frozen surfaces present in this
  tree.

c629 was re-run after the correction; its row structure was re-verified
unchanged. The Cycle-610 family was swept for the same class and is clean.

## Verification: row structure preserved exactly

These runners **exit 1 by design**. c622 in particular is an honest double
falsification (6 PASS / 6 FAIL, exit 1 expected, per its own note), and c662's
two FAILs are estimator/certificate-design failures with neither preregistered
physics falsifier firing. "Zero failures" is the wrong success criterion; the
assertion is row-by-row equality of `(label, verdict)` against the recorded
logs.

| runner | rows | verdicts | exit | result |
|---|---|---|---|---|
| c622 | 12 | 6 PASS / 6 FAIL | 1 | **exact match** |
| c629 | 6 | 4 PASS / 2 FAIL | 1 | **exact match** |
| c662 | 8 | 6 PASS / 2 FAIL | 1 | **exact match** (1676 s) |
| c675 | 2 | 2 PASS / 0 FAIL | 0 | **exact match** |

A row-shape hazard worth recording for future re-runs: **c622's row set is
conditional**. If a certified configuration ever exists it prints three rows
("P4 held size", "P4 held species", "P5 field-ratio law survives"); otherwise
it prints one ("P4/P5: no certified configuration exists to test"). The
recorded log took the else branch, so the assertion is written against the
12-row shape. A future run that certifies would change the shape without any
verdict being wrong.

Comparison is on `(label, verdict)` only: the ` :: <detail>` suffixes embed
numpy 2.x reprs and measured timings that legitimately vary between runs.

## work_history governance (separate, droppable commit)

Same in-lane repair as the two parent PRs, for the same reason (review-loop:
these notes seed nothing into the citation graph): a
`Claim type: bounded_theorem` author hint and citation edges to each note
whose artifacts the runner loads or byte-pins — c622 +2 (610, 611), c629 +3
(610, 611, 622), c662 +3 (610, 611, 622), c675 +1 (662). Manifest delta: 4
added nodes, +9 edges, none removed, none rewired.

Byte-safe: no receipt in this family records a note hash. Claim type is an
author hint the audit lane owns and may override — it is not an audit verdict.
Kept as a separate commit so it can be dropped without touching the science.

## Unenforced upstream edges (flagged, not repaired)

Recording an observation from the pin audit rather than silently fixing it:
c629 and c662 both hash the Cycle-622 runner (`C622_SHA`) but never compare it
against a pinned constant, and c675 hashes Cycle 662 (`C662_SHA`), records it,
and never compares it — so Cycle 675 carries no byte-pin check row at all.
These are real gaps in byte-pin coverage inherited from the campaign. They are
not repaired here because adding a gate row would change the runners'
scientific row structure, which is exactly what this PR asserts it does not
do. Flagged for the lane to decide.

## Claim boundary

Unchanged by this PR. It is a provenance/landing repair, not a new scientific
claim: no result is added, removed, strengthened, or narrowed, and the two
prose corrections above make recorded provenance *more* accurate, not less
conservative. Authority none; audit unset. No claim ID promoted, no audit
verdict applied, no effective-status, ledger, or queue output included.

In particular the vernier 5:4 gold row and the 3:4 / 5:4 comparison surfaces
are carried exactly as recorded; nothing here identifies them as lapse,
redshift, or causal time.

## Execution disclosure (workhorse protocol)

Supervised and reviewed by Claude Opus 5 at maximum reasoning (the model
driving the session, per the repo skill); bounded read-only extraction of the
row-expectation tables and pin maps delegated to Opus 5 workers at maximum
effort. Per the owner's 2026-07-24 direction, workhorse execution may use
fable/opus-5-max or codex-sol-xhigh workers; this run used the former. Every
pin decision, both provenance corrections, the verification design, and all
verdicts are the supervising agent's.

## Integration note: the citation manifest will conflict, by design

The only file this stack and `main` both touch is the generated
`docs/audit/data/citation_graph_manifest.json`. This is the same deterministic
-regeneration case review-loop resolved on #5566. The manifest is a pure
function of the union of notes and the node sets are disjoint additions, so
regenerating after replay onto current main is well-defined.

**This stack is deliberately not rebased.** The substrate runners bake
`ACCEPTED_<cycle>_COMMIT` constants pointing at their own freeze commits, and
receipts record the `definitive_run_HEAD` of the run that produced them. A
rebase rewrites those SHAs and breaks every ancestry gate. Please replay
rather than rebase.


JONbridgerapps:
Landing-mechanism note: this PR is part of the substrate stack and inherits the constraint documented on #5579 — **land by merge, not cherry-pick replay**. The substrate runners bake `ACCEPTED_<cycle>_COMMIT` constants and gate on ancestry; a replay mints new SHAs and fails those rows (tested: c573 goes 7/0 to 6/1). Under `git merge --no-ff` all four substrate cycles reproduce branch-head results exactly. Also: verify each cycle from a clean tree — these runners rewrite their own receipts, so sequential runs in one worktree produce spurious pin mismatches.

JONbridgerapps:
Review-loop provenance — PASS after reviewer fixes.

Reviewed the rebuilt live head `17b7d2ff4e893d35a04ad0262c29ced497c3535a` (the immutable closed-PR ref remained on the pre-rebuild stack), then landed the reviewed delta to `main` at `c1054fa953be968d5f378e0480fa71c4034c93dd`.

Landed commits:
- `96ffa309e3` — rebuilt Cycle 622/629/662/675 science and fresh artifacts
- `e0bcaae225` — reviewer fixes: enforce recorded dependency hashes without new rows; replace Cycle-629 unconditional/overclaiming diagnostics; narrow Cycle-662 false-green global-min labels; correct stale provenance and support-only boundaries; refresh actual receipts/transcripts
- `c1054fa953` — separately droppable governance: bounded claim hints, complete citation edges, citation-graph manifest

Fresh reproduction on the reviewed bytes:
- Cycle 622: 6 PASS / 6 FAIL, exit 1 expected
- Cycle 629: 4 PASS / 2 FAIL, exit 1 expected
- Cycle 662: 6 PASS / 2 FAIL, exit 1 expected
- Cycle 675: 2 PASS / 0 FAIL, exit 0

The Cycle-611 correction does not propagate numerically: its landed change altered only the P-C assertion tolerance/label, not any imported computation. The Cycle-662 raw convergence/falsifier row is explicitly FAIL; the `0.278...` candidate is now only a non-identifying global-min diagnostic, while Cycle 675 supplies the narrow-window numerical completion near `0.313686`.

Validation: vocabulary lint clean; changed runners compile; receipts match normalized `(label, verdict)` transcripts; independent vernier algebra check passed; full audit pipeline and repository invariants passed; strict audit lint had no errors. Authority remains none and Audit remains unset. No audit verdict/status was written or applied.

PR #5597, opened 2026-07-25
Science: direction-set and triangulation covariance are different invariants (Cycle 695)

JONbridgerapps:
## What

Cycle 690 (landed today, `fc568bc0ad`) bounds **triangulation** invariance: no
eight-vertex unit-cube triangulation is invariant under all 24 proper cubic
rotations, the maximum is exactly 12, and the Kuhn/Freudenthal complex attains 6.

If a real-space Regge construction mediates covariance through the **edge
direction set** rather than through the triangulation, the relevant covariance
invariant is different. This PR shows the two come apart.

6 PASS / 0 FAIL, exit 0, exact integer arithmetic, 576-product closure verified.
Self-contained.

## Result

| object | invariant | value |
|---|---|---|
| 0/1 direction set (Kuhn edges) | oriented stabilizer | 3 |
| 0/1 direction set (Kuhn edges) | signed scope | **6** |
| Kuhn six-tetrahedron complex | cube-centred stabilizer | **6** |
| five-tetrahedron edge directions | signed scope | **24** |
| five-tetrahedron complex | cube-centred stabilizer | **12** |

- **Kuhn: the two invariants coincide at 6.** A property of that complex, not an
  entailment — and precisely why a construction reporting "6 of 24" can *appear*
  to quote the triangulation ceiling when it is quoting its direction set.
- **Five-tetrahedron: they diverge by a factor of two.** Its direction set is
  closed under all 24 rotations; the triangulation is invariant under only 12.
- **The relation is one-way**: every complex stabilizer is contained in its
  direction-set stabilizer, and 12 < 24 makes that containment strict for the
  five-tetrahedron fixture. Thus a triangulation-only ceiling does not upper-bound
  direction-set scope.

## Why this matters

**Cycle 690's ceiling of 12 bounds triangulation invariance and does not bound a
direction-set-mediated construction.** Two specific misreadings are blocked:

1. Treating 12 as a universal ceiling for real-space Regge covariance. It isn't.
2. Citing Cycle 690's **triangulation ceiling** as the reason for a measured
   direction-set scope of 6. Cycle 690 separately computes that Kuhn
   direction-set scope, so the separate row can support the value; the ceiling
   theorem cannot.

This is a live issue, not hypothetical: an adversarial review of the SPEC-C
open-coframe compiler found it doing exactly (2) — hard-gating its measured
scope against the Kuhn stabilizer and labelling 6 as "the landed note's
ceiling", when the note's ceiling is 12. That finding is what prompted this
cycle.

It also guards against a plausible but wrong lane decision: switching to the
five-tetrahedron substrate to "buy" covariance is not automatically an
improvement or a loss — it depends entirely on which object mediates the
construction's covariance, and the two move differently.

**This does not assert that any construction achieves 24.** Only that the
ceiling does not forbid it, and that the two invariants must be quoted
separately.

- **V1** — the obstruction is a live mis-citation of a landed theorem, surfaced
  by adversarial review of the gravity-lane compiler; `source_of_blocker_text:
  review_loop`.
- **V2** — new: the five-tetrahedron strict witness `12 < 24` and the explicit
  stabilizer-containment relation. Cycle 690 already contains the Kuhn `3/6/6`
  rows; they are prior art, not claimed as new here.
- **V3** — an auditor can recompute each stabilizer. The content is the strict
  five-tetrahedron witness that separates the invariants despite their Kuhn
  coincidence, together with the correct one-way containment.
- **V4** — non-trivial: it blocks a wrong substrate-swap inference and a wrong
  theorem citation, both live.
- **V5** — this extends 690 rather than restating it: 690 bounds triangulations
  and already records the Kuhn coincidence; this adds the strict five-tetrahedron
  witness showing why that ceiling does not transfer to direction-set scope.

```yaml
trace_class: negative_route_pruning
reachability_to_target: prunes
artifact_role: theorem
```

## Claim boundary and firewalls

`Claim type: bounded_theorem`. Authority none; audit unset; nothing promoted.

- No physics added — two group-theoretic invariants of declared finite fixtures.
- No gravity, metric, dynamics, or observable claim.
- No axiom or primitive proposed or adopted.

Scope: two declared complexes on the eight cube vertices, exact integer
arithmetic, full 24-frame enumeration. Other complexes, larger cells, improper
rotations, and lattice-site rotation centres are untested. N1–N8 remains
reviewer-owned.

## Gates

`repo_invariants_check --check --enforce-links`: PASS, rows=3867,
`graph_delta=acknowledged (enforced)`. `audit_lint --strict`: OK, no errors.

### DIFF
--- a/docs/PHYSICAL_DIRECTION_SET_VS_TRIANGULATION_COVARIANCE_CYCLE695_NOTE_2026-07-25.md
+++ b/docs/PHYSICAL_DIRECTION_SET_VS_TRIANGULATION_COVARIANCE_CYCLE695_NOTE_2026-07-25.md
@@ -1,0 +1,85 @@
+# Direction-set covariance and triangulation covariance are different invariants — Cycle 695
+
+Date: 2026-07-25
+
+Claim type: bounded_theorem
+
+Authority: none. Audit: unset. Constitutional effect: none. No new axiom or
+primitive is proposed or adopted.
+
+Runner: `scripts/physical_direction_set_vs_triangulation_covariance_cycle695_2026_07_25.py`
+(6 PASS / 0 FAIL, exit 0, exact integer arithmetic).
+
+## Why this exists
+
+[Proper-cubic covariance ceiling](PHYSICAL_PROPER_CUBIC_COVARIANCE_CEILING_CYCLE690_NOTE_2026-07-24.md)
+bounds **triangulation** invariance: no eight-vertex unit-cube triangulation is
+invariant under all 24 proper cubic rotations, the maximum is exactly 12
+(five-tetrahedron), and the Kuhn/Freudenthal complex attains 6.
+
+A real-space Regge construction does not always mediate its covariance through
+the triangulation. It may mediate it through the **edge direction set** — asking
+which frames carry every spatial direction class back into the set, up to a
+global sign. That is a **different invariant**, and this cycle shows the two come
+apart.
+
+## Result
+
+| object | invariant | value |
+|---|---|---|
+| 0/1 direction set (Kuhn edges) | oriented stabilizer | 3 |
+| 0/1 direction set (Kuhn edges) | signed scope | **6** |
+| Kuhn six-tetrahedron complex | cube-centred stabilizer | **6** |
+| five-tetrahedron edge directions | signed scope | **24** |
+| five-tetrahedron complex | cube-centred stabilizer | **12** |
+
+**For the Kuhn complex the two invariants coincide at 6.** That coincidence is a
+property of that complex, not an entailment — and it is exactly why a
+construction reporting "6 of 24" can appear to be quoting the triangulation
+ceiling when it is really quoting its direction set.
+
+**For the five-tetrahedron complex they diverge by a factor of two.** Its
+edge-direction set is closed under all 24 rotations while the triangulation it
+belongs to is invariant under only 12.
+
+**Neither invariant bounds the other.** The five-tetrahedron case exhibits a
+direction-set scope (24) strictly greater than the Cycle-690 triangulation
+ceiling (12).
+
+## Consequence
+
+Cycle 690's ceiling of 12 bounds **triangulation invariance**. It does **not**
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Review-loop provenance (Codex GPT-5.6-Sol reviewer seat; cross-family from the Claude Opus 5 author):

- Final review disposition: **PASS WITH BOUNDED CLAIMS**.
- Landed through the fail-closed cherry-pick loop on `main` at `c53a5fab2371473dd3d2e11817bbbe946a999f95` (source `669e011ab7`, citation acknowledgment `2159715fa5`, review fix `c53a5fab23`).
- Five findings were fixed before landing: the false “neither invariant bounds the other” claim became the correct complex-stabilizer subgroup relation; the oriented value `3` is now decisive; the possibly-empty direction accumulator is gated as the exact nonempty 18-vector set; physical direction-set mediation is explicitly conditional; and Cycle 690's existing Kuhn `3/6/6` prior art is distinguished from the new five-tetrahedron `12 < 24` witness.
- Independent recomputation generated the proper cubic group from quarter-turn generators and reproduced `24, 3, 6, 6, 24, 12`, with Kuhn element-set equality and strict five-tetrahedron containment.
- All six runner rows were enumerated; none is structurally always true. No `len(...) >= 0`, literal-`True`, `ACCEPTED_<cycle>_COMMIT`, or `definitive_run_HEAD` shape exists.
- N1–N8: PASS for the narrowed derived negative boundary.
- Validation: runner `6 PASS / 0 FAIL`; committed runner/receipt SHA match; `py_compile`, vocabulary lint, strict audit lint, pipeline validation, portable links, diff check, and enforced repo invariants all passed. The manifest delta is exactly the new node/edge acknowledgment with no rewiring.
- Claim boundary remains `bounded_theorem`, authority none, audit unset. No axiom or primitive is proposed/adopted, no audit verdict was applied, and independent audit remains the only promotion path.

The PR is being closed rather than merged because its reviewed content (including the review fix) is now durably present on `main`.

PR #5602, opened 2026-07-25
science: common-frame pair-generator exchange class registered, with its three limitations (batch R3)

JONbridgerapps:
## What this is

**Phase-2 batch R3** of the repo-state scrub. Registers the 2026-07-14
pair-generator material as a properly registered `bounded_theorem` claim at
`docs/` root, with a **new consolidated runner**, the **three dropped
limitations restored as part of the claim**, and the **ground-sector
degeneracy** as the invariant.

Files (4):

| file | what |
|---|---|
| `docs/COMMON_FRAME_PAIR_GENERATOR_EXCHANGE_CLASS_BOUNDED_THEOREM_NOTE_2026-07-25.md` | new claim note |
| `scripts/common_frame_pair_generator_exchange_class_2026_07_25.py` | new consolidated runner, 46/0 |
| `logs/runner-cache/common_frame_pair_generator_exchange_class_2026_07_25.txt` | SHA-pinned cache |
| `docs/audit/data/citation_graph_manifest.json` | stage-18 graph acknowledgment: **one node, `out_degree` 1** |

## CHURN COST — measured, not estimated

Measured by snapshotting all 3872 tracked ledger shards at the branch point,
running the pipeline, and diffing every row.

| quantity | measured |
|---|---:|
| rows **requeued** (`note_hash` or `runner_hash` moved) | **0** |
| **verdicts risked** (terminal verdict lost or reset) | **0** |
| rows whose `audit_status` changed | **0** |
| rows whose `effective_status` changed | **0** |
| rows whose `deps` changed | **0** |
| rows **added** | **1** (`common_frame_pair_generator_exchange_class_bounded_theorem_note_2026-07-25`, seeded `unaudited`) |
| new strict-lint errors / warnings / notices | **0 / 0 / 0** |

**Phase 1 predicted 0 requeues for R3. Measured churn is 0. It does not
exceed the prediction.** A new file seeds a new row; no existing note hash
moves. The four 2026-07-14 source surfaces and their four runners are
**not modified** — they stay exploratory, which is both correct and free.

The only reason this is not literally a zero-diff-to-existing-files PR is the
`minimal_axioms` row's topology metrics (`direct_in_degree` 511 → 512,
`load_bearing_score`, `transitive_descendants`), which the pipeline
recomputes. That is generated state; it was **restored from `HEAD`** and is
not shipped.

## Why the note could not simply be written where the work was

`docs/audit/data/excluded_source_patterns.txt:19` contains
`docs/work_history/**`, and `seed_audit_ledger.should_gate_node()` drops any
matching node with no audit history. A result written there can **never**
acquire a claim id, note hash, runner pin, queue entry, or verdict —
*regardless of quality*. The four sources also carry `**Type:** meta` and
`**Authority:** none`, so nothing could cite them. Landing at `docs/` root is
the entire content of this batch.

The four sources are referenced in the note **in backticks, not as markdown
links** — deliberately. They are route inputs, not evidence: every algebraic
step is recomputed natively, the runner reads none of them, and backticks
seed zero citation-graph edges, so no dependency edge is created on a row the
seeder would drop. The note's only markdown-link dependency is
`MINIMAL_AXIOMS_2026-06-29.md`, and the seeded row shows exactly
`deps: ["minimal_axioms"]`.

## The claim, with its limitations attached

Under four **supplied** hypotheses — (H1) one `M_2(C)` site domain on the
`Z^3` nearest-neighbour graph, (H2) an autonomous time-independent
self-adjoint pair generator, (H3) a sum of identical two-site terms, (H4)
**common**-frame `SU(2)` covariance — the commutant of the diagonal frame
action on two qubits has complex dimension exactly 2 and equals
`span{I, SWAP}`, so every admissible pair generator is `h = a I + b SWAP`.
The licensed quotient `(h, t) ↦ (α h + β I, t/α)` with `α > 0` removes `a` and
`|b|` and leaves exactly `sign(b)`.

**The invariant is the ground-sector degeneracy: 1 for `b > 0` (singlet), 3
for `b < 0` (triplet).**

**New result, nowhere else in the repo.** The one-excitation band minimum is
**not** a valid separator: `Z^3` nearest-neighbour adjacency is bipartite by
the parity of `x + y + z`, the sublattice relabeling `D = diag((−1)^parity)`
gives `D A D = −A`, hence `spec(A) = −spec(A)`, so the `+J` and `−J`
one-excitation bands are **identical as multisets** after the licensed energy
shift. Gated on `C4`, `Q3`, `K_{3,3}`, `P3` and on an `L = 4` periodic `Z^3`
chunk (64 sites, 192 edges), with a non-bipartite triangle as the breaking
mutation. `FULL_LAW_…:335-337` gets close ("Checking only the minimum energy
does not certify this reversal") but supplies neither the reason nor the
relabeling.

**The three limitations are carried, not smoothed:**

1. **L1 — common-frame covariance is a PREMISE that CREATES the two-point
   menu.** Under **independent** onsite covariance the commutant collapses to
   the scalars alone; solving directly, `a I + b SWAP` is invariant only for
   `b = 0`, so **no nontrivial pair law survives** without a further supplied
   object (connection, link variable, shared frame, symmetry reduction).
2. **L2 — "exactly two" holds only for the two-site edge class under the
   supplied identical-pair-term ansatz.** At three sites
   `H_1 = SWAP_01 + SWAP_02` and `H_2 = SWAP_01 SWAP_02 + SWAP_02 SWAP_01`
   satisfy the *same* Hermiticity, neighbour-exchange, and common-frame
   hypotheses, and `H_1 + η H_2` carries a dimensionless `η` moving the gap
   ratio `(E_1−E_0)/(E_2−E_1)` from **2** at `η = 0` to **1** at `η = 1/3`.
   A gap **ratio** is invariant under both `α > 0` rescaling and `β I` shift,
   so neither clock rescaling nor an energy-zero choice removes `η`.
3. **L3 — the "inert" identity shift is NOT inert on a record-conditioned
   active-edge set.** Two record sectors with different active-edge counts
   acquire a relative phase that moves an interference term on a coherent
   superposition; it is not a removable global phase.

## Why a NEW runner, and why the four existing ones are not reused

Not reused, and **not edited** (editing them would be hygiene on an
exploratory surface with no ledger row — it buys nothing and is out of scope
for R3). They cannot serve as this note's runner:

- the `sign(b)` quotient is "gated" only by the **vacuous literal**
  `{-1, 1} == {sp.sign(v) for v in (-3, 5)}`
  (`relational_…py:198`), which never touches `a`, `b`, `α`, `β`, or `SWAP`;
- **common-frame covariance of `H_1`/`H_2` is claimed but never computed** —
  `single_…py:238` checks only Hermiticity and mutual commutation;
- the active-edge phase is a **hand-built `sp.diag(1, 2)`**
  (`relational_…py:377-382`), with no record-conditioned edge set constructed;
- `qubit_…py:143` and `:144` are **the same check** under two labels;
- ~**34%** of `qubit_…py`'s 44 PASS lines are **prose-needle greps on its own
  note text**;
- and R4 exists in none of them.

**The new runner: `PASS=46`, `FAIL=0`, ~0.7 s.**

- **Exact `sympy` only.** No float is an input to any load-bearing comparison
  and no numeric tolerance is used anywhere; eigenvalue ordering goes through
  an exact three-way comparison that *refuses* anything it cannot decide
  symbolically (this is why the irrational `P3` spectrum is handled exactly
  rather than floated).
- **Zero needle gates.** The runner reads no markdown — not even its own note
  — so the PASS total is entirely mathematical. Nothing is self-referential
  and nothing is vacuous.
- **Ordered label manifest with a drift detector.** Renaming, reordering,
  adding, or dropping a gate fails the run.
- **A CONSTRUCTION-mutation probe for every claimed constant.** Probes rebuild
  the object from a changed construction; they never flip an assertion.

Gate groups: `G1` commutant = `span{I, SWAP}` (both containments + a negative
control); `G2` independent-onsite contrast (commutant = scalars only); `G3`
Hermiticity of both candidates + the licensed quotient (the vacuous-literal
replacement); `G4` ground-sector degeneracy 1 vs 3, invariant across a 3×3
`(α>0, β)` grid and under frame change; `G5` `D A D = −A`, `spec(A) = −spec(A)`
and the band-minimum failure, with the triangle as the breaker; `G6` the
three-site `η` counterexample **including the covariance check the source
runners omitted**; `G7` the record-conditioned active-edge boundary.

### Adversarial mutation battery: 12/12 caught

Because a reviewer previously showed that mutating a headline constant left
gates passing 20/0, I ran a battery that mutates the **constructions** and
requires gates to fail. Every one was caught:

| construction mutated | gates that failed |
|---|---|
| `SWAP` → CZ-like diagonal | G1c G1d G1f G1g G4a |
| diagonal generators → single-site only | G1a G1c G1d G1e |
| independent generators → diagonal | G2a G2b |
| ground-sector operator `SWAP` → `Z⊗Z` | G4b G4c G4d G4e |
| `Z^3` torus + same-parity face-diagonal edge | G5a G5c G5d |
| one-excitation block drops the off-edge term | G5a G5f |
| relabeling `D` → identity | G5d |
| three-site `H_2` → non-covariant `Z_0 Z_1` | G6b G6c G6d |
| `η` term removed entirely | G6f G6h G6i |
| active-edge convention ignores records | G7a G7b G7c G7d |
| evolution time `π/3` → `2π` | G7b G7c |
| a gate label renamed | manifest drift detector |

Two of my *own* first-pass expectations were wrong and the battery exposed
them — `(0,0,0)–(1,1,1)` is a **parity-changing** edge and leaves `Z^3`
bipartite, and `G4a`'s original form read only the eigenvalue multiset.
`G4a` was strengthened to pin `SWAP` as the transposition. That the
invariant reads the `(3,1)` multiplicity split, while it is `G1` — not `G4` —
that pins the operator carrying it, is stated in the note's N3.

## Gate results

| gate | result |
|---|---|
| `python3 scripts/vocab_lint.py --fix` on all branch-modified files | **0 violations**, 0 auto-corrections, 0 needing human review |
| `bash docs/audit/scripts/run_pipeline.sh` (validation) | **exit 0**, all 18 stages, `Pipeline complete.` |
| `python3 docs/audit/scripts/audit_lint.py --strict` | **exit 0, `OK: no errors`**; 23 warnings / 441 notices, **identical to the pre-existing baseline**, **0** referencing this claim |
| `repo_invariants_check.py --check --enforce-links` | **PASS** `link_violations=0 class_f_violations=0 graph_delta=acknowledged (enforced)` |
| `git diff --check` / `git diff --cached --check` | **exit 0** |
| runner | **`PASS=46 FAIL=0`** |
| mutation battery | **12/12 caught** |

Pipeline-output-stripped gate: `git status --porcelain` over
`docs/audit/AUDIT_LEDGER.md docs/audit/AUDIT_QUEUE.md docs/audit/data
docs/publication/ci3_z3/` prints **exactly one line** — the staged
`docs/audit/data/citation_graph_manifest.json`, which is the single allowed
co-landing file per the review-loop drop sequence
(`docs/ai_methodology/skills/review-loop/SKILL.md:915-975`), required because
this landing adds a graph node. Everything else the pipeline regenerated was
restored from the branch tip and is not shipped.

Registration verified in validation mode before restoring: the seeded row
carries `claim_type: bounded_theorem` (`claim_type_provenance: author_hint`),
`note_hash` pinned, `runner_path:
scripts/common_frame_pair_generator_exchange_class_2026_07_25.py`,
`deps: ["minimal_axioms"]`, `audit_status: unaudited`,
`effective_status: unaudited` / `awaiting_audit`, and it entered
`audit_queue.json` with `ready: true`.

## What I refused to do

- **No audit verdict is set or predicted anywhere.** The note writes no
  status value; the seeded row is `unaudited`.
- **No `Status: PASS` in the No-Go Discipline gate.** N1–N8 are recorded as
  **ATTEMPTED** with the explicit line that no closure is claimed and
  judgement belongs to the audit lane. (Two of the source surfaces award
  themselves `**No-go-discipline status:** PASS`; this note does not.)
- **No pairwise hypothesis independence is claimed.** Stated three times —
  in `claim_scope`, in Hypotheses, and in N2 — and the runner attempts no
  independence audit.
- **The exclusion glob was not touched.** Removing `docs/work_history/**`
  from `excluded_source_patterns.txt` would seed ~465 rows at once; that is
  R4, an owner policy decision, explicitly not dispatched.
- **The four source notes and their four runners were not modified**, and the
  four notes were not moved out of `work_history/` (moving them would create
  four ledger rows for prose that correctly carries `Authority: none` and
  would break the absolute paths their own runners read).
- **No new axiom, primitive, vocabulary, or status value.** Every term used
  already appears in the source surfaces or the canonical templates.

## Mandatory framework refresher — read before acting

`docs/MINIMAL_AXIOMS_2026-06-29.md` (full, 194 lines) — `:103-111`
"Admissibility is not a dynamics axiom… It does not choose a Hamiltonian or
transfer operator", which is why H2–H4 are **supplied**, never derived;
`:170` names "source/action and physical-observable identification" as an
open gate outside axiom content, which is where H2–H4 sit.
`docs/ai_methodology/skills/PRIMITIVE_REGISTRY_CHECK.md` (full, 47 lines) —
the three approved primitives supply no pair generator, covariance reading,
sign, or rate; rule 6 makes anything absent from the registry unapproved.
`docs/audit/README.md` (full, 430 lines) — `claim_type` / `claim_scope` /
`audit_status` / `effective_status` are auditor-owned; exactly two premise
types chain-satisfy (axioms and registered primitives); the exclusion
mechanism and its history-preserving drop rule.

Independent audit remains required. This PR asks for none of it in advance.

### DIFF
--- a/docs/COMMON_FRAME_PAIR_GENERATOR_EXCHANGE_CLASS_BOUNDED_THEOREM_NOTE_2026-07-25.md
+++ b/docs/COMMON_FRAME_PAIR_GENERATOR_EXCHANGE_CLASS_BOUNDED_THEOREM_NOTE_2026-07-25.md
@@ -1,0 +1,418 @@
+---
+claim_id: common_frame_pair_generator_exchange_class_bounded_theorem_note_2026-07-25
+claim_type: bounded_theorem
+claim_scope: "Under four SUPPLIED hypotheses -- (H1) one M_2(C) site domain on the Z^3 nearest-neighbour graph; (H2) an autonomous time-independent self-adjoint pair generator; (H3) a sum of identical two-site terms; (H4) COMMON-frame SU(2) covariance, one and the same U acting on both sites of an edge -- the commutant of the diagonal frame action on two qubits has complex dimension exactly 2 and equals span{I, SWAP}, so every admissible pair generator is h = a I + b SWAP with a, b real. The licensed quotient (h, t) -> (alpha h + beta I, t/alpha) with alpha > 0 removes a and |b| and leaves exactly sign(b), and the separating invariant is the GROUND-SECTOR DEGENERACY: 1 for b > 0 (the singlet) and 3 for b < 0 (the triplet). RECORDED EXPLICITLY AS A NEGATIVE RESULT: the one-excitation band minimum is NOT a valid separator, because Z^3 nearest-neighbour adjacency is bipartite by the parity of x + y + z, the sublattice relabeling D = diag((-1)^parity) gives D A D = -A, hence spec(A) = -spec(A), and the +J and -J one-excitation bands are identical as multisets after the licensed energy shift; the separator works only on a non-bipartite graph such as the triangle, and Z^3 is not one. THREE LIMITATIONS ARE PART OF THE CLAIM, NOT CAVEATS TO IT. (L1) H4 is a PREMISE THAT CREATES the two-point menu: under INDEPENDENT onsite covariance the commutant is only the scalars, SWAP is not invariant, and no nontrivial pair law survives without a further supplied object (a connection, link variable, shared frame, or symmetry reduction). (L2) 'Exactly two' holds only for the two-site edge class under the supplied identical-pair-term ansatz H3: at three sites H_1 = SWAP_01 + SWAP_02 and H_2 = SWAP_01 SWAP_02 + SWAP_02 SWAP_01 satisfy the same Hermiticity, neighbour-exchange, and common-frame covariance hypotheses, and H_1 + eta H_2 carries a dimensionless eta moving the gap ratio (E1-E0)/(E2-E1) from 2 at eta = 0 to 1 at eta = 1/3, which a gap RATIO argument shows is removable by neither clock rescaling nor energy shift. (L3) The 'inert' identity shift is NOT inert on a record-conditioned active-edge set: on a supplied convention where an edge is active when neither endpoint carries a record, two record sectors with different active-edge counts acquire a relative phase that moves an interference term on a coherent superposition of those sectors, so the energy shift may not be discarded before the domain is fixed. NOT claimed: any selection of sign(b), of |b|, of a rate, or of a time unit; derivability of H2-H4 from Lattice/Qubit/Admissibility/Record; PAIRWISE INDEPENDENCE OF H1-H4 (none is claimed and none is audited here); a well-defined discrete update on overlapping edges; a strict light cone; any record formation rule, formation trigger, instrument, Born weight, sampling rule, actuality, or realized outcome; the general reversible-record obstruction; continuity of a disagreement minimizer; and any relativistic, chiral, fermionic, gauge, species, matter, clock-metric, or gravitational consequence. This note sets no audit verdict and asserts no PASS."
+upstream_dependencies:
+  - minimal_axioms
+runner: scripts/common_frame_pair_generator_exchange_class_2026_07_25.py
+---
+
+# Common-Frame Pair Generator: The Exchange Class And Its Three Limitations
+
+**Date:** 2026-07-25
+**Type:** bounded_theorem
+**Claim type:** bounded_theorem
+**Scope:** two qubits on one edge of the `Z^3` nearest-neighbour graph under
+four supplied hypotheses; the axioms supply no dynamics.
+**Audit-status authority:** independent audit lane only. This note sets no
+audit verdict and predicts none.
+**Primitive status:** no primitive is approved, registered, edited, or
+enlarged here.
+**Primary runner:**
+[`scripts/common_frame_pair_generator_exchange_class_2026_07_25.py`](../scripts/common_frame_pair_generator_exchange_class_2026_07_25.py)
+**Runner cache:**
+[`logs/runner-cache/common_frame_pair_generator_exchange_class_2026_07_25.txt`](../logs/runner-cache/common_frame_pair_generator_exchange_class_2026_07_25.txt)
+
+**Upstream authority:**
+[`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md)
+
+## Purpose
+
+A symmetry classification that reduces an arbitrary Hermitian two-site
+generator to one physical coefficient has been recomputed more than once in
+this repository without ever becoming citable. This note states it once, with
+its limitations attached rather than dropped, and gates every constant in a
+single runner.
+
+The classification is genuinely useful: an arbitrary Hermitian operator on
+`C^2 ⊗ C^2` carries 16 real parameters, and under the stated hypotheses it
+collapses to two, then to one after the licensed quotient. The limitations are
+equally load-bearing, and they are the part that a summary tends to lose: the
+covariance reading is what *creates* the two-point menu rather than a
+notational convenience; the count does not survive enlarging the support past
+one edge; and the identity term is inert only on a fixed active-edge set.
+
+This note claims a **classification**, not a law. It selects no sign, no rate,
+no formation rule, no instrument, and no actual history.
+
+**Prior exploratory surfaces.** Four working notes dated 2026-07-14 under
+`docs/work_history/repo/review_feedback/` explore this route:
+`QUBIT_SYMMETRY_EXCHANGE_LAW_REDUCTION_PROBE_NOTE_2026-07-14.md`,
+`RELATIONAL_QUBIT_DISAGREEMENT_CANONICAL_LAW_ESCALATION_NOTE_2026-07-14.md`,
+`SINGLE_INVARIANT_ACTION_STEELMAN_ATTACK_NOTE_2026-07-14.md`, and
+`FULL_LAW_INVENTORY_ADVERSARIAL_REDUCTION_NOTE_2026-07-14.md`. They are named
+here as **route inputs only** and are deliberately referenced in backticks
+rather than as markdown links: they each carry `Authority: none`, they are not
+consumed as evidence anywhere below, and nothing in this note inherits status
+from them. Every algebraic step is recomputed natively in this note's runner,
+which reads none of those files. Their runners are **not** reused, for reasons
+stated under Verification.
+
+## Hypotheses (all supplied; none derived)
+
+The Qualification in
+[`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md) requires a
+retained derivation, bridge, or registered primitive before any of these could
+be treated as framework content, and none of them has one. That memo states
+that Admissibility "is not a dynamics axiom" and that it "does not choose a
+Hamiltonian or transfer operator", so H2–H4 cannot be read off the axioms; it
+also lists "source/action and physical-observable identification" among the
+open gates outside axiom content, and H2–H4 sit on that gate.
+
+- **(H1) Domain.** One `M_2(C)` possibility domain per site on the cubic `Z^3`
+  nearest-neighbour graph. *This much is axiom content.*
+- **(H2) Autonomous generator.** Between-record evolution is generated by a
+  time-independent self-adjoint operator. Time-independence is essential and is
+  not implied by strong continuity: the covariant family `h(t) = J(t) SWAP` is
+  pointwise covariant and retains an arbitrary coefficient function.
+- **(H3) Identical two-site terms.** The generator is a sum of identical
+  nearest-neighbour two-site terms. This is an **ansatz**, and L2 below is
+  precisely the cost of it.
+- **(H4) COMMON-frame covariance.** One and the same `U ∈ SU(2)` acts on both
+  sites of an edge. This is the decisive premise; see L1.
+
+**No pairwise independence of H1–H4 is claimed**, and the runner attempts no
+independence audit of them. They are stated as a conjunction and consumed as a
+conjunction.
+
+## Results
+
+**R1 (classification).** Under H1 and H4 the commutant of the diagonal action
+`U ⊗ U` on `C^2 ⊗ C^2` has complex dimension exactly 2 and equals
+`span{I, SWAP}`. Hence under H1–H4 every admissible pair generator is
+
+```text
+h = a I + b SWAP
+  = (a + b/2) I + (b/2)(X⊗X + Y⊗Y + Z⊗Z),
+```
+
+with `a, b` real. The exchange identity `SWAP = (I + X⊗X + Y⊗Y + Z⊗Z)/2` is
+exact. Gates `G1a`–`G1g`; the containment is checked in **both** directions and
+paired with a negative control (`X⊗I` is not in the commutant), so the
+equality is not read off a dimension count alone.
+
+**R2 (quotient).** On a **fixed** active-edge set, the map
+`(h, t) ↦ (α h + β I, t/α)` with `α > 0` leaves the generated channel unchanged
+up to a global phase. Taking `α = 1/|b|` and `β = −α a` carries `h` to
+`sign(b)·SWAP`, so `a` and `|b|` are removed. The sign is **not** removed: the
+two-parameter system `α SWAP + β I = −SWAP` has the unique solution
+`(α, β) = (−1, 0)`, which is not licensed. Positivity of `α` is exactly what
+makes the sign physical — dropping it identifies the two signs. Gates
+`G3a`–`G3f`, with the mutation probe `G3f` supplying the identification that
+the constraint blocks.
+
+**R3 (the separator).** The relabeling-proof invariant separating the two signs
+is the **ground-sector degeneracy**. `SWAP` has spectrum `{+1 (×3), −1 (×1)}`,
+so for `h = a I + b SWAP`:
+
+- `b > 0` has a **one-dimensional** ground sector (the singlet);
+- `b < 0` has a **three-dimensional** ground sector (the triplet).
+
+The degeneracy of the lowest eigenvalue is invariant under every `α > 0`
+rescaling, every `β I` shift, and every common frame change, so it survives the
+whole licensed quotient of R2. Gates `G4a`–`G4f`. `G4f` records the reason a
+weaker invariant will not do: `+SWAP` and `−SWAP` have the **same eigenvalue
+set** `{±1}` and differ only in the multiplicities, so a set-valued spectral
+invariant cannot separate them.
+
+**R4 (the band minimum is NOT a separator — recorded explicitly).** The
+one-excitation band minimum must **not** be used in place of R3. `Z^3`
+nearest-neighbour adjacency is bipartite by the parity of `x + y + z`. Let `D`
+be the diagonal sublattice relabeling `D_xx = (−1)^{parity(x)}`. Then for the
+adjacency matrix `A`
+
+```text
+D² = I,    D A D = −A,    hence  spec(A) = −spec(A).
+```
+
+In the one-excitation sector the exchange sum restricts to `A + (|E|·I − Deg)`,
+which on a regular graph is `A` plus a constant; `Z^3` is 6-regular, so after
+the R2 energy shift the `+J` and `−J` one-excitation bands are **identical as
+multisets**. The band minimum, the band width, and the whole band shape are
+therefore sign-blind on `Z^3` and separate nothing. The separator works only on
+a **non**-bipartite graph — the triangle has `spec(A) = {2, −1, −1}`, no
+diagonal `±1` relabeling reverses it, and its shifted `+J` and `−J` bands do
+differ — and `Z^3` is not one. Gates `G5a`–`G5h`: the one-excitation block is
+**assembled by applying the actual edge permutations**, not asserted from a
+formula, and is verified against the full `2^n` space on two instances;
+`D A D = −A` is checked on the `L = 4` periodic `Z^3` chunk (64 sites, 192
+edges) as well as on `C4`, `Q3`, `K_{3,3}`, and `P3`; the triangle is the
+breaking mutation. `G5h` adds the independent point that the band minimum is
+not even invariant under the `β I` shift, whereas the ground-sector degeneracy
+does not move at all.
+
+## Limitations (all three are part of the claim, not caveats to it)
+
+**L1 — common-frame covariance is a PREMISE that CREATES the two-point menu.**
+H4 is not a notational convenience; it is the physics. Under **independent**
+onsite covariance — separate `SU(2)` actions on the two sites — the commutant
+collapses to complex dimension **1**, the scalars alone; `SWAP` is not
+invariant there. Solving directly, `a I + b SWAP` is independent-onsite
+invariant only for `b = 0`. So under independent onsite covariance **no
+nontrivial pair interaction survives at all** without a further supplied
+object: a connection, a link variable, a shared frame, or a symmetry reduction.
+The axiom sentence "No possibility is privileged" does not choose between the
+two readings. Adopting the exchange class therefore imports a physical premise;
+it does not read one off a symmetry slogan. Gates `G2a`–`G2d`.
+
+**L2 — "exactly two parameters" holds only for the two-site edge class under
+the supplied identical-pair-term ansatz.** The count is a statement about one
+edge under H3, and it does not survive enlarging the support. On three sites
+with a centre and two equivalent neighbours, both
+
+```text
+H_1 = SWAP_01 + SWAP_02,
+H_2 = SWAP_01 SWAP_02 + SWAP_02 SWAP_01
+```
+
+are Hermitian, invariant under exchanging the two equivalent neighbours, and
+invariant under the common-frame diagonal `SU(2)` — they satisfy exactly the
+same symmetry hypotheses. `I`, `H_1`, `H_2` are linearly independent, so the
+family `H_η = H_1 + η H_2` carries a genuinely independent **dimensionless**
+coefficient `η`, and it moves a spectral gap ratio
+
+```text
+(E_1 − E_0)/(E_2 − E_1) = 2   at η = 0,
+                        = 1   at η = 1/3.
+```
+
+A gap **ratio** is invariant under both `α > 0` rescaling and `β I` shift, so
+neither clock rescaling nor an energy-zero choice removes `η`. Symmetry
+licenses the term; it does not fix it. Gates `G6a`–`G6j`. The common-frame
+covariance of `H_1` and `H_2` is **computed** here (`G6c`, `G6d`) — that is the
+property which makes `η` an independent invariant and therefore makes the
+counterexample bite — and `G6e` supplies the negative control (`Z_0 Z_1` is
+Hermitian but fails both symmetry checks) so the covariance gate is not a
+tautology.
+
+**L3 — the "inert" identity term is NOT inert on a record-conditioned
+active-edge set.** R2's removal of `a` is conditional on a **fixed** active-edge
+set. Supply the convention that an edge is active when neither endpoint carries
+a record. Two record sectors then differ in their active-edge count, adding
+`β I` per active edge contributes `β N_active`, and that is not a common scalar
+across sectors of different `N_active`. On a coherent superposition of two such
+sectors the relative phase moves an interference term, which is therefore not a
+removable global phase. The active-edge convention is itself **supplied** — the
+axioms give no formation rule — so L3 is a conditional: *if* record formation
+changes which edges are active, the energy shift may not be discarded before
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
## Correction pushed: L1 was wrong, and the note now states what is computed

This PR previously carried a limitation L1 that said the opposite of what its own algebra supports:

> **(old L1)** common-frame covariance is a PREMISE that CREATES the two-point menu — under independent onsite covariance the commutant is only the scalars, SWAP is not invariant, and no nontrivial pair law survives at all.

Both halves are wrong. The `G2a`–`G2d` gates were correct **as computations** — they compute the collapse for the *continuous* `SU(2)` and are unchanged — but the note generalised them from that supplied group to independent-onsite covariance *as such*, and then drew the wrong conclusion about where the menu comes from.

Every fact below was verified natively and exactly with `sympy` (no float is an input to any load-bearing comparison) before being written into the note. Nothing was taken on trust from the campaign material that raised the objection.

### What is actually true

**1. The scalars-only collapse belongs to the continuous group, not to independent-onsite covariance.**
Replace the continuous `SU(2)` by the order-4 group generated by the spin lift of the `pi/2` rotation about the edge axis, and *independent*-onsite covariance leaves a Hermitian family of **real dimension 4**, with the explicit Hermitian witness `Z (x) Z` — which commutes with `u (x) I` and with `I (x) u` **separately** and is not a multiple of `I`. The rest of the descent is located the same way: common-frame under that finite group leaves **6**, the full order-8 proper edge stabiliser **5**, the bare endpoint exchange **10**. The chain is `16 -> 10 -> 6 -> 5 -> 2`, and every step past 5 is bought with the *continuous* group rather than with a covariance reading. Gates `G8a`–`G8g`, with construction mutation `G8e` (rebuild the same witness against the continuous generators; invariance fails, so `G8c` does not contradict `G2a`). Whether the proper cubic rotations act internally on `M_2(C)` at all is itself supplied and is explicitly **not** claimed.

**2. The covariance premise is two premises with different standing, and the menu does not depend on the supplied one.**
H4 is now split into **(H4a)** correlation — the edge covariance group is a twisted diagonal `{(u, w u w^dag)}` rather than the independent product — and **(H4b)** flatness, `w = I` uniformly. For **every** twist the twisted commutant has dimension exactly 2. This is proved, not sampled: for a *generic* invertible `W`,

```text
(I (x) W)(P (x) I + I (x) P)(I (x) W)^-1 = P (x) I + I (x) (W P W^-1),
(I (x) W) SWAP (I (x) W)^-1              = (W^-1 (x) W) SWAP,
```

are exact symbolic identities (`G9a`, `G9b`), so the twisted diagonal is the `(I (x) W)`-conjugate of the diagonal and conjugation is a linear bijection on commutants. The commutant is `span{I, (w^dag (x) w) SWAP}`, re-derived by native solve in both containments on exact unitary twists (`G9c`, `G9d`). The R3 ground-sector separator is conjugation-invariant, hence `1` for `b > 0` and `3` for `b < 0` for every twist (`G9e`). Construction mutation `G9f`: replace the automorphism twist by the degenerate homomorphism `u -> I` and the dimension moves `2 -> 4`.

**3. Dropping flatness ENLARGES the underdetermination rather than dissolving it.**
Twists transform under per-site re-framing as `w_e -> g_x w_e g_y^dag`, so a loop product is conjugated and its trace is invariant. The 4-cycle assignment `(I, I, I, w)` has loop-product trace exactly `6/5`, neither `2` nor `-2`, so **no** per-site re-framing carries it to the flat assignment (`G10e`, mutation `G10f` applies actual re-framings and finds the trace immobile). Relaxing H4b adds a link field with re-framing-invariant content *on top of* the sign bit. **The two-point menu is therefore a lower bound on the underdetermination, not an artifact of the premise.**

**4. New R5 — "the law is `I - SWAP`" is not a frame-invariant sentence.**
`(w^dag (x) w) SWAP` is Hermitian for every unitary `w` (exact, from `SWAP (w (x) w^dag) = (w^dag (x) w) SWAP`), so it is an admissible pair generator; but for a non-central `w` it is **not** in `span{I, SWAP}` — the span rank rises `2 -> 3` (`G10b`). What is frame-invariant is the separator: the spectrum with multiplicity, hence the ground-sector degeneracy, does not move (`G10d`). Construction mutation `G10c`: rebuild with a central twist `w = -I` and the operator lands back inside `span{I, SWAP}`, rank `3 -> 2`. R1 is a classification **relative to a chosen frame**, and R3 carries the frame-independent content.

**5. Recorded, deliberately ungated.** `MINIMAL_AXIOMS_2026-06-29.md` does **not** force common-frame covariance: the standalone word "frame" occurs **zero** times in it (all six case-insensitive matches are inside "framework"), and its only covariance clause, `:57-58`, reads *"There is one fixed nearest-neighbor admissibility rule, covariant under lattice translations and proper cubic rotations"* — spatial motions, never an internal `SU(2)`. These are textual facts and are left in prose with exact line references: this runner reads no markdown, and a prose-needle gate would be a grep rather than a computation.

L1(iv) records the asymmetry without resolving it. **H4b is supplied outright.** **H4a is the half the Admissibility clause bears on** (`:60-61`, *"determined by, and vary with, the nearest-neighbor conditions"*), and the argument that Admissibility forces it is a **reading** turning on two things the memo does not settle — whether *"No possibility is privileged"* (`:52`) means ray-transitivity of the covariance group, and whether "conditions" is typed as record content or record occupancy. **The note neither gates that argument nor asserts the forcing as a theorem**, and I flag that deliberately: the campaign material states H4a as FORCED, and that is one step stronger than the evidence supports.

L1(v) quotes the prior art that concludes the opposite polarity from the old L1, in backticks as context only, consumed as evidence nowhere:
`COLOR_ARENA_..._2026-07-06.md:76-78` — *"The Qubit axiom itself supplies that identification"*, with frame-relativity at `:96-101` and the leftover registered at `:198` as *"R5 frame transport … not supplied by anything in this note -- open"*; and `work_history/repo/review_feedback/FOUNDATION_LICENSED_PHYSICAL_EQUIVALENCE_WEYL_PAIR_NOTE_2026-07-14.md:29-34,198-216` — *"Neither statement licenses arbitrary site-dependent quantum frames"*, with common `PU(2)` conjugation named the *"narrow safe re-presentation"*. So the corpus treats **independent-onsite** as the horn needing the extra supplied object, not common-frame. The two surfaces do not agree with each other on whether the Qubit axiom *supplies* the cross-site identification; the note takes no side and records the disagreement.

### Gate results

| gate | result |
|---|---|
| runner | `PASS=65`, `FAIL=0` (was 46/0), ~1.4 s |
| cache | regenerated via `precompute_audit_runners.py --force --push-mode none --allow-non-main`; `runner_sha256` matches the file |
| `vocab_lint.py --fix` | 0 violations on both branch-modified source files |
| `run_pipeline.sh` | completed; run as **validation only**, all generated surfaces restored |
| `audit_lint.py --strict` | `OK: no errors`, exit 0 |
| `git diff --check` | clean |

`docs/audit/data/`, `AUDIT_LEDGER.md`, `AUDIT_QUEUE.md`, `MISSING_DERIVATION_PROMPTS.md`, `docs/publication/ci3_z3/*_EFFECTIVE_STATUS.md` and `docs/repo/` were restored before commit, and the newly generated ledger shard was removed. Note on the restore target: those paths were restored to the **branch's committed state**, which is byte-identical to the branch's merge-base for all of them. Restoring literally from the current `origin/main` would have dragged ~1100 unrelated audit-data shards into this PR and reverted the `citation_graph_manifest.json` graph-delta acknowledgment that `repo_invariants_check.py` enforces for a new note. The PR diff against `origin/main` is still exactly 4 files.

**Churn:** 3 files, +577 / -36 in this commit; PR total unchanged at 4 files.

No audit verdict is set or predicted anywhere. No new axiom, primitive, or vocabulary. No `Status: PASS` is awarded in the No-Go gate. No pairwise independence of H1–H4, or of H4a from H4b, is claimed.

JONbridgerapps:
## review-loop: NOT LANDED — returned for targeted repair

Reviewer seat: `codex` / GPT-5.6-Sol, xhigh, one combined pass over all nine
applicable lenses. Orchestration and independent cross-checks: Claude Opus 4.6
in an isolated worktree. Reviewed at head `92cbcdf64f` (both commits), against
`origin/main` at `2b2008a1fa`. The head branch is **retained** — salvage is
deferred, not abandoned.

### Per-lens verdicts

| lens | verdict |
|---|---|
| Code / Runner | **FAIL** |
| Physics Claim Boundary | **REJECT** |
| Proof Obligations | **FAIL** |
| Imports / Support | **FAIL** |
| Nature Retention | **REJECT** |
| No-Go Discipline | **FAIL** (hard gate) |
| Labeling Convention | PASS |
| Repo Governance | FIX |
| Audit Compatibility | FIX |

A `FAIL` from the No-Go Discipline lens blocks a review-loop PASS regardless of
the other lenses.

### What passed, and it is a lot

The mechanical hygiene is genuinely clean and was verified, not taken on trust:

- runner reproduces `PASS=65 FAIL=0` in ~1.4 s; `runner_sha256` in the cache
  matches the file on disk exactly (`517cfa98…c82d6319`);
- the runner reads **no** markdown, greps no prose, and contains **no** float
  literals — the "zero needle gates" claim is true (the old ~34% prose-needle
  problem is genuinely gone);
- `vocab_lint --fix`: 0 violations; no prefilled audit verdict, no
  `audited_clean`, no `effective_status` authoring, no self-awarded No-Go
  `Status: PASS` anywhere in note, runner, or cache;
- the row seeds as `bounded_theorem` / `unaudited` / `ready: true` with
  `deps: ["minimal_axioms"]`, and the manifest delta is exactly one node,
  `out_degree` 1.

The algebra is also mostly right. Independently reconstructed by routes that do
not share the runner's implementation (character theory for the three-site
spectrum, an explicit real-Hermitian-basis rank computation for the commutant
dimensions, hand derivation for the one-magnon block): **R1**
(`span{I, SWAP}`, dim 2), **R3** (ground degeneracy 1 vs 3), **L2** (the
three-site `η`, ratios 2, 1, 3, 4/3, 4/5), **L1(ii)** (twisted commutant dim 2
for every twist), **R5** (span rank 3 → 2), the individual `G8` dimensions
(6, 4, 5, 10, 2), and the `G5a`–`G5f` bipartite obstruction all check out.

### Blocking findings

Six of these were confirmed by two independent computations (the reviewer's and
the orchestrator's), not by one seat.

**1. `OVERCLAIM` — the scalars-only collapse is attributed to the wrong
property. Continuity is not the wall; irreducibility is.**
`docs/COMMON_FRAME_…NOTE_2026-07-25.md:4` (`claim_scope`), `:197-228`,
`:452`, `:462`, `:526-527`, `:410-413`; runner `:969-1023` (`G8b` label,
`G8e` detail, `G8g` label).

L1(i) generalises from a single finite group: the order-4 group generated by
`diag(e^{-iπ/4}, e^{iπ/4})`, the lift of the π/2 rotation about the **edge**
axis. That group is **abelian**, hence reducible on `C^2` — which is why it
leaves real dimension 4. Add the lift of the π/2 rotation about a **second**
cubic axis — still a proper cubic rotation, the only rotations the Lattice
axiom names — and the group is the binary octahedral group, **finite, order
48** (verified by canonicalising enumeration), under which:

```text
INDEPENDENT-onsite, finite two-axis cubic group   dim_R = 1   (scalars only)
COMMON-frame,       finite two-axis cubic group   dim_R = 2   ( = span{I, SWAP} )
```

Both computed with the runner's *own* `herm_commutant_real_dim` helper and
again with an independent implementation. The witness `Z ⊗ Z` does **not**
survive the second axis. So a finite group reproduces the entire collapse, the
`claim_scope` sentence "a property of the supplied CONTINUOUS SU(2)" is wrong,
and `N3`'s hidden-wall line — "the continuity of the internal group, which is
what the scalars-only collapse actually needs" — names the wrong wall. That is
precisely the N3 failure mode the no-go battery exists to catch.

Note the direction of the error: the **superseded** L1 (that independent-onsite
covariance kills the nontrivial pair term) is closer to correct under the finite
group the Lattice axiom actually names than the correction that replaced it.

**2. `BUG` — `16 → 10 → 6 → 5 → 2` is not a nested chain.**
Note `:218-228`; runner `:1016-1023` (`G8g` asserts "the chain … is monotone").

```text
dim comm(SWAP)               = 10
dim comm(rot ⊗ rot)          =  6
dim comm(SWAP AND rot ⊗ rot) =  4     <-- so the 6 is NOT inside the 10
dim comm(rot ⊗ rot, flip)    =  5     ( 5 ⊂ 6 does hold )
dim comm(continuous diag)    =  2     ( 2 ⊂ 5 does hold )
```

The 10 and the 6 come from constraint sets that are not nested. `G8g` checks
only the values 10 and 2; it never checks a single inclusion. The numbers are
individually correct — present them as separate comparison counts, not a
descent.

**3. `NO_GO_OVERCLAIM` — `G5g`'s stated conclusion is false under the note's
own R2.** Note `:157-166`; runner `:673-683`.

`G5g` concludes "the band minimum **DOES** separate there". R2 licenses **any**
`β`. Shifting the triangle's `−J` band by `β = 1`:

```text
+J centred band  [-1, -1, 2]      minimum -1
-J centred band  [-2,  1, 1]      minimum -2
-J band + 1      [-1,  2, 2]      minimum -1   ==  +J minimum
```

So the minimum does not separate on the triangle either, and `G5g` contradicts
`G5h` ("the band minimum is not even a quotient invariant") inside the same
runner. What *is* quotient-invariant is that the two multisets are not
translates of one another — verified, they are not. R4's core claim on `Z^3`
survives; this specific sentence does not.

**4. `PROOF_OBLIGATION` — the quotient omits the `b = 0` orbit its own
hypotheses admit.** Note `:119-140`; runner `:346-379`.

`h = aI` satisfies H1–H4 (verified: it commutes with all three diagonal
generators). At `b = 0`, `α = 1/|b|` is undefined, the shifted generator is the
zero operator, and its ground degeneracy is **4** — not 1 and not 3. The
quotient has three orbits `{0, +, −}`, so "the two-point menu" and R3's
two-case separator need an explicit `b ≠ 0`.

**5. `BUG` — `G10e`/`G10f` do not build what their labels claim, and `G10f`
cannot fail.** Runner `:1152-1174`; note `:249-256`, `:584-586`.

`G10e` asserts "on a closed 4-cycle the twist assignment `(I, I, I, w)` has
loop product `w`" but simply assigns `holonomy = twist` — no cycle, no links,
no product. `G10f`, described in the note as "applying actual re-framings", in
fact computes `trace(g · h · g†) == trace(h)` for unitary `g` — true by
cyclicity for **every** input. It is a tautology presented as a
construction-mutation probe, so the load-bearing step of L1(iii) — that
per-site re-framing acts on the loop product by conjugation — is asserted in a
code comment and never gated. (The claim itself is true; the reviewer built the
four-link product with four independent site frames and confirmed it. It just
is not in this runner.) Related: `G10d` is a strict subset of `G9e`, i.e. the
same check under two labels — the exact defect this note criticises in the
2026-07-14 runners. And since `w` and `−w` define the same twist, the
well-defined invariant is `|tr|` or `tr²`, not signed `tr`.

**6. `SEMANTIC_BRIDGE` — L3 imports supplied structure it does not name.**
Note `:351-365`; runner `:828-891`.

The note supplies the active-edge convention and says so. `G7` additionally
introduces a two-dimensional sector Hilbert space, a **coherent superposition
across two distinct permanent-record configurations**, persistent cross-sector
phase coherence, and an off-diagonal observable connecting them. The Record
axiom says a state *is* a configuration of records and records are permanent;
none of that additional structure is supplied, derived, or listed among H1–H4.
The arithmetic (`1 → 1/2`) is exactly right; the framework reading is not
carried by anything the note supplies.

**7. `PROOF_OBLIGATION` — R4 needs a finite-volume or vacuum-subtracted
normalization.** Note `:119-120`, `:153-163`; runner `:540-696`.
On infinite `Z^3`, `|E|·I` diverges, so `A + (|E|·I − Deg)` and the per-edge
identity shift are not defined. The gated content is finite graphs and an
`L = 4` torus. State R4 on finite periodic boxes, or use
`Σ_e (SWAP_e − I)` whose one-magnon restriction is `A − Deg`.

**8. `NO_GO_OVERCLAIM` — the N1–N8 packet does not meet the battery's own
shape.** Note `:421-532`, against
`docs/ai_methodology/skills/no-go-discipline/SKILL.md:81-137`.
N2 requires a **pairwise** independence table; the note gives a prose control
list (and honestly says it is not an independence proof) whose first named wall
is the false continuity wall from finding 1. N3 names the wrong hidden wall.
N4 is titled "dependency roles, per citation" where the battery requires a
per-citation **residual-match** table. N5 misses the `G5g` overreach. N6 and N7
are adequate, and the note deserves credit for recording every item as
**ATTEMPTED** and awarding itself no `PASS` — but review cannot supply one on
this state.

**9. `REPO_GOVERNANCE` — `claim_scope` pulls in a backticked root note.**
Note `:4`, `:288-319`. The prior-art polarity sentence is inside the YAML
`claim_scope`, which makes
`COLOR_ARENA_…_BOUNDED_THEOREM_NOTE_2026-07-06.md` load-bearing **to the scope
as authored** while it is backticked and currently `unaudited`. Either drop
that sentence from `claim_scope` and keep it in the body as recorded-not-gated,
or make it a markdown dependency and accept the chain condition.

### On the backticked `docs/work_history/**` citations — the conclusion holds, the stated reason does not

This was checked directly. `docs/work_history/**` **is** in
`docs/audit/data/excluded_source_patterns.txt:19`, and all four sources carry
`**Type:** meta` / `**Authority:** none`. They are genuinely not consumed as
evidence — every step is recomputed natively and the runner reads no files. So
**backticks are the right call** and this is not a finding.

But the justification in the PR body is inaccurate: a markdown link would create
a citation-graph *edge*, and the seeder then strips the gated target from ledger
`deps` via `included_cids` — so no dangling ledger dependency would appear. The
practice is correct; the mechanism cited for it is not. Worth fixing in the note
so the next agent does not inherit a wrong model of the seeder.

### Findings recorded but NOT treated as landing blockers

- **Stale PR body.** It still says `PASS=46`, advertises the old 12/12 battery,
  and — more importantly — still states the *superseded* L1 (that common-frame
  covariance creates the menu), which commit `92cbcdf64f` reversed. The
  reviewer called this blocking; the orchestrator does not, because the PR body
  is not landed content. It should still be refreshed so the record is not
  self-contradictory.
- **"H4 is TWO premises with OPPOSITE statuses"** (`:4`, `:189-195`). The body
  ultimately classifies both as supplied and differing in *motivation*
  (`:272-286`). "Opposite statuses" overstates; a wording nit, not a blocker on
  its own.

### Salvage assessment

Run per the skill before closing. The durable, twice-confirmed, runner-backed
core is real and should not be lost: **R1**, **R3** (with `b ≠ 0`), **L2**,
**L1(ii)**, **R5**, the `G8` dimensions as individual counts, and `G5a`–`G5f`
plus `G5h`.

It was **not** salvageable with review-level edits alone. The false attribution
sits in `claim_scope` — the audit-visible field — and repairing it, plus the
chain framing, `G5g`, the `b = 0` orbit, L3's premises, the `G10e`/`G10f`
constructions, R4's normalization, and N2–N5/N8, means re-authoring the note's
entire limitation structure and its no-go packet and adding several new gates.
That is authoring, not narrowing, and the result would carry no reviewer round.
Salvage is therefore **deferred**, and the head branch is kept as the working
handle.

### Suggested repair order

1. Replace "continuous" with the irreducibility condition everywhere
   (`claim_scope`, L1(i), N2, N3, N8, the No-Go block, and the `G8b`/`G8e`/`G8g`
   labels), and add the two-axis finite control as gates — it is two extra
   generators through the helper that is already there, and it turns the
   corrected sentence into a computation rather than prose.
2. Demote `16 → 10 → 6 → 5 → 2` to separate comparison counts, or gate the
   inclusions that actually hold (`2 ⊂ 5 ⊂ 6`).
3. Rewrite `G5g` to the multiset-not-a-translate statement.
4. Add `b ≠ 0` to R2/R3 and the two-point language, or gate the third orbit.
5. Name L3's extended-sector coherence premise as supplied, or demote L3.
6. Build `G10e` from four actual links and `G10f` from four actual site frames.
7. Restate R4 on finite periodic boxes or with `Σ_e (SWAP_e − I)`.
8. Redo N2 as a pairwise table and N4 as a residual-match table; refresh the PR
   body.

No audit verdict was applied, recommended, or predicted by this review. The
independent audit lane remains the sole authority for that.

JONbridgerapps:
Closed by review-loop without landing: No-Go Discipline FAIL (hard gate) plus Code/Runner FAIL and Physics Claim Boundary REJECT. Nine blocking findings, six confirmed by two independent computations. Head branch physics-loop/r3-common-frame-pair-generator-registration-20260725 is deliberately RETAINED as the working handle -- salvage is deferred, not abandoned. See the review comment above for the repair order.

PR #5603, opened 2026-07-25
tooling(audit): refuse runner-bearing notes on excluded paths; capture `## Verification` runners

JONbridgerapps:
Two tooling repairs for the governance defect measured in phase 1 of the
repo-state scrub. Both are toolchain changes: no source note is edited, no
verdict is set or predicted, and no status value is invented.

## The defect

`docs/audit/data/excluded_source_patterns.txt:19` excludes
`docs/work_history/**`, and `seed_audit_ledger.should_gate_node` drops matching
nodes **before** claim typing runs. A result written there can never acquire a
claim id, note hash, runner pin, queue entry, or verdict, whatever `Type:`
header its author writes. Separately, `build_citation_graph.RUNNER_SECTION_RE`
omits `Verification`, which is the dominant house heading for naming a runner,
so notes that follow the convention resolve to `runner_path = None`.

## R1 — `audit_lint`: refuse a runner-bearing note on an excluded path

`audit_lint.py` now emits an **error** for a citation-graph node that has no
ledger row, whose path matches an excluded glob (and is not in
`never_gate_source_paths.txt`), and which names a `scripts/*.py` runner that
exists on disk. The detector is the graph's own runner attribution, so it adds
no independent regex surface and automatically tracks whatever `extract_runner`
understands.

**Ratchet, not sweep — and this was the load-bearing design decision.** The 398
paths that match at this commit are grandfathered in
`docs/audit/scripts/unregistered_runner_bearing_note_baseline.txt` and reported
as `unregistered_runner_bearing_note` notices; only a path **absent** from that
baseline errors. Firing on the existing population instead would put 398 errors
into `audit_lint.py` and make `--strict` red repo-wide on arrival, hard-blocking
every other lane with no drain path — that version is not landable, so it was
not built. The notice form makes the backlog countable and drainable exactly the
way `claim_type_defaulted` already is, while the error form stops the class from
regrowing. A baseline line that no longer describes such a note surfaces as an
`unregistered_runner_bearing_note_baseline_stale` notice, so drains are visible
and the list can only shrink.

The baseline lives beside the scripts rather than under `docs/audit/data/`
because that directory is restored wholesale from `origin/main` before a PR
lands; a baseline kept there would have every drain silently reverted.

The ratchet is live in both enforcement paths already: the pipeline
(`run_pipeline.sh` stage 13) and `pre_commit_audit_check.sh`, which runs
`build_citation_graph` → `seed_audit_ledger` → `audit_lint` on staged notes.

Phase 1 quoted 470 instances from a raw body-mention scan over the 647
unregistered notes. Under the precise rule shipped here — no ledger row, not
never-gated, graph-attributed runner that exists on disk — the population is
**398** at this commit (350 before R2, plus the 48 excluded-path notes R2 makes
visible). The looser body-mention count reproduces at 475 and includes
narrative infrastructure docs (skills, `ROOT_FILE_GUIDE.md`) that merely mention
a script; the graph detector was chosen to avoid grandfathering those under a
noisier rule.

## R2 — `build_citation_graph`: capture `## Verification` runners

`extract_runner` gains a Verification-heading scan **tried LAST**, after every
existing pattern. Measured over the whole docs tree against `origin/main`:

| variant | gains | losses | changes |
|---|---|---|---|
| Verification tried LAST (shipped) | **52** | 0 | **0** |
| Verification merged into `RUNNER_SECTION_RE` in place | 52 | 0 | **5** |

The merged variant moves five notes off the runner they already resolve to,
because a `## Verification` section can precede the `## Artifacts` / `## Runner`
section carrying the note's own runner. Both numbers were re-measured on this
commit before landing and reproduce phase 1 exactly. The two regexes therefore
stay separate, and a unit test pins the ordering.

## CHURN COST

Measured by running `run_pipeline.sh` twice against `origin/main` (once clean,
once with this branch applied) and diffing the generated ledger, graph and
queue. Re-measured after two `origin/main` advances (17 intervening audit
commits); identical each time.

| | |
|---|---|
| Source notes edited | **0** |
| Ledger rows re-seeded | **4** |
| Rows requeued (audit-queue membership change) | **0** — queue byte-identical |
| **Verdicts risked** | **0** |
| Rows whose `audit_status` / `effective_status` / `claim_type` / `claim_scope` / `note_hash` / `auditor` / `previous_audits` / `cross_confirmation` changed | **0** |
| Rows added or removed | **0** |
| `citation_graph_manifest.json` | identical — no manifest acknowledgment needed |

The 4 re-seeded rows are R2 attaching a `runner_path` (and, for one,
`helper_runner_paths`) to rows that previously had none. All four are
`audit_status: unaudited` with `previous_audits: 0`, so no verdict is touched:

- `anomaly_forces_time_admission_iii_note_2026-05-17` (`meta`)
- `ckm_five_sixths_exponent_discriminator_support_note_2026-07-02` (`bounded_theorem`)
- `dm_sigma_hier_closure_packet_note_2026-04-20` (`positive_theorem`)
- `s3_time_theta_to_slice_coupling_factor_rigidity_note_2026-05-17` (`positive_theorem`, `critical`)

The other 48 of R2's 52 gains sit on excluded paths with no ledger row, so they
produce no ledger change at all — they become visible to R1 instead, which is
why the two land together.

Measured churn is at or below the phase-1 prediction (4 requeues, all
`unaudited`, 0 verdicts risked), so no stop condition was triggered.

## GATES

| gate | result |
|---|---|
| `python3 scripts/vocab_lint.py --fix` on branch-modified files | 0 violations, 0 rewrites |
| `bash docs/audit/scripts/run_pipeline.sh` (validation) | complete; `repo_invariants_check` PASS, rows=3872, retained_grade=472, link_violations=0, class_f_violations=0, graph_delta=acknowledged |
| `python3 docs/audit/scripts/audit_lint.py --strict` | **exit 0**; 23 warnings (unchanged), 839 notices (441 baseline + 398 new grandfathered), 0 errors |
| `git diff --check` | clean |
| `python3 -m unittest discover -s docs/audit/scripts/tests` | 746 tests; 12 new tests pass. 7 pre-existing failures (`NoGoDisciplineGateTest` ×4, `FingerprintV1StampingTest` ×3) reproduce identically on unmodified `origin/main` and are untouched by this branch |

Generated audit outputs (`docs/audit/data/`, `AUDIT_QUEUE.md`,
`MISSING_DERIVATION_PROMPTS.md`, `*_EFFECTIVE_STATUS.md`,
`PUBLICATION_AUDIT_DIVERGENCE.md`, `FRONT_DOOR_STATUS.md`,
`RETAINED_BACKBONE.md`, `AUDIT_DISPATCH_QUEUE.md`) were restored from
`origin/main` before committing; the pipeline runs were validation only. The
4-row re-seed lands on the next ordinary pipeline run after merge.

## Tests

12 new unit tests. The ones that matter:

- `test_new_unregistered_runner_bearing_note_is_an_error` — the ratchet returns
  exit **1** on a new instance.
- `test_grandfathered_path_is_a_notice_not_an_error` — exit **0** for a baseline
  path.
- `test_drained_baseline_entry_is_reported_as_stale` — drains are visible.
- `test_verification_never_displaces_an_earlier_artifact_section` and
  `test_verification_never_displaces_a_runner_label` — pin the LAST ordering
  that the 5-note corruption measurement demands.
- Negative cases: a note with a ledger row, a `never_gate` path, a note with no
  runner, and a named runner missing from disk all correctly do not fire.

## Not done here, deliberately

- **Registering the 398 grandfathered notes.** That is phase 1's R4 — a policy
  call about what `docs/work_history/` is for, flagged to the owner and not
  dispatched. This PR records the debt and stops it growing; it does not settle
  it.
- **No verdict is set, predicted, or implied**, and no `audited_*` or
  `retained*` value is written anywhere.
- **No new axiom, primitive, vocabulary, or status value.** Both lint category
  names follow the existing `snake_case` convention and reuse the established
  error / warning / notice trichotomy.
- **No env or flag bypass on the error.** The reviewed baseline file is the
  escape hatch; an env override would make the gate decorative.


JONbridgerapps:
Landed **narrowed** by `/review-loop` (reviewer seat: codex-cli, GPT-5.6-Sol, xhigh; two rounds — full-lens, then confirmation). Closing rather than merging, per the skill.

On `main` as `c8dd46c6ab` (this PR's commit, rebased) + `460e7ed266` (the review fix).

## Landed

- **R2, narrowed.** `## Verification` runner capture, tried LAST. Independently re-measured: three variants of `extract_runner` rebuilt from source and run over all 4600 `docs/**/*.md`. Tried-LAST **46 gains / 0 losses / 0 changes**; merged into `RUNNER_SECTION_RE` in place **52 / 0 / 5**. Your 52/0/0 vs 52/0/5 reproduced exactly before narrowing — the ordering decision was right.
- **R1, demoted to reporting.** The detector, its predicate, its baseline and its negative cases land intact. The error branch does not.

## Why the ERROR did not land

It was **already red**. Re-measured against `origin/main` at review time the population was 403 against a 398-path baseline: five `review_feedback` notes landed in the day after the baseline was frozen. `audit_lint.py --strict` exited **1** on the integrated branch — verified by running it, not predicted.

Refreshing the baseline does not fix that, and this is the part the PR body missed. 369 of the notes the detector finds sit under `docs/work_history/repo/review_feedback/` and were **all** added in the seventeen days from 2026-07-10 to 2026-07-26; 369 of the 461 notes added there since 2026-07-01 name a live runner. The baseline is not a legacy backlog to freeze and drain — it is a snapshot of a stream. And `audit_lint` is a hard gate in two places: `run_pipeline.sh` stage 13 under `set -euo pipefail` (so stages 14-18, including `repo_invariants_check` and the ledger render, never run) and `pre_commit_audit_check.sh`, which refuses **any** commit staging a `docs/**.md` — for every lane, on a whole-graph lint, not just the staged paths.

So the shipped version fails this PR's own stop condition: "red repo-wide on arrival, hard-blocking every other lane with no drain path — that version is not landable, so it was not built." It is red on arrival and re-reds within hours of any drain, with the same absent drain path, since R4 is explicitly not dispatched.

Arming it needs the prior decision about what `docs/work_history/` is for and where a runner-gated result belongs. That is the owner's call and review-loop did not make it. Until then the two notice categories carry the measurement that decision needs: `unregistered_runner_bearing_note` for the population measured when the detector was written, `unregistered_runner_bearing_note_unbaselined` for what the class has grown by since — readable straight off a lint run.

## Second blocking finding: ambiguous runner pins

The Verification fallback took the **first** runner in the section. A Verification section is a reproduction recipe, not a runner label — six corpus sections list several co-equal runners in which position carries no authority. One backs a ledger row: `dm_sigma_hier_closure_packet_note_2026-04-20` lists its three theorem runners in a bash block, its own (`frontier_dm_sigma_hier_upper_octant_selector_theorem_2026_04_20.py`) third, and its helper closure is empty. The pin would have put `frontier_dm_pmns_chamber_spectral_completeness_theorem` into that audit packet as the **sole** runner and left the note's actual runner out of it entirely.

Now requires exactly one distinct runner, matching the uniqueness rule the adjacent top-of-file fallback already applies. Ambiguity is section-local, so an ambiguous section is skipped rather than treated as a veto. Cost: 6 of the 52 gains; the ledger re-seed drops from 4 rows to 3.

## Accuracy corrections

- The comment and test docstring said the LAST ordering exists because a Verification section can precede the `## Artifacts`/`## Runner` section carrying the note's own runner. Traced individually, **none** of the five measured regressions has any runner-section heading, and in all five the displaced stage is the final top-of-file single-path fallback. Corrected, with a test for that actual shape.
- The ambiguity comment claimed `runner_path` is left unset unconditionally; corrected, with a test pinning continue-to-later-singleton.
- Population figures reconciled post-narrowing across all four prose surfaces (the confirmation round caught these mixing pre- and post-narrowing counts).
- Five baseline entries the ambiguity fix drains were pruned (398 → 393), so the landing emits no `..._baseline_stale` notices.

## Confirmed from the PR as submitted

- The detector's predicate matches `seed_audit_ledger.is_excluded_source_path` exactly — same `fnmatchcase` over the same file, same loader.
- No verdict, `audit_status`, `effective_status`, `audited_*`, `retained*` or `previous_audits` value is written or predicted anywhere; no generated audit output ships.
- The re-seeded ledger rows are all `unaudited` with `previous_audits: 0` and `runner_path: None` today, so no ratified verdict is touched.
- The baseline's placement in `scripts/` rather than `data/` is correct, and for the stated reason.

## Gates

`vocab_lint --fix` 0 violations · `run_pipeline.sh` complete 18/18, `repo_invariants_check` **PASS** rows=3883 retained_grade=498 link_violations=0 class_f_violations=0 graph_delta=acknowledged · `audit_lint.py --strict` **exit 0** (21 warnings, 843 notices — 393 baselined + 5 unbaselined + 445 pre-existing — 0 errors, 0 stale) · `git diff --check` clean · portable-links gate clean · `citation_graph_manifest.json` unchanged, so no stage-18 acknowledgment co-lands · 751 tests, the same 8 failures as `main`, reproduced with `main`'s versions of all three changed `.py` files. Generated audit outputs restored; every pipeline run was validation only.

## Open, for the owner

Whether to arm the error, and on what scope. One option the review considered but did not take, because it is a policy call: error outside `docs/work_history/**` (24 grandfathered paths on a near-static surface, where the ratchet would actually hold) while `work_history` stays notice-only pending R4.

Head branch deleted — the content landed.

JONbridgerapps:
Closed by /review-loop: content landed narrowed on main as c8dd46c6ab + 460e7ed266. See the provenance comment above.

PR #5609, opened 2026-07-25
tooling(audit): bind terminal verdicts to the runner source they name (223 unpinned rows, 37 active failures)

JONbridgerapps:
Phase-2 follow-on to the repo-state scrub. Measured against `origin/main`
@ `6154f6d4a3`.

## What "hash-pinned" means here

`invalidate_stale_audits.detect_invalidation` is the only thing that reopens
a live verdict when its runner moves, and it can only compare two snapshot
fields:

- `audit_state_snapshot.runner_hash` — sha256 of `row["runner_path"]`. The
  legacy comparator fires only when this is **non-null**
  (`invalidate_stale_audits.py:473-485`); `blocker_fingerprint_v1` snapshots
  additionally pin `runner_path` and `runner_present`.
- `audit_state_snapshot.helper_runner_hashes` — path → sha256 for every
  transitive-import helper. The comparator **skips the entire helper
  channel** when the key is absent (`:487-501`).

A snapshot missing either field binds nothing on that channel. `runner_sha256`
and `cache_freshness` do not pin: the first is recomputed live, the second
describes today's `logs/runner-cache/` header, not what the auditor read.

## Verified counts (819 terminal-verdict rows)

| snapshot state | rows | verdicts |
|---|---|---|
| `blocker_fingerprint_v1` | 249 | all `audited_conditional`/`audited_failed` |
| legacy, `runner_hash` present | 474 | mixed |
| **legacy, names a runner, `runner_hash` null/absent** | **78** | 77 `audited_clean`, 1 `audited_conditional` |
| legacy, no runner named | 18 | 11 clean, 7 other |
| **declares helpers, no `helper_runner_hashes` map** | **182** | 169 clean, 13 other |

Union of the two unpinned channels: **223 rows — 209 `audited_clean`, 208
retained-grade** (180 `retained_bounded`, 27 `retained`, 1 decoration).
`audit_state_snapshot` is never absent, and no row pins a runner that has
since been deleted.

Phase 1's "95 terminal, 58 clean" reproduces on the total (96 today under
"legacy snapshot with no runner pin", drift from 17 intervening audit
commits) but **not** on the clean split: the true figure is 77 clean naming
a runner, 88 clean in the class overall. No definition tested reproduces 58.
The phase-1 number understated the clean exposure.

## Mechanism

Not helper-vs-declared, not parse-invisibility, not missing caches. **All 223
were audited before the writer recorded the field.**

- 77/78 primary-channel rows carry the exact pre-2026-05-16 snapshot key set
  `{criticality, dep_effective_status, deps, load_bearing_score,
  transitive_descendants}`. `snapshot_audit_state` gained `runner_hash` in
  `62a903eb0e` (2026-05-16) and `helper_runner_hashes` in `7cc2e0526c`
  (2026-07-15).
- 1/78 (`universal_qg_canonical_refinement_net_note`) has the key present and
  null.
- All 78 runners exist on disk today and all 78 have a **fresh cache matching
  the current runner sha** — there is no missing-cache component.
- `FINGERPRINT_STAMP_VERDICTS = {"audited_conditional", "audited_failed"}`, so
  a clean verdict has *never* received a v1 fingerprint. Zero of 450
  `audited_clean` rows carry one.
- PR #5603's `## Verification` capture is still open and does not change the
  count: it adds runner attribution to *unregistered* notes, not to rows that
  already have a `runner_path`.

## Risk: latent vs active

Git history starts at `62a903eb0e` (repo root import, 8461 files), so
pre-2026-05-16 verdicts have no visible pre-verdict source. Counting only
`--diff-filter=M` commits after each verdict, root excluded:

- **Primary runner, 4 rows changed after a clean verdict.** 3 are the single
  line `AUDIT_TIMEOUT_SEC = N` added by `a2f1a67541` — benign
  (`fm_transfer_note`, `gate_b_grown_trapping_transport_note`,
  `persistent_record_matched_compare_note`).
  **1 is substantive: `dm_lepton_synthesis_note_2026-04-19`**
  (`audited_clean` 2026-04-30, `effective_status: retained`) names
  `scripts/frontier_sigma_hier_uniqueness_theorem.py`, which was rewritten on
  2026-07-12 by `99646edd68` (+130/−17, "narrow to supplied-input S_3
  selection table" — withdraws four observational-closure phrases and admits
  three external inputs) and `0bbf5d3be1` (+63/−55). The sibling row
  `sigma_hier_uniqueness_theorem_note_2026-04-19`, which cites the *same*
  runner, was re-audited and is `audited_conditional`. This row kept its clean
  verdict because it pins nothing.
- **Helper runners: 40 rows had a declared helper substantively modified after
  the verdict; 36 are `audited_clean`, 35 retained-grade.** None are
  infra-only. Worst cases: the two
  `gauge_vacuum_plaquette_first_sector_minimal_bulk_completion_3plus1_*` rows
  (16 helpers, 34 substantive edits including "rescope source-sector
  factorization to supplied D" at +608/−161), `teleportation_*` (7-8 edits
  each), and `fifth_family_radial_repaired_positive_packet_note_2026-05-29`.

**Active integrity failures: 37 clean verdicts** (1 primary + 36 helper)
standing on runner source that demonstrably changed after the verdict, with
no pin able to catch it. **Latent: 186** unpinned but source-unchanged.

Caveat stated honestly: `helper_runner_paths` is the transitive-import closure
of the primary runner, so for a row whose primary is unchanged the helper set
is stable — but where a helper itself gained imports, the current closure may
exceed what the auditor saw. That widens the gap, it does not narrow it.

## What this PR ships

**Prevention (writer side).** `docs/audit/scripts/runner_pin_gate.py` is one
predicate executed by both sides. `apply_audit.snapshot_audit_state` calls
`verdict_pin_problems` and raises `RunnerPinIncomplete` rather than write a
terminal verdict that leaves a named runner unbound — mirroring how
`FingerprintV1Invalid` already fails loudly. `audited_clean` on a row naming a
runner absent from disk is refused outright; a non-clean verdict citing the
missing artifact stays legitimate. Non-terminal writes are unconstrained.

**Visibility (reader side).** `audit_lint` classifies every terminal row
against `runner_pin_baseline.json`. Severity follows the note_hash-drift
precedent already in this file — for a retained-grade row an undetectable
source change is a hard error; on a non-retained row it is re-audit-pending
and must not block strict lint:

| class | today | severity |
|---|---|---|
| pin field present but empty (writer regression) | 1 (non-retained) | ERROR if retained, else WARN |
| pre-pin snapshot outside the baseline | 0 | ERROR always — the list is shrink-only |
| recorded prior drift (`source_drifted_since_verdict`) | 44 | WARN `runner_pin_absent_and_source_drifted` |
| source moved *after* the baseline | 0 | ERROR if retained, else WARN |
| unpinned, source unchanged | 178 | NOTICE `runner_pin_grandfathered` |
| entry drained (row re-pinned or reset) | 0 | NOTICE `runner_pin_baseline_stale` |

The baseline is not a bare suppression list: each entry records the runner's
sha **at baseline time**, so the legacy population is frozen — any further
movement under one of these unpinned verdicts raises instead of passing
silently. Entries can only drain. The file lives in `docs/audit/scripts/`,
not `docs/audit/data/`, which is restored wholesale from `origin/main` before
every PR. No env or flag bypass.

A channel counts only when its runner is readable on disk: no pin can be
demanded for source nobody can hash, and that row is refused at clean-verdict
write time instead.

## Refused, deliberately

- **No snapshot is back-filled.** Writing today's sha into a legacy snapshot
  asserts the auditor saw current content — precisely what the missing pin
  makes unknowable — and would erase the 44 measured drifts.
- **No invalidation trigger for the legacy population.** Teaching
  `detect_invalidation` to fire on `runner_hash is None` would reset 209 clean
  verdicts. That is a verdict action and an owner/audit-lane decision.
- **No note edits.** Declaring helpers or repairing runner references in a
  retained note drifts `note_hash`, which this lint already treats as a hard
  error requiring re-audit — a verdict action by another route.
- **No cache regeneration.** All 78 primary runners already have fresh
  matching caches, and generating unrelated caches can fire
  `runner_artifact_issue_resolved` and requeue conditional rows.

## Owner decisions this leaves open

1. **The 37 active failures need re-audit** — only the audit lane can clear
   them. They are named in `runner_pin_baseline.json`
   (`source_drifted_since_verdict: true`) and surface every lint run.
   `dm_lepton_synthesis_note_2026-04-19` is the sharpest: retained, and its
   runner's own scope was narrowed.
2. **Extend `FINGERPRINT_STAMP_VERDICTS` to clean verdicts.** Clean rows are
   the only class that never receives a v1 fingerprint, so they miss
   `runner_path` and `runner_present` pinning. Zero churn on existing rows;
   not shipped here because it changes writer behaviour on the surface whose
   test file (`FingerprintV1StampingTest`) already has 3 pre-existing
   failures on unmodified `origin/main`.

## Churn (measured, not predicted)

- **0 rows requeued**, **0 verdicts risked**, **0 note hashes touched**.
- A full `run_pipeline.sh` on this branch produced **zero tracked-file delta**
  — ledger shards, `AUDIT_QUEUE.md`, `citation_graph_manifest.json` and every
  rendered surface byte-identical to `origin/main`. `last_invalidations: 0`;
  status distribution unchanged (450 clean / 233 conditional / 58 failed /
  34 decoration / 33 renaming / 11 numerical_match / 2 in progress). Nothing
  needed restoring.
- Neither modified file is in the pipeline's write path: `audit_lint` is
  read-only and `apply_audit` runs only in the audit lane.

## Gates

- `python3 scripts/vocab_lint.py --fix` on all five files: 0 violations.
- `bash docs/audit/scripts/run_pipeline.sh`: PASS.
  `repo_invariants_check` PASS, `graph_delta=acknowledged`, 0 link violations.
- `python3 docs/audit/scripts/audit_lint.py --strict`: **exit 0**, 0 errors.
- `git diff --check`: clean.
- 759 tests (25 new). The 7 failures — `NoGoDisciplineGateTest` ×4,
  `FingerprintV1StampingTest` ×3 — reproduce identically on unmodified
  `origin/main`; verified by stashing and re-running.

### DIFF
--- a/docs/audit/scripts/apply_audit.py
+++ b/docs/audit/scripts/apply_audit.py
@@ -48,6 +48,9 @@ import ledger_io
 # used by both the verdict writer and the live comparator.
 import compute_audit_queue
 
+# Single runner-pin predicate shared by this writer and audit_lint.
+import runner_pin_gate
+
 REPO_ROOT = Path(__file__).resolve().parents[3]
 LEDGER_PATH = REPO_ROOT / "docs" / "audit" / "data" / "audit_ledger.json"
 AXIOM_PREMISE_NODES_PATH = REPO_ROOT / "docs" / "audit" / "data" / "axiom_premise_nodes.json"
@@ -839,6 +842,19 @@ def snapshot_audit_state(row: dict, rows: dict[str, dict]) -> dict:
                 f"(baseline problems {problems}) on "
                 f"{row.get('note_path') or row.get('claim_id')}"
             )
+    # Runner-pin gate. `runner_hash` and `helper_runner_hashes` are the only
+    # channels through which invalidate_stale_audits can reopen a verdict when
+    # the runner it cites moves; a terminal verdict whose snapshot leaves
+    # either channel empty on a runner the row names is pinned to nothing.
+    # Fail here rather than write it: the resulting row is indistinguishable
+    # from a bound one until the runner silently changes under it.
+    pin_problems = runner_pin_gate.verdict_pin_problems(row, snapshot)
+    if pin_problems:
+        raise runner_pin_gate.RunnerPinIncomplete(
+            "refusing to write a terminal verdict that binds no runner source "
+            f"(pin problems {pin_problems}) on "
+            f"{row.get('note_path') or row.get('claim_id')}"
+        )
     return snapshot
 
 
--- a/docs/audit/scripts/audit_lint.py
+++ b/docs/audit/scripts/audit_lint.py
@@ -53,6 +53,7 @@ import premise_nodes
 import ledger_io
 import no_go_discipline_gate
 import audit_contract
+import runner_pin_gate
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
 DATA_DIR = REPO_ROOT / "docs" / "audit" / "data"
@@ -1395,6 +1396,67 @@ def main() -> int:
                     "dangling_dependency",
                     f"{cid}: dangling dep {d!r} (no ledger row)",
                 )
+
+    # Runner-pin integrity. A verdict is bound to the runner it names only
+    # through audit_state_snapshot.runner_hash / .helper_runner_hashes; those
+    # are the only fields invalidate_stale_audits can compare. A terminal
+    # verdict whose snapshot leaves a named runner unbound stands even if the
+    # runner is rewritten. Severity follows the note_hash-drift precedent
+    # above: on a retained-grade row an undetectable source change is a real
+    # integrity violation; on a non-retained row it is re-audit-pending and
+    # must not block strict lint.
+    pin_baseline = runner_pin_gate.load_baseline()
+    for cid, row in rows.items():
+        classified = runner_pin_gate.classify_row({**row, "claim_id": cid}, pin_baseline)
+        if classified is None:
+            continue
+        label, detail = classified
+        retained = is_retained_grade(row.get("effective_status"))
+        message = f"{cid}: {detail}"
+        if label == runner_pin_gate.PIN_WRITER_REGRESSION:
+            # The snapshot writer had the field and emitted nothing. Only
+            # apply_audit writes snapshots, and its runner-pin gate refuses
+            # this shape, so a hit here is a regression or a hand-edited row.
+            if retained:
+                errors.append(
+                    message + " (RETAINED-grade row: a verdict that binds no runner "
+                    "source cannot be detected as stale)"
+                )
+            else:
+                add_warning("runner_pin_writer_regression", message)
+        elif label == runner_pin_gate.PIN_BASELINE_MISSING:
+            # Pre-pin snapshot shape on a row the recorded debt does not
+            # cover. The baseline is shrink-only; nothing may enter it.
+            errors.append(
+                message + " (runner_pin_baseline.json is shrink-only; a new "
+                "unpinned terminal verdict must be pinned, not grandfathered)"
+            )
+        elif label == runner_pin_gate.PIN_BASELINE_SOURCE_DRIFTED:
+            add_warning("runner_pin_absent_and_source_drifted", message)
+        elif label == runner_pin_gate.PIN_BASELINE_NEW_DRIFT:
+            if retained:
+                errors.append(
+                    message + " (RETAINED-grade row: do not move a runner that a "
+                    "live unpinned verdict cites — the audit lane must re-look first)"
+                )
+            else:
+                add_warning("runner_pin_baseline_new_drift", message)
+        else:
+            add_notice("runner_pin_grandfathered", message)
+    pin_stale = runner_pin_gate.stale_baseline_entries(rows, pin_baseline)
+    for cid in pin_stale["drained"]:
+        add_notice(
+            "runner_pin_baseline_stale",
+            f"{cid}: runner_pin_baseline.json entry has drained (row is now pinned "
+            "or reset); drop it so the recorded debt keeps shrinking",
+        )
+    if pin_stale["absent"]:
+        add_notice(
+            "runner_pin_baseline_stale",
+            f"{len(pin_stale['absent'])} runner_pin_baseline.json entries name no "
+            f"ledger row (first: {', '.join(pin_stale['absent'][:3])}); drop them "
+            "so the recorded debt keeps shrinking",
+        )
 
     # Effective-status propagation sanity. A retained-grade row's deps must
     # themselves be retained-grade, metadata context, axioms, or approved
--- a/docs/audit/scripts/runner_pin_baseline.json
+++ b/docs/audit/scripts/runner_pin_baseline.json
@@ -1,0 +1,2494 @@
+{
+ "description": "Terminal-verdict ledger rows whose audit_state_snapshot predates the runner-pin fields and therefore binds no runner content. Recorded once, shrink-only: audit_lint errors on any terminal row outside this list, on any retained-grade row whose named runner content moves away from the sha recorded here, and notices every entry that has drained. Entries are never added by tooling and the recorded shas assert repository content at baseline time, never what an auditor saw.",
+ "entries": {
+  "action_power_note": {
+   "audit_date": "2026-05-03T19:47:14.612681+00:00",
+   "audit_status": "audited_clean",
+   "effective_status": "retained_bounded",
+   "runner_path": "scripts/action_power_canonical_harness.py",
+   "runner_sha256_at_baseline": "42edf509eacb80fa08deeffbb58b2d601089e3396dd368783b53f63ae6d4150d"
+  },
+  "action_power_scaling_sweep_note": {
+   "audit_date": "2026-05-19T21:47:36.040389+00:00",
+   "audit_status": "audited_clean",
+   "effective_status": "retained_bounded",
+   "helper_runner_sha256_at_baseline": {
+    "scripts/action_universality_probe.py": "8fd9d0d7d21a8b8e27429666885a8a96f967e4971911cdc8db2d0156d2dfcc4f"
+   }
+  },
+  "alpha_lm_geometric_mean_identity_theorem_note_2026-04-24": {
+   "audit_date": "2026-07-15T18:45:38.154373+00:00",
+   "audit_status": "audited_clean",
+   "effective_status": "retained",
+   "helper_runner_sha256_at_baseline": {
+    "scripts/canonical_plaquette_surface.py": "fd681ecbecda7ab5590b72199ee5bbb4393826f0f8e40ca66f8e2e6531646693"
+   }
+  },
+  "alt_connectivity_family_basin_note": {
+   "audit_date": "2026-05-05T04:20:10.208839+00:00",
+   "audit_status": "audited_clean",
+   "effective_status": "retained_bounded",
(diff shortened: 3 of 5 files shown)


JONbridgerapps:
Landed on `main` via `/review-loop` (review-loop closes rather than merges).

**Landed shas:** `23349ee804` (PR content, cherry-picked onto current `origin/main`) and `834805e717` (review fix pass). Both verified contained in `origin/main`.

**Reviewer seat:** codex / GPT-5.6-Sol at xhigh, two rounds. Round 1 returned `DO NOT LAND` with 9 findings. Round 2 (focused confirmation on the fix pass) returned `LAND` with zero remaining findings.

**Round-1 findings accepted and fixed in `834805e717`:**

1. `classify_row` short-circuited on `source_drifted_since_verdict` before checking `baseline_source_drift`, so every later move on exactly the 44 worst-off entries would have reported as the softer warning forever. Drift evaluation now runs first.
2. The writer gate refused an absent named runner only for `audited_clean`. `audited_renaming` / `audited_decoration` / `audited_numerical_match` receive no v1 presence pin, and `detect_invalidation` returns `None` on an absent→present move under a null legacy hash (verified by direct probe), so those verdicts could still be minted permanently unbindable. The refusal now covers every terminal verdict outside `PRESENCE_PINNED_VERDICTS`, and a test asserts that set stays equal to `apply_audit.FINGERPRINT_STAMP_VERDICTS`.
3. A helper entering an unbound closure after the baseline reported as "source unchanged" — the recorded shas cover only the helpers that existed then. `baseline_source_drift` now takes `check_helper_membership`, passed only when the helper channel is actually unbound.
5. `PIN_BASELINE_MISSING` was an unconditional repo-wide error, contradicting the `note_hash`-drift precedent the same block cites. It now errors on a retained-grade row and warns otherwise under `runner_pin_baseline_missing`.
7. The baseline `description` claimed all 223 rows bind "no runner content"; most are unbound on one channel and still pinned on the other. Reworded, and it now states that the copied `audit_status` / `effective_status` are non-authoritative and read by no classification path.
8. `test_helper_set_growth_under_a_recorded_map_is_new_drift_not_silence` asserted `PIN_GRANDFATHERED` — the opposite of its name — blessing the finding-3 defect. Rewritten plus new contract tests. Round 2 confirmed the new tests fail against the pre-fix implementation (5 failures, 2 errors), so they are not self-confirming.

**Additional review-loop fixes:**

- `helpers_unpinned` treated a recorded helper map with a stale key set as unpinned. `detect_invalidation` returns `helper_runner_paths_changed` on any key-set difference (including an empty map), so that channel is bound; the old predicate would have turned an ordinary import edit under a retained row into a hard `audit_lint` error — in `run_pipeline.sh` stage 13 under `set -e` and in `pre_commit_audit_check.sh` — with a message ("snapshot records the pin field but leaves it empty") that was also factually wrong.
- `docs/audit/README.md` now documents the rule, its six lint dispositions, and its drain path under **The hard rules**, matching the precedent set by the sibling unregistered-runner-bearing-note baseline. The rule shipped undocumented.

**Round-1 findings rejected:**

- *A helper map with matching keys but a null/malformed value is invisible.* Refuted by direct probe of `detect_invalidation`: `None` → `helper_runner_hash_changed:...:missing->...`, `"deadbeef"` → `helper_runner_hash_changed:...:deadbeef->...`, and `None` on an absent helper correctly returns `None`. The channel is bound; draining it is correct. Round 2 reproduced and agreed.
- *`stale_baseline_entries` mis-advises on a partial ledger.* Not actionable: the function cannot know its input is complete, the result is an aggregate NOTICE for exactly that reason, and `audit_lint` always loads the full materialized ledger.

**Narrowed rather than implemented:** the reviewer wanted a dispatcher sidecar queueing the 44 recorded-drift rows. Not landed — dispatching 44 retained-grade rows spends audit capacity, which this PR itself leaves as owner decision #1, and no source note here requests re-audit (the skill's dispatch-manifest rule triggers on source notes). Instead the overclaim was removed: the finding no longer calls the row a "re-dispatch target" and now reads "re-audit candidate; nothing here queues it, and whether to spend audit capacity on it is an owner/audit-lane decision". Round 2 judged this disposition honest.

**On the author's refusal to back-fill snapshots — verified right, not lazy.** Independently: 128 of the 223 rows were audited before the git-history root import (`62a903eb0e`, 2026-05-16), so their pre-verdict source is genuinely unrecoverable; of the 95 that postdate it, 21 have already drifted, leaving at most 74 rows (33%) where a historically accurate back-fill was even possible. Every one of those would still have required writing into `docs/audit/data/` ledger shards, which a framework PR must not ship. The out-of-band `*_at_baseline` shas achieve the same forward protection without the false provenance assertion.

**Independent verification (not taken from the PR body):**

- Recounted the population from the ledger without importing `runner_pin_gate`: 942 terminal rows today, 78 primary-unpinned, 182 helper-unpinned, union 223, 209 `audited_clean`, 207 retained-grade. The PR body's "208 retained-grade" counts one `decoration_under_*` row inside the retained-grade set; 207 is the count under `{retained, retained_bounded, retained_no_go}`.
- Re-derived all 44 `source_drifted_since_verdict` flags from `git log --diff-filter=M` since each row's audit date, root import excluded: 0 discrepancies in either direction.
- Confirmed the two-channel premise by reading `detect_invalidation` and by probing it with a real ledger row as carrier.
- Confirmed all three sha routes (`snapshot_audit_state`, `runner_cache.runner_sha256`, `compute_audit_queue._sha256_file`) are plain sha256 over raw bytes.
- Measured the blast radius of every fix before applying: 0 rows change class on the current ledger. Classification is identical before and after — 44 / 1 / 178, zero errors.

**Gates on the landing set:** `vocab_lint --fix` 0 violations; `run_pipeline.sh` PASS (`repo_invariants_check` PASS, `graph_delta=acknowledged`, 0 link violations); `audit_lint.py --strict` exit 0 both on the branch and re-verified on landed `main`; `git diff --check` clean; portable-links gate clean; no-go-discipline scan produced no matches; 29/29 targeted tests pass; full suite 788 tests with only the 4 failures that reproduce on unmodified `origin/main` (`FingerprintV1StampingTest` ×3, `CleanupWorktreeBranchTest` ×1). No generated audit output, no `docs/audit/data/` content, and no prefilled verdict is in the landing set.

**Open for the owner (unchanged by this landing):**

1. The 44 rows carrying `source_drifted_since_verdict` are re-audit candidates and are not queued. Only the audit lane can clear them; `dm_lepton_synthesis_note_2026-04-19` remains the sharpest (retained, and its runner's own scope was narrowed).
2. Extending `FINGERPRINT_STAMP_VERDICTS` to clean verdicts remains unshipped, as the PR states.
3. 162 runner scripts are now under the frozen retained-grade population: editing one raises a hard `audit_lint` error until the audit lane re-looks or the drift is recorded on the baseline entry. That is the deliberate trade — a loud stop-work in place of a silent integrity failure — and it is the same disposition the `note_hash` rule already takes for retained notes, but it is a real operational surface worth knowing about.

Head branch `tooling/runner-pin-gate-terminal-verdicts-20260725` deleted: durable content landed.

Review-loop reviews and lands; it applies no audit verdict. Every row this touches keeps whatever status the independent audit lane already gave it.

JONbridgerapps:
Closed: content landed on main as 23349ee804 + 834805e717. See the provenance comment above.

PR #5614, opened 2026-07-25
fix(runner): unclip gauge-vacuum plaquette transfer runner stdout

jonathonreilly:
The audit row gauge_vacuum_plaquette_transfer_operator_character_recurrence_note
(fanout 919, load-bearing score 26.3) is held at audited_conditional purely on
evidence completeness, not on the physics:

  "the restricted packet authenticates only a clipped runner-stdout block for
   the load-bearing transfer and source verification ... Repair target: render
   the full SHA-pinned primary-runner stdout and repeat the independent formula
   audit."

  re-audit note: "re-render the complete SHA-pinned primary-runner stdout,
  including all 21 checks and the summary"

The restricted packet renders at most 6,000 characters of runner-cache body.
The committed cache body was 9,169 chars, so the auditor never saw the head of
the execution record. This makes the whole check set unauthenticable for
mechanical reasons.

Printing-only compaction of the runner brings the cache body to 5,176 chars, so
the full record now renders:

- check() gains a fail_detail parameter: a PASS prints only detail, a FAIL
  prints "detail; fail_detail". 16 of 21 call-site detail= strings that merely
  restated their own check label are rerouted to fail_detail=, so they still
  appear verbatim on any failure. The 5 details carrying genuinely new content
  (trace/path-sum, spatial-source, mixed-source, Schur-form, finite-difference
  residuals) stay on the PASS line.
- 4 banner rules ("=" * 88) and 1 blank spacer removed.

No numerical logic, tolerance, assertion, check count, or check ordering is
touched. Verified by an AST-level comparison against origin/main with string
literals and print statements normalized away: the only structural differences
are the check() signature/body and the detail=/fail_detail= keyword renames.

Runner re-run here: exit 0, 21 [PASS], 0 [FAIL], summary line byte-identical
("SUMMARY: THEOREM PASS=0 SUPPORT=21 FAIL=0"). Cache regenerated at the repo's
standard relative path and timeout_sec 120; runner_sha256 matches the on-disk
file and the fresh stdout is contained in the cache body.

The source note is unedited, so note_hash and the dependency graph are
unchanged. The two sibling runners that mention this note reference the note
basename as a registry string, not its stdout, so they are unaffected.

### DIFF
--- a/logs/runner-cache/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.txt
+++ b/logs/runner-cache/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.txt
@@ -1,15 +1,12 @@
 ===== runner cache v1 =====
 runner: scripts/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.py
-runner_sha256: e4da7117aca2f942c8dd862c9e2fd5827f6c235cffca6eb45ec42ffad156690b
+runner_sha256: 56d00d074999bc8fc3ed86eb0b96d4cc8aa8df6d2b5f822fde4e4e378e9b6f0e
 timeout_sec: 120
 exit_code: 0
-elapsed_sec: 0.38
+elapsed_sec: 0.51
 status: ok
 ----- stdout -----
-========================================================================================
 GAUGE-VACUUM WILSON TRANSFER / SPATIAL-MIXED SOURCE SUPPORT
-========================================================================================
-
 Exact SU(3) character data
   tensor levels checked                    = 0..8
   dimension sums                           = [1, 6, 36, 216, 1296, 7776, 46656, 279936, 1679616]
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
review-loop final disposition: **CLOSE AS SUPERSEDED; no PR commit landed**.

Frozen scope was head `6eb930d39f8241094039e4dc14dd6cb0ee08b5a1`, one commit changing only the canonical runner/cache pair. Both paths overlap current `main`: commit `67a71216c21eba03afb714a8cb66fe8a5dc9702f` already landed an independently reviewed complete-cache compaction that satisfies the same audit repair target. At reviewed main snapshot `c094a7e144853c8cfd5792eb56bb5a61bf936c44`, that cache is 5,391 bytes, SHA-pinned, byte-identical to a live run, and contains all 21 verdict lines plus `SUMMARY: THEOREM PASS=0 SUPPORT=21 FAIL=0`.

Review lenses: CodeRunner PASS; PhysicsClaim PASS; ProofObligation CLOSED at the note stated finite-volume boundary; ImportSupport CLEAN; NatureRetention RETAINED SUPPORT for review purposes only; NoGoDiscipline N/A; LabelingConvention N/A; RepoGovernance PASS; AuditCompatibility PASS; MethodologySkill skipped. No source-note claim, dependency, authored grade, or audit verdict changed.

Independent checks: normalized AST comparison to the PR base found the same 21 checks, order, condition expressions, and buckets, with identical non-printing structure; the new fail-detail path was directly exercised. The frozen runner itself also produced 21 PASS / 0 FAIL with exact cache stdout/hash and a 5,176-byte cache. A clean-room implementation importing no candidate code verified both SU(3) dimension recurrences over 81 weights, rebuilt the finite S3 transfer witness with permutation matrices and dense Kronecker products, reproduced trace/path and selected/repeated source identities plus Schur minima, obtained a five-point derivative discrepancy of `8.26e-12`, and verified the S3 character-table convolution eigenvalues are nonnegative. Vocab lint, Python compile, portable-link/no-go scans, and diff checks passed.

The full audit pipeline and `audit_lint.py --strict` passed with no errors. Validation observed `runner_hash_changed:e4da7117->77fa08f6`, reset `gauge_vacuum_plaquette_transfer_operator_character_recurrence_note` to `unaudited`, and placed it in the critical audit queue. Pipeline-generated audit/publication outputs were restored before disposition. The independent audit lane remains the only verdict authority.

The exact fail-closed landing guard was also tested against current main: cherry-picking `6eb930d` conflicts in both the runner and cache, not solely in the citation manifest, so the landing contract forbids pushing it. The salvage pass found no unique science or audit unlock beyond the already-landed main repair. Accordingly this frozen commit is `SALVAGE_REJECT` as redundant, while the current-main repair is review-loop PASS. Per review-loop policy the head branch is intentionally retained because this PR commit itself did not land.

PR #5618, opened 2026-07-25
fix(runner): unclip Wilson SU(3) all-weight positivity runner stdout

jonathonreilly:
The audit row
gauge_wilson_su3_all_weight_positive_coefficient_formal_bridge_note_2026-06-07
(fanout 729) is held at audited_conditional on evidence completeness alone:

  "the restricted runner stdout is clipped before the load-bearing
   Pieri-occurrence and independent-decomposition evidence, so the advertised
   50-pass certificate is not fully inspectable."

  re-audit note: "rerender the complete SHA-pinned runner stdout, including
  Parts 1-3 and all 50 labeled checks"

The restricted packet renders at most 6,000 characters of runner-cache body; the
committed body was 8,726, and the clip fell exactly where the verdict says -
before Parts 1-3. Printing-only compaction brings the body to 5,726, so all 50
labeled checks and every part render.

- check() gains a fail_detail parameter; PASS shows detail, FAIL shows both.
  15 call-site details that restated their own check label are rerouted to
  fail_detail= and still print verbatim on any failure.
- section() reduced from 4 printed lines to 1 across 8 sections (~1.9k of the
  ~3.0k saved); 2 header and 2 footer rules of "=" * 96 and a leading blank
  removed.
- ECHOED_EXPECTATION collapses a rendered detail of the exact form
  "name=V expected=V" to "name=expected=V". It is a regex fullmatch with a
  backreference, so it only fires when the observed and expected values are
  the same string; any disagreement prints uncollapsed. One line collapses
  (found=81 expected=81).
- Part 4 hoists its repeated constant boilerplate ("positive ... c00>0 from
  n=0") into a single stated prefix, with each entry keeping its own c= and d=
  values; "multiplicity=" is abbreviated to "mult=" in 12 Wilson-series lines.
  These two are the only places where a preserved value's spelling changed
  rather than whether it prints; both are trivially revertible.

No numerical logic, tolerance, assertion, check count, or ordering is touched.
Verified by an AST-level comparison against origin/main with string literals and
print statements normalized away: the only structural differences are the
check()/section() bodies, the new collapse regex, and the detail=/fail_detail=
keyword renames. All 30 distinct numeric tokens present in the pre-change stdout
are still present after.

Runner re-run here: exit 0, 50 [PASS], 0 [FAIL], "TOTAL: PASS=50 FAIL=0"
byte-identical. runner_sha256 matches the on-disk file and the fresh stdout is
contained in the cache body.

The source note is unedited, so note_hash and the dependency graph are
unchanged.

### DIFF
--- a/logs/runner-cache/audit_companion_gauge_wilson_su3_all_weight_positive_coefficient_formal_bridge_2026_06_07.txt
+++ b/logs/runner-cache/audit_companion_gauge_wilson_su3_all_weight_positive_coefficient_formal_bridge_2026_06_07.txt
@@ -1,86 +1,68 @@
 ===== runner cache v1 =====
 runner: scripts/audit_companion_gauge_wilson_su3_all_weight_positive_coefficient_formal_bridge_2026_06_07.py
-runner_sha256: 3015a7eeb7fbba47a0222374d0a2f0579c485c9eed629208fe946441d64a07d0
+runner_sha256: 972690018a67c246da26704502768a72b7d4e209b13933ed32af281c0811b19b
 timeout_sec: 120
 exit_code: 0
-elapsed_sec: 0.35
+elapsed_sec: 0.38
 status: ok
 ----- stdout -----
-================================================================================================
 GAUGE_WILSON_SU3_ALL_WEIGHT_POSITIVE_COEFFICIENT_FORMAL_BRIDGE
 Exact Pieri recurrence through level 16; certified coordinate box B_8
-================================================================================================
 
-------------------------------------------------------------------------------------------------
-Part 1: exact SU(3) Pieri identities and full-support tensor recurrence
-------------------------------------------------------------------------------------------------
-[PASS] fundamental Pieri branches conserve dimension locally :: checked all weights in coordinate box B_16
-[PASS] antifundamental Pieri branches conserve dimension locally :: checked all weights in coordinate box B_16
-[PASS] fundamental and antifundamental Pieri rules are exchanged by conjugation :: (p,q) <-> (q,p)
+-- Part 1: exact SU(3) Pieri identities and full-support tensor recurrence
+[PASS] fundamental Pieri branches conserve dimension locally
+[PASS] antifundamental Pieri branches conserve dimension locally
+[PASS] fundamental and antifundamental Pieri rules are exchanged by conjugation
 [PASS] global tensor-power dimension conservation sum m_lambda(n)d_lambda = 6^n :: all levels n=0..16
 [PASS] all computed tensor-product multiplicities are non-negative integers :: 968 exact table entries
 [PASS] full reachable support obeys p+q <= n with no artificial box truncation :: largest table has 153 weights
 
-------------------------------------------------------------------------------------------------
-Part 2: independent exact character decompositions
-------------------------------------------------------------------------------------------------
+-- Part 2: independent exact character decompositions
 [PASS] exact Jacobi-Trudi character decomposition matches the Pieri tables :: independent Q[chi_3,chi_3bar] decomposition at n=0..8
 [PASS] exact Gelfand-Tsetlin torus-character decomposition matches the Pieri tables :: independent basis enumeration and Laurent-character subtraction at n=0..6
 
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
## review-loop PASS — landed direct to `main`

Frozen scope:

- Frozen non-draft head: `ffee8baa48ce58b519540e96a32d28aff0da84c6`
- Original merge base: `452fb7928631e9b36d17acf670795794e8cc9c1c`
- Files: the Wilson SU(3) all-weight positivity companion runner and its paired cache only
- The source theorem note, claim hash, citation dependencies, audit data, and publication surfaces were not edited
- No current-`main` overlap on either frozen path before landing

Review outcome (zero repair iterations):

- `CodeRunnerReviewer`: PASS. Base and candidate emit the same 50 ordered check labels, and every check-condition AST is identical. Live candidate result is `PASS=50 FAIL=0`, exit 0.
- `PhysicsClaimReviewer`: PASS. The positive-theorem claim and its finite-test boundary are unchanged; this is a presentation/transport repair, not a new scientific claim.
- `ProofObligationReviewer`: PASS. The cache body is 5,726 characters and the legacy 6k cache excerpt contains Parts 1–8, all 50 labeled checks, and the final total. Current authoritative live transport is likewise complete.
- `ImportSupportReviewer`: PASS. The only new import is Python stdlib `re`, used by an anchored `fullmatch` backreference; no scientific import, fitted value, premise, or dependency is added.
- `NatureRetentionReviewer`: PASS. All check labels and all numeric token spellings from the old stdout survive while the formerly clipped load-bearing Pieri and independent-decomposition evidence becomes visible.
- `NoGoDisciplineReviewer`: N/A. No source Markdown, no no-go/obstruction artifact, no audit shard, and no negative-claim boundary changed.
- `LabelingConventionReviewer`: PASS. No claim/status labels or authored audit grade changed.
- `RepoGovernanceReviewer`: PASS. `py_compile`, vocabulary (0 violations), cache/source SHA, `git diff --check`, frozen-file-set, and generated-output-stripped gates pass.
- `AuditCompatibilityReviewer`: PASS. The full pipeline passes through stage 18 on refreshed `main`, and strict lint reports zero errors. The changed runner hash requeues the unchanged row as `positive_theorem / unaudited / critical / ready` (queue position 92 in the validation snapshot). No verdict was applied and all generated audit/publication outputs were restored before landing.
- `MethodologySkillReviewer`: N/A; no methodology-skill file changed.

Failure-path and transport checks:

- All 15 `detail` → `fail_detail` migrations preserve the exact expression AST.
- Forced-failure probes show fail-only diagnostics, combined value+diagnostic output, and unequal observed/expected values all remain visible. The equality compactor fires only on the exact full-string backreference case.
- Cache runner SHA equals the on-disk source SHA: `972690018a67c246da26704502768a72b7d4e209b13933ed32af281c0811b19b`.
- Cache stdout matches the fresh live execution modulo the cache delimiter's separating newline.
- Direct consumers also pass: residual-environment companion `33/33`; native positive-class runner `8/8`.

Independent math certificate (does not import the PR runner):

- Symbolically checked both SU(3) Pieri dimension identities, including boundary cases.
- Reimplemented the tensor recurrence independently through level 20; dimension conservation holds and every weight in an unseen `B_10` first occurs at `p+q` with multiplicity `binomial(p+q,p)`.
- Direct Gelfand–Tsetlin pattern counts match the Weyl dimensions on `B_10`.
- Direct Schur index contractions and dimension cancellation pass for dimensions 1–21.
- Finite-packet stabilization passes on 25 arbitrary rational channels.

The first validation run exposed a pre-existing citation-manifest count drift unrelated to this runner/cache-only PR, so landing was held fail-closed. After PR #5621 supplied the responsible manifest acknowledgment on `main`, #5618 was rebuilt on that exact tip and the complete pipeline was rerun successfully. No manifest change was absorbed into #5618.

Landed `main` SHA: `0d9278ea534ebcd6d287ee343e46c843be80e500` (verified contained in fetched `origin/main`). Independent re-audit remains the audit lane's responsibility.

PR #5620, opened 2026-07-25
[physics-loop] readout-limits-forced-local-law block01 — bounded theorem: the readout clause forbids fields and pure numbers, and forces the local law to the Laplacian ray (Cycle 697)

JONbridgerapps:
## What

The physical-observable / source-action / carrier bridge is the blocking cause
in gravity, mass, and causal-time simultaneously. It is a program, not a cycle.
This PR scopes it by settling what the axioms' readout clause **does** decide,
and proving exactly what it **cannot** decide. 11 PASS / 0 FAIL, exit 0, exact
integer and `Fraction` arithmetic, a negative control in every row, runner
imports nothing from the repository.

## The four results

**L1 — the local record-sourced law is forced to two constants.** For a law
acting by a displacement kernel, Record additivity gives linearity, Lattice
gives translation and *proper* cubic covariance, and the invariant kernel space
on a rotation-closed support set has dimension exactly the number of
proper-octahedral orbits in it. Solved exactly over `Q` at six radii and matched
against an independently computed orbit count:

| `\|v\|^2 <=` | 1 | 2 | 3 | 4 | 5 | 6 |
|---|---|---|---|---|---|---|
| points | 7 | 19 | 27 | 33 | 57 | 81 |
| orbits = invariant dimension | **2** | 3 | 4 | 5 | 6 | 7 |

At range 1 the family is exactly `span{I, Delta}`. Adding one named condition —
the law annihilates a constant field — leaves **one** dimension, spanned by the
Laplacian. Dropping the rotations moves the range-1 dimension from 2 to 7; the
anisotropic forward difference `T_e1 - I` is local and is correctly excluded;
the same offset condition at range `sqrt(2)` leaves 2, so the collapse to a ray
is a range-1 fact.

**L2 — a Record readout is position-blind, so it cannot be a field.** "Determined
by record content alone" forbids dependence on the record's site, which forces
the displacement kernel to be constant and the field to take the same value
everywhere. Exact nullity 1 on the blindness system, with the Laplacian's
three-valued point-source response as the negative control. Consequence: every
gravity runner in the lane stipulates its observable because on the current
axiom surface there is nothing else to do. The stipulation is forced, not a
defect of those runners.

**L3 — no nonzero Record readout is dimensionless.** Duplicate a configuration
by a distant lattice translate: additivity doubles the readout, dimensionlessness
fixes it, so it is zero. The load-bearing step is that the duplication is legal,
and it is checked at the level of full nearest-neighbour **content** conditions
(not occupancy), since Admissibility reads conditions. The second reading of
"dimensionless" — invariance under rescaling the scalar unit — is checked
separately and gives the same conclusion, with a vacuous-condition control.

**L4 — the two limits are independent.** A position-blind readout still doubles;
a duplication-invariant ratio is still built from position-blind ingredients.

## The finding the brief asked me to verify

The brief asked whether the gravity five-judge panel and the AC R-eta obligation
describe the same missing object. **On the readout side they do not.** Gravity
needs L2 repaired (a position-dependent readout); the AC eta obligation needs L3
repaired (an intensive readout). L4 shows neither repair supplies the other.
They share only the third named object, the source action. So that part of the
program is **two** sub-objects, not one, and they can be attacked separately.

## What this constrains about the AC live routes

L3 says no single additive readout can carry a pure number. Route (b) — an
occurrence-lane clock/event-rate **ratio** — is exactly the admissible shape.
Route (a), an angle-native identity whose output is `Phi = S_sum = 2/3`, must
therefore itself be a ratio of two readouts or it cannot exist on this surface.

The escape is named rather than dismissed: supplying a unit makes the
record-count readout a candidate reference, which is a convention, not a new
axiom — structurally the same discharge the r=1/2 lane already used for
count-once. But the choice is load-bearing, not cosmetic: on the same fixture the
same numerator gives `6/35` against the count reference and `3/28` against
another admissible one. A normalization derivation has to *select* the
reference.

## Scope — what this deliberately does NOT do

- It does **not** repair the four gravity rows. Their runners operate on
  generated DAGs with a stipulated growth rule, not on `Z^3` record
  configurations, so L1 does not apply to them as written.
- It does not supply the site-anchored readout, the reference normalization, the
  carrier, the source action, a dynamics, a Hamiltonian, or a formation rule.
- It does not derive range-1 locality or offset-insensitivity. Both are named
  conditions, stated as such in the note, the certificate, and the residual
  table.
- It changes no row, lane, or obligation status. Authority none, audit unset.
- It awards itself no N1–N8 verdict. L2 and L3 are negative results; that
  verdict is reviewer-owned.

Answered in writing before this PR at
`.claude/science/physics-loops/toe-retention-source-action-20260725/VALUE_GATE_697.md`.
V1 quotes the three verdict obstructions; the claim is narrowing, not closure.
V2: three results not on `main`; the prior-art sweep found that
`STATIC_SOURCE_READOUT_I1_ACCEPTED_PREMISE_BRIDGE_BOUNDED_NOTE_2026-05-27.md`
*imports* the graph-Laplacian Green's-function identification as accepted-premise
packet P1 — the surface L1 narrows. V3: the mathematics is elementary but which
axiom clause licenses which hypothesis is not available to the audit lane, whose
own verdicts record the opposite. V4: dimension counts with a failing control at
every step. V5: closest landed cycles are 693 (no site index, no symmetry group,
no operator) and 692 (enumeration on one surface, which explicitly disclaimed the
general classification L3 supplies).

## No-Go Discipline (N1–N8)

Full checklist at
`.claude/science/physics-loops/toe-retention-source-action-20260725/NO_GO_DISCIPLINE_CHECKLIST_697.md`:
six attack routes enumerated against each of L2 and L3, wall independence checked
in the runner, hidden-wall scan promoting range-1 locality and offset-insensitivity
to explicit conditions, five-resolution rhetoric audit, two partial-closure paths
found and **named as live routes rather than as new-axiom requirements**, and a
steelman the note is written to survive rather than defeat. Three live escapes are
recorded: infinite collections, a supplied reference, and the site anchor itself.

## Provenance

Cold-run in an isolated worktree detached at the first commit, never in the
working tree. The receipt's `runner_sha256` equals the sha256 of the committed
blob (`4d79044259...`), verified with `git show`.

Authority: none. Audit: unset. No new axiom or primitive is proposed or adopted.
Independent audit is still required before the repo may treat any of this as
retained-grade.


JONbridgerapps:
Review-loop provenance (Cycle 697)

Disposition: rejected as submitted; durable positive core salvaged and landed to `main` at `a89337f0bdf2331b246cd8a75c98230fcbaa8e34`.

Cross-family review: the source package was authored by Claude Opus 5 and independently reviewed in a Codex review-loop reviewer seat. Frozen source head: `56352c5a9323b65655b0c2fc29f2c06425650204`.

The original L2/L3 negative package did not pass review. In particular, L3 conflated dimensionless with duplication-invariant/intensive readout (a pure count is a direct counterexample); C8 did not quantify or implement an arbitrary nearest-neighbour admissibility rule; C9 relabeled the duplication comparison as position freedom; and C10 imposed the scaling law it purported to derive. The N1-N8 reviewer disposition was FAIL. L1 also hid rational/module, linear-convolution, finite-support, and law-covariance premises behind Record language. No gravity row was repaired.

Fix Policy retained only the exact C1-C5 algebraic core, with every extra premise explicit and all negative/readout/gravity claims removed:

- `docs/PROPER_CUBIC_FINITE_SUPPORT_LINEAR_KERNEL_CLASSIFICATION_BOUNDED_THEOREM_NOTE_2026-07-25.md`
- `scripts/proper_cubic_finite_support_linear_kernel_classification_2026_07_25.py`
- the required one-node/one-edge citation-graph manifest acknowledgment

Focused confirmation: exact runner 5 PASS / 0 FAIL; independent Burnside-versus-rational-nullity recomputation agrees for radii 1-6; full audit pipeline PASS; strict audit lint has no errors; repository invariants PASS; vocabulary and portable-link checks PASS; manifest adds exactly one node and one edge with no existing-node rewires. The committed runner has no `ACCEPTED_<cycle>_COMMIT`, `definitive_run_HEAD`, or ancestry constant.

Claim boundary remains authority `none`, audit `unset`, nothing promoted, and no axiom or primitive added. Independent post-landing audit remains required. Closing this PR rather than merging its overclaiming source commits; the reviewed bounded salvage is already contained in `main`.

PR #5611, opened 2026-07-25
methodology(physics-loop): sweep the repo for prior art before running a program

JONbridgerapps:
## Why

Review of a 2026-07-25 physics-loop cycle found that its claimed-new
finite-composition algebraic component was already proved on `main`, for a
strictly more general finite-site statement, in
`GENERATED_FINITE_COMPOSITION_MINIMALITY_THEOREM_2026-07-13.md`.

A second lane showed the complementary no-go failure mode: headlines alone hid
both the exact homogeneous symmetry-only obstruction and the separately live
inhomogeneous readout route.

## What changes

All edits remain in place, so workflow step numbering and existing
cross-references stay stable.

### Step 2: ground, then sweep for prior art

- Refresh `origin/main` immediately before each sweep and record the searched
  commit.
- Search the theorem statement with multiple nouns, reverse order, notation
  variants, synonyms, filenames, derivation obligations, and sharded
  claim/no-go ledger rows.
- Read each matched statement, hypotheses, proof or decisive artifact, current
  audit/effective status, and active review findings.
- Record nonmatching/context-only hits separately, then classify the target as
  already proved on matching premises, already refuted at matching scope, or
  open after matched-hit review.
- If a matched refutation has compatible escapes, hand them to step 4; if it
  has none, stop and choose another target.

### Step 4: read no-go proofs, not headlines

`NO_GO_LEDGER.md` must record exact quantified scope, supplied premises,
current claim/audit/effective status, active review findings, and every named
live route. An unaudited or non-clean no-go is prior-art evidence, not
authority that a route is dead.

Escapes are classified as derivation, new-governance, or new-primitive
approval/registration routes. The last two are unavailable to physics-loop;
already registered approved primitives remain usable within their declared
boundaries after the primitive-registry check.

The h-class example is scoped precisely: homogeneous Record additivity plus
`C3` covariance leaves a free one-parameter family, with
`alpha = 0, 1/9, 1/3, 1, 2/27` as distinct witnesses and only
`alpha = 2/27` the target member. This defeats the repeated symmetry-only pin;
it does not close the separately named inhomogeneous readout theorem.

### Promotion Value Gate

- V2 now requires the searched commit, commands, hits, and matched-hit
  classifications.
- V5 requires a freshly fetched `origin/main`, not only the current campaign.
- The stale internal cross-reference is corrected: the gate lives in workflow
  step 8.

This is a methodology PR, not a science cycle, so it does not claim a V1--V5
science-result certificate of its own.

## Scope and authority

[...]

PR #5624, opened 2026-07-25
[physics-loop] content-pair-kernel-channel-census block03 (stacked on #5623) — bounded theorem: covariance permits exactly five nearest-neighbour two-body couplings, one of them only because the rotations are proper (Cycle 699)

JONbridgerapps:
## Stacked PR

Base is `physics-loop/pair-kernel-minimal-position-extension-20260725` (PR
#5623), which is itself stacked on #5620. The runner depends on neither; the
stacking is for the markdown citations.

## What

Cycle 698 found that the content-blind nearest-neighbour two-body kernel is
exactly one constant. A record does not carry only a position — it locks an
element of the one-site possibility domain. This cycle carries that content
through the same classification and **counts what covariance permits**. It
selects nothing. 7 PASS / 0 FAIL, exit 0, exact integer and `Fraction`
arithmetic, runner imports nothing from the repository.

## The count

Object: a real trilinear form on (6 face displacements) × C × C with
`C = span_R{I, σ₁, σ₂, σ₃}`, under joint covariance
`K(Rv, R.o, R.o') = K(v, o, o')` and, separately, record exchange
`K(-v, o', o) = K(v, o, o')`.

| stage | dimension |
|---|---|
| unconstrained | 96 |
| after joint cubic covariance | **6** |
| after record exchange as well | **5** |

The 6 is computed **twice, independently** — as an exact nullspace over `Q`,
and as the Burnside character average `(1/24) Σ_R fix₆(R)(1 + tr R)²`. A
linear-algebra count and a character-theoretic count agree; neither is derived
from the other. Without the rotations all 96 stay free.

## The census, and an exhibited basis

| channel | dim |
|---|---|
| density–density | 1 (= cycle 698's constant, recovered) |
| density–spin + spin–density | 1 + 1, **identified to 1** by exchange |
| spin–spin | 3 |
| **total after exchange** | **5** |

The whole effect of record exchange is to identify the two mixed channels — it
leaves the pure channels untouched. The three spin–spin invariants are
exhibited explicitly and each verified invariant against all 24 group elements
at all 54 coefficient positions:

```text
isotropic    S · S'
bond-axis    (S · v)(S' · v)
chiral       v · (S × S')
```

These are the familiar isotropic, pseudo-dipolar, and antisymmetric bond forms.
That the framework's own covariance produces exactly this triple and no fourth
is the content.

## What the word "proper" is doing

The chiral form is invariant precisely because `det R = +1`. As a **labelled
counterfactual**, extending to all 48 cubic elements with the Pauli vector
treated as axial kills exactly the chiral coupling and leaves the other two. So
the Lattice axiom's restriction to *proper* rotations is load-bearing for
exactly one of the five.

This defers to
`ADMISSIBILITY_RULE_COVARIANCE_EXTENSION_CLASSIFICATION_..._2026-07-03.md`,
which owns the framework's principled improper-action question for the
admissibility rule and uses a complex-antilinear action re-earned from
`Cl(3,0)`. Different object, different convention; cited in the note.

## Three named conditions, stated not claimed

Flagged in the note because they are load-bearing and none is axiom text:
(1) rotations act on the possibility domain **by spin conjugation** — the
natural identification for the `M_2(C)`/`Cl(3,0)` presentation, but a named
condition; (2) contents are taken in the **Hermitian real form**; (3) record
exchange is imposed as stated, and is reported separately throughout so a
reader who rejects it reads 6 instead of 5.

## Scope — what this does NOT do

- Adopts no source action, selects no coupling, fixes no value, sign, or unit.
  The census is a **budget**.
- Does not claim the physical range is nearest-neighbour.
- Does not derive that the two-body extension is licensed at all — cycle 698
  showed strict additivity forbids it; this counts what it would contain if
  licensed, and says so.
- Repairs no gravity row or AC obligation; changes no status.

## Prior-art sweep (workflow step 2, as landed in #5611)

**Searched commit `ab6e6bd8d96fb5b7f1fe6712a7bb426c8df1c1e1`**, refreshed
immediately before the sweep. Eight searches with commands, hits, and per-hit
classifications are tabulated in
`.claude/science/physics-loops/toe-retention-source-action-20260725/ROUTE_PORTFOLIO.md`,
which also reconstructs the 697 and 698 sweeps in the same format. Title scan
and 6-shell coupling searches return nothing; the one adjacent hit is read in
full and distinguished above.

## Cluster cap — evaluator run, verdict recorded

Third PR in this parent family, so the evaluator was **mandatory and was run
before opening**, as an independent codex `gpt-5.6-sol` xhigh seat in an
isolated worktree at the proposed head, with no write access.

**Verdict: `OPEN`, "though narrowly."** It judged the load-bearing delta real
(96-dimensional representation problem vs 6 scalar coefficients), the claim
type distinct (representation-theoretic census vs minimality result), and
independent review to have substantive new work — namely the three
identifications, which are now named in the note as conditions.

Full V1–V5 at `VALUE_GATE_699.md`.

## Provenance

Cold-run in an isolated worktree detached at the first commit. The receipt's
`runner_sha256` equals the sha256 of the committed blob (`13f2896235...`),
verified with `git show`.

Authority: none. Audit: unset. No new axiom or primitive. Independent audit is
still required before the repo may treat any of this as retained-grade.


JONbridgerapps:
Superseded. The campaign's first block (#5620) was rejected as submitted and only its abstract kernel classification was salvaged and landed (`a89337f0bd`), which deleted this PR's transitive base and auto-closed #5623.

Cycles 698 and 699 have been rebased onto `main` off the rejected commits, revised to apply the #5620 findings, and consolidated into a single PR rather than a rebuilt stack: the two cycles are tightly coupled, and one review pass is a lower burden on this lane than two serialized ones. No content is dropped.

PR #5630, opened 2026-07-25
docs(flavor): scope r=1/2 language to the route, not the setting

jonathonreilly:
## What this responds to

Owner directive, 2026-07-17:

> r=1/2 is not prohibited, just we shouldnt assume we have to force it / if the repo says otherwise its wrong / we can force it if we find a path but its not necessarly our objective

A sweep of the r=1/2 note family classified its language four ways: (A) blanket prohibition on forcing, (B) route-local scoping, (C) universal-forcing falsification, (D) objective-framing that makes a selector a program requirement. (B) and (C) are correct as written and are left alone — (C) in particular is physics, since quarks register Q ~ 0.85/0.73 and neutrinos Q ~ 1/3, so *universal* forcing really is falsified. (A) and (D) are the two defects. This PR repairs one of each.

## Changes

**`docs/KOIDE_Q_DELTA_LINKING_RELATION_THEOREM_NOTE_2026-04-20.md`** — kind (A). The decoration-attachment section explained why it does not attempt the derive-from-retained-premises route, and ended with *"so a derivation forcing it would overreach"*. The bare indefinite subject makes that a standing claim about any forcing derivation rather than about this row's premises. Rewritten to scope the overreach to this row and to state that a future derivation of the registered pattern is not excluded.

**`docs/FLAVOR_TRACE_VS_CENTER_DISSOLVES_NOTE_2026-05-30.md`** — kind (D), two spots. The modulus-selector residual asked *"why `r = 1/2` should be selected"*, presupposing that it is; now *"whether and how"*. The Next Useful Path section called a native selector theorem *"the highest-value closure route"* that records/KMS/matter-action dynamics *"would need to select"* — ranking the selector as the program's top objective and framing it as a requirement. Now one open route, a permitted result rather than a requirement, referencing the landed stationary-point framing.

## Cost

Zero. Both rows are `unaudited` with no retained descendants, so re-opening their audits via the note_hash mechanism costs nothing. `FLAVOR_TRACE_VS_CENTER` has no runner references at all. Three runners reference the KOIDE note; a grep for every phrase in the edited sentences (`repair routes`, `second route`, `spectral ratio equal`, `record-side science`, `would overreach`, `dial setting`) returns zero hits in all three, and all three were re-run:

| runner | result |
|---|---|
| `frontier_koide_q_delta_formal_ratio_repair.py` | `TOTAL: PASS=159 FAIL=0` |
| `frontier_koide_q_delta_linking_relation.py` | `SUMMARY: PASS = 15 FAIL = 0` |
| `frontier_koide_q_delta_linking_relation_theorem_note_2026-04-20_hostile_audit_findings.py` | `1 PASS / 1 FAIL` |

The third one's FAIL is **pre-existing and unrelated**: it expects the note to cite `scalar_selector_remaining_open_imports_2026-04-20`, which the 2026-05-25 formal-ratio repair withdrew from the binding claim. Running that runner against a clean `origin/main` checkout produces byte-identical output. Reported rather than repaired — a stale runner expectation is its own unit, and rewriting a gate to go green here would be exactly the wrong move.

## Dependency graph

Unchanged. The stationary-point note is referenced in **backticks**, never as a markdown link, so `build_citation_graph.py`'s `LINK_RE` produces no new edge. Verified programmatically: the extracted link inventory of both files is identical before and after.

## Scope

N5 rhetoric alignment only. No claim, theorem, numeric, runner, or cache changes. The two notes are independent of each other and can be split if the reviewer prefers one note per PR; they are batched because they are the same defect class under the same directive.

### DIFF
--- a/docs/FLAVOR_TRACE_VS_CENTER_DISSOLVES_NOTE_2026-05-30.md
+++ b/docs/FLAVOR_TRACE_VS_CENTER_DISSOLVES_NOTE_2026-05-30.md
@@ -163,16 +163,20 @@ The current source packet does not close the science. The residuals are:
   square-root readout rather than a singular-value/Yukawa readout;
 - **block selector:** why the physical functional should be trace/dimension,
   block-count, or another native reference state;
-- **modulus selector:** why `r = 1/2` should be selected rather than merely
-  matched by `Q_trace = 2/3`;
+- **modulus selector:** whether and how `r = 1/2` is selected rather than
+  merely matched by `Q_trace = 2/3`;
 - **phase/sign chamber:** how the broader `b = |b|e^{i phi}` chamber is selected.
 
 No new axiom, measured mass input, or audit verdict is introduced here.
 
 ## Next Useful Path
 
-The highest-value closure route remains a native selector theorem: records,
-KMS/modular structure, or matter-action dynamics would need to select both the
-signed/Hermitian readout and the `r = 1/2` modulus. A theorem selecting the
+One open route is a native selector theorem: records, KMS/modular structure, or
+matter-action dynamics that select both the signed/Hermitian readout and the
+`r = 1/2` modulus. Such a theorem is a permitted result rather than a
+requirement on this packet: per
+`FLAVOR_R_HALF_IS_A_STATIONARY_POINT_NOT_FORCED_2026-06-02.md`, the special
+values of `r` are distinguished points of one family, carrying different
+sectors, not one value the framework must select. A theorem selecting the
 dimension/trace reading instead would point the packet to `r = 1`/`Q = 1`, not
 to the charged-lepton Koide value.
--- a/docs/KOIDE_Q_DELTA_LINKING_RELATION_THEOREM_NOTE_2026-04-20.md
+++ b/docs/KOIDE_Q_DELTA_LINKING_RELATION_THEOREM_NOTE_2026-04-20.md
@@ -151,9 +151,11 @@ retained parent with independent content. This section takes the second
 route. The first route is not attempted here: under the framework's measure
 discipline the block-weight setting that makes the spectral ratio equal
 `2/3` is a registered per-sector pattern (a dial setting, alongside the
-other sectors' settings), so a derivation forcing it would overreach; any
-future derivation of the registered pattern itself would arrive through new
-record-side science, not through this formal row.
+other sectors' settings), so forcing it from the premises available in this
+row would overreach. That scoping is about this row, not about the setting:
+a future derivation of the registered pattern itself is not excluded here,
+and would arrive through new record-side science rather than through this
+formal row.
 
 ### Cited parent authority (one hop)
 


JONbridgerapps:
## Review-loop provenance

**Disposition:** `PASS WITH BOUNDED CLAIMS`

- **Landed on `main`:** `b9f983c663cc83b46639bdc55a0d9f3499987c28`
- **Reviewed PR source head:** `49acffb28792faf5b702f0049c84424e20c989a3`
- **Landed source commit:** `b21899405e35338426f627d13a5324d4791e2a72`
- **Landed repair commit:** `b9f983c663cc83b46639bdc55a0d9f3499987c28`
- **Review iterations:** 2 (one narrow repair pass followed by focused confirmation)
- **Review lane:** review-loop, Codex GPT-5.6-Sol at configured maximum reasoning

### Findings repaired

1. The new flavor paragraph consumed the stationary-point note as authority but cited it only in backticks, hiding the load-bearing dependency from the citation graph. It is now a Markdown link, with the graph manifest refreshed.
2. “Carrying different sectors” exceeded the cited note's bounded algebraic scope. The final text distinguishes algebraic points while leaving physical sector/lane assignment and the charged-lepton selector open; it explicitly neither derives nor forbids a future `r = 1/2` selector.
3. The Koide decoration retained admission-era “registered per-sector pattern” language and forecast record-side science as the exclusive future mechanism. The final text is row-local and mechanism-neutral: the formal row contains no selector, while a separately supported matter-action/measure, record/KMS, or other physical mechanism may derive this or another value.

No-Go Discipline N1–N8 was applied to the changed negative rhetoric. The result passes only at the narrow row-local boundary: the edited rows do not themselves supply a selector. No framework-wide no-selector or no-future-derivation claim ships.

### Independent proof and runner checks

- Reconstructed the two-sector entropy independently:
  `dS/dr = -2 log(2r)/(1+2r)^2`,
  `S(1/2)=log 2`, and `S''(1/2)=-1`.
- Rechecked `Delta_d = Q_d/d` exactly for `d=1..100`; the alternative
  `Q'_d=(d-1)/d` coincides with `2/d` only at `d=3`.
- Independently recovered
  `Q_trace=(a²+2b²)/(3a²)` and
  `Q_block=(a²+b²)/(3a²)`, giving `2/3` and `1/2` respectively at
  `b²/a²=1/2`.
- Paired/interacting runners passed:
  flavor `17/0`, formal ratio `159/0`, historical linking `15/0`,
  dependency classifier `10/0`, and selector sidecar `87/0`.

### Repository and audit compatibility

- Vocabulary lint (fix and report-only): zero violations.
- Diff, portable-link, authored-verdict, and mechanical no-go trigger gates: clean.
- The validation pipeline ingested the new dependency and kept both touched rows `unaudited`.
  - `flavor_trace_vs_center_dissolves_note_2026-05-30` remains a `bounded_theorem`; the new stationary-point dependency is visible, and the row remains not-ready behind its pre-existing open dependency.
  - `koide_q_delta_linking_relation_theorem_note_2026-04-20` remains a ready critical `decoration` candidate under its retained parent.
- Strict audit lint: no errors (pre-existing warnings/notices remain).
- Pipeline-generated ledger, queue, dispatch, status, and publication outputs were restored. Only the required citation-graph manifest acknowledgment co-landed: edge count `15200 -> 15201`, with the flavor row's out-degree `1 -> 2`.
- Repository invariants with enforced links:
  `PASS rows=3896 retained_grade=499 link_violations=0 class_f_violations=0 graph_delta=acknowledged`.

No audit verdict, audit-status promotion, effective-status result, auditor transcript, or `apply_audit.py` output was authored or landed. The work supports bounded/formal source claims pending independent audit; it does not confer retained-grade or Nature-ready status.

The PR is being closed rather than GitHub-merged because review-loop landed the reviewed source and repair commits directly through its fail-closed `main` push mechanism.

PR #5631, opened 2026-07-25
fix(unit-singlet): cache the full N1-N8 discipline packet in one default run

jonathonreilly:
## Verdict this responds to

`unit_singlet_overlap_narrow_theorem_note_2026-05-02` — `effective_status: audited_failed`, `criticality: critical`, transitive fanout **936**, `direct_in_degree: 5`.

`notes_for_re_audit_if_any`, verbatim:

> Capture completed hostile and independent mode stdout, add five distinct live alternative-route classes and all five N5 resolution checks, disposition the supplied N6 and N8 candidates, and re-audit.

## Root cause

The failure was evidentiary, not mathematical. The note's `negative_assertion_classes` is `["derived_no_go_boundary"]`, which invokes the N1–N8 discipline gate — but `main()` defaulted `--mode` to `"normal"`, and `runner_cache.execute_runner` invokes the runner with **no command-line arguments**. Only section A ever reached the canonical exact-SHA cache, so the independent and hostile evidence the note cited existed nowhere the auditor could read. A related symptom: bare `PASS:` labels produced `runner_check_breakdown: {A:0, B:0, C:0, D:0, total_pass:0}` despite `load_bearing_step_class: 'A'`.

## Clause-by-clause response

| Verdict clause | Change |
|---|---|
| "Capture completed hostile and independent mode stdout" | `--mode` now defaults to `"all"`; one zero-argv invocation runs A + B + C + D, all in the cache |
| "five distinct live alternative-route classes" | `D1`–`D5`: `positivity_relaxation`, `hermiticity_relaxation`, `centrality_relaxation`, `normalization_substitution`, `parameter_indexed_inference` — each `mech=` / `try=` / `out=BLOCKED` on one contiguous live line |
| "all five N5 resolution checks" | `D7`–`D11`, one per class; `resolution_classes_checked` printed as the exact five-element list; the resolved sentence now exists byte-for-byte in both note and cache |
| "disposition the supplied N6 and N8 candidates" | two N6 lines covering all 7 supplied candidates; one N8 line covering 3 prior audits + 3 retirement entries + 67 physics-loop entries |

Plus `D6` = the N2 wall (`hypothesis_triple_minimality`) as a four-witness bit-pattern gate, and `A<n>/B<n>/C<n>/D<n>` labels so `runner_check_breakdown` resolves a class.

## Gate discrimination

Both new gates fail if the object is wrong, rather than holding by construction:

- `D6` recomputes `(hermitian, psd, central, hs_unit)` for `W1 = -I_6/sqrt(6)` → `1011`, `W2 = i I_6/sqrt(6)` → `0011`, `W3 = diag(1,0,0,0,0,0)` → `1101`, `W4 = I_6/6` → `1110`, and **fails if any single bit differs**.
- `B4` was strengthened from `min_eigenvalue > 0` to `|min_eigenvalue - 1/sqrt(n)| < 2e-10`, so an independently reconstructed **wrong** value now fails instead of merely testing positive.

Section B also reports tolerance bounds against a fixed `2e-10` threshold rather than printing platform-dependent noise digits.

Honest scope note: only four of the five N1 routes are blocked by the N2 minimality wall. `parameter_indexed_inference` is blocked by a **different and independent** wall — the absent physical bridge, recorded as the N4 residual. The note states this explicitly rather than folding it into one wall.

## Numbers

- Cache: **5802 chars** against the 6000 `cache_excerpt_for_audit` ceiling; no `[runner cache excerpt clipped;` marker.
- `TOTAL: PASS=33, FAIL=0` = 8 (A) + 5 (B) + 9 (C) + 11 (D).
- Sibling runners consuming this note's result, both re-run after the note edit: `scripts/yt_ward_step3_same_1pi_construction_2026_05_10.py` → `TOTAL: PASS=74 FAIL=0`; `scripts/frontier_yt_ward_step3_contact_4fermion_vanishing.py` → `PASS: 16, FAIL: 0`. Both exit 0.

## What did not change

The theorem itself. Section A still derives the one-dimensional centralizer from matrix-unit commutators, solves the Hilbert–Schmidt norm equation, and selects `c = 1/sqrt(n)` by positivity at `n = 1, 2, 3, 6`; only the reporting is aggregated to cross-dimension gates, and each gate still recomputes the full per-dimension object. `deps: []` is preserved — the only markdown link in the note points at its own runner, so no note-to-note dependency edge is created and no citation-graph manifest commit is needed.


JONbridgerapps:
`review-loop` terminal disposition: **FAIL — close without landing; preserve the head branch.**

The finite-dimensional theorem and its implementation are sound. Direct proof review and an implementation-independent NumPy/Kronecker reconstruction both recover a one-dimensional centralizer at `n = 1, 2, 3, 4, 6`; the target matrix, hypothesis-necessity witnesses, and the noncentral selected-overlap bypass all pass. The production runner reports `PASS=33, FAIL=0` in its default packet (`normal 8/0`, `independent 5/0`, `hostile 9/0`, `packet 11/0`).

The package is nevertheless non-landable because N7 lacks the independent evidence surface required by the current audit machinery:

- `docs/audit/scripts/no_go_discipline_gate.py::_validate_n7` requires N7 resolution evidence on a path distinct from the N1 runner stdout and authenticated as `runner_stdout_independent` (or as retained/accepted authority).
- The primary runner's `--mode independent` output is still primary-runner evidence at `audit-packet://runner-stdout/...`; placing it in the default cache does not make it a distinct authenticated N7 source.
- `scripts/codex_audit_runner.py` emits `runner_stdout_independent` only by executing a helper runner whose source carries the `N7_STEELMAN_RESOLUTION` marker.
- This claim has no helper runner and no applicable retained/accepted authority that independently resolves the matrix-centrality steelman.

Adding that helper artifact, or changing the audit tooling/control data, would exceed the frozen original three-file scope. Weakening N7 would be dishonest. Five other review findings (N5 semantic relabeling, N2 wall interactions, premise-registry typing, N7 scope, and N8 echo disclosure) were repairable locally, but those tentative repairs do not cure the missing-artifact gate and are not being committed as a partial packet. The theorem core already exists on `main`, so no narrower durable salvage remains.

Checks completed on the diagnostic repair state:

- independent proof/math check: PASS;
- production runner and all four modes: PASS;
- both exact sibling runners: PASS;
- `py_compile`, vocabulary lint, and `git diff --check`: PASS;
- full audit pipeline and strict lint: exit 0; the target remained `unaudited` and was queued for independent audit.

Generated audit/status outputs were restored after inspection. No audit verdict or status promotion was applied. No durable PR content was landed, so the contributor branch is intentionally preserved as the recovery handle.

PR #5632, opened 2026-07-25
feat(oh-schur): derive the boundary source independently of the target trace

jonathonreilly:
## What verdict this responds to

The terminal audit obligation on `oh_schur_boundary_action_note`:

> missing_bridge_theorem: derive the boundary source j independently from the microscopic source/action, then rerun stationarity without reconstructing j from the target trace.

This PR lands the bridge theorem itself. A second PR follows once this lands, adding the parent-side link and the dated downstream-hygiene line that `docs/audit/README.md:348-354` requires of a dependent-side repair.

## What changed

Source-only triple, all three files new:

- `docs/OH_SCHUR_INDEPENDENT_BOUNDARY_SOURCE_BRIDGE_THEOREM_NOTE_2026-07-25.md`
- `scripts/frontier_oh_schur_independent_boundary_source_bridge.py`
- `logs/runner-cache/frontier_oh_schur_independent_boundary_source_bridge.txt`

The interior sites are partitioned into shell `t`, the sites strictly inside it `I`, and the sites strictly outside it `b`. The separation lemma `H_Ib = H_bI = 0` follows from 6-NN adjacency and is gated numerically. Eliminating both `I` and `b` gives

```
j_micro = rho_t - H_tI H_II^-1 rho_I - H_tb H_bb^-1 rho_b
S       = H_tt - H_tI H_II^-1 H_It - H_tb H_bb^-1 H_bt
        = Lambda_R - H_tI H_II^-1 H_It
S f     = j_micro
```

`independent_boundary_source` takes `rho` and the operator blocks and never a trace. Trace-independence is therefore a property of which arguments the construction reads, established by inspection of the source; no gate can detect a source reconstructed from the trace, because `j_micro = S f_true` identically. The runner's legend says exactly this rather than claiming a gate proves it.

## Runner numbers

`TOTAL: PASS=51 FAIL=0`, 21 named gates over box sizes 13/15/17 (fifteen at each size, six at size 15 only), exit 0, empty stderr, byte-identical on repeat runs.

| | check | size 13 | size 15 | size 17 |
|---|---|---|---|---|
| 1 | `nnz(H_Ib)`, `nnz(H_bI)` | 0, 0 | 0, 0 | 0, 0 |
| 2 | `lambda_min(S)` | 6.049262e-01 | 5.507656e-01 | 5.118568e-01 |
| 3 | `joint` = `S` vs single Schur over `I u b` | 8.881784e-16 | 8.881784e-16 | 8.881784e-16 |
| 4a-4d | reconstruction err, prescribed sources | 3.5e-15 .. 1.1e-14 | 5.8e-15 .. 1.4e-14 | 6.5e-15 .. 1.1e-14 |
| 4e-4f | reconstruction err, two exact classes | -- | 1.280216e-14, 1.378019e-14 | -- |
| 5a-5c | rejectors, drop each `j_micro` term | 2.7e-01 .. 1.0e+00 | 2.5e-01 .. 6.5e-01 | 3.1e-01 .. 7.3e-01 |
| 6a-6c | rejectors, `Lambda_R` in place of `S` | 2.387521e-01 | 5.1e-01 .. 7.2e-01 | 3.528884e-01 |
| 7b-7d | rel. interior Schur term | 1.901801e-01 | 5.8e-01 .. 2.8e+00 | 3.987182e-01 |
| 8b | interior perturbation moves `f_pred` | 5.776783e-04 | 4.167462e-04 | 4.542326e-04 |

Row 7a is a labelled tautology exhibit: `||Lambda_R g - (H_0 u_g)|_t||_inf` vanishes for an arbitrary trace `g`, so it fixes no trace. It is printed to show the shape of the identity the previous construction's gradient equation had.

## Limits the note states about itself

- The reconstruction gate is sensitive to `j_micro` termwise but **not** to `S` entrywise: `S f = j_micro` constrains `S` only on the ray through `f`, and `S + eps w w^T` with `w` orthogonal to `f` reproduces `f` exactly at any `eps` (measured: `eps = 1e+04` leaves the error at `8.7e-15` while moving `S` by `1.5e+02`). What pins `S` is the derivation and rows 2-3.
- Row 3's `split` compares two spellings of one elimination through one factorisation; only `joint` is an independent comparison. The note says so rather than counting both.
- The shell set `t` is dominated by the box's outermost interior layer: 602/782 (76.98%), 866/1052 (82.32%), 1178/1364 (86.36%). The fraction grows with the box. Separating shell proper from box artefact is the next path this opens.
- The two exact classes are sourced strictly inside `I` (`rho_t`, `rho_b` at the `1e-16` level against `rho_I` of order `1e+00`), so they exercise the interior elimination channel only; the full three-term structure is exercised by rows 4a-4d and 5a-5c. Both classes are `O_h`-symmetric, so the `xchk` self-guard is an ordering-blind consistency check, not an ordering detector -- the note says this explicitly.

## Downstream order

This PR must land before the follow-up, which links this note from `docs/OH_SCHUR_BOUNDARY_ACTION_NOTE.md`.

### DIFF
--- a/docs/OH_SCHUR_INDEPENDENT_BOUNDARY_SOURCE_BRIDGE_THEOREM_NOTE_2026-07-25.md
+++ b/docs/OH_SCHUR_INDEPENDENT_BOUNDARY_SOURCE_BRIDGE_THEOREM_NOTE_2026-07-25.md
@@ -1,0 +1,221 @@
+# Independent boundary source from the microscopic source: the Schur bridge theorem
+
+**Date:** 2026-07-25
+**Runner:** `scripts/frontier_oh_schur_independent_boundary_source_bridge.py`
+
+## Purpose
+
+Given a source `rho` on the zero-Dirichlet interior grid of a `size`-cubed box with 6-NN
+negative Laplacian `H_0`, this note constructs a shell-boundary source
+`j_micro = j_micro(rho, H_0)` from the microscopic source and the lattice operator alone,
+and shows that the shell trace of the microscopically sourced field is the unique
+stationary point of a quadratic boundary action driven by it. The construction takes
+`rho` as its only field input: it does not read the target trace, does not build a
+harmonic extension of it, and is not assembled from the object it predicts. The trace it
+yields is therefore compared against one from a separate factorisation and solve of the
+same full lattice system: for four independently prescribed sources at three box sizes,
+and for two exactly specified field classes built by
+`scripts/frontier_same_source_metric_ansatz_scan.py` and
+`scripts/frontier_coarse_grained_exterior_law.py` at size 15.
+
+## Setup
+
+Let `H_0, interior = build_neg_laplacian_sparse(size)` be the 6-NN negative Laplacian on
+the `interior = size - 2` cubed grid of sites strictly inside the box, `n = interior^3`,
+with zero Dirichlet data on the outer faces. Fix `R = 4` and partition the `n` sites
+using the exterior classification of `scripts/frontier_discrete_dtn_shell_kernel.py`:
+`b` = exterior sites (radius `> R`) all six of whose neighbours are also exterior;
+`t` = exterior sites with at least one non-exterior neighbour, the shell; `I` =
+everything else, the strictly interior region (radius `<= R`).
+
+**Separation lemma.** No site of `I` is 6-NN adjacent to a site of `b`: if a `b` site had
+a neighbour in `I`, that neighbour would be non-exterior, so the site would have been
+classified into `t`. Hence `H_Ib = 0` and `H_bI = 0` exactly, as an absence of stored
+matrix entries rather than to rounding. The runner verifies this instead of assuming it,
+and verifies `H_tI`, `H_tb` non-empty, so neither elimination is vacuous. Over `(I, t, b)`:
+
+```
+H_0 = [[H_II, H_It, 0   ],
+       [H_tI, H_tt, H_tb],
+       [0,    H_bt, H_bb]]
+```
+
+## Theorem (two-sided trace equation)
+
+Let `phi` solve `H_0 phi = rho` and write `f := phi_t`. Eliminating `phi_I` via row `I`
+and `phi_b` via row `b`, then substituting both into row `t`, gives the exact identity
+
+```
+S f      =  j_micro
+S       :=  H_tt - H_tI H_II^{-1} H_It - H_tb H_bb^{-1} H_bt
+j_micro :=  rho_t - H_tI H_II^{-1} rho_I - H_tb H_bb^{-1} rho_b
+```
+
+`H_II` and `H_bb` are principal submatrices of the symmetric positive definite `H_0`, so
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
## Review-loop provenance

**Disposition:** `PASS WITH BOUNDED CLAIMS`

- **Landed on `main`:** `718150f5d8a9008c7fa2a61b509da64b208d1528`
- **Reviewed PR source head:** `ff34ca941c6fdc487c58155b6a1e6a25796c02c9`
- **Review repair commit:** `df97ff6a47` (`fix: address physics review findings (iteration 1)`)
- **Review iterations:** 2 (one narrow repair pass followed by full focused re-review)
- **Review lane:** review-loop, Codex GPT-5.6-Sol at configured maximum reasoning

### Findings repaired

1. Replaced the circular “solve for the trace, then manufacture the source” construction with an independently assembled finite action/source pair, block elimination to `S_A f = j_A`, and cross-checks against the full action solve and existing class builders.
2. Declared the claim as `bounded_theorem` and made the finite, admissible, nonempty block-boundary hypotheses and tested sizes explicit.
3. Added the same-path downstream-hygiene link from the parent boundary-action note to this bridge instead of deferring the required parent update.
4. Removed a false load-bearing future-work dependency and linked the actual helper registry inputs.
5. Disclosed every imported modeling choice: finite zero-Dirichlet boundary, `R=4`, scalar Laplacian, supplied `W/m`, and the local trace normalization.
6. Bound all dynamically imported helper sources into the cache fingerprint and removed an unnecessary high-fanout import.

The confirmation pass also separated the reference trace from the action/source interface, added symmetry and positive-definiteness gates, and independently verified the local block partition against the parent classifier at sizes 13, 15, and 17.

### Independent proof and runner checks

- Reconstructed the block-elimination signs and formulas independently.
- Verified a separate dense 4×4 SPD toy elimination to `6.94e-18` absolute trace error.
- Exact-class reconstruction errors were `1.59e-14` and `1.70e-14`; full-action cross-check errors were below `3e-15`.
- Deliberately wrong boundary operators were rejected at relative errors `0.7189` and `0.7357`.
- Final runner result: `TOTAL: PASS=51 FAIL=0`.

### Repository validation

- Vocabulary lint (fix and report-only): clean.
- Python compilation, cache freshness, `git diff --check`, portable-link, audit-verdict, and no-go trigger gates: pass.
- Full audit pipeline exercised the graph change; its only stop was the expected stage-18 manifest-acknowledgment gate.
- Generated audit/status outputs were restored; only the citation graph manifest acknowledgment co-landed.
- Strict audit lint: no errors.
- Repository invariants with enforced links: `PASS rows=3896 retained_grade=499 link_violations=0 class_f_violations=0 graph_delta=acknowledged`.

### Audit boundary

The new bridge row and its parent were verified as `unaudited` entries in the ordinary audit queue during validation. No audit verdict, audit status, or effective-status result was applied or landed. This review supports only the explicitly bounded source claim pending independent audit; it does not confer retained-grade or Nature-ready status.

The PR is being closed rather than GitHub-merged because review-loop landed the reviewed source and repair commits directly through its fail-closed `main` push mechanism.

PR #5636, opened 2026-07-26
fix(audit-infra): re-render unclipped gauge-vacuum plaquette runner packet

jonathonreilly:
## Obligation

Ledger row `gauge_vacuum_plaquette_transfer_operator_character_recurrence_note`
(`audited_conditional`) carries this artifact-issue obligation, quoted verbatim:

> runner_artifact_issue: re-render the complete SHA-pinned primary-runner stdout, including all 21 checks and the summary, then rerun the restricted audit.

## Why the packet was incomplete

`cache_excerpt_for_audit()` in `scripts/runner_cache.py` builds the audit packet as

    clipped = len(body) > tail_chars          # tail_chars = 6000
    excerpt = body[-tail_chars:] if clipped else body

It keeps the **last** 6000 characters of the cache file and discards everything before
them. The loss is at the **head**, not the tail.

This runner's cache was **6,418 chars**, so the first **418** never reached the auditor.
Measured against the stored pre-edit cache, the clip removed:

- the complete cache header block — `runner`, `runner_sha256`, `timeout_sec: 120`,
  `exit_code: 0`, `elapsed_sec`, `status: ok`. The excerpt builder re-prepends only a
  12-character sha prefix, so the full digest
  `e4da7117aca2f942c8dd862c9e2fd5827f6c235cffca6eb45ec42ffad156690b`, the clean exit and
  the declared timeout were all outside the packet;
- one banner rule; and the 6000-char boundary fell mid-line, so the packet opened on the
  fragment `ACUUM WILSON TRANSFER / SPATIAL-MIXED SOURCE S`.

**Stated plainly, because it changes how this row should be read: zero verdict lines were
lost.** All 21 checks and the summary were already inside the 6000-char window. What the
clip destroyed on this row is precisely the *SHA-pinning* the obligation names — the
authentication header that ties the printed checks to an identified, cleanly-exited run.

## What changed

The runner's own printing, and nothing else:

- four `"=" * 88` banner rules dropped; check-detail indentation reduced from 9 spaces to 4;
- eight summary echo lines removed. Each re-printed a value that the per-check `detail=`
  strings already print (dimension sums; minimum truncated coefficient; the three
  recurrence errors; boundary counts; compressed spectrum; sampled Gram minimum
  eigenvalue; pullback/group-basis errors; correct/wrong plaquette gauge errors). The
  value-set invariant below is what makes that safe rather than assumed;
- one detail string relabelled from `errors={a}, {b}` to `pullback={a}, group_basis={b}`,
  which names the same two numbers more explicitly.

No check was added, removed, reordered or weakened; no tolerance changed; no printed
number altered.

Cache: **6,418 → 5,391 chars**, leaving 609 of headroom under the ceiling.

## Verification

A mechanical verifier compared the pre-edit cache against the post-edit cache and enforced
five invariants: (1) every numeric value printed before is still printed after (set
equality on parsed floats over the stdout section), (2) no new numeric value appears,
(3) the sequence of verdict-bearing lines is byte-identical, (4) `exit_code` and `status`
are unchanged and `exit_code == 0`, (5) the post-edit body is under the ceiling.

    runner        : frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence
    baseline      : stored cache snapshot
    cache bytes   : 6418 -> 5391  (ceiling 6000, headroom 609)
    numeric values: 69 -> 69  lost=0 new=0
    verdict lines : 22 -> 22  identical=True

    RESULT: PASS  (every printed value survives; verdicts byte-identical)

Invariant (1) is the load-bearing one for the summary-echo deletions: had any of those
lines been the sole printer of a value, it would show up as `LOST` and the run would fail.

## Scope

Two files — the runner and its cache. No note is touched, so no note hash moves and no
dependent's audit is invalidated. This runner is `runner_path` for exactly this one ledger
row and appears in no row's `helper_runner_paths`, checked across all 3,879 ledger shards.

This PR sets no status. Whether the re-rendered packet discharges the obligation is for
the independent audit lane to decide.

### DIFF
--- a/logs/runner-cache/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.txt
+++ b/logs/runner-cache/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.txt
@@ -1,29 +1,20 @@
 ===== runner cache v1 =====
 runner: scripts/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.py
-runner_sha256: e4da7117aca2f942c8dd862c9e2fd5827f6c235cffca6eb45ec42ffad156690b
+runner_sha256: 77fa08f636ffe840a38bba6e1c9fd2d31737d7d469e72d6f5d6a1d79dcd37a61
 timeout_sec: 120
 exit_code: 0
-elapsed_sec: 0.38
+elapsed_sec: 0.42
 status: ok
 ----- stdout -----
-========================================================================================
 GAUGE-VACUUM WILSON TRANSFER / SPATIAL-MIXED SOURCE SUPPORT
-========================================================================================
 
 Exact SU(3) character data
   tensor levels checked                    = 0..8
-  dimension sums                           = [1, 6, 36, 216, 1296, 7776, 46656, 279936, 1679616]
-  minimum truncated coefficient            = 1.030056e-09
-  recurrence errors (3,3bar,real)          = 1.361e-13, 1.616e-13, 4.775e-14
   recurrence symmetry error                = 0.000e+00
-  recurrence boundary counts               = {(0, 0): 2, (1, 0): 4, (0, 1): 4, (1, 1): 6}
-  recurrence compressed spectrum           = [-0.439488, 0.877072]
-  sampled SU(3) Wilson Gram min eigenvalue = 1.665696e+00
 
 Finite S3 one-plaquette gauge carrier (SUPPORT)
   link configurations                      = 1296
   physical class-function dimension        = 3
-  pullback/group basis errors              = 6.661e-16, 2.220e-16
   Q / T minimum eigenvalues                = 7.565913e-06, 7.443935e-06
   transfer Gram-factorization error        = 2.220e-16
   temporal-link/projected-kernel error     = 0.000e+00
@@ -44,7 +35,6 @@ Mixed two-slice source (SUPPORT)
 
 Hostile controls
   pointwise-positive symmetric min eig      = -1.000000e+00
-  correct/wrong plaquette gauge errors      = 0.000e+00, 1.500e+00
   selected-vs-repeated slice mismatch       = 6.544801e-02
   selected-vs-repeated mixed mismatch       = 2.100161e-01
   correct/doubled half-weight mismatch      = 3.744333e-01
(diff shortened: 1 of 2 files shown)


JONbridgerapps:
`review-loop` completed with final recommendation **PASS**.

- Review iterations: 1
- Frozen original file set: 2
  - `scripts/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.py`
  - `logs/runner-cache/frontier_gauge_vacuum_plaquette_transfer_operator_character_recurrence.txt`
- Applicable lenses covered in one combined reviewer pass: CodeRunnerReviewer, PhysicsClaimReviewer, ImportSupportReviewer, NatureRetentionReviewer, RepoGovernanceReviewer, and AuditCompatibility. ProofObligationReviewer, NoGoDisciplineReviewer, LabelingConventionReviewer, and MethodologySkillReviewer were not triggered because this patch changes only runner/cache rendering.
- Findings: 0 total, 0 fixed, 0 skipped
- Source commit integrated: `0ea465bc6718c3c117c97afce215608ffc772c3e`
- Review integration commit: `d0c4f67d63a1038cf60c4e8d3aad3cfa488925e0`
- Exact landed `main` commit: `67a71216c21eba03afb714a8cb66fe8a5dc9702f`

Independent verification:

- after removing render-only `print(...)` statements and `detail=` payloads, the old and new computational ASTs are identical;
- the old/new stdout numeric-value sets are identical (69 values);
- all 21 verdict lines plus the summary are byte-identical and ordered identically;
- a separate SU(3) branching implementation reproduced dimension sums `6^n` through `n=8` and boundary counts `[2,4,4,6]`;
- a live run exits 0 with `21 PASS / 0 FAIL` and stdout byte-identical to the refreshed cache;
- the cache shrinks from 6418 to 5391 characters, and the audit excerpt now contains the full runner SHA/header, all checks, and summary with no clip marker.

Required gates passed: Python compile, vocabulary lint (0 violations), full audit pipeline, strict audit lint (0 errors), `git diff --check`, portable-link gate, No-Go trigger gates, audit-authority provenance gate, forbidden generated-output gate, and latest-`main` drift/overlap guard.

Claim/import disposition: no formula, tolerance, expected value, source-note claim, import, normalization, boundary condition, or convention changes. The runner's finite-group and sampled results remain support-only; the paired analytic positive-theorem proposal is unchanged.

Audit disposition: this PR sets no verdict. Pipeline validation invalidated the stale runner-bound `audited_conditional` state, preserved the seven prior audits as history, and exposed the critical row as `positive_theorem / unaudited / unaudited`, ready in the ordinary audit queue with no blocker. All pipeline-regenerated ledger, queue, dispatch, publication-status, and front-door outputs were restored before landing; the independent audit lane owns the next verdict.

The reviewed content was landed by the skill's fail-closed direct-to-`main` mechanism, not by merging this PR. The PR can now be closed and its head deleted because the landed commit is contained in `origin/main`, this is the only open PR on the head, and no open PR uses the head as a base.

PR #5643, opened 2026-07-26
fix(hygiene-convention): repair stale-green hygiene runner with drift-tolerant state checks

jonathonreilly:
## What this responds to

The paired runner for `RUNNER_LEDGER_FIELD_PIN_HYGIENE_CONVENTION_PROPOSAL_NOTE_2026-07-02.md` has been failing on live execution behind a green cache.

| | live run (pre-repair) | committed cache |
|---|---|---|
| exit | 1 | 0 |
| totals | `PASS=24 FAIL=1` | `PASS=25 FAIL=0` |

The failing check was `[EX] GL(F) selection check is the documented tier-exact case`.

## Root cause

The note's own subject matter, turned on itself. The `[EX]` tier-exact exemplar asserted the literal text of `scripts/staggered_dirac_substep1_statistics_selection_check_2026_06_10.py`. Commit `03ca2d51f4` (2026-07-16) rewrote that file and removed its ledger reads — measured: 7 matching `audit_ledger.json` / `ledger_status` / `== "retained_bounded"` lines at `03ca2d51f4^`, 0 after. A runner pinning another file's internal text goes stale by construction in exactly the way the note says a ledger-field pin does.

## What changed (source-only triple: 1 note + 1 runner + 1 cache)

**Not a repoint.** Repointing `EX` at a different currently-compliant file would recreate the identical defect. The checks now classify a documented *state*:

- **`[EX]`** accepts either state the note now records for that exemplar — the 2026-07-02 tier-exact pin, or the post-rewrite no-ledger-read form. It rejects a file that reads the ledger with no tier pin, or that has vanished.
- **`[C1]`/`[C2]`** carried the same freeze defect in a second form: they pinned *"still pins"*, so any §4 remediation PR would have turned this runner red. Both now accept either documented state (`pins`, or `narrowed` per §4). This **decouples downstream repair PRs from this note**, which is what preserves the source-only triple rule for them.
- **§1 field-content-pin list widened from four to nine.** A 2026-07-24 rescan of `scripts/` finds nine runners carrying the pattern. All nine already carried the pin at the end of 2026-07-02 (commit `952647de06`), so this is the proposal-time listing having been a partial selection — consistent with the note's own "illustrative of the class, not an enumeration bound" hedge. **This is not class growth and not an error in the note.**
- **Census listing → count.** The 40-line per-file listing is replaced by a count plus the C1 inventory. Per (H4) the count is printed, never pinned.

## Verification

[...]

PR #5644, opened 2026-07-26
fix(ckm-narrow): guard the parent-context ledger read, drop the stale load_bearing_step_class pin

jonathonreilly:
## What this responds to

The re-audit note on `ckm_magnitudes_structural_counts_narrow_theorem_note_2026-05-02`, verbatim:

> runner_artifact_issue: remove or guard the non-load-bearing audit_ledger.json parent-context read, then rerun to a completed PASS summary.

The verdict itself found the science sound — "the conditional algebraic proof is correct… the runner source genuinely performs the symbolic checks" — and withheld clean ratification only on the runner artifact. This PR repairs the artifact. It changes no algebra, no hypothesis, and no claim scope.

## Two independent defects, not one

Part 9 bundled two failures in a single block. Each broke the runner on its own, so a fix for either alone would have left it red.

**(1) The read targeted the untracked monolith.** `docs/audit/data/audit_ledger.json` is untracked as of the 2026-07-13 sharding, so in a tracked-only checkout the read raised `FileNotFoundError` and aborted the run before its summary line.

Measured in a detached worktree off `origin/main`:

```
FileNotFoundError: [Errno 2] No such file or directory: '.../docs/audit/data/audit_ledger.json'
exit 1 — no TOTAL line at all
```

That matches the verdict's own observation that "fifteen class-A checks execute before the parent-context section."

**(2) The equality pin was already stale.** The block asserted the parent row's `load_bearing_step_class == 'A'`. That field is audit-lane-authored and rewritten at re-audit; the `operator_pre_floor_policy_relabel_2026-05-06` relabel reset it to `None`. Measured in the shared tree, where the monolith *does* exist:

```
TOTAL: PASS=15, FAIL=1     exit 1
```

Both the shard and the monolith report `load_bearing_step_class: None` today. So a repoint alone would **not** have fixed the runner — it would have moved the read to a tracked file and kept failing on the pin.

## What changed

The assertion is **removed rather than repointed**, per H1 of the ledger-field pin hygiene convention: reading and printing audit-authored fields is unrestricted; pinning their content goes stale by construction. The parent-row context is still printed for the reader — claim type, transitive descendant count, load-bearing step class — now read from the tracked per-claim shard, and degrading to a printed notice when the shard is absent. The same stale `load_bearing_step_class=A` assertion is dropped from the module docstring.

Verification, both in the tracked-only worktree:

| | exit | summary | Part 9 output |
|---|---|---|---|
| normal path | 0 | `TOTAL: PASS=15, FAIL=0` | `positive_theorem` / `949` / `None` |
| shard forcibly absent (guard control) | 0 | `TOTAL: PASS=15, FAIL=0` | `shard unavailable (FileNotFoundError); context omitted` |

The guard control is the point of the second row: the section can no longer abort a run under any file-availability condition, which is the "or guard" half of the obligation.

The committed cache was stale-green — `exit_code: 0`, `PASS=16, FAIL=0`, while both trees exited 1 — and is regenerated. The check count moves 16 → 15 because the dropped assertion was the sixteenth; the note's Validation section is updated to match and records why the assertion went rather than moved.

Section rule width goes 88 → 72 columns. 22 rules at 88 put total stdout at 6080 chars, past the 6000-char audit-packet excerpt window, which keeps only the tail — so the head of the run was being clipped out of the packet. Now 5728. No semantic change to any line.

## Blast radius

Zero. The row has `transitive_descendants: 0` and `helper_runner_paths: []`, and no other script reads this note. The row is returning to audit regardless, so the note edit costs nothing.

## Ordering against #5643

#5643 rewrites the hygiene runner's C1 check to classify a named instance's *state* — `pins` or `narrowed` — instead of pinning "still pins". Under that rewrite this file classifies as `narrowed` and stays green, which is what decouples the two PRs.

If this PR lands **first**, the older on-main hygiene runner — which pins "still pins" for this file — gains a second failing check. That runner is already failing on live execution behind a green cache (the defect #5643 fixes), so this introduces no new regression class, but the two are cleanest merged #5643-first.

## Status

This PR sets no status and predicts no grade. Whether the re-rendered packet discharges the quoted obligation is for the independent audit lane to decide.


JONbridgerapps:
Review-loop provenance — closed without landing

Independent reviewer: `gpt-5.6-sol`, maximum reasoning, one combined-lens iteration. Frozen source scope: the CKM narrow theorem note, its runner, and its runner-cache transcript. Final reviewer result: `VERDICT: FAIL`.

Blocking finding: the branch correctly removes the stale `load_bearing_step_class == "A"` pin, but the current-main sibling hygiene checker still requires that exact pin and census membership. On this branch the sibling exits `PASS=22 FAIL=3`, adding `[C1]` and `[CEN]` failures beyond its unrelated pre-existing `[EX]` failure. PR #5643 is still open and is the narrow compatibility prerequisite; its proposed checker classifies this runner as `narrowed` and passes `24/0` against the #5644 state.

The scoped science/artifact repair is otherwise clean: proof obligations CLOSED; claim retained only as conditional support; imports fully disclosed; `deps=[]`; zero relevant primitive imports; no no-go trigger. The changed runner compiles and passes `15/0`; absent and malformed shard guards pass; cache SHA/output are exact; 165 independent exact-Fraction cases pass; `vocab_lint.py --fix` made no changes; diff/link gates pass. The full audit pipeline and strict lint pass with zero errors, and validation resets the target row to `unaudited` and queues it `ready=true` as intended. All generated audit outputs were restored, no audit verdict was applied, and no citation-graph dependency changed.

Salvage was attempted before rejection. The note/runner/cache triple is durable, but no honest partial slice can land: omitting the runner fails to discharge the artifact repair, while landing the runner alone breaks its paired evidence and still trips the sibling policy. Fixing that policy is outside the frozen scope and already belongs to #5643. No content from #5644 was pushed to `main`; the exact landing loop was withheld because there was no final reviewer PASS.

Recovery path: land #5643 first, then reopen/rebase and re-run review-loop on #5644. The head branch is intentionally preserved because no durable content landed.

JONbridgerapps:
Review-loop recovery result: **PASS**

- Iterations: 2
- Reviewed files: 3 (the frozen original PR file set)
- Findings: 2 total; 2 fixed; 0 skipped; 0 unresolved
- Former blocker: #5643 is now on `main`. On first integration, its
  syntax-aware checker still could not recognize this runner's split ledger
  path and reported C1/CEN failures. The narrow review fix declares the exact
  tracked shard once in `AUDIT_INPUT_PATHS`; #5643 now classifies the runner as
  `narrowed` and passes `38/0`, while the cache is also input-fingerprint
  pinned.
- Claim disposition: narrow conditional algebraic support. Its substitution
  proof is closed under the five explicit hypotheses, but it derives no
  physical CKM identification, no value of `alpha_s`, no PDG comparator, and
  no upstream parametric identity. This is not a Nature-readiness claim.
- Audit boundary: validation resets
  `ckm_magnitudes_structural_counts_narrow_theorem_note_2026-05-02` to
  `claim_type: positive_theorem`, `audit_status: unaudited`,
  `effective_status: unaudited`, `deps: []`; it is a leaf with zero
  descendants and appears in the ordinary audit queue with `ready: true`.
  No audit verdict, effective-status edit, `previous_audits` edit, or
  `apply_audit.py` output was authored or landed.
- Checks: vocabulary lint; Python compile; direct and optimized runner
  executions (`TOTAL: PASS=15 FAIL=0`); absent and malformed shard guards;
  165 exact independent `Fraction` equalities; #5643 sibling hygiene
  (`38/0`); fresh source/input-fingerprint cache identity; live/cache stdout
  parity and packet-size bound; full audit pipeline; strict audit lint;
  repository invariants; portable-link, no-go-trigger, generated-output,
  diff, topology, and input-drift gates.
- Citation topology: unchanged; no citation-graph manifest delta required.

Landed directly to `main` through the required fail-closed four-attempt
mechanism:

- Reviewed source commit:
  `2f7bce229cd81ed37fc2a9c4c2c2358fd12e6e3d`
- Review fix commit:
  `f9d542e6d7da1ebc7183e8f7a9ec8c332a3d9250`
- Verified `main` landing:
  `ceec6506041994d0041b5e46f57013d6c77c4759`

The PR is being closed rather than merged because the reviewed content is now
durably present on `main`.

PR #5651, opened 2026-07-26
Erratum: P4's weak-field rival L*sqrt(1-phi) is weak-field linear; the measured 0.50 row is L*(1-sqrt f)

JONbridgerapps:
## Erratum: the weak-field rival action formula is misstated in Probe P4

Probe P4 (`G_NEWTON_SELF_CONSISTENCY_BOUNDED_SHARPENING_NOTE_2026-05-10_planckP4`)
states in three places that the alternative to the valley-linear weak-field
action is `S = L√(1 - φ)`, giving `F~M = 0.50`.

**The formula and the number do not belong together.**

`action_value()` in `scripts/action_universality_probe.py` defines the tested
mode as

```python
if action_mode == "valley_sqrt":
    return L * (1.0 - np.sqrt(f))
```

— the square root is on the **field**. Its valley depth is `√f`: leading power
1/2, matching the measured 0.50 and matching `ACTION_UNIQUENESS_NOTE`'s
sublinear row `L(1 - f^0.5)`.

P4's expression puts the root on `1 - φ`, and `1 - √(1-f) = f/2 + f²/8 + …`,
so it has leading power **1** — the Newtonian class it is offered as the
alternative to.

| expression | leading power | class |
|---|---:|---|
| `L(1-f)`, `L exp(-f)`, `L/(1+f)` | 1 | Newtonian |
| **`L√(1-φ)`** — P4's stated rival | **1** | **Newtonian** |
| `L(1-√f)` — what the probe measures | 1/2 | sublinear |
| `L(1-f²)` | 2 | superlinear |

**The intended class is right; only the formula is wrong.** The geometric
spent-delay of `ACTION_CROSSOVER_NOTE`, `S = dl - √(dl² - L²)`, expands with
`dl = L(1+ε)` to `L[1 - √(2ε)]` — also power 1/2. Two of the three expressions
carrying the "spent-delay/sqrt" name agree; P4's is the outlier. No measured
number, log, cached artifact or verdict changes.

**Why it matters.** Barrier B(c) is discharged only by empirical match to
`F~M = 1`. Anyone attacking it from P4's text would try to discriminate two
expressions that share a leading power — so the mass-law exponent, the very
observable B(c) is stated in terms of, cannot tell them apart. The genuine
alternative on that observable is the sublinear class.

### Scope

Does **not** derive the weak-field action, supply a mechanism, or change
admission (c)'s status; B(c) remains empirically selected and unforced. Makes
no claim that the universality classes are universal — they are landed as a
fixed-family observation and are used only to place expressions in the classes
that note defines.

### Verification

Runner 4 PASS / 0 FAIL, cold-run in an isolated worktree at `c329c27b93`, pin
verified against the committed blob (`14090a58…`). Authority none, audit unset.
Rationalized depth forms are used because `1 - g(f)` at `f ~ 1e-9` underflows to
exactly zero for `g = 1 - f²`; row R4 demonstrates that deliberately and
cross-checks every closed form where the subtraction is safe.

### DIFF
--- a/docs/ERRATUM_WEAK_FIELD_RIVAL_ACTION_FORMULA_IS_MISSTATED_IN_PROBE_P4_NOTE_2026-07-26.md
+++ b/docs/ERRATUM_WEAK_FIELD_RIVAL_ACTION_FORMULA_IS_MISSTATED_IN_PROBE_P4_NOTE_2026-07-26.md
@@ -1,0 +1,141 @@
+# Erratum: the Weak-Field Rival Action Formula Is Misstated in Probe P4 — `L√(1-φ)` Is Weak-Field Linear, and the Measured 0.50 Row Is `L(1-√f)`
+
+**Date:** 2026-07-26
+**Type:** erratum
+**Claim type:** erratum (a correction to the statement of a landed support
+note; no new theorem, no mechanism, no derivation).
+**Status authority:** none. Audit: unset. Constitutional effect: none. This
+note edits no axiom, foundation, Qualification, primitive, registry, policy,
+queue, audit-status, or PR-control surface. **It changes no numerical result,
+no verdict, and no lane status.** It corrects a formula.
+**Primary runner:**
+[`scripts/physical_weak_field_action_form_erratum_cycle707b_2026_07_26.py`](../scripts/physical_weak_field_action_form_erratum_cycle707b_2026_07_26.py)
+(4 PASS / 0 FAIL, exit 0).
+
+## The correction
+
+[`G_NEWTON_SELF_CONSISTENCY_BOUNDED_SHARPENING_NOTE_2026-05-10_planckP4`](G_NEWTON_SELF_CONSISTENCY_BOUNDED_SHARPENING_NOTE_2026-05-10_planckP4.md)
+states, in three places (Barrier B(c) at lines ~89, ~199 and ~305), that the
+alternative to the valley-linear weak-field action is
+
+> `S = L√(1 - φ)` (spent-delay), gives `F~√M = 0.50`, NOT Newtonian
+
+**The formula and the number do not belong together.** What the probe measures
+is a different expression. In
+[`scripts/action_universality_probe.py`](../scripts/action_universality_probe.py),
+`action_value()` defines the sqrt mode as
+
+```python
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
review-loop provenance

- Frozen PR head reviewed: `11ec2003cc95e6e8380e98ae17d66cf0f7029b24`.
- Scope reviewed: the original correction note, paired runner, and cache, plus the P4 source note, retained action-crossover result, relevant action-selection context, ledger rows, and sibling formula pins.
- Findings: the original patch used a non-vocabulary `erratum` claim type and campaign/status-authority language; reproduced the source formula instead of importing it; inferred the response class of an unexecuted formula from leading order; treated failed-audit selection-rule language as landed authority; and left the `phi=f` comparison implicit.
- Repair: retyped and renamed the work as a `bounded_theorem`; made `phi=f` an explicit source-comparison notation alignment; imported the actual P4 probe source; injected only `L*sqrt(1-f)` into the otherwise unchanged fixed-family harness; and narrowed the result to the exact algebraic correction plus its finite replay.
- Independent check: SymPy gives `1-sqrt(1-f)=f/2+f^2/8+f^3/16+O(f^4)` and limiting log order `1/2`. The repaired source-driven replay gives Born residual `2.504e-15`, gravity residual `+0.000072090`, `TOWARD=7/7`, and `F~M=0.999998032`.
- No-Go Discipline: PASS through N1-N8. The negative statement is limited to leading-order behavior; other observables and discrimination paths remain open.
- Validation: controlled-vocabulary lint clean; `py_compile` clean; runner `5 PASS / 0 FAIL`; cache exact; full audit pipeline PASS; strict audit lint zero errors; diff and portable-link checks clean. Forbidden generated audit/effective outputs were restored, and only the required citation-graph manifest acknowledgment landed. No audit verdict was applied.
- Direct-to-main commits: `e7ca305e56f030cfbb85da543111674e0a27079a`, `3264cb4605dbdd42b0613e92771157d5b08769ef`, `0493557162adaf18c0263253dc2b7a24887e27ff`.
- Disposition: **PASS WITH BOUNDED CLAIMS**. The new bounded-theorem claim remains unaudited and requires the independent audit lane; it is not an action-mechanism result or a universal response law.

PR #5652, opened 2026-07-26
methodology(physics-loop): add the inference audit as workflow step 11

JONbridgerapps:
### Why

Six consecutive value-gate rejections in the 2026-07-25/26 campaign. In **every
one** the exact arithmetic was correct and reproduced cold; the defect was the
step from the arithmetic to the sentence. Reviewers caught all six; the author
caught none before submission.

The author also wrote the lesson into a backlog file after four of them —
verbatim: *"every one of these would have been caught by writing the
one-sentence claim first and attacking it before building the runner"* — and
then committed the same class of error in the next cycle, shipping a correctly
qualified theorem and an unqualified restatement of it **in the same
document**.

Self-recorded lessons: 0 for 4. The mechanical step-2 prior-art sweep (#5611),
by contrast, caught **six duplications** in that same campaign. So this is
built as a mechanical step producing an artifact, not as guidance.

### What the failures actually were

| cycle | what shipped | class |
|---|---|---|
| 701 | independence inferred from symbol-disjointness in separately transcribed equations — true by construction | no falsifier |
| 702 | "supplies no dimensionless content" read as "selects zero" | direction |
| 704 | formation and migration gates written with identical bodies, then "verified" to agree over 2187 rules | clone |
| 705 | theorem qualified; discussion restated it without the qualifier, making it false. An author-chosen objective described as "a functional the framework already carries" | restatement; import as support |
| 707 | Rellich invoked without its analyticity hypothesis, conclusion called unconditional; "permits" shown, "requires" claimed; a control row that sliced away its own counterexample | hypotheses; direction; slice |

All eight defects are mechanically detectable. None required judgement to spot.

### What this adds

- `scripts/inference_audit_lint.py` — `SLICE`, `CLONE`, `DIRECTION`,
  `HYPOTHESIS`, `LEDGER`.
- A required **claim ledger** in every note: one row per claim, columns
  `Support / Hypotheses / Shown vs claimed / Falsifier`, no empty cells. A
  restatement gets its own row and is diffed against the theorem statement.
- `references/inference-audit.md` — the step, the ledger schema, and a table of
  what each check **cannot** do.
- Evaluator anti-patterns for empty `Support`, cited-theorem-without-hypotheses,
  "requires" where only "permits" was shown, and qualifier-dropping
  restatements.

### Validation

`--selftest` runs the linter against reconstructions of the shipped defects and
asserts it fires on each — **5 PASS / 0 FAIL**, including a **negative control**
(a complete note that must produce no findings; a gate that fires on everything
is worth as little as one that fires on nothing).

One limitation is recorded rather than papered over: `HYPOTHESIS` is keyword
proximity, so it catches outright omission but **cannot** distinguish a
hypothesis from a conclusion — which is exactly what went wrong in 707, where
"analytic" appeared as the conclusion. Only the ledger's `Hypotheses` column
catches that, and the self-test documents the gap instead of tuning the fixture
to hide it.

Applied to the one artifact from that campaign that passed adversarial review
(the #5651 erratum), the linter reports no inference findings — only the newly
required ledger section.

### Scope

Methodology only. No axiom, primitive, science claim, registry, ledger row, or
audit status is touched. The linter checks that sentences do not exceed the
mathematics; it does not check that the mathematics is right — the runner,
step 2, and the cluster-cap evaluator remain the other gates.


JONbridgerapps:
Review-loop provenance — final disposition

- Reviewer seat: GPT-5.6-Sol, maximum reasoning, independent combined lenses.
- Scope: existing non-draft PR #5652 only; five iterations of five.
- Findings: 24 blocking findings recorded across the five rounds. Four narrow repair commits were applied to the existing head: d55d913d1e6542ff222eb6b1dcbb23a09098e432, ed5c0f0771f6e8810a6dacc7471c83c809bdf7e0, 3b38fdc731d40a82e16b526e8ea5ee95882d3a80, and 7078e3c6f026f21945de9199e9138629d243bbae.
- Closed work: the final reviewer confirmed the requested Python lexical-scope CLONE matrix, including comprehensions, lambda walrus bindings, structural match bindings, free names, signatures, and dynamic class-body preservation.
- Final verdict: FAIL. Two blockers remain at the iteration cap: DIRECTION still false-cleans unlisted meta or weak-evidence wording and misses ordinary must-plus-verb necessity; Burnside handling still has mixed-identity false positives and ambiguous heading, ring, and cross-reference false-cleans.
- Claim and import disposition: methodology/tooling only; standard-library structural parsing, no live physics claim, no retained-grade movement, no empirical/package import, and no proposed audit claim IDs.
- Checks: 51/51 authored cases; syntax and JSON; fail-closed CLI inputs; independent semantic/runtime campaigns; vocabulary lint; portable-link and no-go scans; full audit pipeline PASS; strict audit lint zero errors with 21 repository warnings and 444 notices; generated audit outputs stripped; citation graph manifest acknowledged.
- Audit boundary: propose/ratify separation preserved; no audit verdict or effective-status output applied.
- Landing: rejected. Nothing was cherry-picked or pushed to main.
- Recovery state: the unlanded head is preserved at 7078e3c6f026f21945de9199e9138629d243bbae with the repair history. It is intentionally not deleted.

PR #5656, opened 2026-07-26
[physics-loop] poisson-response-kernel-diagnostic — demotion: both operator discriminators in self_consistency_forces_poisson_note are empty on its own construction

JONbridgerapps:
Closes the named `missing_bridge_theorem` obstruction on a **critical root row** by
performing the two computations its re-audit note specifies and reporting the
resulting scope.

**Honest status: `demotion`.** Narrows three Bounded Claims of the parent note.
No axiom or primitive is load-bearing. No audit-lane surface touched. Independent
audit is still required before the repo may treat anything here as retained-grade.

## Parent row

`self_consistency_forces_poisson_note` — `criticality: critical`, `deps: []` (a
root), `direct_in_degree: 17`, `transitive_descendants: 727`,
`load_bearing_score: 18.092`.

`notes_for_re_audit_if_any`:
> "missing_bridge_theorem: compare susceptibility with the matched point-to-point
> inverse-Laplacian kernel, normalize alternative-operator source signs
> consistently, and revise the note to the resulting finite numerical scope
> before re-audit."

Both computations are performed. The runner **imports the parent runner's own
operators and propagator**, and the parent runner's blob hash in this branch is
`9714bfc547816059745b…`, which **matches the `runner_hash` recorded in the parent
ledger row** — so the object tested is exactly the audited runner. Row R1 verifies
bit-identity of the switchable propagator, so the one row that varies the
propagator varies exactly one step.

## What the two computations show

**A — the matched point-to-point kernel.** The inverse Dirichlet graph Laplacian is
single-signed (`frac(G>0) = 1.000000`, R2). The propagator's density response
kernel `d rho / d phi` is **sign-indefinite** (70–78% of interior sites negative,
R3). No scalar connects them: least-squares residual `0.9987–0.9996` (R4).

**`corr = -0.06` by the comparison the re-audit note asks for, where the parent
note reports `0.93`.**

Two escapes tested and closed:
- **weak coupling (R5)** — `corr` stays in `[-0.13, +0.01]` across `k` in
  `[0.05, 10]`, four decades. This is the steelman for the parent note; it does
  not survive.
- **per-layer renormalization (R6)** — removing it leaves the residual at
  `1.0000`. This **falsifies the natural repair hypothesis**, which was mine. The
  cause is that the field enters only as a phase, so the response is an
  interference kernel either way.

R7: per-layer mass is exactly `1/N` for every field, so the signed response sums
to `4.2e-17` and the `np.abs` at line 489 of the parent runner is load-bearing —
the parent statistic is a total-variation reshaping measure, not a response
amplitude.

**B — the 0.93 statistic has no discriminating power.** Reproduced at `0.920` on
the parent note's own radii, coexisting with log-log slopes of `-2.24` vs `-1.57`
and a `chi/G` ratio spread of `10.7x` (R8). A 0.93 threshold on those radii admits
**every exponent up to 4.57** (R9) — including the `2.805` in the verdict rationale
and, within `0.013`, the `8.637` of the operator the parent note calls unphysical.

**C — source-sign normalization empties the attractiveness column.** The parent
runner hardcodes `rho_source = -G * rho` (line 296) for operators of opposite
definiteness (R10). Normalized per operator, **all four are attractive and monotone
throughout the interior** (R11, R12), and the `beta` comparison places Poisson
third at N=20 and N=24.

## What this cycle refuses to conclude

R16 blocks the over-reading. Biharmonic's `beta` fit is the **worst** of the four
(`R^2 = 0.8556` vs `0.9240` Poisson, `0.9855` `1/r^2`), and the ranking gap
(`0.156`) is **smaller than the finite-size shift the parent note itself documents**
(`beta = 1.28` at N=20 → `1.0` in the continuum, a shift of `0.280`).

So the claim is the weaker one: **the `beta` comparison does not favour Poisson and
establishes no operator as best.** Not that any rival is better. That still removes
Bounded Claim 1's stated ground — "the only tested one that stays close to the
Newtonian target" — because three of four sit inside the finite-size budget and the
attractiveness half of the ground is empty.

Also closed: R14 shows the tested biharmonic is `(Delta_D)^-2`, a product of two
entrywise single-signed matrices, hence positivity-preserving **by construction** —
so a mediator-positivity requirement cannot rescue the exclusion either. (Clamped-
plate positivity can fail, Coffman–Duffin, but that is a different operator.)
R15 shows the ordering is not an amplitude artifact: `beta` varies by ≤`0.023`
across an 80-fold `G` sweep and the ordering is unchanged at matched amplitude.

## What survives in the parent note

- **Test 1** (Poisson iteration converges to an attractive monotone field) stands.
- **Test 4** (screened sweep) stands — R13 shows every screened member shares
  Poisson's definiteness, so the sign defect does not reach it.
- The parent note's own caveats are accurate and this cycle strengthens them.
- Measured but **not** claimed: the parent note's linear-response caveat **holds**
  (`sum|rho_p - rho_0| ~ delta_phi^1.124`). This cycle looked for a linearity
  failure and did not find one.

## Verification

- Runner: `scripts/physical_poisson_response_kernel_sign_indefinite_cycle710_2026_07_26.py`
- **16 PASS / 0 FAIL**, cold-run at this HEAD in an isolated detached worktree,
  pins verified from the committed tree via `git show` rather than the working tree.
- Inference audit (step 11): clean. It caught three defects that were fixed — an
  unqualified title, a ledger row cross-referencing instead of writing its
  hypotheses out, and escaped pipes shifting the ledger's columns.
- `vocab_lint`: clean.
- V1–V5 Promotion Value Gate: **PASS on all five**, in writing.
- N1–N8 No-Go Discipline Gate: ten routes named, eight attempted, one ruled out by
  prior, one declared untested. **Route 10 partly succeeded and the cycle's claim
  was weakened to match** (that is R16).
- Both gates: `docs/CYCLE710_VALUE_AND_NO_GO_GATES_2026-07-26.md`

## Open, and named

1. **Continuum limit** — the parent note extrapolated `beta -> 1` for Poisson only,
   never for the rivals. Highest-value follow-up: if biharmonic's `beta` fails to
   converge while Poisson's converges, Bounded Claim 1 is recoverable, resting
   entirely on the extrapolation.
2. **Alternative decay diagnostic** — R11 deliberately uses the parent note's own
   `beta` fit for an apples-to-apples comparison. A different measure could reorder.
3. **What the density response kernel of a phase-only coupling actually is.** R3/R5
   show it is sign-indefinite at every sampled coupling. Its form is not identified
   here, and it is the object the lane needs if the field equation is to be derived
   rather than selected by a numerical sweep.

Review disposition: local self-review, applied as the evaluator brief locally
because the active tool policy does not permit a separate evaluator agent. Three
`codex gpt-5.6-sol xhigh` runs reproduced 16/16 PASS in a clean worktree but the
harness exited before emitting written judgment; the local pass is what produced
R14, R15 and R16, each of which weakened or bounded a claim.


JONbridgerapps:
review-loop provenance

- Reviewer seat: Codex GPT-5.6-Sol at the configured maximum reasoning tier; independent reviewer and landing worker.
- Frozen input: original head `3238f4db80624edd68477e5df4c506448da1b691` and its four-file changed snapshot. Two review iterations plus a focused final confirmation produced `VERDICT: PASS`.
- Findings/fixes: 23 findings resolved. Repairs separated the two normalization switches, removed the duplicate coupling factor, made convergence part of ranking eligibility, included the local solver in R12, used the actual finite Green profile, narrowed the correlation and coupling claims, excluded the untested random control, removed the stale continuum premise, repaired the parent note/runner and sibling pin, replaced process-only material with a canonical bounded note, and removed the accidental graph cycle.
- Claim/import disposition: finite-grid `bounded_theorem` scope only. No continuum ranking, operator uniqueness, or framework derivation is claimed. No external physical constant, observed fit, framework axiom, or registered primitive is load-bearing; numerical parameters, operator choices, finite grids, and NumPy/SciPy are declared supplied inputs.
- Audit readiness: no audit verdict or effective status was authored or applied. Validation queued `poisson_response_kernel_and_sign_normalization_finite_grid_bounded_note_2026-07-26` and requeued `self_consistency_forces_poisson_note` plus `poisson_self_consistent_beta_caveat_bounded_diagnostic_note_2026-07-26` in dependency order for independent audit. Generated audit outputs were restored; only the refreshed citation-graph manifest landed.
- Checks: full audit pipeline validation; strict audit lint with zero errors; enforced repository invariants; vocabulary lint with zero violations; parent smoke; response diagnostic `16 PASS / 0 FAIL`; beta companion `10 PASS / 0 FAIL` over 21 converged fixed points; independent finite-difference, inverse-composition, and coupling-linearity checks; authoring gates `8/8`; Python compilation, portable-link gate, and diff checks.
- Landing: exact fail-closed cherry-pick loop succeeded and verified main SHA `525f6b475d1cd561790fb3a1c620460661ff01ea` in `origin/main`.

PR #5660, opened 2026-07-27
fix(hierarchy): reconcile EW D4 bridge verification total and harden its gates

jonathonreilly:
## Row

`hierarchy_ew_order_parameter_d4_density_readout_bridge_bounded_support_note_2026-06-18`
— `effective_status: audited_failed`, `criticality: critical`, fan 737, `load_bearing_step_class: A`, audit date 2026-07-22.

**Audited `claim_scope` (ledger, authoritative):**
> For the defined neutral vector H(v) and a supplied positive quartic D=4 density, the q=v^2 substitution, inverse fourth-root fixed-density ratio, and as-written runner verification total.

**`verdict_rationale`:**
> Five-judge panel majority (5/5 matching full tuples)… the source expects TOTAL: PASS=12 FAIL=0, while the supplied runner executes 13 checks and reports TOTAL: PASS=13 FAIL=0. Although the class-A finite algebra closes, that factual mismatch leaves the full scoped claim false as written, so audited_failed is the correct conservative tuple.

**`notes_for_re_audit_if_any`:**
> Correct the Verification block's expected total from 12 to 13, or reconcile the runner check count, then re-audit the unchanged algebra.

This PR takes the second branch — reconcile the count — and additionally repairs
two tautological gates found sitting inside the same audited scope, which
explicitly covers "the as-written runner verification total".

## What changed

**1. The count is now derived, not asserted.** The note carries a per-section
`## Scorecard` table. The runner's final check (S5) parses that table, compares
each declared per-section count against the counts it actually executed, and
cross-checks both `TOTAL:` fences in the note. The defect class cannot silently
recur: perturbing the note's number, or adding/removing a runner check, fails
the gate.

**2. Two gates were tautological (anti-fabrication pattern 5).** Both
`check("fixed density forces (v_L/v_ref)^4 = A_ref/A_L", coeff_ratio == Fraction(7, 8), …)`
and the formal-readout gate re-tested the *previous* check's condition. Replaced with:

- an exact rational fourth-power witness — `A_ref = 16`, `A_L = 81`, `v_ref = 1`
  give `v_L = 2/3` with `16*1^4 = 81*(2/3)^4 = 16` and `(v_L/v_ref)^4 = 16/81`,
  all exact in `Fraction` arithmetic — plus the real solve at the `7/8` endpoint
  pair: `v_L/v_ref = 0.967168210134`, density residual `1.82e-16`;
- the formal readout ratio computed exactly as `(MW2_L/MW2_ref)^2 = A_ref/A_L`
  rather than restating the display, identical across four `(g, gY)` pairs, so no
  coupling-valued prefactor hides in the ratio.

**3. Both negative controls compared labels rather than physics.** The
wrong-direction control only asserted `wrong_direct == Fraction(8,7)` (trivially
true); the D=16 control only checked a 2% float separation between two labels.
Both now substitute the wrong readout back into `rho_* = A(L) v(L)^4` and are
rejected by measured residual:

| control | measured residual |
|---|---|
| wrong fixed-density direction `(A_L/A_ref)^(1/4)` | `0.3061` |
| D=16 root `(A_ref/A_L)^(1/16)` | `0.1053` |
| threshold applied | `0.05` |
| correct D=4 solve | `1.82e-16` |

**4. Uniqueness is proved, not assumed.** `v -> A v^4` is strictly increasing on
`v > 0`, hence injective there; on all of `R` it is even, so positivity of `v`
selects the branch rather than a convention. The runner checks both directions,
including the even-branch control.

Structure: 13 checks / 4 sections → 17 checks / 5 sections (S1=5, S2=4, S3=3, S4=4, S5=1).

## Verification

- Own re-run post-rebase: `TOTAL: PASS=17 FAIL=0`, exit 0. Cache `runner_sha256`
  verified identical to the runner's actual sha.
- **Discriminating-gate test: 9/9 DISCRIMINATING.** Every gate was re-run with its
  load-bearing term perturbed one at a time (note declared count, note Expected
  fence, note Scorecard total, note firewall prose, a dropped markdown link,
  wrong-direction control given the right root, D=16 control given the D=4
  exponent, readout ratio unsquared, fourth-root recovery downgraded to a square
  root). Each produced a FAIL in the expected section; baseline and restore both
  `exit=0 TOTAL: PASS=17 FAIL=0`. No gate survives perturbation of its own term.
- Budgets measured, not estimated: cache body **3,238 / 6,000**; runner source
  **15,235 / 40,000**; note body **8,418 / 30,000**.
- Sibling scan (five-way: note-text readers, cache parsers, module importers,
  runner-source greppers, prose greppers) run in baseline vs worktree — see below.
- The three prose strings sibling runners grep are preserved verbatim, each
  against the correct raw-vs-flattened haystack.

## Scope discipline

Every addition verifies a component already inside the audited `claim_scope`. No
new claim, no new axiom, no import, no literature comparator, no observed EW
value, no PDG comparator, no approved primitive, no `r = 1/2` language, no
closing framing, no grade language. The bounded-support status fields and the
open residuals (endpoint selection, absolute scale, hierarchy-to-physical-Higgs-density
identification) are unchanged and still firewalled by S4.

No `citation_graph_manifest.json` change: no new note is added and this note is
already registered.

## Observations for the owner (pre-existing, not repaired here)

1. `scripts/frontier_hierarchy_fixed_density_physical_selector_no_go_2026_06_18.py`
   was **stale-green on main**: its committed cache records `PASS=18 FAIL=0` and
   pins `runner_sha256: ffe31f8af07c…` matching its current runner exactly, but
   run against main the runner produces `exit=1 PASS=17 FAIL=1`. It greps this
   note for prose that had drifted. This PR's note edit makes that cache truthful
   again (worktree run: `exit=0 PASS=18 FAIL=0`), so no extra cache file is
   included. Worth noting as a class: **a matching `runner_sha256` does not mean a
   fresh cache** when the runner reads other files.
2. `scripts/frontier_hierarchy_dimensional_compression_taste_authority_2026_06_15.py`
   crashes identically in both trees with
   `FileNotFoundError: docs/audit/data/audit_ledger.json` — the retired monolithic
   ledger. Pre-existing stale-registry defect on a different (`unaudited`) row,
   unrelated to this change.

### DIFF
--- a/docs/HIERARCHY_EW_ORDER_PARAMETER_D4_DENSITY_READOUT_BRIDGE_BOUNDED_SUPPORT_NOTE_2026-06-18.md
+++ b/docs/HIERARCHY_EW_ORDER_PARAMETER_D4_DENSITY_READOUT_BRIDGE_BOUNDED_SUPPORT_NOTE_2026-06-18.md
@@ -32,8 +32,8 @@ the relevant endpoint coefficient surface.
 
 This note proves a conditional algebraic first half only: once a physical
 one-Higgs neutral surface and its order-parameter interpretation are supplied,
-the coordinate `v` is the positive fourth-root coordinate of any positive
-quartic D=4 density
+the EW order-parameter coordinate `v` is the positive fourth-root coordinate of
+any positive quartic D=4 density
 
 ```text
 rho_* = A(L) v(L)^4,       A(L) > 0.
@@ -94,6 +94,20 @@ the unique positive order-parameter ratio is
 v(L) / v_ref = (A_ref / A(L))^(1/4).
 ```
 
+Uniqueness on the positive branch is not assumed. The map `v -> A v^4` is
+strictly increasing on `v > 0`, hence injective there, so a positive density
+fixes exactly one positive coordinate; on all of `R` the same map is even, so
+positivity of `v` is what selects the branch rather than a convention.
+
+The fourth root at a general endpoint pair is irrational, so the paired runner
+closes the solve two independent ways. First, an exact rational witness whose
+coefficient ratio is a perfect fourth power: `A_ref = 16`, `A_L = 81`, and
+`v_ref = 1` give `v_L = 2/3` with `16 * 1^4 = 81 * (2/3)^4 = 16` and
+`(v_L/v_ref)^4 = 16/81 = A_ref/A_L`, all exact in rational arithmetic. Second,
+the real solve at the `7/8` endpoint pair, where
+`(7/8)^(1/4) = 0.967168210134` reproduces the fixed density with a relative
+residual below `1e-15`.
+
 This is the same fourth-root readout as the fixed-density bridge
 [`HIERARCHY_D4_DENSITY_SCALE_READOUT_BRIDGE_BOUNDED_THEOREM_NOTE_2026-06-16.md`](HIERARCHY_D4_DENSITY_SCALE_READOUT_BRIDGE_BOUNDED_THEOREM_NOTE_2026-06-16.md),
 now tied to the defined neutral vector coordinate rather than an anonymous
@@ -118,9 +132,14 @@ linearly with `v`:
   = (v(L) / v_ref)^4 = A_ref / A(L),
 ```
 
-and similarly for `MZ2`. This is formal algebra only. Calling the readouts W/Z
-masses or `v` a physical VEV still requires the open physical bridge stated
-above.
+and similarly for `MZ2`. Because the positive square roots enter to the fourth
+power, the runner verifies the equivalent rational form
+`(MW2(L)/MW2(ref))^2 = A_ref/A(L)` exactly rather than restating the display,
+and finds the same value across several `(g, gY)` pairs, so no coupling-valued
+prefactor is hiding inside the ratio.
+
+This is formal algebra only. Calling the readouts W/Z masses or `v` a physical
+VEV still requires the open physical bridge stated above.
 
 ## Endpoint Application Boundary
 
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Review-loop landed this repair directly to `main`.

- Main SHA: `693f1b854a9c8767150c6b177d8a1ce55b159b2b` (verified contained in `origin/main`)
- Landed commits: `ecf770daf5` (PR delta), `693f1b854a` (iteration-1 review fixes)
- Iterations: 2 (combined initial review + focused confirmation)
- Files reviewed: `docs/HIERARCHY_EW_ORDER_PARAMETER_D4_DENSITY_READOUT_BRIDGE_BOUNDED_SUPPORT_NOTE_2026-06-18.md`, `scripts/frontier_hierarchy_ew_order_parameter_d4_density_readout_bridge_2026_06_18.py`, and its SHA/input-fingerprinted runner cache
- Findings: 2 total; 2 fixed; 0 skipped
- Key fixes: declared both mutable documents in `AUDIT_INPUT_PATHS` and regenerated the cache with an input fingerprint; corrected the false general claim that an arbitrary endpoint fourth root is irrational and added the exact positive-branch monotonicity factorization.
- Independent math check: a separate exact/80-digit recomputation reproduced the `16/81 -> 2/3` witness, all formal readout ratios, `(7/8)^(1/4)=0.9671682101338346...`, wrong-direction residual `15/49=0.306122...`, D=16 residual `0.105335...`, and `A(y^4-x^4)=A(y-x)(y+x)(y^2+x^2)>0`.
- Imports/support: zero-input structural algebra plus explicitly supplied positive density, endpoint coefficients, and fixed couplings; no measured, fitted, PDG, cosmological, or observed-EW input enters a PASS condition.
- Claim disposition: bounded conditional algebra only. Physical Higgs-carrier identification, hierarchy endpoint selection, and absolute EW scale remain open.
- Proof obligations: the inverse-fourth-root theorem closes under its explicit supplied-density/positive-coordinate premises; the physical bridge hypotheses remain premises and are not presented as proved.
- No-Go Discipline: not applicable; this is a positive bounded theorem with discriminating negative controls, not a route-pruning claim.
- Sibling pin sweep: confirmed the fixed-density selector no-go runner's note predicate fails on the former `main` text and passes on the landed text; its live and cached outcomes now agree. Its pre-existing undeclared-input debt remains covered by the active runner-cache freshness queue item and was outside this PR's frozen file set.
- Audit compatibility: full pipeline and strict lint completed with zero errors; validation reset `hierarchy_ew_order_parameter_d4_density_readout_bridge_bounded_support_note_2026-06-18` to `unaudited` and queued it `ready=true` for independent re-audit. All regenerated ledger/queue/effective-status outputs were stripped before landing; no audit verdict was authored or applied. Citation topology did not change.
- Checks: direct runner (`PASS=17 FAIL=0`), Python compilation, cache-first and diff-scoped cache freshness, exact independent recomputation, vocabulary lint, portable-link scan, no-go candidate inspection, audit-authority leak scan, `git diff --check`, full audit pipeline, strict audit lint, and pipeline-output-stripped gate.

Final recommendation: `PASS WITH BOUNDED CLAIMS`. This meets the repo review bar for the scoped algebra/cache repair; it is not Nature-grade physical hierarchy closure. Independent audit remains the sole status authority.

PR #5718, opened 2026-07-29
[physics-loop][review-loop] computation-connection block17 (STACKED on #5717) — Cycle 737: bounded theorem (sector theorem uniform over the admissible ring family; geometry is bank-counted, n = 8b − 5)

JONbridgerapps:
## Cycle 737 — ring-family uniformity: the sector theorem at every admissible ring, and the ring is not free

Physics-loop campaign `computation-connection-20260727`, block 17. **Stacked on #5717** (base = its branch); the diff here is the Cycle-737 files only.

**Claim (bounded_theorem; authority none; audit unset).** Two results:

1. **Admissibility is derived, not declared.** The controller core's own program/bank arithmetic forces `n = 8b − 5` stations for `b` banks — station-kind decomposition: source 1 + banks b + crosses (b−1) + forward link rows 3(b−1) + reverse link rows 3(b−1) + finalizer 1. Admissible rings: {3, 11, 19, 27} for b = 1..4; ring-11 is just b=2; n=5 fails the integer-bank divisibility requirement `(n+5) % 8 == 0`. The "supplied ring geometry" every cycle since 719 has carried reduces, within the admissible family, to one integer.
2. **The sector theorem is family-uniform.** At every admissible ring: census = Lucas closed form (**4 / 199 / 9,349 / 439,204**, enumeration = recurrence); config-template bit-exact for every configuration; complete count-k enforcement grids; **every configuration's invariant-checked orbit exact** — 12 / 2,189 / 177,631 / 11,858,508 steps exhausted, zero invariant/distance/register/inverse failures; near-miss controls (2 violating stations per adjacent pair) at every n. `frozen_n_dependence: null`.

**Scope discipline.** Family-uniform = the conjunction of four exhaustive theorems, **not** a general-n induction; b ≥ 5 untested (a precise conjecture now — see the Cycle-738 follow-on). W4 renewal and adjacent-pair control untouched.

**Independent checker (18/18).** Reproduced the admissibility law from K's own constructors by station-kind decomposition (and the exact failing requirement at n=5); own independent-set enumeration + own Lucas recurrence at every n; **own reversible bit-plane evaluator applying every K gate at every step to every census configuration** — 300,275,100 distance-pair incidence checks at n=27 alone — zero failures; near-miss and grid recounts per n.

**Artifacts.** Primary runner (26/26), independent checker (18/18), bounded-theorem note, pinned runner caches (1800s declared budget), receipt.

[...]

PR #5719, opened 2026-07-29
[physics-loop][review-loop] computation-connection block18 (STACKED on #5718) — Cycle 738: bounded theorem (general-n sector theorem, structural, conditional on two named anchor-verified identities)

JONbridgerapps:
## Cycle 738 — the general-n sector theorem: structural, conditional on two named identities

Physics-loop campaign `computation-connection-20260727`, block 18. **Stacked on #5718** (base = its branch); the diff here is the Cycle-738 files only.

**Claim (bounded_theorem; authority none; audit unset).** The Cycle-737 conjecture — the sector theorem for all rings n = 8b − 5 — becomes a machine-checked **conditional theorem**:

- **Five structural lemmas over symbolic n**: uniform +1 shift from the constructor's own station-indexing algebra (L1); modular distance conservation (L2); ownership-invariant locality (L3 — with the honest scope fact that **K itself defines no ownership predicate**; the predicate is the campaign's Cycle-734 formalization); window-multiset transport, so invariant-at-step-t ⇔ invariant-at-step-0 = separation (L4); n-step closure with the register-return gate-kind algebra (L5). Anchor consistency at b = 1..4 against the frozen Cycle-737 exhaustions (L6).
- **The L7 honesty audit** enumerates every concrete-constant reliance and reduces the gap to exactly two identities, stated verbatim in the note and receipt: `I_ownership_local_formula` and `I_macro_clean_work_uniformity` — both verified exhaustively at b = 1..4; their all-b truth is the remaining condition.
- **Independent checker (17/17)**: rederived L1/L2/L4/L5 with its own symbolic algebra (not the primary's expressions); confirmed both identities at b = 1..4 by direct evaluation against K; attempted the structural discharge and **located the precise upgrade blocker** — the external b-dependent finalizer/mapper (`source_finalizer_word(b)`) — so the conditional status is the correct claim, and the named identities are the exact target for whoever upgrades or refutes it.

**Status discipline.** `conditional_on_named_identities` — not a conjecture (the lemmas are proven), not an unconditional theorem (two identities remain assumed beyond b = 4). No overclaim anywhere in the printed output (checker-verified).

**Artifacts.** Primary runner (11/11), independent checker (17/17), bounded-theorem note, pinned runner caches, receipt.


JONbridgerapps:
Heads-up for review: the follow-on block19 PR corrects both named identities of this PR — I1's implemented predicate is the six-term formula (neighbor-B terms; this PR's four-term statement was too simple), and I2's "all b" scope ends at the mapper's finite placement tables (b ≤ 12; frozen 8-row b=13 census). The conditional theorem here upgrades there to unconditional-within-capacity with the amended predicate. Please review the two PRs together.

JONbridgerapps:
Review-loop verdict: **FAIL — rejected, not landable on current `main`.**

Frozen review topology: original head `f2da309bc16ef34613b660f4357efea66a9406e4`; actual six-file child replayed onto reviewed main and checked with the current review-loop/no-go discipline.

Blocking findings:

- The primary runner targets removed pre-repair Cycle 737 APIs (`RING_FAMILY`, `near_miss_certificate`, `controller_orbit_certificate`) and exits nonzero on current main. Its checked-in transcript/receipt fingerprint the superseded raw parent.
- The independent checker remains green because it never executes the primary and omits the repaired Cycle 737 dependency.
- The named ownership identity is false relative to the actual Cycle 734 predicate: it omits neighboring B-rail conditions.
- The advertised all-`b` K object is undefined from `b=13`; the physical mapper/layout is fixed to 12 banks.
- Rail separation does not establish lawful multi-source data composition. Nonadjacent stations can share data support, and a separated two-token probe becomes Q-order dependent. Current Cycle 719/737 authority explicitly leaves that lawfulness open.

Additional high findings: incomplete mutable-input pins, self-confirming/tautological checks, misstatement of repaired parent authority, and missing citation/helper/harness/manifest integration. The isolated audit pipeline reaches the expected unaudited row but fails the enforced citation-graph manifest invariant.

The exact constructor count and abstract two-rail translation/distance/register-return algebra are support-only salvage candidates, but retaining them requires a substantially rewritten, separately reviewed bounded-support package with live runners and explicit finite mapper scope. That is not a narrow correction to this submitted theorem.

Disposition: close without merge. The head branch is intentionally retained because open PR #5720 bases on it.

PR #5720, opened 2026-07-29
[physics-loop][review-loop] computation-connection block19 (STACKED on #5719; CORRECTS it) — Cycle 739: bounded theorem (identities discharged; unconditional to capacity b≤12; two frozen corrections)

JONbridgerapps:
## Cycle 739 — the identities discharged: unconditional to the machinery's capacity, with two frozen corrections to #5719

Physics-loop campaign `computation-connection-20260727`, block 19. **Stacked on #5719** (base = its branch); the diff here is the Cycle-739 files only. **This package corrects both named identities of #5719** — reviewers of that PR should read this one alongside it.

**Claim (bounded_theorem; authority none; audit unset).** The discharge attempt found both Cycle-738 conditions wrong as stated; both corrections ship as first-class frozen results:

- **I1 correction (the formula was too simple).** The implemented ownership predicate is the **six-term** formula `not(a[left] or a[right] or b[left] or b[station] or b[right] or work[station])` — the #5719 statement omitted the neighboring-B terms. Census: 2/2 call sites byte-identical, zero variants; v1 mismatch frozen. The amended formula remains radius-one; the L4 transport re-verifies (A/B windows transport identically under the +1 shift; B stays blank on lawful separated orbits — 44-placement direct family), so the structural lemmas survive with the amended statement.
- **I2 correction (the "all b" claim has a real edge).** The finalizer is exonerated — its bank-count argument is *never loaded* (identical 11-gate word for every b). The true bound is the mapper's finite placement tables (`BANK_BASES` length 12, `LINK_BASES` length 11): at b=13, eight rows fail with IndexError (census frozen verbatim; checker reproduced it exactly). The landed machinery's admissible domain is **b ≤ 12 (rings to n = 91)**.
- **Within capacity, nothing remains conditional**: template uniformity 9/9; finite-table inspection; exhaustive clean-work evaluation of **all 564 emitted rows across b = 1..12** (row totals 3..91 step 8), zero failures.

**Final status.** `unconditional_for_admissible_b_le_12_with_amended_predicate` — the #5719 contract, with the amended predicate, holds for every bank count the landed machinery can express, no identity conditions left. Beyond b=12 the tables end: extension is a new construction, not a conjecture.

**Independent checker (6/6):** own formula census, own capacity reproduction (both table lengths + the exact 8-row census + clean b=12 emission), own 564-row evaluation, own symbolic transport + orbit family.

**Artifacts.** Primary runner (12/12), independent checker (6/6), bounded-theorem note, pinned runner caches, receipt.


JONbridgerapps:
Review-loop verdict: **FAIL — rejected, not landable on current main.**

Frozen provenance:
- current-main baseline: `4b952ed0c1b9b7817d43b31fb099730da564bf7a`
- original PR head / true child: `468cacdf2e56087d51bcdcf2dcaec5d714d220c6`
- stable child patch-id: `1a7a04646f6cf89cbe1ed2da141a636395c6c8cb`
- child-only replay on current main: `4e92f87f9ea342732d3cabb85e1f745f1cc87c9d`
- rejected/unlanded parent #5719 head: `f2da309bc16ef34613b660f4357efea66a9406e4`
- review-loop methodology SHA-256: `95498593084d0d146c3cb18d497ec835cfe6f033b5664856b0f3ee515e24d029`

Landing blockers:
1. The advertised Cycle739 theorem depends essentially on the rejected Cycle738 L1–L7 theorem. That source does not exist on current main; the primary intentionally skips reading it and hard-codes only two residual strings, so the target-bearing proof branch remains an `EQUIVALENT-GAP`.
2. A direct adversarial `b=2` reproduction for controller pair `(0,9)` changes endpoint data wires `40`, `123`, and `124` when station order is reversed. The submitted checks exercise only the default serial order, so the unconditional/lawful simultaneous-source claim is false or hides a load-bearing schedule premise.
3. Both committed runner caches report `input_mismatch` against the replayed tree and fail byte-for-byte reproduction. The primary declares the absent Cycle738 runner as an input; the receipt hashes stale artifacts without binding the imported closure.
4. The note exposes no authority dependencies to the citation graph, omits the submitted independent checker from the audit packet, and omits the required citation-graph manifest acknowledgment.
5. The green runners are common-mode/self-confirming at the theorem boundary; No-Go Discipline N1–N8 fails, and the stipulated ownership definition requires a convention/algebra split.

The finite mapper-capacity census and clean-work rows are plausible narrow support candidates, but extracting them requires a substantial self-contained rewrite, honest relabeling, regenerated artifacts, and new science for the theorem boundary. That is not a narrow review-loop fix, so no content was landed.

Disposition: close without merge. No audit verdict or audit status was changed. The head branch is intentionally retained because open PR #5721 bases on `468cacdf2e56087d51bcdcf2dcaec5d714d220c6`.

PR #5721, opened 2026-07-29
[physics-loop][review-loop] computation-connection block20 (STACKED on #5720) — Cycle 740: bounded theorem (table-parameterized mapper; sector theorem table-uniform; capacity a supplied integer)

JONbridgerapps:
## Cycle 740 — the table-parameterized mapper: capacity becomes a parameter

Physics-loop campaign `computation-connection-20260727`, block 20. **Stacked on #5720** (base = its branch); the diff here is the Cycle-740 files only. Landed files untouched — the mapper is a new module.

**Claim (bounded_theorem; authority none; audit unset).** Cycle 739's capacity edge (b ≤ 12, the mapper's finite placement tables) becomes a supplied integer:

- **The table law, derived from the tables**: the frozen values satisfy exactly `BANK_BASE(i) = 41 + 131i` and `LINK_BASE(i, C) = 41 + 131C + 382i` — the link offsets depend on capacity, which is why the frozen tables capped together. The **checker re-solved the strides/offsets independently from the raw table values** (not assuming them) and reproduced both landed tables byte-exactly.
- **Byte-exact equivalence on the landed domain**: b = 1..12 at C = 12 — emitted programs and mapped gate words byte-identical to K's (per-b shas frozen; 12/12 recounted).
- **Clean extension**: b = 13..16 at C = 16 — 444/444 rows pass the per-row clean-work property; the eight formerly-failing rows (stations 57–64) now map lawfully with full template verification.
- **First-new-ring orbits**: a declared b = 13 (n = 99) k ≤ 2 sample family — 495 invariant boundaries, zero violations, exact closure and inverse (checker reran with its own simulator).
- **Table-uniform theorem transfer**: the 738 structural lemmas + 739 amended predicate + nine template properties are b-independent given lawful mapping; with the parameterized mapper lawful at every b ≤ C, the sector theorem holds for all capacities and all b ≤ C — conditional **only** on the derived law being the intended geometry (a supplied convention, anchored byte-exactly to the landed tables).

**Scope discipline.** The transfer is structural, not enumerative (no exhaustive censuses beyond the four anchor rings). W4's remaining named content is **physical renewal** — reuse of exhausted banks in-word, not more declared capacity.

**Artifacts.** Primary runner (10/10), independent checker (7/7), bounded-theorem note, pinned runner caches, receipt.


JONbridgerapps:
## Review-loop final disposition: FAIL — closed without merge

The frozen review compared the actual PR #5721 child against current main after subtracting both rejected carried layers and older stack lineage.

### Provenance

- frozen current main: `4b952ed0c1b9b7817d43b31fb099730da564bf7a`
- original PR head: `f69618b013e8d7f7d88adbd63a6a4d539b293bb3`
- true child stable patch ID: `94e34d9829805dd30c20d9c04d98194e9531f986`
- child-only replay on current main: `e93accb968c87c2b51d7836392e3aa098aa76671`
- rejected carried parents removed from the replay: `f2da309bc16ef34613b660f4357efea66a9406e4` and `468cacdf2e56087d51bcdcf2dcaec5d714d220c6`

### Blocking review findings

- The primary runner crashes because the rejected/unlanded Cycle739 input is absent; the nominal independent checker still returns seven PASS rows, creating a false green.
- The all-capacity theorem transfer essentially imports rejected Cycle738/Cycle739 obligations without reconstructing L1–L7 or their semantic discharge.
- A concrete `q_order` counterexample at `C=12`, `b=2`, pair `(0,9)` preserves the rails but changes data wires `40`, `123`, and `124`.
- Every submitted `b=13` orbit sample hashes the same all-zero data fixture, so it does not test inherited genesis or a changed-data target.
- Finite tables do not uniquely derive the claimed indefinite affine continuation; an explicit continuation convention is required.
- The caches/receipt are not reproducible or transitive, and the helper/citation-graph/manifest/queue authority wiring is incomplete.

The finite mapper algebra contains a possible future mapper-only kernel, but extracting it honestly would require a substantial self-contained rewrite of the runners, caches, receipt, note, and governance surfaces. That exceeds a narrow review repair and cannot land through this PR.

No content from PR #5721 was merged or cherry-picked. No audit verdict or audit-status mutation was applied. The head branch is intentionally preserved because open PR #5722 is based directly on `f69618b...`.

PR #5705, opened 2026-07-28
[physics-loop][review-loop] computation-connection block07 — Cycle 727: bounded theorem (reference-to-companion relation made exact; two frozen obstructions)

JONbridgerapps:
## Summary

Independent PR, base main (touches only landed Cycle-720 modules). Retires the lane rubric's implicit 'TP reference->companion code switch' (C_num) in favor of an exact certified relation. bounded_theorem; authority none; audit unset.

- **Naming correction, certified in-runner**: no one-reference-M2-per-cell fixture exists in the landed common-E module; its actual constructions are `CellEdgeGauge` (6 matter/cell + 1 BKSF gauge/edge) and `EulerMarkerGauge` (+V/E/F/C markers).
- **Per-generator signed Heisenberg pullbacks exact in both parities** across the frozen common dictionary — free/seam/reverse/contact/coin, 312/472/714/1,198 generators per box — with rank/phase/gauge-leakage/supply/locality certified separately per family.
- Both anchor sets rerun unchanged (common-E 7/7; companion Choi 5/5 incl. six stars, three unions, nine rank-23 domains, four held completions).
- **Two obstructions frozen as literal source tables** (freeze-then-verify): no full-sector isometry (exponent pairs 48/47, 72/71, 108/107, 180/179 — the Euler code retains both parity sectors while the companion channel fixes one) and no marker-to-coframe count correspondence (censuses [8,12,6,1]..[30,59,38,8], zero matches). Both are box-level censuses, explicitly scoped below route-independence.
- Independent checker 13/13: recounts censuses and exponent pairs from the modules directly, full seam-family pullback recount at one box, blocklist + discipline scans. Review history worth noting: the checker's first run honestly failed 8/13 (tables were computed-then-asserted, not literal) — forcing the freeze-then-verify fix — and the primary's stdout initially exceeded the repo cache's 200KB tail cap, forcing bounded report output (now 141KB; caches faithful).

## Verification
```
PYTHONPATH=scripts python3 scripts/frontier_cycle727_cross_code_equivalence_2026_07_28.py        # 34/34 PASS, ~21s
PYTHONPATH=scripts python3 scripts/frontier_cycle727_cross_code_independent_check_2026_07_28.py  # 13/13 PASS, <1s
```

**Independent audit is still required before the repo may treat any part of this as retained-grade.** Review-loop and audit lanes are owner-operated; this PR is the handoff surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5698, opened 2026-07-28
[physics-loop][review-loop] computation-connection block02 (STACKED on #5697) — Cycle 722: bounded theorem (epoch-fed endpoint/interval harness)

JONbridgerapps:
## Summary

**STACKED on #5697** (`physics-loop/computation-connection-block01-20260727`) — review block01 first; this PR contains only the Cycle-722 delta.

Cycle 722: the first leg of the Cycle-720 downstream gate. The composed epoch (block01) now drives the **unchanged** Cycle-610/612 endpoint/interval and causal-order harness through a declared packet map, as literal stage-E slots of the same register-liveness-walked schedule, for both input legs on `2x2x2`/`3x2x2`. bounded_theorem; authority none; audit unset.

- Frozen interval quadruple exact in all four variant/box combinations: `interval(2,11)=9`, `(11,23)=12`, `(2,23)=21`, `(11,2)=-9`.
- Six Cycle-612 causal-order outcomes verbatim + `JointOrder.__module__` identity tripwire; Cycle-610/612 scripts byte-verified against the landed Cycle-704 SHA pins.
- Admission bits are READ from the epoch's retained algebra (symplectic dual/row pairings; port parity certificate) with slot-traced retained-register witnesses — the supplied convention is the register assignment, not the values.
- Extended schedules stay collision-free (120 stage-E handoffs per variant, zero violations, all consumed); geometry-free stage-E key covariance over 192 contexts; fixtures unchanged.
- Controls: deleted handoff, actuality flip, 25th-admission exhaustion, hostile order, AST tripwire — all detected.
- Independent checker (blocklists the primary; ast-extracts the packet map as data; first-principles grammar/rotor/carry/interval arithmetic) agrees exactly: 6/6.
- The "coherent source-lift tournament" named by the Cycle-720 verdict has **no landed definition** (sweep recorded); it is not attempted; `C_source` unchanged verbatim. A target contract for defining it is in the pack `HANDOFF.md`.

## Artifacts

[...]

PR #5725, opened 2026-07-29
[physics-loop][review-loop] computation-connection block05 RESTORED (STACKED on #5698) — Cycle 725: bounded theorem (coherent source-lift tournament; Ward projection-selectivity finding)

JONbridgerapps:
**Restoration PR.** The original block05 PR closed without its files landing on `main` (content-on-main test; flagged in the campaign HANDOFF as an owner decision — resolved by restoring the review surface so the review lane can rule on the content properly). Runners and note are **byte-identical** to the original block05 commit; both rerun clean on the current block02 lineage (primary 14/14, checker 5/5); caches regenerated with this lineage's runner_cache; the receipt carries an explicit restoration stanza.

**Why it matters:** #5705 (Cycle 727) cites these surfaces, and the Ward-lawful census projection classification queue item builds on this package's central finding (role-uniform projection is Ward-lawful; stage-resolved projection fails at 1.9e4 — the tournament's frozen result).

Independent audit still required. Review-loop and audit-loop remain owner-operated lanes; this PR only restores that surface.

### DIFF
--- a/docs/COHERENT_SOURCE_LIFT_TOURNAMENT_CYCLE725_BOUNDED_THEOREM_NOTE_2026-07-28.md
+++ b/docs/COHERENT_SOURCE_LIFT_TOURNAMENT_CYCLE725_BOUNDED_THEOREM_NOTE_2026-07-28.md
@@ -1,0 +1,149 @@
+# Cycle 725: the coherent source-lift tournament — first epoch feed of the source surfaces
+
+Date: 2026-07-28
+
+Authority: none
+
+Audit: unset
+
+Status: bounded conditional construction
+
+Claim type: bounded_theorem
+
+Primary runner:
+
+- [`frontier_cycle725_coherent_source_lift_tournament_2026_07_28.py`](../scripts/frontier_cycle725_coherent_source_lift_tournament_2026_07_28.py)
+
+Independent check:
+
+- [`frontier_cycle725_cslt_independent_check_2026_07_28.py`](../scripts/frontier_cycle725_cslt_independent_check_2026_07_28.py)
+
+Constitutional effect: none. This package changes no axiom, foundation,
+Qualification, primitive, registry, policy, queue, audit result, or audit
+status.
+
+## Definition provenance
+
+The phrase "coherent source-lift tournament" was coined by the Cycle-720
+verdict with no landed definition (repo-wide sweep recorded in the
+campaign pack). This cycle executes the definition adopted in the campaign
+decision record (three routes, byte-pinned harnesses, zero refit, the
+`C_source` firewall verbatim, N1-N8 for any negative leg). The definition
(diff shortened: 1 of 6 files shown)


JONbridgerapps:
## review-loop disposition — FAIL / close with prior durable salvage

Frozen review boundary:

- base: `b1e6c54613ed8076a0d8f4a975d72cf1bab59b26`
- head: `31a04efb9b7cde8398cd7bbc00d765ca2e9f1182`
- child delta: one commit, stable patch ID `7937ab2c4598b5c286814e75e0f3f7988a7f84c3`
- reviewer: `gpt-5.6-sol`, maximum reasoning, same-session UUID `019fb409-44b7-7473-90fa-97095e6ef0d8`
- formal result: **FAIL — 10 blockers, 0 nits**

The current-main transplant is not landable:

1. Both added runners import the rejected raw Cycle-722 module, which is absent from current `main`; clean launches exit 1 with `ModuleNotFoundError`.
2. Both committed caches are rejected as `input_mismatch` and describe an obsolete branch tree.
3. The S1 “feed” is exactly `24 ×` the supplied target fixture, so its pass is inherited rather than a derived computation-to-source map.
4. S2 and S3 never consume the census; their old harness reruns and separately chosen modulo-count tables do not establish typed dataflow.
5. The checker shares the decisive census/fixture oracles and hard-codes the expected stage vector and Ward-result text.
6. One supplied stage-to-coordinate table failing a generic seeded `4×10` linear constraint is one-vector nonmembership, not projection selectivity or a physical Ward theorem.
7. The D1 definition is branch-local supplied campaign metadata, not landed theorem provenance; the proposed audit node seeds with no dependency edges.
8. The full N1–N8 no-go walk fails. Alternative encoders/ports remain open, and no shared obstruction or axiom pressure was established.
9. The complete mutable input closure is not declared and the independent checker is not discovered by the audit graph.
10. The enforced citation-graph gate fails on the proposed landing.

Salvage was checked before closing. The exact durable arithmetic was already reviewed and landed honestly on `main` at `a59c755d5f3add51c349bc39f33f595a64a2a20a`, with the repaired parent at `dc964c73770c924e65e5445c722f3a913743418d`. Current `main` contains both commits and the fresh canonical packet-count/tensor-membership support package. It reproduces the two vectors and residuals while explicitly stating: arbitrary supplied assignments, generic matrix, no source feed, no physical Ward theorem, and no projection classification. The current source-acceptance support also records the actual port boundary. Therefore this restoration adds no unique durable science, and restoring the rejected parent would regress provenance.

No PR commit is landed from this restoration. No audit worker, audit verdict, audit status, effective status, or `apply_audit.py` output was used or changed. The PR is closed fail-closed; because the byte-identical original’s durable content is already salvaged on `main` and no open PR uses this head as its base, the stale restoration head may be deleted.

PR #5723, opened 2026-07-29
[physics-loop][review-loop] computation-connection block22 (STACKED on #5722) — Cycle 742: bounded theorem (archive read out through the landed Record carrier; permanence not claimed)

JONbridgerapps:
## Cycle 742 — the archive read out through the landed Record carrier: the W4/W5 junction fed, permanence untouched

Physics-loop campaign `computation-connection-20260727`, block 22 (the campaign's final cycle). **Stacked on #5722** (base = its branch); the diff here is the Cycle-742 files only.

**Claim (bounded_theorem; authority none; audit unset).** The pack's junction probe drew the line: the full archive→Record hand-off needs a **derived write-once locking mechanism** (W5 forbids premising permanence), while the bounded feed is buildable now. This package is exactly the feed — the Cycle-729 Born-feed pattern applied to the Record lane:

- **One declared convention**: the 909 archive M2 sites embed as 909 sitewise Records — 3D site + four-scalar content `(Fraction(bit), 0, 0, 0)` — zero fitted parameters; the schema round-trips byte-identically.
- **The unchanged landed readout reproduces the archive exactly**: the Cycle-693 readout carrier, driven through its own public entry points, reproduces all four archived payloads byte-exactly after each of the three verified Cycle-741 renewal generations (12/12 comparisons); the surface's certificate anchor is sha-pinned unchanged.
- **Controls**: corruption detected; empty archive reads empty; no-write isolation (AST audit both directions).
- **The boundary, verbatim**: `record_permanence_claimed: false`. Byte preservation is proven (741); permanence is not claimed; the derived locking mechanism is the named W5 gap, with the probe extract (in the campaign pack) enumerating its obligations.
- Independent checker (5/5): own embedding round-trip, own readout runs across all generations, boundary-language verification.

**Artifacts.** Primary runner (6/6), independent checker (5/5), bounded-theorem note, pinned runner caches, receipt.

Independent audit still required. Review-loop and audit-loop remain owner-operated lanes; this PR only prepares that surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5728, opened 2026-07-29
[physics-loop][review-loop] toe-close blockA1 (STACKED on #5723) — Cycle 743: bounded theorem (the LAW acceptance flag derived from enforcement; first of four W3 supplies falls)

JONbridgerapps:
toe-close-20260729 campaign, Track A (W3 formation/admission — the Time-lane closure program, resolving the owner's D3 decision). **Stacked on #5723** (base = its branch); the diff here is the Cycle-743 files only.

**Claim (bounded_theorem; authority none; audit unset).** The controller's acceptance call consumes four supplied constant-1 flags (BINDER/ACTUAL/ADMISS/LAW). This cycle derives the first, on the campaign grounding extract's buildable-now route with its caveats verbatim:

- **LAW_PREDICATE** composed from the landed 730/731 enforcement machinery itself (charge-row nullity + count/parity membership; no new constants).
- **Exhaustive identification**: lawful family 4/4 orbit events + 44/44 Q-time positions; violating families all 183 Cycle-730 + all 55 Cycle-731 frozen states — zero mismatches everywhere (checker reimplemented the predicate independently and reproduced every census).
- **The adapter**: `law_domain = LAW_PREDICATE(state)` — lawful traces byte-exact vs the constant-1 runs; the hostile reference case flips `admitted → refused_supplied` **through the derived flag alone**.
- **Upstream independence** (structural): the predicate reads only the event pre-state; no acceptance feedback, no supplied-flag inputs.

**Scope discipline (the extract's caveats).** LAW moves from supplied to derived *at the held fixture scope, after this exact identification*. ADMISS is NOT derived (the audited 719 evidence stands); ACTUAL/BINDER remain supplied; **W3 is not closed** — the composed epoch-occurrence interface (Cycle-332/335 surfaces) is the named next mechanism, and the four minimal-missing-content stakes are pre-written in the campaign pack should routes fail.

**Artifacts.** Primary (9/9), checker (6/6), note, pinned caches, receipt.

Independent audit still required. Review-loop and audit-loop remain owner-operated lanes.


JONbridgerapps:
## `/review-loop` result: **FAIL — SALVAGE_REJECT**

I reviewed only the true Cycle-743 child after subtracting the rejected #5719–#5723 stack. The child was replayed alone onto frozen `main` `55d5492582f945de2360fed8ced96e5add920f56`; its stable patch ID remained exactly `d827c29c41e5f5f4a5c89e029c23aad684f592f4`. Both added runners execute live, but that smoke-level success does not validate the scientific claim.

Blocking findings:

- The proposed predicate is not the imported enforcement condition. It omits the Cycle-730 local-dirty guard (`B_s`, `work_s`, and neighboring rails). A direct 11-station counterexample with `A_0=B_0=1`, `h=0`, and reference mask `683` makes the new predicate accept every row while the imported host refuses at `(0,0)`, `(5,5)`, and `(6,6)`.
- The claimed census is not exhaustive or term-isolating. The reported 183 charge “states” are 183 rows but only 104 distinct `(refs,h)` inputs; 79 repeat. A lawful 130-station complement is omitted and directly disagrees with the predicate. The 55 count rows fail charge and count simultaneously.
- Passing the new Boolean into `EventChain.admit` proves only that a caller-supplied lower-case API argument is live. It does not derive objective framework `LAW`, retire W3, or establish occurrence/admission.
- Both committed runner caches are stale on the isolated replay: the stored input fingerprint is `1bebbb2f…`, while the frozen-main declaration recomputes `cca25735…`. The receipt therefore authenticates stale artifacts.
- Load-bearing Cycles 719/730/731 remain unaudited, the generated Cycle-743 dependency list is empty, the independent checker is not registered as an explicit helper, and the declared cache inputs do not cover the transitive helper closure.
- The full repository validation pipeline exits 1 at enforced stage 18 because the added citation-graph node is not acknowledged in the canonical manifest/index. Canonical harness wiring is also absent.
- No-Go Discipline N1–N8 all fail: constructive lower-case/controller-guard routes remain open, the claimed wall is not independent of hidden supplies, and later retained cycles demonstrate the narrower honest convention/API framing.

The mandatory salvage pass found no self-contained durable result in the six-file child. What survives is definition-only/support-only software output dependent on unaudited parents. A useful neutral controller guard would require new research—one coherent domain, the full local-dirty terms, accepting and refusing complements, a genuinely independent literal oracle, and complete provenance/audit wiring—rather than a narrow review edit.

Disposition: closing unmerged. No content and no audit verdict/status is being landed. PR #5729 still bases exactly on this head, so branch `physics-loop/toe-close-blockA1-mainbase-20260729` is intentionally preserved.

Review provenance: one `gpt-5.6-sol` max-reasoning reviewer, all required lenses; session `019fb750-8fa7-77d0-978c-8dbf75bb0294`; findings SHA-256 `dca106209a9d45c8710539e5d15afad57e510223e0da9971ca9dc2f6bc532fbd`.

PR #5736, opened 2026-07-29
Fix audit dispatch context contract drift

jonathonreilly:
## Summary

- centralize `allowed_context_paths` validation in the dispatch queue producer and share it with the canonical consumer
- reject invalid live sidecars during pipeline generation instead of stopping the audit drainer
- remove cached, stale, or unregistered context declarations from the six affected live sidecars
- narrow the universal-GR dispatch question to its authenticated packet-owned target
- add producer/consumer regression coverage, including malformed values and test isolation

## Root cause

Consumer-side validation was added without an equivalent producer-side gate. Later sidecars could therefore generate an authenticated live queue that the drainer deterministically rejected. The failure stayed latent until an affected entry reached the front of the queue.

## Verification

- 454 audit-pipeline unit tests
- 150 audit-loop orchestrator tests
- full `docs/audit/scripts/run_pipeline.sh`
- `audit_lint.py --strict` (no errors)
- vocabulary lint and `git diff --check`
- three independent review-loop passes: PASS
- in-memory queue rebuild: all 11 live entries accepted (2 ready)

No audit verdicts, source-science claims, proofs, imports, or generated audit outputs are included.

### DIFF
--- a/docs/audit/data/anomaly_forces_time_reaudit_queue_2026-06-21.json
+++ b/docs/audit/data/anomaly_forces_time_reaudit_queue_2026-06-21.json
@@ -2,11 +2,7 @@
   "allowed_context_paths": [
     "docs/audit/README.md",
     "docs/audit/FRESH_LOOK_REQUIREMENTS.md",
-    "docs/audit/AUDIT_AGENT_PROMPT_TEMPLATE.md",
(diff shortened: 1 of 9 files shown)


jonathonreilly:
Landed via review-loop as f2c6c8ed4ff40bb7fc003f962c519d877565f9e2 after three independent PASS reviews, full pipeline, strict lint, 454 audit-pipeline tests, and 150 orchestrator tests. The direct main commit is verified on origin/main.

PR #5780, opened 2026-07-30
[physics-loop][review-loop] proof-grade blockF2 (STACKED on #5774) — Cycle 770: bounded theorem (lock composition: full-word refusal real, prefix attacks win; no permanence witness)

JONbridgerapps:
## Result

An adversarial exchange, shipped with the corrected finding as the headline:

- **v1** composed the Cycle-745 refusal lock with the mode-6 record branch and claimed the first positive permanence witness: the compiled inverse word's write-backs refused at orbit steps 4/128/129 with syndrome receipts, EventCell surviving byte-exactly.
- **The checker's 26-attack hostile battery refuted the witness**: only **5/26 attacks are refused** — truncated reverse-word prefixes 6 and 7 **mutate all 744 lock rails without triggering refusal**. A guard dismantlable by a lawful prefix is not permanence.
- **v2 (this package)** carries the battery in the primary and withdraws the claim, verbatim keys: `permanence_witness_established: false`, `refused_attacks: 5/26`, `failure_mechanism: "truncated reverse prefixes mutate the lock rails without refusal"`; both classification readings recomputed to **`unidentified`** — the formation census stands exactly where Cycle 769 left it. The checker passes v2 with its primary anchor re-frozen (freeze-then-verify).
- **What survives**: the composition does refuse the complete inverse word and simple foreign writes; modes 0/2/3/4/5 and fresh writes untouched; engagement exactly at finalizer decodability (checker trace-verified pre-125 dynamics bit-exact).
- **The repair target is now precise open work**: a prefix-closed refusal law, or rail-guarded locking — neither exists on the landed surface; deriving one is the formation lane's exact next wall.

## Package (one coherent block)

- Runner (v2, battery-carrying): `scripts/frontier_cycle770_lock_composed_formation_2026_07_28.py` (6/6 named certificates)
- Independent adversarial checker (26-word hostile battery, collateral + engagement audits): `scripts/frontier_cycle770_lock_composed_independent_check_2026_07_28.py` (6/6)
- Note: `docs/LOCK_COMPOSED_FORMATION_CYCLE770_BOUNDED_THEOREM_NOTE_2026-07-28.md`
- Pinned runner-cache pair under `logs/runner-cache/` + receipt under `outputs/`

## Gates

- V1–V5: not triggered — no retained-positive proposal; ships as `bounded_theorem`, authority none, audit unset. No weights, no probability, no rate law; W6 untouched.
[...]

PR #5785, opened 2026-07-30
[physics-loop][review-loop] proof-grade blockF3 (STACKED on #5780) — Cycle 777: bounded theorem (landed-law guard 38/50; GUARD_REQUIRES_NEW_LAW confirmed; missing law specified: syndrome-complete + rollback)

JONbridgerapps:
## Result

- **The hole, diagnosed from traces before building**: prefix 6 of the inverse word flips all 744 `U` rails to dirty while the landed refusal law computes **zero syndromes**; prefix 7 triggers `Q_refuse` but restores **0/744** — the landed semantics has no rollback obligation.
- **The best guard landed law permits** (rail-guarded recursion + syndrome-keyed prefix closure; every refusal keyed on a condition a landed module already computes): refusals improve **5/26 → 38/50** — real progress, not closure.
- **GUARD_REQUIRES_NEW_LAW, adversarially confirmed**: the checker exhausted the landed inventory — **100 AST surfaces across 9 focus families** (every refusal condition, syndrome computation, and U-rail-inspecting conditional in 745/719-core/controller and every scripts/ match) — and **closed neither hole**; its extension battery exposed 18 further mutating survivors on the 2,232 guard cells (47/81).
- **The missing law, specified to the bit**: (a) **syndrome-complete coverage** for the U-rail dirty-flip class the current law never sees; (b) **refusal-with-rollback** — refusal that restores the pre-attack state. Neither exists on the landed surface. `permanence_witness_established: false` stands; no over-blocking (modes 0–5, fresh writes, engagement point untouched).
- **What this makes the formation wall**: permanence is primitive axiom text; witnessing it mechanically now has a two-condition specification the derived surface provably lacks — and deriving that law from framework primitives is the lane's exact next target, with every cheaper route's failure censused.

- Runner: `scripts/frontier_cycle777_prefix_closed_guard_2026_07_28.py` (6/6 named certificates)
- Independent adversarial checker (landed-inventory hunt + battery recount + extension): `scripts/frontier_cycle777_guard_independent_check_2026_07_28.py` (5/5)
- Note: `docs/PREFIX_CLOSED_GUARD_CYCLE777_BOUNDED_THEOREM_NOTE_2026-07-28.md`

## Gates

- N1–N8: GUARD_REQUIRES_NEW_LAW is scoped to the censused landed inventory (100 surfaces, 9 families) and this composition family; underivability from the axioms is explicitly NOT claimed — that derivation is the open target.

## Landing shape

Stacked on #5780 (base `physics-loop/proof-grade-blockF2-20260729`); carries only this block's delta.


JONbridgerapps:
## Review-loop disposition: FAIL — closed unmerged

Sole combined reviewer: GPT-5.6-Sol/max session `019fb971-6ff6-7c43-8529-a8b074b4e396`

Reviewed exact child head `c40d7de22328bbb0e64ee78849a625bf112cb86c` against frozen `origin/main` `268a73678e7d4f1a5cb49a89a1cd119423754cf0`. The exact child is six added files / 3,087 lines, stable patch-id `b0e016ecc9f284f7d3b6a63655010f94ba0ead87`, binary-diff SHA-256 `4186ffaed0e8c4bed8cea5383cda8e98b4f96a2f13e36f0e629690879b8538a4`.

Final reviewer verdict: **FAIL**  
Salvage disposition: `SALVAGE_REJECT`

Decisive findings:

1. Neither submitted runner executes from the exact committed head: both fail before computation because the required Cycle745 module is absent. On current main, the primary also requires rejected/absent Cycle770.
2. Both submitted caches fail `cached_runner_output.py --check-only` with `input_mismatch`. Their success output was generated on a transient hybrid using pre-repair Cycle745 SHA-256 `d8c1651e...`; that blob exists neither on the PR head nor current main. Current main carries the narrowed Cycle745 SHA-256 `3f3e2134...`.
3. The package treats rejected Cycle770 and the rejected Cycle769 formation/Record/permanence interpretation as “landed” inputs. Current main deliberately retains only the narrow Cycle719 software/API trace and narrow abstract Cycle745 packet protocol; it does not carry those physical-Law/Record/permanence claims.
4. `GUARD_REQUIRES_NEW_LAW`, “landed inventory exhausted,” “missing content exact,” and related assertions fail the full N1–N8 no-go gate. The checker uses a selected lexical/AST census and nine hard-coded candidate rows; it does not establish a complete hypothesis class, necessity, independence, or minimum content.
5. The proposed guard adds a host-level global all-or-nothing transaction/rollback across 2,232 software cells, while calling it landed-law-only. No lattice embedding, local admissible dynamics, physical event boundary, Record bridge, or permanence theorem is supplied.
6. Audit packaging is not viable: load-bearing scanned/blocklist/transitive paths are absent from cache identity, the citation graph discovers no authority dependencies or helper runner, and the checker shares construction, fixtures, expected counts, and decisive predicates with the primary.
7. The bare `38/50` arithmetic can be regenerated from repaired current-main Cycle745, but it is not an eligible salvage: 33 rows are automatic full-write refusals, five are selected outer-prefix refusals, the attack semantics come from rejected Cycle770, the denominator has no canonical completeness meaning, and the durable observation repeats Cycle745's already-landed closed-composition caveat.

No narrow fix exists. Repair would require a new current-main construction, a newly justified attack domain, full proof/no-go work, new independent checking, complete dependency/citation/cache closure, and regenerated evidence. That is new science/methodology work rather than a review-loop correction. No source fix, audit verdict, audit worker, or landing attempt was performed.

The head branch is intentionally retained because open PR #5788 still uses it as its base.

PR #5797, opened 2026-07-30
[physics][review-loop] Cycle 789: compile the three-register companion input circuit

jonathonreilly:
## Claim

Cycle 789 constructively compiles a bounded three-register even-CAR companion input circuit and an explicit fixed coframe schedule on the landed Cycle-720 physical grammar. It diagnoses the two-bank collision, repairs it with a third local bank, and verifies collision-free fixed scheduling under all 24 proper-cubic frames and all 576 products.

## Exact boundary

This is a conditional input-side compiler with supplied already-encoded L bank, clean ancillas, fixed finite ROM/schedule, and local parity sector. It does not execute the recurrent G in the same epoch, derive genesis, provide physical time, establish a Born law, or prove a minimum/no-go/axiom-pressure result.

## Evidence

- two-bank pairwise residual 7.850462293418876e-17 but identity failure 0.7071067811865476
- third-bank identity residual 7.771561172376096e-16
- zero nearest-neighbor, return, collision, or G-firewall failures through held 6x5x4
- 24 frames x 8 origins and all 576 products checked
- hostile reversal and deletion controls active
- N1-N8 discipline applied; the broad no-go gate fails

Audit authority: none. Audit status: unset. No audit verdict is requested or applied.


jonathonreilly:
Review-loop confirmation on clean current-main reconstruction:

- independently read the theorem note, receipt, and all three runners
- cold-ran all three runners: exit 0/0/0; report hashes match the packaged receipt
- independently reconstructed the register collision, signed character channel, held-box route schedule, hostile-order residual, and inversion-CZ repair
- full `docs/audit/scripts/run_pipeline.sh`: PASS
- `audit_lint.py --strict`: PASS (zero errors; pre-existing warnings/notices only)
- authority remains none; audit remains unset; no audit verdict applied

The bounded claim and supplied/derived/open boundary are review-compatible.

PR #5835, opened 2026-07-30
Fix forensic route marker tokenization

jonathonreilly:
## Summary

- treat underscores and hyphens as lexical separators for N1 route-class marker matching
- accept ordinary plurals for alternate carrier/sector markers
- document the accepted syntax in the auditor template and retry guidance
- add regressions for the three identifiers that opened the live mechanics circuit

## Root cause

Python regex word boundaries do not split on underscores. The validator therefore rejected documented markers inside verbatim evidence identifiers such as oriented_cubical_boundary and POST_RECORD_DYNAMICS. The alternate-carrier expression also accepted singular carrier but not the ordinary plural carriers. Three deterministic false rejections reached the mechanics-circuit threshold and stopped the drainer.

The evidence manifest, verbatim-field checks, route semantics, and fail-closed N1 gate remain unchanged.

## Validation

- 319 relevant unit tests passed
- full audit pipeline passed
- strict audit lint passed with no errors
- vocabulary lint reported zero violations and zero changes
- Python bytecode compilation passed
- git diff --check passed

No audit verdict, ledger row, publication status, or science claim is changed.


jonathonreilly:
Review-loop iteration 5: PASS from a fresh GPT-5.6-Sol/max reviewer, with zero actionable findings. Landed via the fail-closed exact replay procedure onto current main as 0814ded20c0b0c9f0dd4da8291f1cca8c8162777. Verification included 257/257 adversarial marker expectations, compatibility over 379 stored packets / 2,333 routes, 473/473 tests, full audit pipeline, strict lint with 0 errors, and final main/PR freshness checks. Generated review residue was removed. This was review-only; no audit verdicts were applied. All findings from iterations 1–4 were closed.

PR #5837, opened 2026-07-31
Cycle 823: compile companion full-seam endpoint instrument

jonathonreilly:
## Claim\n\nBounded theorem: on four held finite boxes, the complete four-factor Cycle-720/822 companion seam swaps its declared matter endpoints on all 1,312 reachable physical/target phase classes. Three neutral physical M2 per edge coherently expose p = a xor b, return both scratch registers, preserve one rebuilt P_ext, and use only returned nearest-neighbour routes.\n\n## Exact evidence\n\n- 82 edges; 328 matched physical/target row pairs; 1,312 exhaustive signature replays\n- zero row-mask, monomial, endpoint-swap, scratch, pointer, collision, typing, untyped-use, elementary-parity, or prefix-parity failures\n- 14 returned CNOT route macros per edge; maximum held route distance 43\n- exact 15-factor one/two-M2 Toffoli decomposition, residual 7.35e-16\n- 24 proper-cubic frames, eight origins, 576 ordered products under transported-program covariance\n- one-particle mass residual unchanged at 5.55e-17\n- factor deletion, single-endpoint mutation, OR deletion, pointer-cleanup deletion, dirty-domain, held-size, and route-return controls\n\n## Boundary\n\nThe retained bit is a reversible endpoint-opportunity pointer, not occurrence, actuality, admission, Record, Born weight, source, clock, duration, or rate. Pointer consume/reset or a fresh bank, clean genesis, autonomous occurrence, translation-local atlas generation, target-native frame recompilation, source/gravity response, Record/Born selection, and prediction attachment remain open. This is a one-use finite supplied-program result, not a recurrent reset law or unbounded compiler theorem.\n\nAuthority: none. Audit: unset. Please use review-loop; independent audit remains separate.


jonathonreilly:
Landed atomically to main at a1d5403465 after review-loop PASS WITH BOUNDED CLAIMS; audit remains unset and independent.

PR #5886, opened 2026-08-01
feat(gravity): cycle 703 — finite-range law tournament finds no matching competitor to the supplied nearest-neighbour law

jonathonreilly:
## Summary
- Cycle 703 in the gravity evidence-ceiling lane: executes the single next discriminating experiment declared by the landed Cycle 700 note — the range-mutation row generalized from the one face mutation to a supplied three-shell finite-range family (12 face, 8 body-diagonal, 6 doubled-axis offsets; nonnegative weights).
- One new bounded_theorem note + one paired deterministic runner (33 PASS / 0 FAIL, exit 0, stdout 3186 bytes, wall 107–154 s across four runs, byte-identical output modulo elapsed).
- Result: within the supplied family and under the executed protocol, no competing law matched the executed double-relational ratio suite. Every Cycle 700 anchor reproduces at 1e-12 through independent shell machinery before any new claim is made.

## What the runner establishes
- Exact projective scale-blindness of the ratio suite and exact isotropy of all three shells' second moments, so a single continuum scale carries no discrimination — all identification content is beyond the leading continuum symbol.
- A nonsingular 3x3 first-order response of the three ratio rows to the three shell weights (analytic Jacobian confirmed by a central-difference convergence gate; smallest singular value 6.66e-3). The near-null direction has mixed signs, so it exits the nonnegative family in both orientations.
- The conic first-order minimum is computed exactly (half-plane vertex enumeration, cross-checked by an independent grid-plus-refine upper bound): minimum separation rate 0.855 per unit weight, 228x the box-19 convergence residual.
- Grid tournament over 63 supplied laws: minimum separation 0.1276, 34x the threshold (10x the box-19 residual — the landed Cycle 700 discriminator form).
- Adversarial 12-start ghost search over the continuous weight domain: best separation-per-unit-weight found 0.344, about 9x the threshold; the optimum's separation persists on four held-out box sizes that took no part in selecting it.
- Saturation finding: the face-shell response is sublinear (secant magnitude falls strictly, factor 1.93 by weight 0.1), so the direct measured separations, not the linearization, carry the identification content.

## Scope kept
- This is a comparison between supplied model laws; Cycle 700's stance sentence is preserved verbatim. No empirical falsification claim, no observed target, no uncertainty model.
- LATTICE axiom unchanged; no coupling value, sign, or scale is selected or derived; every such object is named as supplied.

## Landing convention
- Tracked artifacts: note (docs/), runner (scripts/), cold-run log + runner-written receipt (outputs/) — no logs/runner-cache entry, per the c69x-family convention.
- Separate audit-infra commit acknowledges the new note in the citation-graph manifest; the pipeline-seeded ledger shard is removed, ledger-row creation left to the audit lane. repo_invariants_check.py --check --enforce-links: PASS, graph_delta=acknowledged.

## Test plan
- [x] Runner re-run independently by reviewer: 33 PASS / 0 FAIL, exit 0
- [x] Cold run without --no-receipt produced the tracked log + receipt; receipt's embedded runner sha256 verified against the runner file
- [x] Banned-substring and preserve-string greps on both files (Cycle 700 quotes byte-exact)


JONbridgerapps:
## Review-loop result: REJECT — closing unmerged

Reviewed the exact submitted head `c0f2264f219ae8de5053178275d9a7a28aee28c7` and replayed its science artifacts onto frozen current main `f4979c872ed79379367fd6fff19d598c2da46a37` (replay head `0a30fe972940bc6cd4c7ffa1a1a87809f295f18c`). The runner is mechanically reproducible (`33 PASS / 0 FAIL`), and an independent implementation reproduced the shell moments, bulk symbol, Jacobian, conic LP, 63-point grid, local-search transcript, and secants. Those positive results do not support the submitted headline.

Blocking findings:

1. The note defines a match using absolute separation `THR = 0.0374578316797`, but the continuous search optimizes `sep(w)/|w|_1` and compares that normalized quantity directly to the absolute threshold. The admissible point `w=(0.001,0,0)` has independent ladder differences approximately `(0.0009192983, 0.0022820752, 0.0023839332)`, hence `sep=0.0023839332 < THR`. This directly falsifies the claimed continuous-domain non-match.
2. The 12-start coordinate pattern search evaluates 1,106 finite probes but supplies no global-optimizer certificate over `[0,1]^3`; it cannot establish an “adversarial optimum” or a continuous-family no-go.
3. The audit graph omits two dynamically imported load-bearing helper runners, so the restricted audit packet is incomplete.
4. The projective scale-blindness statement omits required invertibility and nonzero detector-denominator hypotheses.
5. PR-local identifiers leak into the live science note and must be removed before landing.

The NoGoDisciplineReviewer found `N1`–`N8` all failing. The sole full pipeline invocation used `AUDIT_APPLY=0` and stopped at the source-stability checkpoint because the reviewer findings file appeared during its fingerprint window; it was not rerun, and no audit worker or verdict application occurred. Pipeline cleanliness therefore also did not pass.

Salvageable after narrow repair: conditional projective scale-blindness, exact shell moments and the bulk symbol, the numerical local Jacobian/conic LP, the exhaustive stated 63-point grid result, the face-ray secants, and the finite 1,106-probe transcript. The strongest honest repaired package is bounded algebraic/numerical support—not a continuous-family no-go or identification theorem.

Final disposition: `REJECT` / `NEEDS MANUAL SCIENCE`. Closing without merge and retaining the remote head branch for repair.

PR #5887, opened 2026-08-02
feat(gravity): cycle-704 record-register symplectic dilation of the cycle-700 readout channel

jonathonreilly:
## Summary

Cycle 704 in the gravity evidence-ceiling lane. It responds to the one PARTIAL row in the landed Cycle-700 tally that names a missing construction:

> | exact reversibility with conservation, deletion, sign, scale, and range mutations | PARTIAL | C4–C4b; no Stinespring or unitary dilation of the open-system readout channel is constructed |

This cycle constructs the classical reversible completion that row says is missing — for the channel Cycle 700 actually executed — and measures exactly what the readout discards.

## What lands

- `docs/PHYSICAL_RECORD_REGISTER_SYMPLECTIC_DILATION_CYCLE704_NOTE_2026-08-01.md` (claim type: bounded_theorem)
- `scripts/physical_record_register_symplectic_dilation_cycle704_2026_08_01.py` (paired runner)
- `outputs/` cold-run transcript + runner-written receipt (c69x landing convention: no `logs/runner-cache/` entry)
- `audit-infra:` citation-graph manifest acknowledgment (sole change under `docs/audit/data/`)

## The construction (five claims, all gated)

1. **Non-disturbance, bit for bit.** The augmented system — the Cycle-700 field plus a four-site record register riding the same leapfrog flow — reproduces the executed Cycle-700 trajectory by array equality on the physical slice: final field, final momentum, and the whole hold average, no tolerance.
2. **The register carries the readout.** The register reproduces the published detector ratio against the field-side average and against the landed Cycle-700 anchor.
3. **Every factor is exactly symplectic.** Kick, drift, and record coupling are each S = I + JH with H symmetric and (JH)^2 = 0, so S^T J S = J holds identically rather than to a tolerance; the composed 1100-step map is gated dense at small size and cross-validated against the live integrator.
4. **Un-registration is exact inversion.** The explicit factor-wise inverse returns the augmented trajectory to its initial data at the same floor as Cycle 700's dynamical reversal and drives the register back to zero — records are removed by running the flow backwards, not by a separate deletion rule.
5. **Exact conservation.** Register momentum is conserved at exactly 0.0 (no update ever writes it), including under an active back-reaction check that proves the bit-identity gates can fail off the physical slice.

Openness is measured rather than asserted: the executed zero-source channel has rank exactly 4 with kernel dimension 1454 at box side 9, witnessed through the real integrator (a kernel state that moves at order one yet registers at the float floor; a row-space state louder by fifteen orders of magnitude; two states with identical records and visibly different environments). The full Cycle-700 mutation family — deletion, sign, scale, detector swap, range, conservation with reversibility — keeps its gates through the dilation.

Honesty boundary, stated in the note: the construction is classical; a quantum Stinespring or unitary dilation on the site algebra M_2(C) is not constructed and no claim about one is made — it is named as the next route this opens. One raw entrywise inversion residual sits above a naive entrywise bound and is disclosed plainly, with a normalization-free round trip corroborating it.

## Verification

- Runner: 54 PASS / 0 FAIL, exit 0; deterministic 3826-byte transcript, byte-identical across four independent runs; cold wall 6.9–8.1 s observed.
- Dependencies: markdown links to the minimal axioms and the Cycle-700 chain note only; cycles 696, 701, 702, 703 appear backticked as context (no citation edges — their PRs are open/unlanded).
- Pipeline: `repo_invariants_check.py --check --enforce-links` prints PASS with `graph_delta=acknowledged`.


JONbridgerapps:
Review-loop disposition: **PASS after one narrow repair iteration; landed**.

Reviewed exact submitted head `cdaeda69fa2309813881b04336dc18232c5bbbca` against frozen fresh `main` `f4979c872ed79379367fd6fff19d598c2da46a37` using exactly one GPT-5.6-Sol/max adversarial reviewer. The submitted algebraic cotangent lift was sound, but the physical/Record/open-system framing, audit-helper discoverability, and transcript/receipt provenance were not acceptable as submitted.

The narrow repair:

- scopes the result to a generic real cotangent-lift bookkeeping extension and the declared numerical fixture;
- makes clear that `B=0` is a supplied invariant subspace, not a selected physical sector;
- keeps Cycle 700's Stinespring/framework-site-unitary residual explicitly `PARTIAL`;
- exposes the full six-helper frozen-main execution closure to audit packets;
- regenerates a repeat-byte-identical 54 PASS / 0 FAIL transcript and binds its bytes/SHA plus the final runner SHA in the receipt.

Validation completed once on the repaired replay:

- `AUDIT_APPLY=0 bash docs/audit/scripts/run_pipeline.sh`: pass (1/1)
- `python3 docs/audit/scripts/audit_lint.py --strict`: pass (1/1; no errors)
- frozen-scope vocabulary fixer: pass (1/1; zero violations)
- independent exact algebra and numerical-fixture reconstruction: pass
- audit compatibility: pass; generated row remained unaudited, non-authoritative, and carried no verdict

The reviewed four-commit sequence landed directly on `main` as:

1. `872e10804d6fe7af945d9cd2ff9f129989b030b4`
2. `762895c829571dd3e25cb41acba5516c86fd1255`
3. `7f191400a90a2261254a3f5f2abdf769f0665f25`
4. `48c70fa8da7940053574d7b405775c94d19a74e0`

Final reviewed report: 198 lines, SHA-256 `99a8a5ab0cde15fa5800f70b80ce31f315fde238280827cf2ac8fd66045515e5`, terminal marker `TERMINAL: PASS`.

No audit verdict was applied and no authority was promoted. Open-PR ancestry checks found no descendant of this head and no open PR based on this branch, so the landed head branch is safe to delete while closing this PR.

PR #5884, opened 2026-08-01
feat(gravity): cycle 702 — the equivariance lattice of partial orbit averages is the body-diagonal subset-stabilizer lattice

jonathonreilly:
## Summary

Cycle 702 in the gravity evidence-ceiling lane. Cycle 700 landed the two-carrier split with the coframe carrier sitting at exactly the body-diagonal stabilizer `D3` (frames `[1, 4, 9, 15, 18, 23]`). The in-flight sister cycle (`PR #5879`) proved the full 24-frame orbit average exactly `O`-equivariant and the one-coset conjugation law, and deliberately scoped its prose to one-coset partial averages. This cycle supplies the multi-coset statement, with its own gates, self-contained off landed Cycle-696/700 machinery:

- **Coset collapse + containment (exact for every source):** `D3`-covariance makes the summand of `X_S(rho) = (1/|S|) sum g^{-1}.P(g.rho)` depend only on the right coset `D3 g` (four cosets), and the translation identity `X_C(u.rho) = u.X_{C u}(rho)` gives `Stab(C) <= Equiv(X_C)` with no tolerance — change of variables in a finite group sum.
- **The lattice:** the coset action is the standard permutation action on the four body diagonals (dictionary `phi([D3 g]) = axis(g^{-1}.(1,1,1))`); subset stabilizers by union size are `6, 4, 6, 24` with complement duality. The profile is non-monotone: adding a second coset shrinks the symmetry from order six to order four, and a three-coset average is exactly as symmetric as the one coset it omits.
- **Order bound:** exhaustive subgroup census of the 24-frame group — 30 subgroups, order histogram `{1:1, 2:9, 3:4, 4:7, 6:4, 8:3, 12:1, 24:1}`, with a minimal-counterexample completeness certificate gated over all 720 subgroup-element adjunctions. Exactly five subgroups have order above six and each is transitive on the four cosets, so no proper nonempty coset union has a stabilizer above order six; this excludes in particular any accidental order-eight case.
- **Measured equality at the pin:** for all 15 nonempty unions the measured equivariance set equals `Stab(C)` exactly (stabilizer-side defect `3.330669e-16`, off-side minimum `1.223444e-01`, genericity witnesses `3.670333e-01` and `7.340666e-01`), at `L = 3` with an `L = 7` two-coset spot check (`1.665335e-16` vs `6.421907e-01`). Containment is exact for every source; equality is measured at the pinned source, amplitude, and sizes, and the note says so.

[...]

PR #5879, opened 2026-08-01
science: orbit-averaged all-24 coframe carrier, cycle 701

jonathonreilly:
## Summary

Cycle 701 in the gravity evidence-ceiling lane. Addresses Cycle 700's PARTIAL contract item #1, whose landed row reads:

> | real-space all-24 carrier | PARTIAL (two-carrier split) | B5–B6: prediction carrier all-24; coframe carrier exactly D3; the energy scalars are re-pinned to a stated convention and one reference row is withdrawn |

Cycle 696's coframe compiler is covariant under exactly the body-diagonal stabilizer `D3` (6 of the 24 proper cubic rotations). This cycle builds the orbit-averaged carrier `X(rho) = (1/24) * sum_g g^{-1} P(g rho)` over all 24 proper rotations and proves + measures:

- **Theorem 1 (exact equivariance):** `X` is exactly `O`-equivariant for any pipeline `P` and any source, by change of variables in the group sum; the `1/24` weight is group-determined (zero new constants). Measured: `L=3` equivariance defect `2.220e-16` under all 24 frames vs unaveraged violations up to `0.4581`; `L=7` defect `2.220e-16` vs `1.3036`.
- **Theorem 2 (one-coset conjugation law):** `D3`-covariance makes all six terms of a right coset `D3.g0` equal (measured spread `1.456e-11`, c696's own float floor); a one-coset average equals a single term and is exactly equivariant under the conjugate subgroup `g0^-1 D3 g0`. The predicted conjugate set equals the measured small-defect set (both-ways set equality), and the four-term right-transversal average reproduces the full orbit average to `3.640e-12`.

The PD-margin scalars survive averaging: `s_pd = 0.422836427078498` vs averaged onset `0.42303983651076393` (gain `+2.034094e-04`); `lambda_min` at amp `0.40` is `0.03227752380131965`; the averaged coframe field's cubic-invariance spreads drop from `~0.33` to `<= 7.105e-15` across all four amplitudes, with the `L=7` tier at `3.539e-16`.

Claim type: `bounded_theorem`. The c690 no-go is untouched (it quantifies over one-fixed-triangulation constructions; the average repeats four). c696's scope=6 rows are untouched. Improper rotations out of scope. Odd `L` only (lattice-site centre).

## Files

[...]

PR #5888, opened 2026-08-02
feat(gravity): orbit-averaged K transport with derived D3 selection (cycle 705)

jonathonreilly:
## Summary

Cycle 705 in the gravity evidence-ceiling lane. It responds to two named openings: Cycle 700's two-carrier covariance split (scalar six-neighbour readout covariant under all 24 proper cubic frames; coframe/K chain covariant only under the six-frame body-diagonal stabilizer D3) and Cycle 702's boundary item (f), which left transport of the K chain through an orbit-averaged carrier unexamined.

The note proves and measures, through the landed Cycle-696 machinery applied to a centred 24-frame pullback average of the metric perturbation field:

1. **Exact parts-level transport law, all 24 frames.** The per-axis open finite-difference parts of the averaged coframe obey `P'_i(sigma(x)) = s_i * P_{p(i)}(x)` with the frame's row permutation `p` and signs `s` — exact at the float floor (~1e-15 at L=3 and L=7), for the executed source and an edited source alike. The averaged chain is fully equivariant at the parts level; none of the 24-frame group is lost there.
2. **The trace contraction is where D3 enters.** Summing the parts law gives covariance of the declared trace K exactly on constant-sign frames. Determinant +1 sorts those into the three all-positive even permutations (K a scalar, floor ~2e-15) and the three all-negative transpositions (K exactly odd, `K(sigma(x)) = -K(x)`); the 18 mixed-sign frames break at order the field amplitude. Both identities are law identities — they hold on the edited source at the same floors.
3. **Derived endpoint selection = the measured Cycle-700 list.** Because the endpoint excitation is even in K, the odd frames rejoin the scalar ones: the excitation row through the averaged carrier is invariant under exactly the six frames `[1, 4, 9, 15, 18, 23]` — the same D3 Cycle 700 measured — now derived from the sign structure rather than tabulated. Defect ~2e-16 on the six, broken on all 18 others; the quadrature row splits by parity (invariant on the cyclic frames, sign-flipped on the order-2 frames).
4. **The average sharpens rather than blurs.** The single-complex excitation row is D3-invariant only approximately (4.2e-12 at L=3, 4.0e-11 at L=7); the averaged row is exact to ~2e-16 — four to five orders below, with the gap growing with box side.
5. **Amplitude-1.0 positivity structure.** At the landed failure amplitude (L=3), the single chain reproduces the landed 6-of-27 coframe positivity failure and minimum perturbed length; the averaged carrier is positive-definite everywhere with margin 0.025, and its margin field is itself 24-frame invariant — a new measured fact, stated at L=3 with L=6 named as out of reach.
6. **Rejectors.** The column-form transport law breaks at 0.30; the single-complex parts law breaks; a D3-subset average does not deliver the all-24 parts law (the full average is load-bearing); an edited source moves the carrier; the domain-key cache is exact against a cache-off recomputation.

## Verification

- Paired runner: 74 checks PASS, 0 FAIL, at L=3 and L=7; deterministic across runs (byte-identical modulo wall time); stdout 5264 chars; ~6 s wall.
- Cold log and runner-written receipt under `outputs/`.
- Citation-graph manifest refreshed; the authority-link guard passes with the delta acknowledged.

## Boundary

No all-24 K-endpoint row is claimed; the six-frame scope of the K rows is located in the trace contraction, not in the averaged carrier or the parts. Even contractions of the parts are named as a next route, not adopted. Nothing here reunifies Cycle 700's two carriers; the cycle locates their difference in one contraction and derives which six frames survive it.

## Downstream

Independent of open PRs #5879, #5884, #5886, #5887 — no file overlap; citation edges go to landed notes only.

### DIFF
--- a/docs/PHYSICAL_ORBIT_AVERAGED_K_ENDPOINT_TRANSPORT_D3_SELECTION_CYCLE705_NOTE_2026-08-01.md
+++ b/docs/PHYSICAL_ORBIT_AVERAGED_K_ENDPOINT_TRANSPORT_D3_SELECTION_CYCLE705_NOTE_2026-08-01.md
@@ -1,0 +1,231 @@
+# Orbit-averaged K-endpoint transport and the derived D3 selection rule (Cycle 705)
+
+Date: 2026-08-01
+Claim type: bounded_theorem
+Status: unaudited (fresh note, paired runner below).
+
+## Dependencies
+
+Landed and unaudited:
+
+- [Minimal axioms](MINIMAL_AXIOMS_2026-06-29.md)
+- [Cycle 700 operational source-response-readout chain](PHYSICAL_OPERATIONAL_SOURCE_RESPONSE_READOUT_CHAIN_CYCLE700_NOTE_2026-07-25.md)
+- [Cycle 696 joined compiler tournament](work_history/repo/review_feedback/PHYSICAL_OPEN_COFRAME_K_ENDPOINT_JOINED_COMPILER_TOURNAMENT_NOTE_2026-07-23.md)
+
+Backticked context only, with no links: the compiler module
+`physical_open_coframe_k_endpoint_compiler_cycle696_2026_07_25.py` and its own
+dependency `c576`, from which the 24 proper cubic frames are taken; and the open,
+unlanded `Cycle 701`, `Cycle 702`, `Cycle 703`, `Cycle 704`.
+
+## Setting
+
+The Cycle-696 compiler executes one chain end to end: a supplied source is turned into
+a response, the response into a site metric `I + h`, the metric into a site coframe by
+the principal symmetric square root, the coframe into per-axis open finite-difference
+parts, the parts into the declared trace `K`, and `K` into an endpoint Hamiltonian whose
+one-excitation unitary is read out per site. Every stage used below is that landed
+function; nothing in the chain is reimplemented here.
+
+Cycle 700 measured that this chain does not carry one covariance scope but two. Its
+carrier table reads
+
+> | coframe carrier | 6 of 24 | 18 explicit out-of-scope witnesses |
+
+against, for the scalar prediction carrier,
+
+> | scalar six-neighbour carrier | 24 of 24 | anisotropic spread `0.05650461860052066` |
+
+and it identified the smaller scope explicitly:
+
+> The Cycle-696 compiler's well-posed covariance scope of six frames is exactly the
+> body-diagonal stabilizer D3, isomorphic to S3: the six frames are
+> `[1, 4, 9, 15, 18, 23]`, they are closed under product and inverse, their orders are
+> `{1:2, 4:2, 9:2, 15:3, 18:3, 23:1}`, [...]
+
+Quoted from an open, unlanded PR, the Cycle-702 boundary names the opening executed
+here:
+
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
## Review-loop disposition: REJECT (exact submitted head)

Reviewed head: `09e1f260f314ef7cb5640201ca348f2470cd77b4`

This exact head is not safe to land. The narrow finite-group algebra is salvageable, but the submitted package crosses its evidence boundary and is not audit-compatible.

Blocking findings:

1. **Audit runner misbinding.** The generated Cycle 705 ledger row binds the Cycle 696 compiler as `runner_path`; the new Cycle 705 runner is absent from both primary and helper runner fields, and the generated runner breakdown is `total_pass: 0`. The `## Paired runner` heading is not recognized by the current citation-graph parser.
2. **False source-independent exclusion.** The derivation proves D3 is the subgroup guaranteed by the declared linear contraction and even endpoint functional on unrestricted formal parts. It does not prove that every particular source breaks the other 18 frames. The admissible zero field has `K=0`, hence all 24 endpoint readouts are invariant; symmetric/cancelling fields and `sin^2` periodic aliases can also enlarge the stabilizer.
3. **Non-forensic numerical artifacts.** The runner deliberately removes a constructed digit pair from printed measurements and stores those censored strings in the receipt. The receipt contains no hashes for the note, runner, helpers, source, environment, or cold output.
4. **Unsupported cross-cycle bridge.** Cycle 700's coframe/direction-set residual and Cycle 705's downstream trace residual happen to select the same six frame indices but are different objects; the former is not rederived through `X24`.
5. **Incorrect even-size claim.** With the stated center `(L-1)/2`, signed permutations map integer boxes bijectively for even and odd `L`. Odd `L` is only a restriction of the runner's floored-center implementation.
6. **Underpowered green gates.** Direct `X24` equivariance samples only frames 0 and 2; the landed minimum is not equality-pinned; amplitude-order language is based on one amplitude; and two finite fixtures cannot establish a universal source-independent theorem.
7. **Load-bearing provenance gap.** The Cycle 696 work-history compiler supplies the source, response, coframe, contraction, and endpoint chain but is audit-excluded and not represented as a dependency in the generated Cycle 705 row.

Verification receipt:

- One full pipeline invocation only, with `AUDIT_APPLY=0`: exit 0, checkpoint `af2204953fef`.
- Standalone Cycle 705 runner: 74 PASS / 0 FAIL. This confirms the two shipped fixtures, not the universal claim.
- Strict vocabulary lint: exit 0 (report-only warnings/notices).
- Independent reconstruction closed the 24-frame enumeration, six-frame subgroup/closure, Reynolds equivariance, square-root conjugation, mirrored open-difference law, formal trace parity, and endpoint `sin^2` formula.
- No audit worker was launched and no audit verdict/state was applied.

Salvage path:

- bind Cycle 705 under a recognized `## Runner` section and make all load-bearing helpers audit-visible;
- narrow the theorem to the guaranteed D3 formal stabilizer, with the other-18 results explicitly labeled numerical witnesses for the two declared fixtures;
- remove source-independent/physical-selection language unless a new attainable-range/source-classification theorem and exact nonzero certificates are supplied;
- remove digit-dependent formatting, retain raw numerical values, and hash the full proof packet/environment;
- correct the even-`L` theorem/implementation distinction and remove unlanded/open-PR provenance language.

Final reviewer verdict: `REJECT`. The remote head branch is being retained for repair. No audit verdict was applied.

PR #5892, opened 2026-08-02
feat(gravity): cycle-708 signed-stabilizer edit-set classification

jonathonreilly:
## Summary

Cycle 708 in the gravity source-response lane. The note proves a **signed-stabilizer
classification theorem** for how the 24 proper cubic frames act on the endpoint `K`
field of a physically sourced edit set, and the paired runner confirms every branch of
it against the full real chain on four edit domains at two lattice sizes.

- **Classification (derived).** For an edit domain `D`, the stabilizer of `D` inside
  the 48 signed permutation matrices (a bookkeeping set only — the physical frame scope
  stays the 24 proper rotations) determines a sign-set for every proper frame `g`: the
  checkerboard-sign elements lying in `g·Stab(D)`, read through the multiplicative sign
  character. The sign-set is one of four values — empty, plus, minus, or both — and
  membership forces, by exact state collapse and pointwise transport, the measured
  endpoint relation `K_g = K`, `K_g = −K`, both, or neither. The converse direction
  (measured class implies predicted class) is measured, not derived.
- **Counting law (derived).** The lawful-frame count per domain follows the subgroup
  product formula `|CS·Stab| = |CS||Stab| / |CS∩Stab|`: 24 lawful frames on the
  single-edit domain, 12 on the two-edit domain, 6 on the three-edit domain, 24 (all
  double-signed) on the symmetric pair.
- **Collision criterion (derived, both directions).** A domain carries both signs on a
  lawful frame **iff** its stabilizer contains a minus-sign checkerboard element. The
  symmetric-pair domain realizes it, with corollaries all measured: `−I` in the
  stabilizer forces a palindromic `K` spectrum, pointwise antisymmetry under inversion,
  and an exactly-zero centre value.
- **Battery.** Predicted vs measured class profiles agree 24/24 on all four domains at
  both sizes: single-edit (12,12,0,0), two-edit (6,6,0,12), three-edit (3,3,0,18),
  symmetric pair (0,0,24,0) in (plus, minus, both, broken) order. Lawful floors sit at
  1e-15..1e-13 (plus branch) and 1e-12..1e-10 (minus branch); broken rows and wrong-sign
  rejectors sit 8+ orders above them.
- **Rejectors.** A proper-rotations-only classifier misses half the lawful frames of the
  two-edit domain (6 of 12); a determinant-sign model misclassifies 12 frames of the
  single-edit domain; wrong-sign checks on the sign-asymmetric domains separate by
  1.5e+08 or more; the sign-set is measured single-valued on every lawful frame.

## Honest boundary

- The **minus-branch floor values are measured, not derived** (2.5e-12..1.6e-10 across
  domains and sizes); deriving the response-stage floor constant is a named next path.
- The **class separation lands at 8.2 orders at L=7** (1.5e+08, from the single-edit
  wrong-sign row), short of the ~9 anticipated from L=3 (1.5e+09); the note names the
  mechanism (minus-floor growth with size plus wrong-sign shrink 5.0e-02 → 1.2e-02) in
  its honest-boundary section. Gate bound is the platform-stable 1.0e+07.
- Scope: L ∈ {3, 7}, open boundaries, response amplitude 5.0e-02, the four named edit
  domains. Improper signed permutations enter as computational identities only, never
  as claimed symmetries of the framework.

## What lands (5 files, 3 commits)

| commit | files |
|---|---|
| `feat(gravity)` | note `docs/PHYSICAL_SOURCE_EDIT_SET_SIGNED_STABILIZER_CLASSIFICATION_CYCLE708_NOTE_2026-08-02.md` + runner `scripts/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02.py` |
| `chore(outputs)` | cold stdout + receipt JSON under `outputs/` |
| `audit-infra` | `citation_graph_manifest.json` acknowledgement of the new note |

## Runner

`TOTAL: PASS=95 FAIL=0`, stdout 3734 bytes (under the 6000-char ceiling), well inside
the 900 s budget. Receipt floats are `{:.1e}` strings. My own triple-run at the final
working point is byte-identical in both stdout and receipt; the committed cold file is
the stdout of a fresh run of the reviewed runner.

## Review notes

Line-by-line review of the executor draft caught and fixed, pre-landing:

- One **tautological gate**: the rho bit-equality check read the collapsed pair through
  a chain cache keyed on the same domain equality it was checking, so it compared an
  object to itself. Patched to recompute rho fresh from the collapsed link state (with a
  comment pinning why the cache must not be used there). Printed values unchanged.
- Two receipt rejector fields were **hardcoded literals** (gate-verified, but against
  receipt-carries-measured discipline); now computed from the run.
- A **frame/coset conflation** in one stdout line, one receipt key, and one note bullet:
  the single-valuedness counts (24/12/6) count lawful frames, not stabilizer cosets —
  the single-edit domain's 24 lawful frames sit on 6 distinct cosets. Wording fixed in
  all three places.
- Adopted the lane's **clip discipline** pre-landing: clipped coframe values are
  diagnostics, not theorem gates. The runner asserts, per size, that every chain the
  battery evaluates — lawful, broken, and wrong-sign rejector rows alike — reports the
  principal coframe unclipped, and records the minimum metric positivity margin in the
  receipt. This gate then **caught the draft working point**: at amplitude 2.0e-01 the
  multi-edit domains (and the symmetric pair at L=3) drive the metric non-positive and
  the compiler's clip guard fires — while every theorem gate still passes, because the
  clip commutes with the frame action, so collapse and floors stay exact on silently
  clipped coframes. The amplitude moved to 5.0e-02, the largest sweep point at which no
  battery chain clips at either size; the margins are healthy (minimum metric positivity
  margin 4.8e-01 at L=3, 4.2e-01 at L=7).

The working point deliberately differs from cycle 707's: that note's single-edit
domains never clip at amplitude 2.0e-01, but this battery's multi-edit domains do, so
cycle 708 runs at 5.0e-02 and its measured floors are not line-by-line comparable to
the cycle-707 numbers.

## Relation to open work

Extends the cycle-700 evidence ceiling and the cycle-703 range-law tournament (both
cited as dependencies, landed/unaudited); sits beside the cycle-707 pointwise-transport
note (referenced as context only — nothing in this note depends on it). Named next
paths: derive the response-stage floor constant; off-centre / mixed-axis / same-order
different-subgroup edit families; the L-scaling law of the floors.

### DIFF
--- a/docs/PHYSICAL_SOURCE_EDIT_SET_SIGNED_STABILIZER_CLASSIFICATION_CYCLE708_NOTE_2026-08-02.md
+++ b/docs/PHYSICAL_SOURCE_EDIT_SET_SIGNED_STABILIZER_CLASSIFICATION_CYCLE708_NOTE_2026-08-02.md
@@ -1,70 +1,178 @@
-# PHYSICAL_SOURCE_EDIT_SET_SIGNED_STABILIZER_CLASSIFICATION_CYCLE708_NOTE_2026-08-02
+# Signed-stabilizer classifier for a supplied source-response compiler: conditional algebra and a four-domain finite battery — Cycle 708
 
-Date 2026-08-02. Paired runner
-`scripts/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02.py`;
-paired outputs
-`outputs/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02_cold_2026-08-02.txt`
+**Date:** 2026-08-02 (revised 2026-08-11, review-loop iteration 1)
+
+**Type:** bounded_theorem
+
+**Status:** proposed_retained
+
+**Primary runner:**
+[`physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02.py`](../scripts/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02.py)
+
+**Independent checker:**
+[`physical_source_edit_set_signed_stabilizer_classification_cycle708_independent_check_2026_08_02.py`](../scripts/physical_source_edit_set_signed_stabilizer_classification_cycle708_independent_check_2026_08_02.py)
+
+**Run artifacts:**
+[`cold stdout`](../outputs/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02_cold_2026-08-02.txt)
 and
-`outputs/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02_receipt_2026-08-02.json`.
+[`receipt`](../outputs/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02_receipt_2026-08-02.json),
+plus the source-bound
+[`primary cache`](../logs/runner-cache/physical_source_edit_set_signed_stabilizer_classification_cycle708_2026_08_02.txt)
+and
+[`independent-check cache`](../logs/runner-cache/physical_source_edit_set_signed_stabilizer_classification_cycle708_independent_check_2026_08_02.txt).
+
+## Trace gate
+
+```yaml
+trace_class: frontier_discovery
+target_claim_id: null
+target_blocker_text: "classify the proper-frame K-sign behavior of a supplied finite source-response compiler"
+source_of_blocker_text: frontier_question
+reachability_to_target: supports
+artifact_role: runner_certificate
+next_trace_action: "test the conditional classifier on broader edit families and derive, rather than sample, the constant-sign transport premise"
+```
+
+## Status fields
+
+```yaml
+actual_current_surface_status: bounded-support
+target_claim_type: bounded_theorem
+trace_class: frontier_discovery
+reachability_to_target: supports
+conditional_surface_status: conditional-support
+hypothetical_axiom_status: null
+admitted_observation_status: null
+claim_type_reason: "an exact finite-group implication conditional on a supplied compiler transport premise, with that premise and its numerical converse tested only on four edit domains at L in {3,7}"
+audit_required_before_effective_retained: true
+bare_retained_allowed: false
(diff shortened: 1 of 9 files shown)


jonathonreilly:
Review-loop PASS after one repair iteration; landed direct-to-main at `f7f7559a7475784d9b998c70402e742ba377c474` from exact current-main base `591b4364071e82de78ef6230dbeb00107688f9e2`.

Scientific disposition: BOUNDED; proof obligations CONDITIONAL; imports DISCLOSED; audit compatibility PASS; No-Go Discipline NOT APPLICABLE. The exact finite-group implication is valid for a supplied edit domain when its K-map satisfies the stated constant-sign transport premise. For the four tested domains, the proper-frame profiles are `12/12/0/0`, `6/6/0/12`, `3/3/0/18`, and `0/0/24/0`; the inversion pair consequently has both signs on every proper frame and the tested K-map is palindromic/centrally antisymmetric to numerical tolerance. Numerical premise/converse evidence is limited to those four domains at L=3 and L=7; this is not a theorem for arbitrary integer edit sets.

Repaired before landing: narrowed the arbitrary-domain overclaim; corrected the collision proof sidedness; proved determinant balance in the subgroup-product count; enumerated every lawful representative rather than one representative per sign; minimized wrong-sign distances over all frames; declared transitive runner inputs/timeouts; replaced bare gate/domain shorthand; removed a non-load-bearing dependency; added audit metadata, caches, and a pure-integer independent checker; registered that helper in the citation graph.

Verification: primary `95 PASS / 0 FAIL`; independent checker `59 PASS / 0 FAIL`; hostile right/left-coset mutation rejected with 8 label failures; hostile transitive-input mutation produced `input_mismatch`; full exact-current-main pipeline PASS; strict audit lint no errors; changed-science evidence readiness PASS; enforced repository links/invariants PASS; vocabulary lint 0 violations; `git diff --check` PASS. Generated audit ledgers, queues, status views, and verdict surfaces were stripped; audit status remains independent-lane `unaudited`.

PR #5900, opened 2026-08-02
feat(gravity): the frame group factors through the zero-defect stabilizer, and the minimal covering subgroup has order four — Cycle 715

jonathonreilly:
## What this responds to

Cycle 714 measured that one supplied source's pairing collapses onto four
values under the 24 proper rotations, and Cycle 707 measured that the endpoint
sign law has an order-4 source stabilizer. Two four-fold structures,
apparently unrelated. This cycle identifies them as complementary halves of a
single exact factorization and converts the count four from a measurement into
an index.

## What the delta claims

Three results over the landed Cycle-696 static spatial assembly at `L = 3` and
`L = 4`. The group arithmetic is recomputed inside the runner; three membership
literals are carried as declared fixtures, and every property claimed of each
one is gated.

**I — exact factorization.** The zero-defect set `S` (frames that reassemble
the operator unchanged, measured at tolerance `1.0e-09`) has order 6 and is
closed under composition in 36/36 ordered products. The quartet generated as
powers of the x-axis rotation coincides with the transcribed Cycle-707 source
stabilizer, meets `S` in the identity alone, and the product map `S x C4` is a
bijection onto all 24 proper rotations. The index of `S` is `24 / 6 = 4`.

**II — the operator is a function on the four right cosets.** The coframe
relabelling composes as an anti-homomorphism, so frames fixing the operator act
on the left of the frame label and the invariant is the right coset. Measured
within-coset deviation `1.2e-10` at `L = 3` (n = 98) and `L = 4` (n = 279)
against an operator scale of `29.5`; the four representatives are separated by
`4.0000`; clustering all 24 reassembled operators returns exactly four. These
rows read no source at all.

**III — covering subgroups are sufficient, and the minimal COVERING order is
four.** Averaging a supplied source over a subgroup `H` makes its pairing
frame-blind whenever the product set `S . H` is the whole group. The
product-set count turns that into arithmetic: `|H| = 4 |H meet S|`, hence
`|H| >= 4` with equality exactly for the complements of `S`. Nine of the 30
subgroups cover, with orders `[4, 4, 4, 4, 8, 8, 8, 12, 24]`, and the Cycle-707
stabilizer is one of the four minimal covering subgroups.

This is the sufficiency direction only. No converse is claimed, and no
classification of frame-blinding subgroups is claimed. What is measured in the
converse direction is a finite fact about declared fixtures: on the six seeded
Gaussian source fixtures, blindness and covering agree in 30/30 subgroups at
both sizes, separation factor `2.2e+09` at `L = 3` and `5.3e+08` at `L = 4`.

**Published witness.** The normalized constant source is fixed by every
relabelling permutation, so it is already frame-blind under the trivial
order-one subgroup — measured spread `3.3e-16` at `L = 3` and `7.4e-14` at
`L = 4`, gated as `III.constwitness`. It is published as an existence witness,
not as a refutation: the covering claim above is scoped to what the runner
measured.

## Source scope of the measured rows

The runner computes and gates its own row inventory (`III.rowscope`): of the 14
measured blindness rows, 12 are properties of the six seeded Gaussian fixtures
(`III.scan`, `III.gap`, `III.sextet`, `III.subset`, `III.c707blind`,
`III.all24` at each size) and 2 are not — the two `III.constwitness` rows are
measured on the normalized constant source, declared as the seventh fixture.
The Theorem II operator rows use no source.

## Rejectors

Left cosets are NOT constancy classes (deviation `4.0000`, so a construction
that mistook the anti-homomorphism for a homomorphism fails the row); a
non-complement order-4 subgroup meets `S` twice and covers only 12 of 24; the
sextet `S` alone does not blind the seeded fixtures (`3.9e-02` / `6.2e-01`);
and a four-element subset drawn from inside `S` is gated to lie inside the
measured `S`, is not a subgroup, and does not blind the seeded fixtures
(`5.2e-02` / `8.4e-01`).

## Mutations run against the gates

- `FOUR_IN_S = (1, 4, 9, 23) -> (0, 4, 9, 23)`, frame 0 outside the measured
  `S = [1, 4, 9, 15, 18, 23]`: `III.fourinS` FAIL, exit 1, `PASS=39 FAIL=1`.
- drop `all24` from the declared seeded-row list: `III.rowscope` FAIL (10 vs 12
  seeded rows), exit 1, `PASS=39 FAIL=1`.
- misclassify one constant-source row as absent: `III.rowscope` FAIL (1 vs 2
  constant-source rows), exit 1, `PASS=39 FAIL=1`.
- `BAD4 = (0, 4, 19, 23) -> (0, 4, 18, 23)`: `I.rejectorsubgroup` and
  `I.rejector` FAIL, exit 1.
- replace the constant source with a seeded Gaussian draw: both
  `III.constwitness` gates FAIL, exit 1.
- an exactly antisymmetric source under a nonidentity involution (zero-norm
  average): the spread routine returns NaN and the enclosing gates FAIL, exit 1
  — the degenerate branch fails closed rather than substituting a denominator.

Independent non-shared-path check: explicit dense permutation matrices with
`numpy.linalg.solve` instead of `argsort` plus an inverse reproduce the
constant-source spreads.

## Landing set and runner

Four files: the source note under `docs/`, the runner under `scripts/`, its
canonical cache under `logs/runner-cache/`, and the paired receipt under
`outputs/`.

`scripts/physical_frame_group_factorization_and_minimal_blinding_cycle715_2026_08_02.py`
— `TOTAL: PASS=40 FAIL=0`, exit 0, a few seconds. It declares
`AUDIT_TIMEOUT_SEC = 120`, and the cache was written through
`scripts/runner_cache.py` `execute_and_write_cache` at that declared timeout,
so the cache header carries `timeout_sec: 120` and an
`input_fingerprint_sha256` over the declared `AUDIT_INPUT_PATHS`.

The Cycle-696 compiler chain is a load-bearing CODE dependency, not context:
its five modules are the runner's declared inputs, and the compiler is loaded
through a module-level path constant so the packet builder resolves the whole
chain into this claim's restricted audit packet. The Cycle-707 note is a
load-bearing markdown dependency (one citation-graph out-edge) and is also
content-pinned as a declared input because the stabilizer quartet is
transcribed from it.

## Boundary and audit

The order of `S` is measured, not derived — "the index is four" is exact given
`|S| = 6`. Every floating-point row is conditional on the audit-excluded
Cycle-696 compiler chain, and the `C4` identification is conditional on a
transcription from the unaudited Cycle-707 note. The operator is indefinite in
this sector, so no positivity reading is licensed. Sizes are capped at `L = 4`
for the full 30-subgroup scan by host memory. Static spatial sector only.

This branch introduces one `bounded_theorem` row, unaudited: independent audit
is required before any retained grade, and this PR writes no `audit_status`,
`effective_status`, or audit verdict.

Hard landing condition recorded in the note: co-land a refreshed
`docs/audit/data/citation_graph_manifest.json`, regenerated on the actual
landing tree. Built on a disposable tree at `origin/main` `323d7fc32d` plus
this delta, the topology delta is exactly one added node — this note — with
out-degree one, its single edge pointing at the Cycle-707 note, and nothing
removed or rewired.


jonathonreilly:
Review-loop PASS and landed on exact current main.

Scientific repair: retained the exact finite frame-group complement, right-coset, and covering-subgroup results conditional on the L={3,4} numerical near-zero-defect sextet; withdrew the universal minimum-blinding/necessity claim. The all-ones source is an explicit identity-subgroup accidental-blindness witness. The Cycle-707 stabilizer is independently reconstructed, transported-Hessian equality is correctly framed as four numerical clusters, and source-generic/continuum/arbitrary-L claims are excluded.

Evidence: primary runner PASS=44 FAIL=0; independent non-importing checker PASS=20 FAIL=0; hostile orientation/non-complement/matrix-mutation/noncovering tests pass; canonical caches are exact-input fresh; full exact-main pipeline, strict audit lint, changed-evidence readiness, vocabulary lint, py_compile, invariants/links, and diff checks pass. No-Go Discipline is not applicable after removing the universal negative boundary. No audit verdict was applied.

Original head: 36b538b7162ca8fdd648fa08d73f2c40734258a1
Validated base: d2a116d919a529fbc42ba45a39db124c8ef639e1
Reviewed/landed head: 71a25bb17fbb7c57d823244ee03988d740f75962

PR #5906, opened 2026-08-02
feat(gravity): the frame dependence is non-spectral — the Rayleigh transfer and the level-set law (cycle 718)

jonathonreilly:
## What this responds to

Cycles 715–717 established *what* the reassembled static operator's frame
dependence is a function of: the body diagonal `delta(g)` that the frame carries
onto `(1, 1, 1)`, whose four fibres are the right cosets of the measured
stabilizer sextet. They did not say **where in the operator** that dependence
lives, nor why averaging over a subgroup transitive on the diagonals should make
a value agree when the objects being averaged plainly do not.

## What changed

A new bounded-theorem note, its paired runner, cold stdout and receipt, and the
citation-graph manifest acknowledgement.

**Reassembly is an orthogonal relabelling.** `Q_g = P_g Q P_g^T` holds at
deviation `0.0e+00` over all 24 frames at `L = 3` and `L = 4`. Permutation
matrices are orthogonal, so the 24 reassembled operators are isospectral: sorted
eigenvalues agree to `1.8e-13` / `4.3e-13`, and trace, Frobenius norm and
log-determinant to `1.7e-13` / `1.8e-12`. Meanwhile the entries differ by at
least `4.0e+00` across body diagonals against `1.2e-10` within one. The frame
dependence is therefore entirely non-spectral. The comparison resolves: a
rank-one `1.0e-03` bump moves the spectrum by `2.3e-04` / `2.1e-04`, nine orders
above the isospectrality residual.

**The conjugator index is derived and the wrong one is rejected.** All 576
ordered frame pairs conjugate by the inverse-left product at deviation
`0.0e+00`; the reversed product fails on the 456 pairs where it differs, by up
to `4.0e+00`.

**The frame moves onto the source.** `v_A(g) = R(bbar_{gA})` with `R` the
Rayleigh quotient of the single fixed identity-frame operator — agreement
`2.2e-16` / `6.0e-16` at `L = 3` and `5.9e-14` / `2.0e-14` at `L = 4` over 120
evaluations each. `R(u)` is a weighted mean of inverse eigenvalues with
non-negative weights (`3.1e-15` to `7.9e-14`), so every attainable value lies in
their convex hull — measured `-7.1e-01` to `2.1e+00` at `L = 3` and `-2.3e+00`
to `2.7e+01` at `L = 4`, with the observed values strictly inside. The hull
straddles zero: the operator is indefinite and the functional is not a norm.

**The level-set law.** The 24 pulled-back sources are pairwise at least
`1.2e+01` / `2.1e+01` apart, yet `R` takes exactly four values on them and its
level sets are exactly the four body-diagonal fibres of six — within `3.0e-12`
against between `2.7e-03` (ratio `9.2e+08`) at `L = 3`, within `1.9e-10` against
`5.6e-03` (ratio `3.0e+07`) at `L = 4`.

**Blindness is a coincidence of levels.** A minimal blind coset has left
stabiliser of order 4, so its 24 left translates collapse to 6 averaged sources.
Those 6 are at least `2.4e+01` / `3.7e+01` apart with spectral weights differing
by up to `9.2e-01`, and still share one value to `2.8e-12` / `6.6e-11`. A
four-frame set drawn from a single diagonal, by contrast, has 24 distinct
translates and spreads `3.9e-02` at `L = 3`, `5.2e-01` at `L = 4`.

**Size-dependent census, reported as such.** Of the 15 coincidence patterns on
four diagonals, the 98 single slots at `L = 3` realise 14 — the missing one is
the all-distinct pattern, so no single-slot record at `L = 3` separates all four
diagonals. At `L = 4` all 15 are realised and 48 of 279 slots separate all four.
Slots that are frame constant with no averaging number 6 at `L = 3` (direction
classes 1, 3, 7) and 19 at `L = 4` (classes 5, 9, 11, 13). Generic sources
realise the finest pattern 20 of 20 at both sizes, which localises cycle 716's
"structured sources blind more" observation in the source rather than the
assembly.

## Runner

`scripts/physical_spectral_blindness_rayleigh_level_set_cycle718_2026_08_02.py`
— `TOTAL: PASS=65 FAIL=0`, re-run by the author. Cold stdout and receipt are
landed under `outputs/`; every floating-point value in the note is printed by
the runner itself.

## Downstream order

Follows the cycle-715, cycle-716 and cycle-717 branches, which are open in
parallel; this note cites them by name without a citation edge, so it applies
independently of their order.


jonathonreilly:
Review-loop PASS; landed fail-closed on exact current main.

Provenance:
- reviewed submitted remote head: d7bc518a39462e724a3beb441a0c00d68f81c26d
- replay base / contained predecessor: b9f3a4975dfb3aa3dbfde08bd94fc093badda01a
- repaired and landed head: 757f60665f085311d1aefad2099d7580be732b32

Scientific correction: the submitted individual-eigenvector weight comparison was not invariant inside the many numerical doublets. Re-deriving the measure with eigenspace projectors shows the stronger, correct finite result: the six distinct regular-coset averaged sources are symmetry-related and share the full eigenspace-resolved spectral measure, so their common Rayleigh value is symmetry-enforced rather than an accidental scalar collision. The note and runners now state only exact finite permutation/spectral-measure identities and numerical orbit/census results for the supplied Cycle696 compiler at L=3,4 with two named probes. No generic-source, arbitrary-L, continuum, or source-independent necessity/sufficiency claim is made.

Validation:
- primary exact-main runner: PASS 82 / FAIL 0
- independent dense-permutation/eigenspace reconstruction: PASS 26 / FAIL 0
- hostile mutations: eigenspace clustering, action order, and census count all fail detectably
- full repository pipeline: PASS
- strict audit lint: PASS, zero errors
- changed-evidence readiness: PASS
- caches fresh; vocab lint, py_compile, diff/link portability, and graph invariants PASS
- graph delta: one node, four predecessor edges

Claim physical_frame_orbit_spectral_measure_and_rayleigh_census_cycle718_note_2026-08-02 remains bounded_theorem / unaudited and requires the independent audit lane. No audit verdict was applied.

PR #5912, opened 2026-08-03
feat(gravity): derive the frame group and box-centre centrality from the assembly stencil (cycle 721)

jonathonreilly:
## What this responds to

Cycles 717 through 720 established, by direct evaluation of the assembled static
open-box form, that a twelve-element set of signed axis permutations leaves that
form fixed, that its six proper members are a sextet inside the twenty-four
proper rotations, and that the box-centre point reflection commutes with all
twenty-four. In those cycles the group was **measured, not derived** — found by
relabelling the assembled form and reading off which relabellings changed
nothing. The path left open at the close of cycle 719, and again at the close of
cycle 720, was whether that structure is derivable from the assembly stencil
instead. This PR derives it.

## What changed

A new source note, a paired class-A runner, the runner's cold standard output
and receipt, and a citation-graph manifest refresh acknowledging the new note.

The derivation is a four-link chain, and the assembled form is consulted only
afterwards, to confirm a prediction that was already fixed:

1. **Affine identity.** The compiler's own centre-conjugate site map has
   translation part exactly `(I - R)c`, an integer box-corner offset taking only
   `0` and `L - 1` per axis. Deviation `0.000000e+00` over 48 signed axis
   permutations and box sizes 3, 4, 5, with `0` site mismatches against the
   landed map. Dropping the offset sends `7938` site images out of the box, and
   four unimodular non-permutation rejectors give non-integer offsets — the
   identity is special to the signed axis permutations.
2. **Homomorphism.** The induced edge-slot relabelling composes exactly:
   `0` mismatches over all `2304` ordered pairs at both L = 3 and L = 5, against
   `175104` and `1094400` under the reversed composition order.
3. **Centrality as a corollary.** The box-centre point reflection is the scalar
   `-I`, hence central; the homomorphism transports that to slot maps, worst
   commutator `0`, with no assembled form evaluated anywhere in the argument.
   A single-axis sign flip fails to commute with `32` of 48, so centrality is
   not automatic. The cycle-719 closed form for the box-centre relabelling is
   **recovered** at `0` mismatches rather than remeasured.
4. **The stencil fixes the group.** All `24` path simplices of the base cell
   carry the body diagonal of the four-cube, so a signed axis permutation
   preserves the stencil exactly when it preserves the body-diagonal line —
   agreement on 48 of 48. Order `12`, determinant split `6` proper and `6`
   improper, proper half exactly the landed sextet `(1, 4, 9, 15, 18, 23)`,
   and `24 / 6 = 4` cosets. The four-valued frame label of cycles 717 to 720 is
   therefore the coset count, as a theorem about the stencil.

Two further results. The improper half is supplied by the periodic tick fold and
by nothing else — holding the tick fixed drops the stabilizer to order `6`, and
that residual six is exactly the proper half. And the twenty-four local pieces
of the assembly are pairwise identical at spread `0.000000e+00` despite carrying
twenty-four distinct class tuples, which is what makes the transport
combinatorial.

## Runner numbers

`TOTAL: PASS=46 FAIL=0`, exit code 0, 4411 characters of standard output, two
consecutive runs byte-identical in both stdout and receipt.

Prediction: at L = 3, 4, 5 (98, 279, 604 slots) the derived order-12 group
predicts assembled-form invariance on all 48, within `1.243450e-10` against a
floor of `4.000000e+00` outside the group and an entry maximum of
`2.945214e+01`. Measured frame classes `4` of block size `6`, matching the
derived cosets, at every box size.

Discriminating gate: on six deliberately mutilated stencils the derived and
measured groups are **equal**, not merely nested — orders 12, 1, 1, 12, 2, 1
across the full stencil, two single templates, one full orbit, a non-orbit half,
and a pair, at floors from `1.221442e+00` to `1.116895e+01`. The derivation
moves with the template set.

## Boundary

The chain derives the group from the supplied stencil and box geometry; it does
not derive it from the axioms. The stencil, the tick fold of length 2, and the
slot storage convention are all supplied by the landed cycle-696 compiler, and a
different admissible stencil gives a different group — which is exactly what the
counterfactual gate shows. The six improper members remain registered as
computational identities of the assembled form, not lattice symmetries: the
Lattice axiom names proper cubic rotations only. Every physical statement is
carried by the proper sextet alone.

The combinatorial rows — offset identity, homomorphism, centrality corollary,
stabilizer order and split, sextet identification, coset count — are exact
integer facts. The invariance and separation figures are machine-precision
measurements of one fixed compiler at three box sizes.

## Ordering

Independent of the other open cycle-71x and cycle-72x PRs; it reads only landed
main-resident notes and the landed cycle-696 compiler. Cycles 717, 719 and 720
are referenced as backticked context with no authority edge, since they are in
flight.


jonathonreilly:
Review-loop PASS and contained on exact main `afa5b2654164a25eed50cd9e54749df799734425` (reviewed original remote head `00d7fd182c7f86a4e63a518e8d97ef50d591fcd0`, replayed against predecessor main `da860072f92de3e9be2325213a4cb03b8efba479`).

Substantive repair: the submitted `46/0` runner checked only that the tick-fixed stabilizer had order six and was a proper subset, then incorrectly identified it as the proper half. Independent reconstruction shows the tick-fixed group is the unsigned coordinate-permutation `S_3`, with determinant split `3+3`. The modulo-two fold adds its disjoint globally sign-reversed coset, also split `3+3`; the folded order-12 stabilizer has proper sextet `(1,4,9,15,18,23)`, drawing three members from each coset. The note, primary gates, independent checker, receipt, canonical caches, and claim boundary were repaired accordingly.

Bounded result: exact finite enumeration of 48 signed axis permutations and 24 supplied path templates; exact site/slot action and central `-I` at `L=3,5`; supplied-Cycle696 assembled-form agreement at `L=3,4,5` (inside residual at most `1.243450e-10`, outside floor `4`); and six altered-template controls with derived/measured orders `(12,1,1,12,2,1)`. This is conditional on the supplied compiler/stencil conventions, not an arbitrary-`L`, continuum, framework-improper-symmetry, or axiomatic stencil derivation.

Validation: primary `49/0`; independent `30/0`; four hostile mutations each failed; canonical caches fresh on exact input closure; full 18-stage pipeline PASS; strict audit lint zero errors; changed-evidence `checked=1/failures=0`; invariants `link_violations=0`, `class_f_violations=0`, graph delta acknowledged; vocabulary zero violations; portable links, bytecode compile, and diff checks PASS. Canonical No-Go trigger was false. Generated audit/status projections were stripped. No audit verdict was applied; the row remains `bounded_theorem` / unaudited pending the independent audit lane.

PR #5914, opened 2026-08-03
fix(audit): prevent redundant batch attempts

jonathonreilly:
## Summary
- suppress same-policy stdout-transport re-audits only while the terminal audit's complete blocker fingerprint is also unchanged
- requeue when cache, dependency, runner, or other authenticated blocker state moves under the same packet policy
- select parallel audit waves as an antichain over transitive mutable ordinary dependencies while preserving safe sibling and accepted-premise parallelism
- exclude presentation-only `generated_order` from alternate-source row identity so unrelated candidate renumbering does not stale independent deliveries

## Why
A production audit window exposed avoidable repeat work and packet invalidation across concurrent targets. The original patch removed immediate same-policy repeats, but review found that policy equality alone could strand a real artifact repair and that direct-dependency collision keys both serialized safe siblings and missed ancestor/descendant collisions. The reviewed implementation uses the canonical blocker comparator and the full ordinary-dependency relation without changing evidence, packet, or verdict standards.

## Validation
- 189 audit-loop orchestration tests passed serially
- 485 audit-pipeline tests passed
- 62 focused changed-path tests passed
- 21 dispatch-shadow tests passed
- 22 audit-science-fingerprint tests passed
- full audit pipeline passed
- strict audit lint passed with zero errors
- changed audit evidence: checked=0 failures=0 control_failures=0
- vocabulary lint passed
- Python compilation passed
- git diff check passed

Generated audit-state outputs were restored before publication per review-loop policy. No audit verdicts or status artifacts are included.

### DIFF
--- a/docs/audit/scripts/compute_reaudit_candidates.py
+++ b/docs/audit/scripts/compute_reaudit_candidates.py
@@ -319,8 +319,9 @@ def build_payload(rows: dict[str, dict]) -> dict:
             "hash has changed since the audit_state_snapshot was taken; or "
             "(c) the audit requested complete runner stdout and a usable "
             "source-bound cache falls above the legacy 6k limit but within "
-            "the current 20k head+tail transport budget, and the terminal "
-            "audit predates the current packet-policy fingerprint."
+            "the current 20k head+tail transport budget, unless the terminal "
+            "audit used the current packet-policy fingerprint and every "
+            "recorded blocker fingerprint remains unchanged."
         ),
         "packet_policy_fingerprint": packet_policy,
         "eligible_claim_types": sorted(ELIGIBLE_CLAIM_TYPES),
--- a/docs/audit/scripts/compute_reaudit_candidates.py
+++ b/docs/audit/scripts/compute_reaudit_candidates.py
@@ -23,6 +23,7 @@ import sys
 from datetime import datetime, timezone
 from pathlib import Path
 
+import audit_science_fingerprint
 import premise_nodes
 import ledger_io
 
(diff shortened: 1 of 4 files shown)


jonathonreilly:
## Review-loop disposition: PASS — landed

- Reviewed PR head: `c4dfb4b6f53360b76b67c66b04bce5a452b2e854`
- Landed on `main`: `fadd2aeca0d442ae242a02c2c829f9ac2e11977d`
- Reviewed/landed tree: `584c2b3b0662c82c30b63832e85d0156d192990f` (exact match)

### Findings fixed

1. Same packet-policy equality could suppress a real transport re-audit after blocker evidence moved; suppression now also requires the canonical complete blocker fingerprint to remain unchanged.
2. Direct dependency collision keys serialized safe siblings but missed transitive ancestor/descendant invalidation; batch selection now uses a transitive ordinary-dependency antichain with accepted-premise/meta stops.
3. `generated_order` made alternate-source identity change when an unrelated earlier candidate disappeared; it is now excluded as presentation-only metadata while substantive fields remain authenticated.
4. The generated policy summary now states the actual packet-policy-plus-blocker conjunction.

### Confirmation

- Full 18-stage audit pipeline: PASS
- Strict audit lint: PASS, zero errors
- Changed audit evidence: `checked=0 failures=0 control_failures=0`
- Audit-loop orchestrator: 189/189 PASS (serial)
- Audit pipeline tests: 485/485 PASS
- Focused changed-path tests: 62/62 PASS
- Dispatch shadow: 21/21 PASS
- Audit science fingerprint: 22/22 PASS
- Vocabulary lint, Python compilation, and diff check: PASS
- Committed scope: exactly the original four script/test paths; no audit verdict, ledger, queue, status, or publication outputs landed

Science/no-go/proof/import lenses were not applicable to this audit-tooling-only change. One optional legacy fixture module, `test_fingerprint_v1_stamping`, remains red on unchanged `main` because its temporary axiom-premise registry omits the now-required `nodes` object; the fixture, writer, and fingerprint reader were unchanged by this PR, so no out-of-scope repair was folded into this landing.

PR #5920, opened 2026-08-03
feat(gravity): adjacency-admissible assembly and the excess-slot trade (cycle 723)

jonathonreilly:
## What this responds to

The cycle 722 note named four next paths; this is the second of them — attack the
slots where the assembly stencil exceeds the LATTICE axiom's 6-NN adjacency. If a
construction could be assembled from axiom-adjacency slots alone, the frame-label
structure of cycles 721–722 would become entirely axiom-internal.

## What changed

New source note, paired runner, cold run and receipt, plus the citation-graph
manifest acknowledgement stage 18 requires for a new note.

**Combinatorial half — exact integer arithmetic, complete enumerations.**

- Counting by **spatial footprint weight** (the L1 weight of a slot's spatial part)
  rather than by "purely spatial", the stencil's 240 slot-uses split 24/108/72/36.
  So **108 slot-uses, and 21 of the 65 distinct slot variables per cell, exceed the
  axiom's adjacency** — the seam is larger than the 48 purely spatial ones.
- Affine independence bounds 6-NN pairs at 2 among three corners and at 3 among four
  independent corners; reaching 4 needs a dependent quadruple, carried in the runner
  as an explicit rejector.
- Over all 4368 corner 5-subsets of the tick-extended cell, **3008 are nondegenerate,
  and every one carries at least 3 exceeding slots at each of the four tick splits**.
  No corner assembly stencil is adjacency-only. The Kuhn path stencil carries 5/4/4/5
  against that floor, and attains the purely spatial floor 3/1/1/3 exactly.
- Facet forcing (a corner 4-simplex meets the tick-0 hyperplane in a 3-face exactly
  when 4 corners lie there, and then in exactly one), a verified cone relation
  (24·vol = 6·vol of the base facet), and the facet census — **58 nondegenerate corner
  tetrahedra, 182 corner dissections** by exact integer separating-axis tests, floor
  18 — together with the volume bookkeeping give a **global floor of 48 exceeding
  slot-uses** for any corner stencil, against the Kuhn stencil's 108.
- The note states plainly, and the runner gates, that this floor of 48 and the
  stencil's purely spatial count of 48 are counts of two different things whose
  agreement carries no content.

**Assembly half — the open-coframe endpoint compiler, tick-resolved, open box.**

- The assembled form is **cell-local**: largest entry past one cell exactly
  0.000000e+00.
- The exceeding block carries exactly **one flat direction per cell per tick** (16, 24
  and 54 at the three sizes), annihilated by the mixed block, with a well-conditioned
  live part — so elimination is well defined.
- **Deletion horn:** dropping the exceeding variables discards **0.529** of the form
  at box size 3 and **0.547** at box size 4.
- **Elimination horn:** the Schur complement keeps the stationary value (deviation
  below 1.0e-12; a shifted form breaks it at 8.608e-04) but generates range — **0.235**
  past one cell at box size 3, and **0.160** at range 2 plus **0.304** at range 3 at
  box size 4, the full box diameter, with largest past-cell entry 6.318 against the
  assembled form's exact 0. The range is generated, not inherited.
- **The frame label survives both horns:** symmetry count 6 and 8 frame labels for the
  full, deleted and eliminated forms at both box sizes.

## Runner

`scripts/physical_adjacency_admissible_assembly_trade_cycle723_2026_08_03.py` —
**34 PASS / 0 FAIL, exit 0**, ~9 s, ~285 MB peak. Re-run in the worktree reproduces
the committed cold file byte for byte.

## Claim type

`bounded_theorem`. Authority: none. Audit: unset. No new axiom or primitive is
proposed or adopted. Dependencies cited as landed and unaudited.


jonathonreilly:
Review-loop PASS and direct-main containment complete.

Provenance: original remote head `2884b16dd6649dd6ac655832776fd3488686e60b`; repaired/confirmed PR head `1938aeb0ae6bf0f8144f417d8dcdb243a5eb621b`; contained main SHA `8ab9ed0bc3a1abe89d8a25ac2bddcaad0a3dae9c`.

Science boundary: exact exhaustive conservative floor 48 only for a supplied complete one-tick-unit-cell dissection into nondegenerate corner 4-simplices with all vertex pairs counted as slot-uses (the later Cycle725 result on main strengthens the supplied-model floor to 68/all pieces and 108/minimal pieces). The Cycle696 deletion/cutoff-pseudoinverse/stabilizer-partition results are finite numerical rows only: matrices at L=3,4 and the stated L/LT flat-count scan. The submitted exact-stationary language was refuted and removed; the retained cutoff construction has nonzero D-gradient norm `3.147189e-06`. No physical all-construction, exact stationary, arbitrary-L, boundary-free, or continuum claim.

Evidence: primary 36/0, independent exact checker 10/0; deterministic fresh stdout/receipt; local threshold and Cycle696 semantic mutations fail closed and invalidate both caches; full exact-main pipeline 18/18; strict lint zero errors; changed evidence checked=1/failures=0; invariants rows=4784/link=0/Class-F=0/graph acknowledged; vocabulary/diff clean. Generated audit/status outputs stripped; only source packet, canonical caches/receipt, helper registration, and manifest acknowledgment landed. Audit remains unaudited; no audit verdict was applied.

PR #5933, opened 2026-08-03
feat(gravity): scale-free adjacency ceiling and the dissection cost bracket (cycle 724)

jonathonreilly:
## What this responds to

The preceding cycle measured that no **corner** assembly cell of the tick-extended unit cell can be built from the six nearest-neighbour directions the LATTICE axiom names, and left the obvious escape open: a refined cell, or vertices at other rational positions, might be adjacency-only after all. If that escape were real, the frame-label structure of the two cycles before it would rest on a choice of cell rather than on the axiom.

## What changed

**The escape is closed off by a rank argument, not by enumeration.** Any set of lattice sites pairwise at L1 distance at most one occupies at most **two distinct positions** (verified: zero offending triples over a 125-site box). An adjacency-only vertex set therefore lies in a two-site slab, its affine rank is 1 or 2, and every five-subset has volume exactly zero. So **no nondegenerate 3- or 4-simplex is adjacency-only at any lattice resolution, any box, any vertex choice.** The corner restriction is not doing any work. Rejector: the tight corner cell has affine rank 4 at adjacency cost 3.

The cell count then becomes a cost question, and this PR brackets it.

| quantity | value |
| --- | --- |
| per-cell adjacency floor, corner census | 3, attained by exactly 64 cells (all doubled stars, all minimal volume) |
| floor for an **arbitrary** corner dissection | **56** (volume-weighted, at cost 7 / volume 3 of 24) |
| floor for a **unimodular** corner dissection | **96** |
| achieved by the monotone path stencil | **108** |

The 56 supersedes the previous cycle's 48 by a single line of volume-weighted arithmetic over a complete census, replacing a facet-census plus cone-relation plus volume-bookkeeping chain. The 96 comes from writing the cost as 120 minus the saving and capping the saving by two interior-disjointness clique maxima (8 over floor-cost cells, 16 over below-average-cost cells, the latter witness containing no floor-cost cell). The bracket moves from 48-to-108 to **96-to-108**.

Refining the resolution once gives 2449800 nondegenerate cells, the same per-cell floor of 3, and a region floor of 80. The note and the runner both state against interest that 80 sits **below** the achieved 108, so refinement raises the guaranteed minimum and does not raise the achieved count.

## Runner

`scripts/physical_scale_free_adjacency_dissection_bracket_cycle724_2026_08_03.py` — 23 PASS / 0 FAIL, exit 0.

Adversarial hardening carried in the runner itself, because the clique numbers are the attackable part:

- the interior-disjointness predicate is an exact integer separating-hyperplane scan, cross-checked against a structurally different linear-programme formulation with **0 disagreements over all 2016 floor-cost pairs**, plus a self-pair rejector;
- the floor-cost maximum is recomputed with the colour bound **removed entirely**;
- the below-average maximum is recomputed under a **reversed vertex order**;
- both witnesses have their pairwise disjointness re-derived straight from vertex coordinates rather than from the cached graph.

An earlier reading of this graph took candidates in increasing rather than decreasing colour order, which silently under-reports; it produced a below-average maximum of 14 and an inflated floor of 98. The note names that error and points an auditor at the below-average maximum of 16 as the one number with no colour-free confirmation.

## Landing

Note, runner, cold output, receipt, and the citation-graph manifest acknowledgement. The cycle 721, 722 and 723 results are cited as context in backticks without dependency edges, since none is resident on main; the results here are independent of all three.


JONbridgerapps:
LANDED to main by review-loop cherry-pick: d8be77c034b668dc39166e851d750ba7087b037e (CLOSED means landed). Review provenance: Sol review iteration 1 -> FIX_THEN_PROCEED (8 findings); fix iteration 1 (worker; lower-bound-only restatement, supplied-model section with open bridges, disjoint_lp fail-closed, exact integer arithmetic, narrow affine-rank demotion with Review record); confirmation: PASS. Landing note: the branch's manifest edit was superseded by a fresh regeneration on the landed tree.

PR #5942, opened 2026-08-03
Exact adjacency-cost bracket for tick-box dissections (cycle 725)

jonathonreilly:
## What this responds to

Cycle 723 measured the seam at 108 slot-uses per cell for the monotone-path stencil.
The open worry was whether 108 is a property of that stencil or of the box itself.
Cycle 724 bounded it from below at 96 and left a gap.

## What changed

**The gap is now zero, and the 96 is explained.**

- **Minimum 108, exact**, over all dissections of one tick-box into minimal-volume
  corner pieces. Two independent integer dual certificates on two independent
  sample-point families reach it (fixed-weight: support 211, tight at 2093 points;
  invariant: eight orbits at denominator 2, tight at 2224 points), and dissections
  attain it.
- **Not a stencil property.** Three structurally unlike dissections sit at 108, with
  cost profiles `{4: 12, 5: 12}`, `{3: 1, 4: 11, 5: 11, 6: 1}` and
  `{3: 1, 4: 13, 5: 7, 6: 3}`.
- **Maximum 128**, closed by an invariant certificate at denominator 3 (tight at 1136
  points) and attained.
- **Cost parity is forced** over the minimal pieces (rank 465 on both sample families,
  seven single-piece perturbations refuted), so the attainable set is exactly the eleven
  even values 108 through 128 — every one realised by a checked dissection.
- **Scope boundary: 68.** Admitting coarser corner pieces drops the floor to 68, closed
  by a certificate at denominator 6 (tight at 296 points) and attained by a genuine
  sixteen-piece dissection with volume profile `{1: 8, 2: 8}`. Minimality of the pieces
  is exactly what pins 108. The parity relation does not extend past the minimal pieces.
- **A denominator law** governs certificate sharpness: every sample point-orbit has size
  24, so an invariant certificate at denominator D yields a multiple of 24/D, predictable
  with no optimiser. Sixteen rungs predicted, sixteen attained: floors 96/108/104/108 and
  48/60/64/68, ceilings 144/132/128/128 on both brackets.
- **Cycle 724's 96 is named:** it is the largest multiple of 24 at or below 108, hence
  exactly the ceiling of integral covariant certificates. Integrality alone is not the
  obstruction — the fixed-weight family carries an integral certificate reaching 108 —
  it is covariance together with integrality that caps at 96.
- **A single point-orbit provably caps at 84**; eight orbits suffice.
- Every carried bound is locally maximal: all 114 single-orbit unit strengthenings and
  the uniform strengthening break feasibility, on five representative certificates.
- A cost-72 family whose volumes sum to 24 is carried as a **negative control** and
  correctly rejected (86 overlapping interior pairs, cover-once false on both families).

## Runner

`scripts/physical_exact_adjacency_dissection_bracket_cycle725_2026_08_03.py` —
**TOTAL: PASS=26 FAIL=0**, about two seconds, exact integers throughout, no solver in
the loop. Bounds are carried as integer multiplier vectors and checked directly;
attaining families are carried as piece lists and verified genuine by a complete
separating-direction decision procedure (all vertices are zero-one corners, so a
supporting direction may be taken orthogonal to three ternary vectors, bounding its
entries by four and making the sweep a decision rather than a sample); the parity
statement is derived in-runner by elimination over the two-element field.

## Cross-checks run before opening this

Each headline number was re-derived by different machinery: a third sample family on
unrelated weights reproduces every cover-once verdict at zero boundary incidences;
pairwise interior intersection re-decided by linear-programming feasibility gives
verdicts identical to the direction sweep; all four bracket endpoints recomputed by
simplex on the primal relaxation return 108, 128, 68 and 128. Each load-bearing gate was
shown to discriminate — corrupting a multiplier drives the least slack negative, swapping
one piece breaks cover-once, a degenerate sample family fails the boundary check, and the
parity test rejects parity-changing perturbations while accepting parity-preserving ones.

## Boundary

108 is exact **for minimal-volume corner pieces** and is not a bound on dissections in
general; downstream use must carry the piece class. The statement covers one lattice
cell through one tick — longer tick runs and larger spatial blocks are open work. The
acting group is the axiom's proper spatial rotations with the tick fixed. Nothing here
derives a metric or a field equation; it fixes an exact combinatorial cost the geometry
lane's constructions must pay.


JONbridgerapps:
LANDED to main by review-loop cherry-pick: ca8b983a3f6dd178e746e409762a2fc091672eb2 (CLOSED means landed). Review provenance: Sol review iteration 1 -> FIX_THEN_PROCEED; fix iterations 1-3 (worker; universal negatives narrowed with a full No-Go Discipline Gate — every N1 route genuinely ATTEMPTED by bounded in-runner computation, 26->30 gates incl. all nineteen certificates strengthened; 15-row pairwise wall table; the 84-vs-108 single-orbit/joint-halves distinction stated on every surface); confirmations: FAIL, FAIL, PASS. Citation-graph manifest refreshed.

PR #5944, opened 2026-08-04
[physics-loop][review-loop] toe-time blockG10 (STACKED on #5943) — Cycle 885: bounded theorem — GBW1 split and re-priced; the disjoint window is impossible

JONbridgerapps:
STACKED on the blockG9 branch (PR #5943); only the blockG10 commits are new here. Cluster-cap evaluation (10th gravity-family PR): **OPEN** — answers exactly the door G9 named and sized (GBW1, covering 5/10); new load-bearing content absent from all nine prior PRs (permanence refuting the boundary-shell window; the disjoint-window impossibility theorem; the GBW1a/GBW1b re-scoping); distinct claim type vs G9's anatomy; recorded in the campaign pack.

## Process disclosure

Authored across two Claude Opus 5 workers under supervisor spec (codex quota exhausted; substitution disclosed in the note). The first worker was stopped mid-block by the owner; its near-complete draft was secured as a WIP commit, then a second worker reviewed the draft as UNTRUSTED, found nine defects (three load-bearing: capped exhibit lists reported as failure counts; uniqueness counting a rival its own REQ4 certificate refuted; a hard-coded centre classification), repaired them, and reran. One supervisor spec error is on record: the relaunch brief wrongly asserted a pin path was broken; the worker verified digests and refused the repoint (two distinct Cycle-884 receipts share a basename in different directories — the pin was correct). Checker independence is cross-context; the checker additionally audited the one science-moving repair (the admissibility filter) against the draft's pinned blob and cleared it of outcome-motivation (REQ4 predates the filter, byte-identical).

## Honest status

Bounded worked result. GBW1 is NOT closed: it is split, re-priced, and mostly discharged, with the checker WIDENING one residual the primary understated. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **barrier — DERIVED**: B(R) = supp(R), equivariant 1,440/1,440, permanence-monotone; the landed fixed central barrier refuted on 1,152/1,440 checks (checker: reproduced twice, all translational, rotation-only 0/288).
- **D — GAUGE**: stationary above the last formation depth 12/12; checker strengthened to EVERY depth, not sampled ones.
[...]

PR #5945, opened 2026-08-04
[physics-loop][review-loop] toe-time blockG12 (STACKED on #5944) — Cycle 887: bounded theorem — the window freedom sized; containment derived; the extent is gauge

JONbridgerapps:
STACKED on the blockG10 branch (PR #5944); only the blockG12 commits are new here. Cluster-cap evaluation (11th gravity-family PR): **OPEN** — re-prices the residual its parent block named, with structure theorems, a positive axiom derivation, a gauge theorem, and an explicit correction of the parent's inference; distinct claim type vs G10's coordinate classification; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; codex quota exhausted; substitution disclosed in the note). Checker independence is cross-context: independent group/orbit/harness implementations, a zero-overlap fresh 8-configuration family, quote-needle tampering among the teeth. Supervisor line-by-line review on record in the campaign pack.

## Honest status

Bounded worked result. This block RE-PRICES Cycle 885's window-extent residual — "a dilation scale" was an undercount — and corrects one of 885's inferences in the same stroke. It refutes a closure claim and claims none of its own. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **Q1 — the freedom sized (closure claim REFUTED)**: Minkowski sufficiency theorem (every finite nonempty rotation-invariant S gives an admissible W_S = supp ⊕ S) + injectivity lemma (S ↦ W_S injective) give exact counts **15 / 1,023 / 2,097,151** admissible Minkowski windows at box radius 1/2/3 — 885's dilations are 3 of the 1,023. Necessity FAILS: erosions, bounding-box closures, size- and content-keyed inflations, unions (primary), plus rank/threshold filters (an infinite family; dilations are its t=1 members), morphological closing, bitcount/diameter keying (checker). The annular (a,b) chart collapses 1,023 set behaviours to 113 (family-dependent ratio; strict coarseness is not).
- **Q2 — selection**: every selector byte-quoted and fidelity-checked. Count-once: NOT GROUNDED (multiplicity bound on record-carrying sites; silent on zero-record sites). Minimality: ABSENT from all three pinned texts (grep reported honestly; would be a NEW supplied convention). Gate-B texts: grounded AND negative (they declare the window supplied). **The one derivation: support containment.** The additivity clause forces a dichotomy (contain the records or miss them entirely); the disjoint branch is refuted computationally (16/23 additivity violations for the origin-free inflation). Lower bound derived; upper bound provably unaddressed.
- **Correction of 885**: permanence forbids RETRACTION, not DISPLACEMENT — the boundary shell failed REQ4 as a difference, not because the support is selected; containment fails as a REQ-consequence by 511 explicit witnesses (checker-generalized). Containment holds, but from additivity.
- **Readout-gauge theorem**: 9/12 admissible maps give an IDENTICAL axiom-level readout on every configuration — the window extent is a gauge direction of the only observable the axioms supply; the annular chart moves with the gauge, so (a,b) carries information the axioms cannot certify.
- **Residual, priced**: one no-interaction convention (what would re-collapse the space to a single S); one structuring set S ∋ 0 (bounded below by the derived containment, above by nothing); the centre convention (unchanged from 885). The extent question is re-addressed from the derivation lane to the owner surface.

## Verification

Primary: 12/12 certificates, deterministic double-build, exit 0, 47.6s; all seven of 885's headline numbers reproduced before anything new ran (family digests match — results compose). Checker: exit 0 independent of claim survival, 14/14 claims survive, fresh-family transfer holds, 6/6 teeth. Note: docs/WINDOW_FREEDOM_SIZED_CYCLE887_BOUNDED_THEOREM_NOTE_2026-07-28.md; ship receipt with per-file sha256+git-blob pins in outputs/.

### DIFF
--- a/docs/WINDOW_FREEDOM_SIZED_CYCLE887_BOUNDED_THEOREM_NOTE_2026-07-28.md
+++ b/docs/WINDOW_FREEDOM_SIZED_CYCLE887_BOUNDED_THEOREM_NOTE_2026-07-28.md
@@ -1,0 +1,234 @@
+# The window freedom, sized: a thousand windows, one derived containment, and a gauge nobody ordered — Cycle 887
+
+Date: 2026-08-04
+
+Authority: none
+
+Audit: unset
+
+Status: bounded worked result (owner-directed gravity axiom-up ladder;
+no axiom surface touched). This block RE-PRICES Cycle 885's residual —
+the "dilation scale" was an undercount, not a wrong sign — and the
+chain of blocks correcting each other's readings is functioning as
+designed.
+
+Claim type: bounded_theorem
+
+Runners:
+
+- [`frontier_cycle887_window_freedom_2026_07_28.py`](../scripts/frontier_cycle887_window_freedom_2026_07_28.py)
+- [`frontier_cycle887_window_freedom_independent_check_2026_07_28.py`](../scripts/frontier_cycle887_window_freedom_independent_check_2026_07_28.py)
+
+Receipt:
+
+- [`window_freedom_cycle887_receipt_2026_07_28.json`](../outputs/window_freedom_cycle887_receipt_2026_07_28.json)
+- [`window_freedom_independent_check_cycle887_receipt_2026_07_28.json`](../outputs/window_freedom_independent_check_cycle887_receipt_2026_07_28.json)
+
+Constitutional effect: none. This package changes no axiom, foundation,
+Qualification, primitive, registry, policy, queue, audit result, or audit
+status.
+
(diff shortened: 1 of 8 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-5945 (archive/ledger/45/pr-5945.json). Carrier: PR #5952 (quadratic gauge break supersedes the linear gauge claim) and PR #6009 (ledger window row, exact-0 ablation strengthening). Branch retained; science line in the ledger entry.

PR #5949, opened 2026-08-04
gravity c726: the four-box facet charge splits into a tick half and a mixed half, superadditive by one at the floor

jonathonreilly:
## What this settles

Cycle 723 measured the adjacency cost of the monotone-path stencil at 108 slot-uses per
cell; cycle 725 showed 108 and 128 bracket that cost exactly over all minimal-volume
dissections of the tick-box. Both counted vertex pairs *inside* a piece. This cycle counts
what the pieces show on the box's own eight three-dimensional facets, and finds that the
count carries an internal structure the interior cost does not.

The **facet-visible charge** of a dissection is the summed cost of the tetrahedra its
pieces induce on the eight facets. It splits in two:

- the **tick half**, on the two equal-tick facets, which keep all three spatial directions;
- the **mixed half**, on the six facets that trade a spatial direction for the tick.

Every four-vertex facet slice of a minimal piece has volume 1, so a minimal dissection of
the box induces a *minimal* dissection of each facet, and the per-cell charge sums agree
with the per-facet sums. Enumerating the 180 minimal dissections of one facet gives
tick charge in `[18, 21]` and mixed charge in `[8, 10]`, hence a facet-wise bracket
`2 x [18, 21] + 6 x [8, 10] = [84, 102]` for the total.

**The ceiling is exactly additive; the floor is not.** A dissection attains 102 with both
halves maximal at once. But the floor 84 is unreachable: one witness attains the tick
minimum 36 alone (paying mixed 55), another the mixed minimum 48 alone (paying tick 41),
and the joint floor is **85** — superadditive by exactly one. The 85 is carried by an
integer certificate at denominator 2, support 411, least slack 0, and met by an exhibited
dissection at tick 37, mixed 48.

**Structure costs charge, and the two charges are independent.**

| configuration class | minimum facet-visible charge |
|---|---|
| all minimal dissections | `85`, exact — certificate plus attaining witness |
| square-consistent configurations | at least `86` — complete sweep of the 24 square diagonals |
| face-to-face triangulations | in `[86, 88]`, with `88` attained |

Because `85 < 86`, every configuration at the facet floor fails square consistency, so a
fortiori is not face-to-face: the cheapest facet behaviour is forced off the face-to-face
class. Meanwhile face-to-face does **not** move the adjacency bracket — face-to-face
witnesses sit at box charges 108 and 128, both already reached without the restriction.
The two charges also separate the other way: two dissections at box charge 108 carry facet
charges 85 and 96, and three at box charge 110 carry 88, 91 and 89.

**The stencil orbit is doubly extremal.** The 24 monotone-path simplices sit at the tick
floor `36 = 2 x 18` and the mixed ceiling `60 = 6 x 10` simultaneously, at box charge 108.
The pairing is not forced — a different witness attains tick 36 with mixed 55.

**A charge law that holds on one half and fails on the other.** On an equal-tick facet the
charge is a function of the facet's square-diagonal pattern alone, with a single parity
class; on a mixed facet it is not — 36 patterns carry two different charges and both parity
classes appear. The mechanism is geometric: an equal-tick facet keeps all three spatial
directions, so every pair its pattern distinguishes is a pair the cost weighs, while a
tick-carrying facet trades a spatial axis away and its third kept direction enters the
pattern but not the cost. 58 of the 64 diagonal patterns are realizable; the six absent
ones all carry three set bits.

## Method

No solver in the runner. Bounds are carried as an integer multiplier vector checked
directly against the sample points; attaining families are carried as piece lists and
checked to be genuine dissections. Interior disjointness is **decided** by a
separating-direction sweep over the full integer range that a four-dimensional supporting
direction can require, not sampled. Both sample families are gated to have zero boundary
incidences before any bound resting on them is read. Every integer in the note is the
runner's own count from the run that produced its `TOTAL` line.

`TOTAL: PASS=28 FAIL=0`, exit code 0, byte-identical across two consecutive runs.

## Independent cross-checks performed before writing the note

Each headline number was re-derived by machinery the runner does not use, in a scratch
script not carried in this change; 22 cross-checks, 0 misses.

- Both facet brackets were recomputed from global coordinates rather than per-cell sums,
  and every witness's facet charge was recomputed by direct summation over the box's own
  facets rather than from the per-cell charge table.
- The floor 85, the ceiling 102, and the square-consistent floor 86 were all recomputed by
  mixed-integer programming, a route sharing no code with the carried certificate and the
  complete sweep. All three agreed.
- The same programme pins the **face-to-face floor at exactly 88**, matching the exhibited
  witness, and gives the free facet-charge range at box charge 108 as `[85, 96]`. Both are
  reported here as cross-checks only; the note states the face-to-face floor as the gated
  bracket `[86, 88]`, since pinning it to a value needs a solver and the runner has none.
- Containment was re-decided barycentrically rather than by half-spaces, with least
  absolute barycentric coordinate 1.0 on every sample.
- The certificate was shown to discriminate: raising any one weight drives the least slack
  negative, lowering one weakens the bound below 85, and it binds at zero slack on 1174 of
  the 2672 minimal pieces. The separating-direction test rejects a cell against itself, and
  the cover-once test rejects a single-piece substitution.
- The pattern law was re-tested on both halves and confirmed tick-specific, so the note
  claims it for the equal-tick facets alone.

## Honest boundary

- The middle rung is a **lower bound**. The square-diagonal sweep relaxes each facet
  independently, so 86 is not shown to be attained by any configuration.
- The face-to-face floor is stated as a bracket, not a value, because the solver-free
  argument reaches 86 and the exhibited witness reaches 88.
- The certificate is carried **at** denominator 2. No claim is made that 2 is the smallest
  denominator that carries 85.
- Which half carries the superadditive unit at the joint floor is a property of the
  exhibited witness, not a gated law.
- Nothing here derives a metric, a curvature, or a field equation. It fixes a second exact
  combinatorial charge, on the boundary rather than the interior, that the geometry lane's
  constructions must pay.

## Files

- `docs/PHYSICAL_FACET_CHARGE_TICK_MIXED_SPLIT_CYCLE726_NOTE_2026-08-04.md`
- `scripts/physical_facet_charge_tick_mixed_split_cycle726_2026_08_04.py`
- `outputs/physical_facet_charge_tick_mixed_split_cycle726_2026_08_04_cold_2026-08-04.txt`
- `outputs/physical_facet_charge_tick_mixed_split_cycle726_2026_08_04_receipt_2026-08-04.json`
- `docs/audit/data/citation_graph_manifest.json` (one node, out-degree 1)

Status is unset and no audit language is authored anywhere in this change. The related
in-flight cycle notes of this lane are referenced as context only, in backticks, so this
note carries a single dependency edge, to the minimal axioms note.


jonathonreilly:
Review-loop PASS and landed by exact-head fail-closed replay.

Provenance: submitted head `e4e1aad7bb0af416f9dd68ebe18e27b136d70189`; repaired PR head and contained main `e99815b6b4c3c12c01e2efe492d122559bf43660`.

Scientific repair: this is an exact finite theorem of a supplied one-box, equal-grained corner-simplex/24-minimal-piece/disclosed-pair-charge model. It is not a framework derivation of that model or a physical slot-cost law. Within that domain the facet-charge floor is 85 (denominator-two certificate plus attaining witness), ceiling 102, square-consistent lower bound 86, and face-to-face bracket [86,88]. Tick-facet spectra are 18..21 and mixed-facet spectra 8..10; the six-bit diagonal pattern fixes tick charge but has exactly 36 mixed-pattern collisions. The monotone-path family is a finite witness. Cycle 725 is a direct, receipt-pinned, unaudited dependency only for the exact 108–128 box-charge endpoints; Cycles 723/724 are provenance only and Cycle 873 ordering only.

Open boundary: no framework model selection, physical cell/slot-charge identification, tick–Admissibility realization, coarser/non-corner/nonsimplicial pieces, multi-box/tick, other boundary, continuum, metric, action, curvature, or field-equation conclusion.

Repairs also made the runner fail closed, strengthened the mixed-pattern gate to exactly 36, bound all witness tuples, added hostile certificate/substitution controls, generated the live receipt, added primary and independent input-bound caches, landed the N1–N8/N5 record, and exposed the independent checker in the citation-graph packet.

Final gates: primary 32 PASS / 0 FAIL; independent exact checker 8/0; hostile dependency/certificate/pattern mutations rejected; full pipeline 18/18; strict lint 0 errors; changed evidence 1 checked / 0 failures with forensic readiness; invariants and links clean; vocabulary clean. No audit verdict was applied.

PR #5953, opened 2026-08-04
feat(geometry): bound declared two-slab corner charge [216, 256] (Cycle 727)

jonathonreilly:
## Summary

Cycle 725 measured the exact bracket `[108, 128]` on the adjacency cost of a minimal-volume corner dissection of one lattice cell carried through **one** tick, and its own boundary named the next question verbatim: *"The bracket is a statement about one lattice cell carried through one tick. Extending it to longer tick runs or larger spatial blocks is open work, not a corollary."*

This PR answers the tick half. **Over every minimal-volume corner dissection of one lattice cell carried through two ticks, the adjacency cost is exactly `[216, 256]` — twice the one-tick bracket at both ends.**

Extensivity was not a safe guess. Stacking two one-tick dissections always gives a valid two-tick dissection, so the cost is subadditive and the floor could only have come in at or below 216 — nothing forced it not to come in strictly below. Nothing at all bounded the ceiling from above by 256: a two-tick dissection may use pieces straddling the seam between ticks, and 11936 of the 17280 minimal pieces do exactly that. Both ends land on the doubled value anyway.

## Results

- **Floor.** Integer certificate at denominator 2 with `sum u + Z = 9`: `48 * 9 / 2 = 216.000000` on the nose. Least slack 0, equality on 13392 of 17280 minimal pieces.
- **Ceiling.** Integer certificate at denominator 288 with `sum u + Z = 1536`: `48 * 1536 / 288 = 256.000000` on the nose. Least slack 0, equality on 6336 of 17280.
- **Both ends attained.** 48-piece witnesses at cost 216 and 256, each with all 1128 piece pairs carrying an exhibited separating integer normal.
- **The one-tick values measured in the same run.** 24-piece one-tick dissections at cost 108 and 128, 276 of 276 pairs each — so `216 = 2 * 108` and `256 = 2 * 128` are both sides measured, not one side supplied.
- **Not a counting artefact.** Per-piece adjacency cost ranges over `(3, 7)`, so counting alone gives only 144 below and 336 above. The true bracket sits strictly inside.
- **A denominator law.** Every one of the 364 point orbits carries exactly 48 points, so a certificate at denominator `D` has value `48 (sum u + Z) / D`: reaching 216 needs an even denominator and 256 a multiple of three. Certificates are homogeneous, so this is a statement about the value, not about the scale chosen.
- **A charge that only exists from two ticks on.** Tick-span charge — vertex pairs whose tick separation exceeds 1 — is identically zero on the one-tick box, verified on all 2672 minimal one-tick pieces. Its zero set is exactly the 5344 = 2 * 2672 slab-confined pieces; the other 11936 cross the seam and still reach the same least adjacency charge 3.

## Method

Sample points are **generic by construction, not by draw**: the barycentric numerator of a point built with weights `W` is the `W`-combination of its corners' numerators, so with those integers bounded by `C = 6` and `W` superincreasing past `C`, no combination can vanish. Verified as well as constructed — 0 boundary incidences over all 17280 pieces. Witnesses are certified without a solver: volume 1 per piece, volumes summing to the box, and every pair shown interior-disjoint by an exhibited separating integer normal. The runner contains no linear or integer programme.

- Census, all three spectra, and all four witness charges re-derived with an independently written corner ordering and explicit pair loops.
- Cover-once corroborated by random-point classification with a hard assertion; every point off every wall covered exactly once.
- **The bracket reproduced on a second generic weight family** in a different superincreasing base — a genuinely different linear system, giving `[216.000000, 256.000000]` again.
- Both certificates shown maximally tight: nudging any one of the 364 multipliers in the helpful direction breaks the inequality on some piece, 364 of 364, for floor and ceiling alike; value-preserving compensated shifts all drove least slack negative.
- Gates shown wired to the data: corrupting each load-bearing constant one at a time turns the run red every time.
- Every number in the note checked mechanically against the runner's own stdout.

## Test plan

- `python3 scripts/physical_tick_extensive_adjacency_bracket_cycle727_2026_08_04.py` reports `TOTAL: PASS=49 FAIL=0`, exit 0, about 19 seconds, 4359 characters of output.
- Re-run from a clean worktree checkout reproduces the cold output byte for byte.

## Status

Unaudited source note. The independent review and audit lanes own grading and landing.


jonathonreilly:
Review-loop PASS on repaired exact head `5078cc77a8b7a7ba091373bbce529a8f887e4f86` against exact current main `e99815b6b4c3c12c01e2efe492d122559bf43660`.

Scientific boundary: exact finite theorem only for the supplied equal-grained two-slab unit box, normalized-volume-one corner 4-simplices, and a declared spatial-pair charge assigning zero weight to slab-coordinate separation. Exact bracket `[216,256]`, both endpoints attained; one-slab controls `[108,128]`. No framework piece/charge selection, physical tick realization, coarser/noncorner/nonsimplicial, larger-block, longer-run, boundary, or continuum claim.

Evidence: primary `PASS=50 FAIL=0`; independent exact reconstruction `PASS=15 FAIL=0`; four source-level hostile mutations fail cache freshness, primary direct execution, and source-pinned checker. Full exact-main pipeline PASS 18/18; strict lint zero errors; changed evidence checked 1/failures 0/forensic ready; invariants rows 4784, link violations 0, class-F violations 0, graph acknowledged; vocabulary 0; diff clean. Generated audit/effective-status outputs stripped; only the required citation-graph manifest acknowledgment lands. No audit verdict was applied.

PR #5962, opened 2026-08-04
geometry: two cells side by side, and what the seam costs when the long axis is spatial (cycle 728)

jonathonreilly:
## What this cycle does

Cycle 727 charged a box that was long in the tick. This cycle takes the affinely identical
box — two lattice cells side by side, carried through one tick — and moves the long axis
onto a **spatial** direction, so the adjacency charge now sees it. Same 24 corners, same
17280 minimal pieces, same 48 pieces per dissection. Only the direction the charge reads
changes. That makes it a controlled experiment on what the charge is actually sensitive to.

## Results

**A seam-respecting dissection costs between 216 and 256, and both ends are attained.** A
piece with zero span along the long axis sits inside one closed cell; there are 5344 such
pieces, exactly two translated copies of the 2672 one-cell pieces with the charge unchanged.
So a seam-respecting dissection *is* a pair of one-cell dissections. The one cell brackets
exactly at [108, 128]: a floor certificate at denominator 2 and a ceiling certificate at
denominator 3, each checked piece by piece over all 2672 pieces with least slack zero, and
each attained by an exhibited dissection. Stacking gives block witnesses at exactly 216 and
exactly 256, both with zero seam crossings.

**The dearest dissection cannot respect the seam.** An explicit 48-piece dissection costing
318 is exhibited and verified — every piece of unit volume, all 1128 pairs separated by an
integer normal produced on the spot — and 318 sits above 256. Thirty-one of its 48 pieces
cross the seam. The minimum 216, by contrast, is attained without crossing at all.

**The maximum is bracketed at [318, 324].** A ceiling certificate over the block's own 1080
piece orbits, carried at denominator 12, holds on every one of the 17280 minimal pieces with
least slack zero and equality on 1200 of them; its value 3888 over 12 gives 324. The bound
available from the charge alone is 432, so the certificate removes most of the trivial slack.

**No dissection of this box carries the box's symmetry.** The group has order 16, all 1080
piece orbits have size 16, so a symmetric dissection is exactly 3 orbits. Only 23 orbits can
appear in any exact cover, and a direct sweep over triples of those 23 finds none.

**The two charges agree on the cheap dissection and part company on the dear one.** The same
48 pieces read 318 spatially and 238 transposed. On the stacked monotone stencil both charges
read 216, because the stencil is built from all 24 orderings of the four coordinates and the
exchange of the long spatial axis with the tick permutes its pieces among themselves. That
agreement is a property of that dissection, not of the charges: the dear one-cell dissection
is moved by the same exchange and reads 128 against 116.

**A denominator law.** A block certificate value is `16 T + 48 Z`, hence a multiple of 16, so
a bound of `c` needs 16 to divide `c D`: 216 forces an even denominator, 324 a multiple of
four. The certificate carried here has denominator 12.

## Method

No solver anywhere in the artifact. Certificates are integer data verified by direct integer
arithmetic against every piece; witnesses are certified by unit volume plus a separating
integer normal exhibited for every pair. Sample points are built superincreasing from a
measured barycentric bound, and the artifact checks the consequence — zero boundary
incidences over all 17280 pieces — so every point lies in the interior of exactly one piece
of any dissection and the bounds hold with no symmetry assumption. Symmetry only compresses
the certificate.

## Honest read

- There is **no block-level floor certificate**. The claim carried is that 216 is the least
  cost found and that no seam-respecting dissection costs below it — not that 216 is optimal
  over all dissections of the block.
- The ceiling window [318, 324] is open by six units. The gap is not a search artifact a
  finer denominator removes: rescaling weights, constant and denominator together leaves the
  bound exactly at 324, re-checked at three rescalings. Sharpening needs a different
  certificate shape.
- The 318 witness is a regular triangulation, found by lifting, so 318 is a certified
  attained cost, not a certified maximum.

Every headline number was re-derived by a route the artifact does not use, and every
load-bearing gate was perturbed to confirm it fails when the object it tests is wrong: the
eligible-orbit count was recomputed point by point against all 17280 sample points with no
representative argument, the certificates each break under a one-unit change to a tight entry
or to the constant, and the block certificate's orbit-level slack was independently confirmed
to lift piece for piece to the full check. The note's prose was then audited sentence by
sentence against the gated numbers; three numbers that came from ungated search rather than
from the artifact were removed rather than kept.

## Artifacts

- `docs/PHYSICAL_SPATIAL_BLOCK_SEAM_DICHOTOMY_CYCLE728_NOTE_2026-08-04.md`
- `scripts/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04.py` — `TOTAL: PASS=65 FAIL=0`
- `outputs/..._cold_2026-08-04.txt` — the runner's stdout verbatim, re-run in the landing
  worktree and byte-identical
- `outputs/..._receipt_2026-08-04.json` — transcribed from that stdout, every field pinned

Status is `unaudited source note`; the independent lanes own review and grading.

### DIFF
--- a/docs/PHYSICAL_SPATIAL_BLOCK_SEAM_DICHOTOMY_CYCLE728_NOTE_2026-08-04.md
+++ b/docs/PHYSICAL_SPATIAL_BLOCK_SEAM_DICHOTOMY_CYCLE728_NOTE_2026-08-04.md
@@ -15,7 +15,7 @@ Canonical primary cache:
 Primary receipt:
 [`outputs/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04_receipt_2026-08-04.json`](../outputs/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04_receipt_2026-08-04.json).
 
-Independent reconstruction:
+Independent checker:
 [`scripts/physical_spatial_block_seam_dichotomy_cycle728_independent_check_2026_08_04.py`](../scripts/physical_spatial_block_seam_dichotomy_cycle728_independent_check_2026_08_04.py)
 (15 PASS / 0 FAIL, fail-closed), with canonical cache
 [`logs/runner-cache/physical_spatial_block_seam_dichotomy_cycle728_independent_check_2026_08_04.txt`](../logs/runner-cache/physical_spatial_block_seam_dichotomy_cycle728_independent_check_2026_08_04.txt)
--- a/docs/PHYSICAL_SPATIAL_BLOCK_SEAM_DICHOTOMY_CYCLE728_NOTE_2026-08-04.md
+++ b/docs/PHYSICAL_SPATIAL_BLOCK_SEAM_DICHOTOMY_CYCLE728_NOTE_2026-08-04.md
@@ -1,6 +1,50 @@
 # Two cells side by side: what a seam costs when the long axis is spatial
 
-Status: unaudited source note. Cycle 728 of the emergent-geometry lane.
+Date: 2026-08-04
+
+Claim type: bounded_theorem
+
+Authority: none. Audit: unset. Constitutional effect: none. No new axiom or
+primitive is proposed or adopted. Cycle 728 of the emergent-geometry lane.
+
+Primary runner:
+[`scripts/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04.py`](../scripts/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04.py)
+(65 PASS / 0 FAIL, fail-closed).
+Canonical primary cache:
+[`logs/runner-cache/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04.txt`](../logs/runner-cache/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04.txt).
+Primary receipt:
+[`outputs/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04_receipt_2026-08-04.json`](../outputs/physical_spatial_block_seam_dichotomy_cycle728_2026_08_04_receipt_2026-08-04.json).
+
+Independent reconstruction:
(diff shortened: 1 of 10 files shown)


jonathonreilly:
Review-loop PASS and contained on exact main `9c428c93189a695068097a78606f2a3526ad5a14` (reviewed repaired PR head identical to landed main).

Scientific result retained after repair: in the supplied finite 2x1x1-spatial by one-tick normalized-volume-one corner-simplex model, the seam-respecting sector has exact attained bracket `[216,256]`; a checked 48-piece crossing dissection attains spatial charge `318`, transposed charge `238`, with 31 crossing pieces; and an exact ceiling certificate gives the global maximum window `[318,324]`. Because `318>256`, every global maximizer crosses the seam. The carried order-16 proper-spatial-stabilizer-times-tick-reversal action has 1,080 size-16 piece orbits, 23 eligible orbits, and zero exact-cover triples, excluding invariance under that entire action.

Review repairs: rebased oldest-first onto exact predecessor main `5078cc77a8b7a7ba091373bbce529a8f887e4f86`; declared the supplied-model, minimal-axiom, kinetic-isotropy, and Cycle725 premise boundary; removed the contradictory global-minimum wording, full-box-group mislabel, uncarried regular/lift-search claim, and failed-certificate impossibility; made the runner fail closed and receipt-generating; replaced raw cold output with canonical input-bound caches; added an independent 15-gate exact reconstruction with a second sample chamber and hostile mutations; landed the full N1-N8 No-Go Discipline record and N5 resolution certificate; registered the independent checker as an audit helper; regenerated the three-edge citation topology.

Final gates PASS: primary 65/0, independent 15/0, hostile primary mutation exits nonzero, canonical caches fresh, full exact-main pipeline 18/18, strict lint 0 errors, changed evidence checked 1/failures 0/forensic ready, invariants link violations 0/class-F violations 0/graph acknowledged, portable links, vocabulary lint, compile/JSON, and clean diff. Generated audit/status outputs were stripped. No audit verdict was applied.

Boundary: the global floor remains open; the six-unit maximum gap remains open; partial-subgroup symmetry is unclassified. No physical tick realization, physical simplex/cell selection, regular-triangulation/lift-search result, certificate-family no-go, arbitrary-box/repeated-run/boundary/thermodynamic/continuum/metric/curvature/field-equation claim.

PR #5968, opened 2026-08-04
feat(gravity): pin the two-cell block cost interval at [216, 320] and show the maximiser is not a lower hull

jonathonreilly:
## What this settles

Two lattice cells side by side, carried through one tick, cut into minimal pieces, each
piece charged for the pairs of its corners more than one step apart in space.

**The cost of a dissection of this box lies between 216 and 320, and both ends are
attained.** The previous cycle bracketed it at [216, 324] with a best witness of 318;
this pins the interval exactly.

- **Floor 216** — integer certificate at denominator 512, least slack 0 on all 1080 piece
  orbits and again on all 17280 minimal pieces, tight on 30; attained by the stacked
  monotone stencil.
- **Ceiling 320** — integer certificate at denominator 49, checked the same two ways,
  tight on 53; attained by an exhibited 48-piece dissection.
- Charging every piece the least or the most its charge allows gives only 144 to 432, so
  both ends sit well inside what counting alone delivers.

**The dearest dissection is not the lower hull of any lift.** Sixteen of its facets are
carried by a single piece and lie away from the box boundary, so it is not face-to-face;
a lower hull always is. The stencil and the earlier 318 witness are lower hulls, each with
an exhibited integer height clearing all 912 lower-face inequalities. That answers a
question the previous cycle left open: its search generated dissections by lifting, so the
maximiser was outside its reach by a structural obstruction rather than by bad luck.

**Two statements from the previous cycle are corrected upward.** Its best witness of cost
318 is superseded by the verified dissection of cost 320; and its boundary read that a
sharper block ceiling would need a different certificate shape does not hold — the shape
is the same and the denominator changes.

## Method

No solver runs inside the artifact. Both certificates arrive as integer data and are
verified by direct integer arithmetic; each witness is certified by volume plus an integer
separating normal produced on the spot for every one of its 1128 piece pairs; regularity
is settled in both directions by supplied integer heights and by a single-count interior
facet. The sample points are forced generic by construction — the largest barycentric
integer is measured first, then the corner weights are chosen superincreasing above it —
and the consequence is checked as zero boundary incidences over all 17280 pieces, so the
bound holds with no symmetry assumption; symmetry only shrinks the program from 17280 rows
to 1080, and the artifact confirms that shrinking is lossless.

Every headline number was re-derived by a route the artifact does not use: the three costs
by a plain double loop over corner pairs, the slacks in unbounded integers, the bounds in
exact rational arithmetic, and both endpoints again from a fractional relaxation solved
independently of every certificate. The membership matrix was rebuilt from exact integer
cofactor adjugates and agrees entrywise. Non-regularity was proved a second time by an
exact integer vector of nonnegative multipliers summing the 912 inequalities to zero, and
regularity a third time by reading lower facets off the convex hull of the lifted corners.

Each gate was then damaged on purpose. Three perturbations did not break their gate; all
three were chased down and traced to the probe, not the objects. Two were the stencil
height moved by one at a single corner — being a lower hull is an open condition, and the
inducing heights form a bounded cone whose measured radius is exactly 16 in either
direction at every one of the 24 corners, so jitter smaller than that is supposed to
survive. The third corrupted a membership entry in a column whose weight is zero, which
changes no product; retargeted to a live column on a row with no slack it breaks the
certificate every time. A first version of the exact-cover cross-check demanded that no
sampled point land on a face, which random rational sampling cannot deliver; restating the
predicate correctly cleared all three witnesses.

## Artifacts

- note `docs/PHYSICAL_BLOCK_COST_INTERVAL_LIFT_OBSTRUCTION_CYCLE729_NOTE_2026-08-04.md`
- runner `scripts/physical_block_cost_interval_lift_obstruction_cycle729_2026_08_04.py`
- cold output and receipt under `outputs/`

Runner prints `TOTAL: PASS=43 FAIL=0` in 5.1 s; re-run inside this worktree reproduces the
landed cold output byte for byte. Status is `unaudited source note`; no audit language is
authored here.


jonathonreilly:
Review-loop PASS and contained on exact main `d12d5922750ba579780eee647e50b5be3622acf8` (submitted head `8ca899138d7a5c1370e02cfe30660073b0db6103`). Repaired the package to an exact finite supplied-model theorem: direct Cycle 728 dependency and equal-graining premise are input-bound; unsupported Cycle 728 lift-search/certificate-shape history and any all-maximizer implication were removed; primary is fail-closed with generated receipt; an independent exact checker, hostile mutations, canonical caches, explicit helper closure, and N1-N8/N5 packet were added; cold output removed. Science independently reconstructed: all 42,504 five-subsets / 17,280 unit pieces, both all-piece integer certificates, costs 216/318/320, all 1,128 pair separations for each of three witnesses, regular margins 16/32, and 16 unpaired internal facets for the carried cost-320 witness. Primary `44/0`; independent `13/0`; full exact-current-main pipeline `18/18`; strict lint `0`; changed evidence checked `1`, failed `0`, forensic-ready; invariants/links/class-F `0`; changed-surface vocab and diff clean; generated audit/status outputs stripped. Boundary: exact only for the supplied equal-grained 2x1x1-spatial x one-tick, normalized-volume-one corner-4-simplex 48-piece model with declared spatial-L1 pair charge. No physical model/tick/charge selection, no all-maximizer regularity classification, and no arbitrary-block/tick/boundary/continuum claim. No audit verdict was applied.

PR #5978, opened 2026-08-04
feat(geometry): cell dissection cost is 108 plus its off-support pieces

jonathonreilly:
## What this is

Cycle 731 of the emergent-geometry lane, on the single cell of the doubled lattice.
Cycle 725 measured the adjacency cost of dissecting that cell into minimal pieces and
found the interval 108 to 128 with both ends attained; cycle 730 showed each end
reduces to a membership test on one piece at a time. This cycle exhibits a floor
certificate whose per-piece slack takes only two values, and that turns the bound into
an identity.

## The result

The certificate is valid on all 2672 minimal pieces at denominator 216, its value
23328 is exactly 108 times 216, and its slack spectrum over every piece is exactly
0 and 216. Since the per-piece slacks over a dissection sum to the denominator times
the cost minus the value, this gives, for every 24-piece dissection whatever its cost,

    cost = 108 + (number of its pieces outside the certificate's support)

Checked on four exhibited dissections at costs 108, 114, 108 and 128, against 0, 6, 0
and 20 pieces outside. The cost is a count.

Consequences measured here:

- The ceiling caps cost at 128, so at most 20 pieces lie outside and every dissection
  of the cell holds at least 4 pieces inside the support. The cost-128 dissection
  attains exactly 4.
- The support is settled at 38 orbits from both sides: every piece of a cost-108
  dissection is tight, and each of the 38 sits in a cost-108 dissection drawn entirely
  from the support. Six forced completions already realize all 38, and a complete
  sweep of all 501942 five-subsets of those orbits shows no five of them do.
- The ceiling admits no indicator, by an exact integer dependency among five orbit
  rows with coefficients 3, 1, -1, -1, -2 at residual 0. Four of the five are tight in
  every zero-gap ceiling certificate, which leaves 3x = 2 with no 0/1 answer. The same
  relation pins the least ceiling denominator at 3 and it is attained. The floor's
  own pattern on those five orbits gives -2 and does have an answer.
- The mechanism is general. The 57 orbit rows span 13 dimensions and the constant
  column does not raise that span, so every dependency has coefficient sum zero and is
  therefore an exact slack identity independent of which certificate was chosen. A
  complete sweep of all 1906884 five-element supports of the 49 row classes finds 185
  such dependencies, all of coefficient sum zero, with charge combinations
  {-2: 22, 0: 136, 2: 26, 4: 1}, of which 49 carry a nonzero identity.
- Reading the support locally: adjacency charge at most 4 forces support and charge 7
  forces outside, deciding 752 of 2672 pieces; the remaining 1920 split both ways, and
  none of 6 local invariants swept separates the 38 support orbits from the other 19.

## Adversarial review performed before opening this PR

- Every headline number was re-derived by a route different from the runner's at each
  step: rational elimination against fraction-free integer elimination, popcount
  against restricted L1 separation, breadth-first closure against the stored orbit
  labelling, permutation-expanded determinants against stored inverses, and an exact
  rational nullspace of the whole 57-row matrix. All agreed.
- Discrimination was tested rather than assumed. 0 of 116 single-step moves in the
  floor certificate stay valid at zero gap; 0 of 31 orbits occurring in the exhibited
  dissections can be moved across the support and keep the identity; 10 of 10
  single-coefficient changes destroy the row dependency; 0 of 36 single substitutions
  in the six-member family still realize all 38 orbits.
- Two findings forced changes to the artifact. A claim that the five-orbit relation is
  distinguished by its support was refuted by the 185-dependency census and replaced
  by the measured facts, and a reading of a truncated listing which suggested every
  nonzero charge combination is plus or minus 2 was refuted by the single combination
  of 4. Both corrections are recorded in the note's cross-check section.
- Floor validity on its own survives half the perturbations tried and is called out in
  the note as weak by itself; the load-bearing claim is the conjunction of validity,
  zero gap and the two-value spectrum. The ceiling's not-an-indicator spectrum is a
  negative property of one certificate, so the note rests that conclusion on the
  positive integer identity instead.

## Artifacts

- Note: `docs/PHYSICAL_COST_IDENTITY_INDICATOR_CERTIFICATE_CYCLE731_NOTE_2026-08-04.md`
- Runner: `scripts/physical_cost_identity_indicator_certificate_cycle731_2026_08_04.py`
- Recorded output and receipt under `outputs/`

The runner carries no dissection literals: the stencil is rebuilt from the
permutations of the four coordinates and the rest by exact cover with no cost
objective anywhere in the search. It reports `TOTAL: PASS=45 FAIL=0` in 5275
characters, re-run from its landed path with output identical to the recorded file.
The note is unaudited; the independent lanes set its status.


jonathonreilly:
Review-loop PASS on repaired exact head `18d5b694c94ebc25e60c7fd884e21c872605f81f`, replayed against exact current main `57739e3e85b24d9736901c618bd2c2e37016b4ce`.

Scientific result retained, with scope repaired: in the supplied finite one-cell normalized-volume-one corner-simplex model, every declared-class dissection obeys `cost = 108 + off-support piece count`. The exact floor indicator has support 1,792 pieces / 38 carried orbits, and both implementations verify genuine cost-108 dissection completions through all 38 orbits. In the fixed generic-incidence integer ceiling-certificate family only, the exact row identity forces least positive integer denominator 3 and excludes a binary ceiling indicator. The six-invariant non-separation statement is limited to the six named scalars on 57 orbits. No physical Admissibility/tick/simplex selection, other certificate-family, piece-class, arbitrary-domain, boundary, or continuum conclusion is claimed.

Review repairs: declared the supplied-model and premise boundary; removed the physical defect-density promotion and false coarser-piece exclusion; made terminal exact-cover search require genuine pairwise-separated dissections; added fail-closed receipt generation; added an independent AST-data-only reconstruction with hostile controls; replaced raw cold output with canonical input-bound caches; registered the helper in the citation graph; and added the full No-Go Discipline N1–N8 packet. No audit verdict was authored or applied.

Evidence: primary `45 PASS / 0 FAIL`; independent `15 PASS / 0 FAIL`; all 38 floor completions and all 4 forced ceiling completions independently verified; full repository pipeline `18/18`; strict lint zero errors; changed evidence `checked=1 failures=0`; invariants `rows=4807 link=0 class-F=0 graph=acknowledged`; vocabulary zero; diff clean. Generated audit/status outputs stripped.

PR #5981, opened 2026-08-04
feat(emergent-geometry): cell dissection cost is always even, spectrum exactly 108 to 128

jonathonreilly:
## What this settles

Earlier cycles of this lane bracketed the adjacency cost of a dissection of the single
cell between 108 and 128, with both ends attained. Twenty-one integers sat inside that
bracket with nothing said about which of them a dissection can realise.

Exactly eleven of them occur: the even ones.

The evenness is forced, not observed. The note exhibits a set of 228 sample points that
meets every one of the 2672 pieces of least volume in a number of points congruent,
modulo 2, to that piece's adjacency charge, with no constant term. Because no sample
point lies on a piece boundary, each one is interior to exactly one piece of any
dissection whatsoever, so summing the row congruence over a dissection turns it into a
statement about cost — and 228 is even.

Three further results come with it.

**The argument never uses the piece count.** The certificate carries no constant term, so
the conclusion follows from the covering property alone, with no appeal to how many
pieces a dissection has or to their volumes.

**The symmetry ladder is sharp at index 4.** A complete sweep of all 98 subgroups of the
cell's 48 symmetries finds that among the 12 of order at least 12, exactly one admits an
invariant certificate, and it has order 12 — a pure rotation subgroup carrying the tick
flip on none of its elements. All 7 subgroups of order above 12 fail. The symmetry the
argument gives up is index 4, exactly and no more.

**Modulo 2 is sharp, by two independent routes.** The eleven exhibited costs differ by a
greatest common divisor of 2. And modulo 3 there is no certificate at all — not as a
failed search, but by an exhibited geometric obstruction: four pieces of least volume
inside the single six-corner set `[0, 1, 3, 7, 8, 10]`, counted with multiplicities
`[1, 2, 2, 1]`, cover 266 sample points exactly three times each and the remaining 2470
not at all. Their charges `[4, 5, 6, 5]` sum to 31, which is 1 modulo 3, and the
multiplicities sum to 6, so the obstruction survives any constant shift. That obstruction
is not a freak: of the 8008 six-corner subsets of the cell, 1104 carry one of exactly
this kind using only the pieces of least volume they contain themselves.

## Method

Certificates and witnesses, checked — not searched for. The artifact calls no optimiser.
The parity certificate is derived inside the runner by exact Gaussian elimination over
the field of two elements, so no list of 228 point indices is transcribed by hand. The
eleven dissections are pinned as explicit 24-tuples and each is verified from scratch:
24 distinct pieces of least volume, pairwise disjoint by an exhibited integer separating
direction, and every one of the 2736 sample points covered. Volume plus disjointness make
each cover exact without asking a solver anything.

## Numbers

Runner reports `TOTAL: PASS=39 FAIL=0` in 13.67 s, 4859 characters of output. Every
number in the note appears in that output, apart from the cross-check counts, which the
note names as such.

## Adversarial review performed before opening this PR

Every headline was re-derived by a method the runner does not use, and every gate was
tested against a perturbed object. Three findings changed the work:

- The barycentric cross-check confirmed the integer incidence, but only on a sample of
  pieces. The note now says so rather than implying full coverage.
- Nudging the floor weights by a single unit breaks a row in only 12 of the 24 possible
  directions, so those weights are not the only integer system certifying the floor. The
  note says "the exhibited floor rows" throughout and never claims a unique certificate.
- A draft sentence claimed a shuffled charge vector is still certifiable modulo 2,
  reading that as evidence parity is a common condition. Re-measurement refuted it: the
  point-level system has rank 465 among its 2672 rows, and ten random shuffles and ten
  uniformly random targets were all infeasible. A shuffle keeps how many pieces carry an
  odd charge and still fails. The note now states the stronger, correct reading —
  certifiability is a fact about how charge sits on the pieces of this cell.

## Boundary

The parity theorem is about this cell, proved by exhibiting a certificate for this cell's
2672 pieces. An earlier cycle's attempt at a parity law across objects was refuted and
nothing here revives it. The floor of 108 and ceiling of 128 are inherited from earlier
cycles and re-verified, not re-derived. The eleven dissections are exhibited, not
classified — the note does not say how many dissections realise each cost.

## Artifacts

- `docs/PHYSICAL_PARITY_CERTIFICATE_COST_SPECTRUM_CYCLE732_NOTE_2026-08-04.md`
- `scripts/physical_parity_certificate_cost_spectrum_cycle732_2026_08_04.py`
- `outputs/physical_parity_certificate_cost_spectrum_cycle732_2026_08_04_cold_2026-08-04.txt`
- `outputs/physical_parity_certificate_cost_spectrum_cycle732_2026_08_04_receipt_2026-08-04.json`
- `docs/audit/data/citation_graph_manifest.json` (separate `audit-infra:` commit)


jonathonreilly:
Review-loop PASS and landed on exact current main at `b29866c6152fc90bd5d2a42348e3d7cf3760c5ce` (submitted head `c1e6e249073a1b9b909228545bbff199b6dfc4f9`). Repaired before landing: demoted the construction to supplied finite data; made Cycle 731 a direct input-bound dependency; narrowed the symmetry and modulo-3 negatives to the fixed 2,736-point incidence ansatz; removed the geometric triple-cover/universal-certificate overclaim; added a separate exact checker, hostile controls, generated receipt, canonical caches, fail-closed exits, and an N1-N8/N5 packet. Scientific result: exact spectrum `{108,110,...,128}` for the supplied one-cell corner-simplex model. Primary `44/0`; independent `12/0`, including a distinct 168-point GF(2) certificate and broader 2,928-normal validation of all eleven witnesses. Full exact-main 18-stage pipeline, strict lint, changed evidence (1/0), invariants/links, vocabulary, and diff gates PASS. No audit verdict was created or applied.

PR #5984, opened 2026-08-05
[physics-loop][review-loop] toe-time blockM4 (STACKED on #5982) — Cycle 917: bounded theorem — the chain certifies: branching was never necessary; geometry buys a ceiling

JONbridgerapps:
STACKED on the blockM3 branch (PR #5982); only the blockM4 commits are new here. Cluster-cap evaluation (4th M-family PR): **OPEN** — falsifies a recovered no-go's geometric mechanism while confirming its home-system validity, replaces the slogan with three measured laws, and names the bracket's successor; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). The block's one design freedom (the geometry set) is declared; everything else inherits the frozen route-C gates verbatim (21/21 constants byte-verified; the partition rule verified by reproducing the frozen memo's own six cube fragments exactly). Disclosed: the spec's G3 description is arithmetically inconsistent — BOTH readings were run and reported; a receipt-key overstatement about the xi_reg estimator's provenance was corrected by the checker (the recovered d=1 note contains no formula; the estimator is the frozen memo's, adapted — booked as an import). Supervisor line-by-line review on record.

## Honest status

Bounded worked result at the frozen gates and fields (the graded field-ceiling diagnostic is a declared non-claim outside the frozen field set; the cube row is imported from the pinned receipts, not recomputed). No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **THE CHAIN CERTIFIES** — the headline falsification: the 9-site chain (max degree 2, zero branching, zero loops — the geometry the recovered d=1 note declares impossible) reaches permanence-grade redundancy at λ=0.05: R_ind=2 over its two disjoint arms, C_ab=0.00337, persisting four samples, first hit Jt=0.6. The no-go-inconsistency detector fires on the REAL chain row — with the planted-certification control proving the detector works. **Branching is not necessary** in this Ising comparator.
- **The mechanism's death as a generalization**: ξ_reg = 1 on EVERY geometry at both fields, certifying or not (including the cube) — the recovered note's stated mechanism cannot predict failure anywhere; it was comparator-specific (its own diagnosis: the Gauss-law blanket + ring zero-mode of the Schwinger chain, a different system per the 916 dictionary). **The no-go stands in its own comparator; the geometric slogan dies.** The operative variable is the COUNT of conditionally independent registers within range 1, not the range.
[...]

PR #5985, opened 2026-08-05
[physics-loop][review-loop] toe-time blockM6 (STACKED on #5984) — Cycle 920: bounded theorem — the measurement was right and the floor was a grid line

JONbridgerapps:
STACKED on the blockM4 branch (PR #5984); only the blockM6 commits are new here. Cluster-cap evaluation (5th M-family PR): **OPEN** — vindicates a never-checked measurement while retiring its reading, with an exact cancellation identity, a bisected true crossing, and a corrected sibling premise; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). All recovered artifacts digest-verified against the 916 record before use; ZERO imports from the historical engine — the primary reimplements from the note's equations and the checker supplies a genuinely THIRD construction (full operator algebra; nnz matching entrywise). Supervisor line-by-line review on record.

## Honest status

Bounded worked result at the note's own declared parameters (none alternative-named — parameter-pinned; g=0.0 unusable, reported not swept). Eight audit rows emitted; the 916 B2 correction flagged for a follow-up edit rather than left as two disagreeing receipts. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **Q1 — REPRODUCED, exactly**: 22/22 compared quantities agree across three implementations (every count vector identically; baselines at 1e-12; energies at 9e-15; the decisive crossings at 9e-14). **The historical measurement is correct as published.**
- **Q2 — the floor is a GRID ARTIFACT**: the true suffix crossing is **theta = 0.140516818611** (bisected to 1e-14; monotone); the declared 0.20 overshoots by ×1.4233 and is simply the smallest of six hand-chosen thresholds above it — nothing in (0.1, 0.2) was ever examined. Sharper: **the namesake quantity cancels out of its own criterion** (fill = kappa·(A/N) = N/12 identically, dev 4.4e-16 — "deposition per activity" plays no role in its floor; the criterion is a quantized crossing count); **11 of the 12 gated cells are zero by construction** (the "transient-complete cascade" rests on ONE event); and the mass parameter moves the true crossing +32% while the grid floor never moves — the grid conceals the dependence it was meant to measure. **Not a convention artifact**: both baselines coincide bond-by-bond at 3.3e-16 (both kicks are occupation-diagonal) — **correcting 916's bridge premise B2** from a baseline-kind mismatch to a preparation difference.
- **Q3 — the lineage verdict**: **0.20 is a grid label, not a measured threshold** — the d=3 lineage's floor citation is now dead from both sides (916's category error + 920's source-side artifact). Audit rows sharpened: the API drift re-verified (B still runs only from history); the grid mislabel confirmed OUTCOME-BEARING (the cache prints a 6-unit protocol while executing 10 units; all four published count vectors would differ on the printed grid — the floor happens to survive, but any citation of B's COUNTS cites numbers the stated protocol does not produce).

## Verification

Primary: exit 0, 27s; restriction gates hard-fail against the recovered cache bytes — all pass value-for-value; all falsifiers fire (incl. a planted cascade that relocates the floor and a single-cell disagreement flipping Q1 to REFUTED); double-run bitwise identical. Checker: exit 0, 21s; 6/6 claims survive; the declared system attacked against the note's, runner's, and engine's own bytes (all ten engine terms, both kicks); the crossings independently found at 9e-14; 11/11 teeth. Note: docs/DEPOSITION_REPRODUCED_FLOOR_GRID_ARTIFACT_CYCLE920_BOUNDED_THEOREM_NOTE_2026-07-28.md; ship receipt with per-file pins in outputs/.


JONbridgerapps:
Consolidated to the archive: ledger id pr-5985 (archive/ledger/85/pr-5985.json). Carrier: this ledger science line + closed-PR record. Branch retained; science line in the ledger entry.

PR #5990, opened 2026-08-05
feat(gravity): local move structure of the single cell's least-cost cuttings (cycle 734)

jonathonreilly:
## What this establishes

The object is one cell of the lattice carried through one tick: the four-cube on `{0,1}^4`, nearest-neighbour adjacency only, proper cubic rotations alone kept. Earlier cycles measured the floor of the adjacency cost on this cell. This cycle asks what freedom a cutting has once it sits at that floor.

**The least volume forces the piece count.** Over all 4368 five-corner subsets the volumes run `[(0,1360),(1,2672),(2,320),(3,16)]` in units of one over 24. The least nonzero value is one such unit, so a piece is one part in 24 and a cutting has exactly 24 pieces. A hole left by removing `k` pieces can only be refilled by `k`.

**A corner inside a piece is a corner of that piece — and the scope of the argument was measured, not assumed.** Whole-number readings are what the least volume buys: all 336 five-corner sets of volume two or three read some corner in fractions. The conclusion is not so bought — 0 of those 336 reach past their own corners. The note claims the narrower fact.

**The cuttings really cut the cell.** Each of the 15168 pairs of pieces sharing a cutting is pushed apart by a plane exhibited and verified in whole numbers, so the 24 pieces meet only on boundaries and, carrying the volume of the cell between them, fill it.

**No move on two pieces keeps the cost.** Of the 15168 pairs, exactly 288 admit a second fill, and that set is exactly the pairs meeting in four corners. Each carries one relation of type `(2,2,2)` with unit weights, so the four moving corners are the corners of a flat square and the move re-cuts it along the other diagonal. It always costs more, by `[(1,192),(2,96)]`.

**No move on three pieces keeps the cost, stated positively.** Of 649600 co-occurring triples, 40512 admit a second fill and 0 admit one by three pieces of least cost. The fills that do exist cost `[(19,27264),(20,14592),(21,384)]`, every one strictly above the three-piece floor 18.

**The smallest cost-keeping change replaces four pieces.** Over all 124812100 pairs of cuttings at the floor the number of differing pieces takes `[4,6,7,...,24]` — one, two, three and five absent. The note keeps the two absences apart: one, two and three are ruled out by the pair and triple measurements above; five is ruled out by a census complete because its entries sum to the number of pairs.

**That move is a rigid flip on one of five regions.** The 46128 smallest moves re-cut 120 regions of 8 corners, 5 shapes up to the cell symmetry in families `[12,12,24,24,48]`. A region holds `[8,32]` pieces and cuts into four in `[2,24]` ways, of which exactly 2 reach its own floor 24; swapping those two carries a floor cutting to a floor cutting 46128 times, exactly the number of smallest moves.

**Local rigidity and global width coexist.** Under moves of up to four, five, ... ten pieces the 15800 cuttings sit in `[349,349,157,61,61,13,1]` groups, reaching one only at ten — while 29069284 pairs share no piece whatsoever.

Every headline was re-derived by a method the runner does not use: fraction-free elimination for the volumes, a solve-and-check for the corner statement (13360 corners, no mismatch), a complete search with the opposite pivot rule (496849 nodes rather than 502838, identical 15800 cuttings), one sixth of the trace of the cube of the co-occurrence matrix for the triples, posting lists for the distance census, depth-first labelling for the ladder, stabiliser orders for the region families. The 15800 cuttings decompose into 391 families of sizes 8, 12, 24, 48 — matching the count measured in the previous cycle by a different route.

Every check was also perturbed to confirm it would fail on a wrong object. A cost reading one column alone varies inside 40 of the 57 piece families. Admitting the mirror turns gives 96 elements and 36 families rather than 57, so the count depends on keeping only proper rotations. The boundary test flags 37680 of 42752 corner readings. The fraction test flags 0 of the 2672 pieces of least volume. The separation test refuses all 159 sampled pairs that genuinely share an interior point. Dropping the most-used piece (1975 cuttings) leaves 13825. Of the 1248 pairs meeting in three corners, 0 admit a second fill. Of 12862 eight-corner sets with full spread, only 120 are regions.

Two findings changed the work rather than confirming it. A gate asserting the largest difference was below 24 **failed**, refuting a claim of mine that no two floor cuttings are disjoint; the complete census was then measured three independent ways and the honest result — 29069284 disjoint pairs — is now its own gate. And the perturbation of the corner statement inverted its framing: the containment conclusion is not special to the least volume, only the route through whole numbers is. Both halves are now measured and printed.

## Files

- `docs/PHYSICAL_LEAST_COST_CUTTING_FLIP_AND_MOVE_LADDER_CYCLE734_NOTE_2026-08-04.md`
- `scripts/physical_least_cost_cutting_flip_and_move_ladder_cycle734_2026_08_04.py`
- `outputs/..._cold_2026-08-04.txt`, `outputs/..._receipt_2026-08-04.json`
- `docs/audit/data/citation_graph_manifest.json` (separate `audit-infra:` commit)

Runner: 28 gates, `TOTAL: PASS=28 FAIL=0`, 65 s, no solver. Claim type `bounded_theorem`; audit status unset, for the independent lane to set.


jonathonreilly:
Review-loop PASS on repaired exact head ed1386dfd1aad3432d087e799a3e0d85df532e45, replayed onto exact main 46c96c71798745575b12126309a7784d93b96031.

Science correction: the submitted 649,600 count is the number of triangles in the pairwise cooccurrence graph, not genuine triples sharing one floor minimizer. Exact membership intersection gives 636,032 genuine shared triples and 13,568 spurious pairwise cliques. The repaired claim keeps the exact finite supplied one-cell, 24-piece, four-coordinate charge model: floor 144; 15,800 floor dissections; minimum minimizer distance 4; 46,128 four-piece edges; threshold component counts 349,349,157,61,61,13,1; 120 eight-corner regions in carried-family sizes 12,12,24,24,48; exactly two floor refills per region and every four-edge an involutive swap. Two-piece and four-piece replacements now pass exact complement cover/separation checks. The three-piece refill spectrum is explicitly candidate-only; the exact absence of a three-piece floor-preserving move follows independently from the complete distance census.

Boundary: no physical tick/Admissibility or assembly-cell-to-simplex bridge, no nonfloor smallest-move result, no other piece/cost classes, noncorner/coarser/multicell/arbitrary-domain or continuum extension, and graph connectivity is not physical dynamics or ergodicity.

Evidence: primary PASS=31 FAIL=0; independent reconstruction PASS=11 FAIL=0 with opposite-pivot cover and hostile controls; both canonical caches fresh and source-closure bound. Full exact-main repository pipeline PASS 18/18; strict lint zero errors; changed evidence checked=1 failures=0 control_failures=0; invariants rows=4807 with link_violations=0, class_f_violations=0, graph_delta=acknowledged; vocabulary and diff checks clean. Generated audit/status outputs stripped. Review-only: no audit verdict was applied.

PR #5986, opened 2026-08-05
[physics-loop][review-loop] toe-time blockM5 (STACKED on #5984) — Cycle 919: bounded theorem — degree 5 certifies at the frozen field; the threshold is located, not bracketed

JONbridgerapps:
STACKED on the blockM4 branch (PR #5984); only the blockM5 commits are new here (parallel sibling of blockM6 #5985, same base). Cluster-cap evaluation (6th M-family PR): **OPEN** — executes the ladder's named successor measurement and converts the (4,6] bracket into a located integer threshold with published margins, cross-degree controls, and a checker-corrected claim surface; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). The spec's "at least 2 branches of depth 2" tree family was run at both extremes, neither privileged. The checker corrected the primary's rationale for the 0.075 design extension (adopted; strengthens the claim). Supervisor line-by-line review on record.

## Honest status

Bounded worked result at the frozen route-C gates verbatim (21/21 constants quote-identical to the pinned 917 receipt; Cycle 917 reproduced value-for-value — 156 rows, maximum deviation exactly 0 — before any new number was produced). The threshold claim is frozen-grade; the graded ceiling table is diagnostic-grade (flagged in the receipt — not to be cited at the same grade). No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **Degree 5 certifies at 0.10 — all twelve cells YES** across four geometries sharing nothing but degree: K_{1,5}, a 16-site all-branches-depth-2 tree, a 10-site two-branches-depth-2 tree, and 917's cube with one face deleted (degree 6→5 at fixed loop number). n runs 6–16, depth 1–2, loops 0–4: degree is isolated from every confound the 917 checker named. **On integer degrees (4,5] = {5}: the field-ceiling threshold is LOCATED at degree 5.** Graded ceiling now complete and non-decreasing: degree 2 → 0.05, degrees 3–4 → 0.075, degrees 5–6 → 0.10.
- **The ceiling law confirmed at degree 5**: max R_ind = 5 on every loop-free geometry at every field; the loopy cube-minus drops to 3 above the low field while still certifying — loops cost redundancy (the ceiling drop and certification are different things). Third confirmation of the 917 law at a new degree.
[...]

PR #5991, opened 2026-08-05
[physics-loop][review-loop] toe-time blockM7 (STACKED on #5986) — Cycle 921: bounded theorem — the loop cost is a pair-cycle law, and loop count was never the variable

JONbridgerapps:
STACKED on the blockM5 branch (PR #5986); only the blockM7 commits are new here. Cluster-cap evaluation (7th M-family PR): **OPEN** — converts two blocks' measured slogan ("loops cost redundancy") into an exact mechanism law with a pre-registered predictor, a 120-cell adversarial confirmation, and an honest exception promoted to the next measurement; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). Both readings of "loop count" were run (linear and fitted-half cyclomatic), neither silently dropped. One receipt defect the checker found is documented rather than hidden: fragment-pair keys appear under two orderings (one inherited from the pinned 917 receipt, unchangeable without breaking the value-for-value gate); content verified identical by canonical comparison — downstream readers key on the unordered pair. Supervisor line-by-line review on record.

## Honest status

Bounded worked result at the frozen route-C gates verbatim (21/21 constants quote-identical to both pinned receipts; **917 AND 919 reproduced at deviation exactly zero — 156 rows each — before any new number**). Every candidate predictor is a pre-registered integer function of the graph computed before any propagator ran. The crossover-field ladder is probe-grade (outside the frozen field set). No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **The pair-cycle law.** For each pair of pointer-fragments, let d be their anchors' distance in the graph with the pointer deleted (shortest pointer-through cycle = d+2). Over the frozen window: **d=1 (triangle) kills both fragments' content at every field; d=2 (plaquette) kills the pair's mutual independence at λ≥0.075 but not at 0.05; d≥3 is free** — and max R_ind is the independence number of what survives. Loop-free geometries have no finite d, recovering 917's ceiling-equals-degree exactly; 917's high-field loopy failures are exactly its d=2 pairs.
- **Loop count refuted, in both readings**: a ten-loop geometry pays nothing; four internal loops pay nothing; six ladders hold the pair set fixed while loop count rises and the ceiling never moves. **Position** reduces to the binary through-the-pointer gate (once through, distance IS length).
- **Why no counting statistic can work**: two matched pairs of designed geometries are identical on every statistic the machinery tracks — sites, bonds, loops, seams, degrees, depth, components, fragment-size multiset — and still split their ceilings 4 vs 3. Only anchor distance differs.
- **Discrimination**: 42/42, 42/42, 41/42 exact at the three fields (32/32 on all designed geometries) against nine mechanisms and five degeneracy probes; best rival 34/42. Margins are wide (+0.016/−0.010 bits on the gate), unlike the 919 persistence razor.
- **The exception, promoted not buried**: 917's degree-2 chain at 0.10 (predicted 2, measured 1) — a second, size-driven, loop-independent channel (the 4-site arms cross the dependence gate at 0.0217). Unexplained here; launched as Cycle 927.
- **Scope rules that travel with the law**: under the frozen tie-break, anchor distance 2 is constructible only in cube coordinates (the graded regime cannot appear on tie-free geometries — a property of the frozen rule); length-3 cycles force +X/+X preparation structurally.

## Verification

Primary: exit 0, 218.5s; routes Chebyshev/Bessel vs Taylor at 6.8e-14, vs dense eigh at 6.9e-14; deterministic double-run with stable content digest. Checker: exit 0, 11.6s; fully independent machinery (Floyd–Warshall partitions parsed from the frozen memo's bytes, Pauli-Kron Hamiltonians, Lanczos/Krylov with full reorthogonalisation vs Padé, opposite site-ordering convention, Bron–Kerbosch MIS); **SUPPORTED, zero disagreements; 12/12 teeth; adversarial hunt 120/120 on 60 out-of-roster sub-lattices; model-degeneracy attack on 102 geometries — no rival within 8 cells**. Note: docs/LOOP_COST_PAIR_CYCLE_LAW_CYCLE921_BOUNDED_THEOREM_NOTE_2026-07-28.md; ship receipt with per-file pins in outputs/.


JONbridgerapps:
Consolidated to the archive: ledger id pr-5991 (archive/ledger/91/pr-5991.json). Carrier: this ledger science line + closed-PR record. Branch retained; science line in the ledger entry.

PR #5996, opened 2026-08-05
[physics-loop][review-loop] toe-time blockT6 (STACKED on #5951) — Cycle 922: bounded theorem — the P=32 miss derives from the short arc; two carrier labels change hands

JONbridgerapps:
STACKED on the blockT5 branch (PR #5951); only the blockT6 commits are new here. Cluster-cap evaluation (3rd T-family PR): **OPEN** — closes the lane's named carrier miss with a sealed dominating rule, corrects two carrier labels via a classifier defect nobody had looked for, and dissolves the residuals into the same condition read backwards; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). The spec's alignment-forbidden guess was declared a non-premise and is falsified — followed the data. **Seal honesty, disclosed in full:** the worker saw 891's label-level B=8 output during scoping before the condition was written; the condition was fitted only to clock-local B=4..7 incidence and its B=8/B=9 predictions were pre-registered by SHA-256 before any clock-local B=8 attribution was computed (the prediction contradicted the 891 label on two cells; the pre-registration recorded the risk cell in advance, so no retro-fit was possible). B=8 is label-informed but clock-locally blind; **B=9 is blind on every level — the primary never builds a B=9 corpus; only the independent checker does.** The checker's over-budget first run (which did rebuild B=8 and reproduced the primary's row exactly) is preserved in the receipt as superseded, disclosed as not-the-shipped-run. Supervisor line-by-line review on record.

## Honest status

Bounded worked result. **RC-2 is FITTED-THEN-SEALED, not derived** — the mechanism sketch holds for single-token orbit words only, and the note says so on its claim surface. Three rival closed forms fit the corpus equally well (the entry-gap values are multiples of 8 and N grows by 8 per bank — no threshold inside one 8-wide band is distinguishable; reported, not argued away). Anatomy dumps are sampled at 8 rows per cell with exhaustive classification. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **The miss, explained:** bank 2 OWNS the value 32 at B=7 through all three entry-gap row pairs — the geometry is intact — but **no clock containing bank 2 reads period 32 anywhere**. All 276 P=32 episodes exhaustively accounted (bank 4's same-edge complement; bank 0's cross-token); the bookkeeping identity at zero violations on 407 pairs.
- **The label theft (the larger finding):** 891's classifier is value-based with entry-gap priority, so the ENTRY_GAP label is stolen wherever a bank's entry gap equals its own edge complement. Clock-locally: **B=6 P=24 and B=8 P=32 have zero bank-owned entry-gap episodes** (both are same-edge complements — two 891 carrier labels corrected); B=7 P=24 splits (20 true entry-gap episodes; the rest complement). The value-level sealed holdout of #5951 stands untouched. Dated qualifications executed on the blockT5 branch with pin refreshed (updates #5951).
- **RC-1 (derived, B=3..12, zero disagreeing rows):** the entry-gap value is realised by exactly THREE ordered pairs, not one — and the handoff-carrying pair is the dominant carrier at several B; 891's swap-only census would have missed it.
- **RC-2 (short-arc necessity, fitted-then-sealed): a bank-owned entry-gap reading needs 2P < N.** At B=7 this excludes exactly P=32 — the miss, predicted. Sealed B=8 (532,176 stretches): necessity 6/6, one sufficiency failure at the pre-flagged b=B−2 cell. Fully blind B=9 (checker-built): necessity violations none; the one sufficiency failure is again the pre-flagged b=B−2/P=8 cell. Across 27 cells RC-2 strictly dominates 891's ring-alignment rule (wrong on 1 sufficiency-side vs 7).
- **The 40/48 residuals are not a fourth shape:** ordinary same-edge complements whose 2P ≥ N forbids ring-periodic readings — only stretch-local finite-form readings survive, which is why they come in ones and twos (and why #5951's finite form was exact on all four witnesses while its ring form refused them). 891's labels were correct; it simply had no rule for the route.
- **Named cheap successor:** the third entry-gap pair h_f(b)→r(b−1) exists geometrically at every B and never fires anywhere — no rule yet. RC-3 sufficiency has a named discriminating measurement (equal-width runs, ≥8 clean stable ticks), deliberately not forced into a rule.

## Verification

Primary: exit 0, 540s; 10/10 gates; the 891 restriction gate at zero failed checks; pre-registration digests printed with holdout-free build logs, byte-identical after the holdout; double-run identical modulo the runtime field. Checker: exit 0, 792s; SUPPORTED_WITH_PART_REFUTATION, 9/9 gates, 10/10 teeth — own tick generator validated tick-for-tick against the pinned kernel (three tiers, zero mismatches), a third-route set-based detector, the rule re-derived from the primary's stated text alone and determinate on all 55 cells; the part-refutation and the model degeneracy are carried on the claim surface as results. Note: docs/P32_CARRIER_SHORT_ARC_LABEL_THEFT_CYCLE922_BOUNDED_THEOREM_NOTE_2026-07-28.md; ship receipt with per-file pins in outputs/.


JONbridgerapps:
Consolidated to the archive: ledger id pr-5996 (archive/ledger/96/pr-5996.json). Carrier: this ledger science line + closed-PR record. Branch retained; science line in the ledger entry.

PR #5998, opened 2026-08-05
feat(emergent-geometry): a two sided charge on the pieces of the cell's least-cost cuttings (cycle 735)

jonathonreilly:
## What this responds to

Earlier cycles in this lane measured the floor of the adjacency cost on the single
cell of one lattice step and one tick, and the smallest cost keeping change a
cutting at that floor admits: it replaces four pieces and flips one region between
its two cuts, so the local freedom of a floor cutting is a switch for each region
it fills. Whether those switches can be thrown independently was left open. This
cycle answers that, and finds what the population carries in place of independence.

## What changed

A new source note, its paired runner, the recorded output, and the receipt. Nothing
else in the repo is touched except the citation-graph manifest, in its own commit.

**The switches are not independent.** If they were, a group of cuttings joined by
the smallest move would be a cube on its switches. That happens 276 times, covering
480 of the 15800 cuttings, and every such group has dimension 0, 1 or 2; no other
group has a number of cuttings that is a power of two at all. The obstruction is
exhibited piece by piece: of the 273936 pairs of switches a cutting offers at once,
54912 are pairs whose two moves share a piece. The largest group holds 9320 of the
15800 cuttings, with 33216 moves inside it and its farthest member 16 smallest moves
from the first.

**A two sided charge, read off the pieces.** Asking for a weight on the 192 pieces
in play, over the field with two elements, whose sum over a cutting's 24 pieces
reverses under every smallest move gives 120 demands of rank 86; they are
consistent. Adding the demand that the sum be kept by every move on six pieces
raises the rank to 87 and cuts the labels to a single two sided split, of sizes 7704
and 8096. Asking instead for reversal at six pieces has no solution at all. One
exhibited weight sits on 56 of the 192 pieces and splits all 120 regions; drop any
single piece from it and at least one region is left unsplit.

**Locally unpinned, globally rigid.** Drop any single one of the 120 region demands
and the rank is still 86, so no region pins the charge by itself. The 120 regions
fall into 5 families under the cell symmetry, and dropping a whole family does lower
the rank: sizes 12, 12, 24, 24 and 48 against surviving ranks 84, 84, 75, 83 and 64.

**The behaviour on larger moves was never demanded, and is mixed.** Of the 60096
moves on seven pieces 26880 reverse the charge, and of the 151704 on eight pieces
28608 do.

**A hard core.** 48 cuttings admit no cost keeping move on eight pieces or fewer.
They form a single orbit of the cell symmetry, and all 48 sit on the smaller side of
the charge.

## Runner

`TOTAL: PASS=26 FAIL=0`, exit 0, recorded output 5277 characters. No solver: a
complete backtracking search over the 400 pieces at the cost floor, complete
enumeration over explicit finite sets, and exact elimination over the field with two
elements. Every number in the note appears in the runner's own output, except the
cross-check section, which carries an explicit carve-out sentence naming it as
computed outside the paired runner.

## Cross-checks run before opening this PR

Each headline was re-derived a second way, by a method the runner does not use: the
move census by packed-bit exclusive-or against a popcount table over all 124812100
pairs rather than by the Gram product; the groups by breadth-first search rather
than union find; the two demands met a third way by colouring the move graph
directly; the ranks by dense elimination on integer arrays rather than big-integer
bit masks, agreeing with the runner's weight on every one of the 192 pieces; and the
120 regions rebuilt from the move census alone, by reducing each smallest move to
the unordered pair of four piece sets it exchanges, which returns the same partition
and the same 5 families.

Three checks were run to break the gates rather than confirm them: moving one piece
out of the exhibited weight drops the smallest moves it reverses from 46128 to
44206; applying the free sign turns the charge into its own complement and meets
both demands with the same two sizes, so the freedom names nothing new; and of the 4
labels the demand at four pieces alone allows, 2 survive the demand at six.

## Boundary

Every count is scoped to the single cell of one lattice step and one tick, with the
adjacency cost and the least volume as defined in the note. The pool of 192 pieces,
the 120 regions and their 5 families are measured by the search, not derived from a
symmetry statement. The charge is a two valued function on cuttings; nothing here
identifies it with a physical quantity or says it is conserved by any process. Time
enters only as the fourth column of the cell, and the tick flip is kept in the
symmetry group throughout, so no result depends on an arrow.


jonathonreilly:
Review-loop PASS and contained on `main` at exact SHA `3359b01a207724b3766f7d40d542d220bc713491`.

Provenance:
- submitted head independently verified: `2d1e2c0f8a612ee984d04c31b80c570fe8577e47`
- exact replay base: Cycle734-contained `ed1386dfd1aad3432d087e799a3e0d85df532e45`
- repaired/reviewed head: `3359b01a207724b3766f7d40d542d220bc713491`

Material scientific correction:
- The submitted headline that free switch behavior survives only through dimension two “anywhere” was false. It measured only entire connected components that are labelled cubes.
- Exhaustive labelled extension finds whole-component cubes only through dimension 2 (276 components, 480 vertices), but embedded commuting switch cubes through exact maximum dimension 6: 160 dimension-6 cubes, each with 64 vertices; no dimension 7 in the supplied 15,800-vertex/120-label graph.
- Retained finite results include the 400 cost-6 pieces; 15,800 exact 24-piece floor cuttings using 192 pieces; 46,128 four-piece move edges; component census; exact GF(2) charge split; fixed-system size-4/size-6 inconsistency; family deletion ranks; and the frozen 48-cutting orbit.
- Cycle734 is a direct, receipt-bound dependency. Cycle733 is context/order only.

Claim boundary: exact supplied one-cell normalized-volume-one corner-simplex model and declared four-coordinate L1 pair charge/move graph only. The no-dimension-7 result is confined to the enumerated graph/label language; the additive inconsistency is confined to the fixed 192-piece GF(2) system. No physical tick/Admissibility/simplex-selection, other piece/charge/move languages, arbitrary domain, or continuum claim.

Verification:
- primary `32/0`
- independent reconstruction `15/0`
- hostile cache/source controls fail closed
- full repo pipeline `18/18`
- strict lint: 0 errors
- changed evidence: 1 checked, 0 failures, forensic-ready
- invariants: link violations 0, class-F violations 0, graph delta acknowledged
- vocabulary, links, bytecode, and diff checks clean
- generated audit/status projections stripped; no audit verdict applied and `apply_audit.py` was not run

PR is MERGED and the remote topic branch is absent.

PR #6011, opened 2026-08-05
AXIOM (owner-approved): Admissibility determines the local probability distribution from nearest neighbors

JONbridgerapps:
Owner-approved constitutional wording change, 2026-08-05. The file path is unchanged so existing links and runner references remain stable.

## Governing change

Admissibility second sentence changes from:

> For each site, the available possibilities are determined by, and vary with, the nearest-neighbor conditions.

to exactly:

> For each site, the probability distribution over the possibilities is determined by, and varies with, the nearest-neighbor conditions.

This remains one Admissibility rule, not a new axiom.

## Forced reading with Qubit and Record

- The distribution ranges over the full Qubit one-site possibility domain.
- Available/admissible denotes the support of that distribution. On a finite atomic menu, support is exactly the nonzero-probability possibilities. On a continuous domain, a supported exact point may have zero singleton measure.
- Probability distribution carries normalization. Record locks exactly one supported realization, conditional on formation at that site.
- The memo does not specify the distribution form or values, realized draw, formation site, formation probability/rate, context selection, update law, or Born bridge.

These are interpretive consequences of the existing Qubit and Record axioms; no additional governing support sentence is added.

## Landing package

Stacked PR #6013 repairs every changed semantic consumer, strengthens the memo/registry guard, refreshes the canonical evidence caches, and removes stale campaign artifacts. The validation pipeline correctly invalidates 887 prior judgments on the changed framework-premise epoch, produces a 3,600-row re-audit queue, and leaves zero retained-grade rows until the independent audit lane re-ratifies them. Those generated audit/status outputs are validation-only and are not part of the landing.

Generated with Claude Code; independently reviewed and repaired through the repository review-loop.

PR #6010, opened 2026-08-05
[physics-loop][review-loop] toe-time blockQ14 (STACKED on #6006) — Cycle 940: bounded theorem — R1-A3 negative: no mirror in the substrate; A3 stands at its measured price

JONbridgerapps:
STACKED on the blockQ13 branch (PR #6006); only the blockQ14 commits are new here. Cluster-cap evaluation (13th Born-family PR): **OPEN** — answers the lane's decisive remaining question with a non-existence proof, a grammar theorem, and an exact logical placement of naturality vs A3, closing the exercise's wall-1 program; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). The prior-art sweep ran first; its dominant find — the stranded envariance note — was pinned **by blob out of the git object store** (it exists in no worktree) and byte-quoted: its own text concedes A3 is the only premise outside {Quantum, Record}, and its symmetry step fires only once A3 is granted. The r-lane w=1/2 uniqueness argument was checked and excluded (zero symmetry content). One checker bug was caught by its own sensitivity tooth *failing to fire* (a tautological semantic test, rewritten to real commutation). One checker finding adopted mid-block strengthens the result (multiset preservation is insufficient for a symmetry of the law, so the non-existence is proved at the loosest end of the spectrum and holds a fortiori); the science digest was unchanged by the adoption. μ stayed strictly parametric throughout. Supervisor line-by-line review on record.

## Honest status

Bounded worked result; window-relative (the 175-orbit tree; restriction gates at the full horizon) and arena-relative (the six declared sites). Theorem A2 is grammar-contingent (reported as a tooth). Colour refinement is sound-not-complete — exactly the direction a non-existence proof needs, corroborated by four independent checker routes. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **Theorem A1 — no substrate automorphism swaps the menu items at any site.** A non-existence proof: 1-WL colour refinement separates the two endpoint wires under every labelling *including the coarsest* (invariant under all lane and station permutations). The wiring witness: the pair is mirror-symmetric one gate deep and asymmetric two gates deep (divergent downstream wires decoded); the natural repair chain is irreducible, each breaking gate exhibited. The one genuine multiset-level involution the substrate owns **fixes both endpoint wires** — corroboration from the opposite side. Transfers verbatim to the 918 M_A arena.
[...]

PR #6012, opened 2026-08-05
[physics-loop][review-loop] toe-time blockQ15 (STACKED on #6010) — Cycle 943: bounded theorem — the menu asymmetry is one wire wide; the owner vindicated in structure

JONbridgerapps:
STACKED on the blockQ14 branch (PR #6010); only the blockQ15 commits are new here. Cluster-cap evaluation (14th Born-family PR): **OPEN** — corrects a shipped block's scope from its own bytes, localises the lane's asymmetry to a single named wire with the mirror structure exhibited, proves the encoding and axiom layers innocent, and names the substrate modification that would convert the newly adopted axiom into the first derived Born values; recorded in the campaign pack.

## Process disclosure

Worker-authored primary and independent checker (Claude Opus 5 under supervisor spec; substitution disclosed in the note). The block exists because the OWNER challenged the supervisor's reading of #6010 — and the challenge is vindicated in structure. Killed once by a process restart; resumed from transcript with everything recomputed from disk. **Six checker refutations, all adopted**, including two that hit the primary's own published claims (a commutation IFF that was only an IF, with the commuting-wire count now honestly ensemble-dependent — sign only; timing glosses falsified and withdrawn). μ stayed strictly parametric; no weight value appears anywhere. Supervisor line-by-line review on record.

## Honest status

Bounded worked result; window- and arena-relative as its parents. The affine intertwiner the checker found occupying the primary's declared scope limit is fitted, validated out-of-sample, and **not a law symmetry** (fails off-orbit; licenses no weight equality) — the honest claim is "no law-commuting value symmetry swaps the menu," not "no map relates the branches." The block ran against the pre-adoption memo; its findings fed the owner's adoption deliberations as context. No axiom/primitive/registry/audit surface touched. Independent audit still required.

## Results

- **The #6010 family delta, from its own bytes**: 940 swept wire-relabellings only — zero value-space maps — while the stranded envariance note's swap is a state-space operation. The owner's question was genuinely untested.
- **The localisation**: the endpoint wires are controls of nothing but the record apparatus (exactly six TOF patterns); the source-pointer pair is exactly menu-symmetric (net contribution invariant under the swap; checker-confirmed embedded in the full 3,108-gate chunk). Branch divergence over all eight atoms confines to **8 of 5,815 wires in one lane** — six mirror-paired and related exactly by flip{1,6}+swap(124,125)+swap(255,256). **The sole obstruction is the record cell's orientation field** — the one record datum with no mirror partner — diverging at the five genuine-menu atoms and never at the three lock-timing atoms. The bulk dynamics is menu-swap invariant. Two independent non-existence routes kill all eight atoms.
- **Timing**: every atom breaks strictly before its lock (choice-to-break as low as 2 boundaries) — the pre-record symmetric era is real but short; the structure is symmetric, the record apparatus is fast.
- **The encoding and the axioms are innocent**: #6010's A2 grammar theorem is **gauge** (a two-nonzero-word encoding leaves the leaf digests byte-identical; live-wire sensitivity control confirms the test sees); and the axioms do not orient the menu (six byte-quotes, a zero-hit ordering scan, the Open Gates' explicit disclaimer of formation rules). **#6010's negative is model-contingent; the class-level question is open.**
- **The named successor, now live under the adopted axiom (PR #6011)**: mirror-partner the orientation field. In that kernel the menu swap becomes structure-preserving, the two branches' neighborhoods coincide — and the adopted Admissibility distribution clause plus its covariance **forces the symmetric-site distribution: 50/50 as a theorem, not a registration**. Launching as Cycle 946.

## Verification

Primary: exit 0, 263.7s; 10/10 certificates, 67/67 restriction gates (936's tree and 940's results value-for-value), 14/14 teeth; double-run identical. Checker: exit 0, 332.8s; 16/16 certificates, 14/14 teeth; fully independent machinery (own AST replay, commutation engine, GF(2) solver, own gauge wire); import firewall over 936/940/943-primary at zero hits. Note: docs/PRERECORD_SWAP_ORIENTATION_WIRE_CYCLE943_BOUNDED_THEOREM_NOTE_2026-07-28.md; ship receipt with per-file pins in outputs/.

### DIFF
--- a/docs/PRERECORD_SWAP_ORIENTATION_WIRE_CYCLE943_BOUNDED_THEOREM_NOTE_2026-07-28.md
+++ b/docs/PRERECORD_SWAP_ORIENTATION_WIRE_CYCLE943_BOUNDED_THEOREM_NOTE_2026-07-28.md
@@ -1,0 +1,195 @@
+# One wire wide: the menu asymmetry localises to the orientation field — Cycle 943
+
+Date: 2026-08-05
+
+Authority: none
+
+Audit: unset
+
+Status: bounded worked result (owner-directed — the block exists
+because the OWNER challenged the supervisor's reading of Cycle 940,
+and the challenge is VINDICATED IN STRUCTURE, refuted in timing; no
+axiom surface touched — the block ran before the 2026-08-05
+Admissibility adoption and consumes the pre-adoption memo). The
+940-family delta is established from 940's own bytes: it swept
+RELABELLINGS ONLY (zero value-space maps — while the stranded
+envariance note's swap is a state-space operation, so 940 tested
+the wrong family for the owner's question). The value-space family
+is now swept, and the finding is sharp: the endpoint wires are
+controls of NOTHING but the record apparatus (exactly six TOF
+patterns; the source-pointer pair is EXACTLY menu-symmetric,
+mechanically verified and confirmed by the checker under a harder
+embedded test); the branch divergence over all eight atoms confines
+to EIGHT of 5,815 wires in one lane, SIX of them mirror-paired and
+related EXACTLY by the swap flip{1,6} + swap(124,125) +
+swap(255,256) — **the sole obstruction is the record cell's
+ORIENTATION field, the one record datum with no mirror partner**,
+which diverges at exactly the five genuine-menu atoms and never at
+the three lock-timing atoms. The bulk dynamics is menu-swap
+invariant; the asymmetry belongs to the record apparatus and is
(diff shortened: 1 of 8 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6012 (archive/ledger/12/pr-6012.json). Carrier: PR #6014 (twelve-gate census subsumes the sketch; automorphism proof completes it). Branch retained; science line in the ledger entry.

PR #6016, opened 2026-08-06
Cycle 736: the space of charges on the single cell's least-cost cuttings

jonathonreilly:
## What this responds to

Cycle 735 landed one two-sided charge on the 15800 cuttings of the single cell that
reach the floor of the adjacency cost, and left open whether that charge is one of many
or the one the cell carries. This cycle measures the whole space it sits in.

## What is measured

The 15800 cuttings span 88 dimensions over GF(2) and their differences 87.

**Which sizes of move a weight on the 192 pieces can reverse.** A weight is exhibited
whose reading changes across every one of the 46128 four-piece moves. At six, seven,
eight, nine and ten a family of 7, 5, 5, 7 and 5 exchanges summing to zero is exhibited,
each family of odd size, so at those sizes no weight can change across every move. The
four-piece exchange span sits inside both the six- and the seven-piece span, neither of
which holds the other; their sum is the eight-, nine- and ten-piece span, their meet has
dimension 85, and the four-piece span is exactly that overlap. Of the 63 non-empty sets
of move sizes exactly 3 admit more than one answer pattern, each admitting 2.

**The charge space.** Weights answering the smallest move uniformly form a space of
dimension 107; on the cuttings they induce 8 charges spanning dimension 3, of which 2
are constant. Up to swapping the two sides there are 3: one keeping the six-piece move,
splitting 7704 to 8096 with all 48 rigid cuttings on the 7704 side — the charge the
earlier cycle measured, reproduced here including the 26880 of 60096 it flips at seven;
a sister keeping the seven-piece move, splitting 7424 to 8376; and their sum, keeping
the four-piece move, splitting 5664 to 10136. Asking a weight to answer six and seven
uniformly at once leaves a space of dimension 105, every charge of which is constant.

**Where the four-keeping charge comes from.** The smallest move leaves 349 groups of
sizes 1, 2, 4, 7, 236 and 9320 with 144, 96, 36, 48, 24 and 1 groups at each size. No
symmetry splits a group; the 349 fall into 14 orbits, and the 24 of size 236 form a
single orbit with stabiliser of order 2. The four-keeping charge is exactly the
indicator of that orbit, holding 5664 cuttings, and the 48 cuttings carrying no move on
at most eight pieces all sit off it.

**Canonicity.** Every one of the 8 charges is left where it is, point by point, by all
48 symmetries of the cell, while the indicator of a single group of 236 cuttings is
carried to 24 different functions.

## Runner

`scripts/physical_cell_cutting_charge_space_cycle736_2026_08_05.py` — `TOTAL: PASS=14
FAIL=0`, about 14 seconds, under 4000 characters of output. It runs the complete search
over the 400 pieces of least cost itself rather than reading a cache.

- Every headline number re-derived by a second method: the reversal ladder by rank
  comparison of the exchange matrix against the same matrix with an all-ones column
  adjoined, independent of the dual-witness route the runner uses; the groups by a
  breadth-first walk instead of union-find.
- Each gate fed the object it should turn away — a perturbed weight, a wrong split, a
  charge moved off its orbit — and confirmed to fail.
- Every sentence of the note read against the printed numbers in both directions: no
  claim without backing, no gate without a sentence. Twelve wordings were corrected and
  one over-long line repaired.
- The population itself rebuilt from the sixteen corners upward sharing no step with the
  runner — volumes from a permutation sum instead of an expansion in two-by-two minors,
  faces from three-by-three minors instead of an inverted edge matrix, sample points
  chosen against an explicit list of facet planes, the search settling the last
  uncovered point instead of the first, and the symmetries by permuting and flipping the
  bits of a corner label. It returns the same 4368 five-subsets, the same 2672 pieces of
  least volume, the same least cost of six with 400 pieces at it, and the same 15800
  cuttings using 192 pieces between them, agreeing with the cached table row for row.

## Boundary

Claim type is `bounded_theorem`. Everything is a complete finite measurement on one cell
of the lattice carried through one tick; nothing here extends to more than one cell, and
the note says so in its own boundary section. Status is `unset` for the independent
audit lane to decide.


jonathonreilly:
Review-loop PASS (review preparation only; not an audit verdict). Submitted head `2027f16c60da42d151dd32d53ca4eea76844867e` was repaired and independently reconfirmed at PR head `50318f355dfd43056c9974f61c9d1ffdbe0830ca`, then landed fail-closed on exact current main as `00f72f2bc3a1ebc85c0777ecfad95f750cb227e5`.

Resolved blockers: fail-open/unbound evidence; absent executable independent checker; sample-mask covers mislabeled as geometric cuttings; missing N1–N8/N5 packet; premise/dependency overreach; empty size-five vacuity; and missing checker helper closure. Final bounded result is exact only for the supplied unit four-cube/corner-simplex/four-coordinate-L1/48-action model: 15,800 geometrically certified cuttings, the six nonempty move sizes `{4,6,7,8,9,10}`, rank-three induced GF(2) charge space with eight functions including two constants, and four-keeping support equal to the 24-by-236 component orbit. No size >=11, nonlinear/nonbinary, other model/action/cost, multi-cell, physical, arbitrary-L, boundary, or continuum claim.

Fresh primary `PASS=15 FAIL=0`; independent reconstruction and hostile controls `PASS=14 FAIL=0`; full exact-main pipeline 18/18; strict lint zero errors; changed evidence checked 1/failures 0 before generated-output stripping; invariants rows=4807/link_violations=0/class_f_violations=0/graph acknowledged; diff/vocabulary/portable links clean. Pipeline-generated audit/status outputs were stripped; only the citation manifest acknowledgment lands. No audit verdict was applied.

PR #6020, opened 2026-08-07
Quark-lane dial comparators are mixed-convention; corrected common-scale values (meta)

JONbridgerapps:
**Final reviewed source package.** Earlier sibling PRs #6019, #6021, and
#6022 were closed; the durable source content is consolidated here. The note is
intentionally `meta`/support, not a theorem or retained-grade proposal.

## Comparator correction

`QUARK_MASS_SPECTRUM_KOIDE_SCHEME_OPEN_GATE_NOTE_2026-05-26` displays
`r_up = 0.773642` and `r_down = 0.597141` from a mixed reference-scale
convention; the up-sector list also mixes a pole top mass with MSbar masses.
The corrected common-scale MSbar comparator is

```text
                                   r_up          r_down
open-gate note, displayed        0.773642      0.597141   mixed scale; pole top
same inputs, MSbar top           0.767630      0.597141   mixed scale
canonical common scale           0.83097057    0.62108951
                                 ±0.002211     ±0.007511
```

The central values are evaluated on the canonical `μ = 162.5 GeV ≥ m_t`
row. They preserve and widen `r_lep < r_down < r_up`; no existing consumer
claim is overturned. The kappa bookkeeping near-hit at `5/3` is display-only
and should not be carried forward.

## Exact boundary and prior art

The only new exact algebra is the sharpness witness

```text
2(μ₁² + μ₂²) − (μ₁ + μ₂)² = (μ₁ − μ₂)².
```

It is **not a converse**. The literal converse is false: `x = (4,2,1)` with
unequal factors `(1/4,1,4)` on `x` (equivalently `(1/16,1,16)` on the masses)
permutes the multiset and leaves `Q` and `r` exactly fixed. Common positive
rescaling, flavour-universal fixed-flavour running, the mixed/common-scale
distinction, and the already-known cross-sector spread are explicitly
deferred to their landed prior-art notes.

## Numerical provenance and residuals

- Mass central values/errors use the PDG 2022 quark summary-table vintage;
  `α_s(M_Z)=0.1180±0.0009` uses the PDG 2023 QCD review. Asymmetric errors are
  symmetrized by averaging upper/lower magnitudes, then rounded to the
  precision displayed by the runner.
- Error propagation is genuinely symmetric: centered `±1σ` finite
  differences give `σ(r_up)=0.002211` and `σ(r_down)=0.007511`.
- Finite heavy-threshold matching applies to flavours light at a crossed
  heavy-quark threshold. The comparator proxy therefore shifts `u,c`
[...]

PR #6021, opened 2026-08-07
[CLOSED — folded into #6020] housekeeping: lane 6 y_τ entry point

JONbridgerapps:
Pure housekeeping. One file, no science.

`docs/lanes/open_science/06_CHARGED_LEPTON_MASS_RETENTION_OPEN_LANE_2026-04-26.md`
still nominates **6B**, the `y_τ` Ward identity construction, as both its
"First parallel-worker target" and its Phase-1 entry point. That surface was
exhausted three months ago by
`CHARGED_LEPTON_Y_TAU_WARD_COMBINED_NO_GO_NOTE_2026-05-10`, which states that no
Ward identity of the form `y_τ_bare = (retained content) × (structural
sqrt-rational constant)` is constructible from the framework surface it defines.

As written, the lane doc points new workers at a closed construction.

## What changed

Annotated **in place**, not deleted, so the lane's history stays readable:

- a dated status-correction block at the top;
- strikethrough + superseded marker on the "First parallel-worker target" line
  and on the Phase-1 recommendation;
- a `CLOSED SURFACE (2026-05-10)` marker on §6B itself.

It also records that **6C/6D/6E are blocked as routed, not refuted** — they are
written as "automatic once 6B lands," and the no-go explicitly does not bar
closures via parallel-lane progress or new structural primitives, so a `V₀`
route that does not pass through a Ward identity of that form is untouched by
it.

## Deliberately not done

**Names no replacement target.** Choosing the lane's new entry point is science,
and this is housekeeping. The lane is left with no nominated first target rather
than a made-up one.

No status is set, predicted, or re-graded. No claim, runner, registry row, or
audit verdict is touched. Nothing outside this one file is modified.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EEnGMUdwTHWaaAH43odZqx)_

### DIFF
--- a/docs/lanes/open_science/06_CHARGED_LEPTON_MASS_RETENTION_OPEN_LANE_2026-04-26.md
+++ b/docs/lanes/open_science/06_CHARGED_LEPTON_MASS_RETENTION_OPEN_LANE_2026-04-26.md
@@ -10,13 +10,38 @@ active lane; the Koide flagship lane covers ratios but not absolute scale.
 1–3 months of focused work on existing Ward-identity scaffolding).
 **Primary closure targets:** retained absolute `m_e`, `m_μ`, `m_τ` derived
 from one axiom — no PDG observational pin.
-**First parallel-worker target:** construct the y_τ Ward identity on the
+**First parallel-worker target:** ~~construct the y_τ Ward identity on the
 charged-lepton carrier (analog of the retained `y_t(M_Pl)/g_s(M_Pl) = 1/√6`
 top-quark Ward identity). With the Koide ratios closure in flight providing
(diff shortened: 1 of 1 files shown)


JONbridgerapps:
Closing — the change itself is kept, just not as a standalone PR.

The lane-6 annotation is correct and worth having (lane 6 still nominated 6B, the `y_τ` Ward identity, as its first parallel-worker target three months after that surface was exhausted), but it is housekeeping rather than science and does not warrant its own review cycle. The identical diff is now folded into #6020, which is the single surviving PR from this session's work.

Superseded by #6020.

---

PR #6022, opened 2026-08-07
[CLOSED — novelty below the bar] Sector-independent dial exclusion

JONbridgerapps:
> **BLOCKED ON #6020 — do not land or audit this first.** Every number in T1 and
> T2 is consumed from the common-scale comparator note produced there. That file
> is **not present in this branch**, so the markdown link to it **does not
> resolve** and the citation graph cannot seed the edge that carries all of this
> note's inputs. **Land #6020 first.** If those comparators move, T1 and T2 must
> be re-run.

## Review pass (2026-08-07) changed the claim — read this before the checklist below

A review pass added a fourth wall, credited prior art, and narrowed the novelty
claim. **The N1–N8 checklist further down predates that pass**; where they
disagree, the note is authoritative. Deltas:

| # | Finding | Disposition |
|---|---|---|
| 1 | **The observation is prior art.** `FOURTH_AXIOM_RG_SCALE_DYNAMICS_SCOPING_2026-06-05` §2.1 already states T1 qualitatively ("the best single universal value … still misses a sector by `0.137`; a single scale-invariant fixed point gives one number and is falsified by the quark sectors") and §5 already states T1's forward constraint. The max-record-entropy and gauge-representation-channel no-gos carry the `r = 1/2` special case. **N1's route table and N3's hidden-wall scan were run without a prior-art sweep.** | Fixed: prior-art section added, all three cited, novelty narrowed to *the quantification only*. The prior note is `Type: meta` / non-claim, so this is a promotion-with-error-bars, not a duplication — but it must not be read as a new observation. |
| 2 | **A fourth wall was missing.** The comparators are computed from PDG masses through the C₃-circulant (Brannen) ansatz and the `√m` one-leg-amplitude identification, both recorded as **non-retained** by `QUARK_C3_CIRCULANT_SOURCE_LAW_BOUNDARY_NOTE_2026-04-28`. | Fixed: **W4 (mass-to-dial dictionary)** added. T1 escapes W3 but rides W4. "T1 stands if T2 falls" no longer implies T1 is dictionary-free. Wall count 3 → 4. |
| 3 | **Comparators silently superseded four landed notes** (`0.597`/`0.773` → `0.621090`/`0.830971`), and moved in the direction that *strengthens* T1. | Fixed and stated plainly. `Q = Σm/(Σ√m)²` is homogeneous of degree zero, so a common rescaling cannot move the dial; the whole shift is the mixed-mass-convention artifact. Downstream reconciliation of the four notes is **not** done here. |
| 4 | **The 149.5σ minimax was a rational grid scan**, which gives an *upper* bound on the minimax — the wrong direction for an exclusion claim. | Fixed: exact convex piecewise-linear envelope-breakpoint enumeration. Exact value **149.4807σ**. T1 additionally shown **K-independent** for every `K < 149.48`. |
| 5 | **Symmetrised errors were an unquantified modelling choice** (`m_u`, `m_d`, `m_s` have strongly asymmetric PDG errors). | Fixed: exact σ-inflation break-even factors now computed. Binding case is **down-type at 3.30×**; published arm asymmetries are far below that, so the down-type conclusion survives — but 3.3× is a stated condition, not a harmless convenience. |
| 6 | Three of the 29 checks were `check(..., True, ...)` fiat assertions. | Fixed: replaced by a real distinctness test or demoted to prose. |
| 7 | **Comparator firewall verified clean.** Every load-bearing comparison is `Fraction`; floats appear only inside `print` formatting. | No change needed. |
| 8 | **N4 residual-matching judgement verified correct.** `ACPHILAMBDA_RECORD_OUTCOME_ORBIT_OCCUPANCY_NON_SUPPLY_NO_GO_NOTE_2026-07-04`'s residual really is the `F_C`/`F_R` non-selection; it is rightly not cited as a witness. The W3 quote is verbatim and in context. | No change needed. |

Runner **37/37** (was 29/29), exact `Fraction` arithmetic downstream of the
comparator input declaration; cache regenerated and fresh.

## Claim (as reviewed)

**T1 (exact).** A *candidate sector-independent supply* is one whose output is a
single numerical `r` shared by every fermion sector. No such value exists within
5σ of all three common-scale dial comparators — the intervals are pairwise
disjoint, the triple intersection is empty, and the exact minimax still sits
**149.4807σ** from at least one sector. The σ figure is large mainly because the
charged-lepton dial is pinned to `1.0e-5`; the *physical* separation at the
minimax is `Δr ≈ 0.33`.

| sector | comparator `r` |
|---|---|
| charged lepton | 0.499990767 ± 0.0000102 (0.9σ from exactly 1/2) |
| down-type quark | 0.621090 ± 0.007335 |
| up-type quark | 0.830971 ± 0.002204 |

**T2 (exact, conditional on W3 and W4).** Uniform occupancy over `n` atoms gives
`w = 1/n`, hence the ladder `r = (n−1)/2`. Leptons sit **on** the `n = 2` rung
(0.9σ); both quark dials sit **strictly between** rungs, 16.5σ and 76.7σ from
the nearest.

## The four walls

- **W1 — carrier universality.** That all three sectors register the same C₃ `hw=1` carrier is not derived; the species bridge lists across-fermion-type alignment as an open residual.
- **W2 — registration scale.** The companion invariance theorem covers flavour-universal QCD running only; SM Yukawa contributions to the mass anomalous dimension are flavour-*dependent*.
- **W3 — energy dictionary (T2 only).** The weight-to-dial coordinate is supplied only through an explicitly unadopted, support-only energy dictionary.
- **W4 — mass-to-dial dictionary (T1 and T2).** The comparators ride the non-retained C₃-circulant ansatz and `√m` one-leg-amplitude reading.

All four are pairwise independent. T1 carries W1, W2, W4; T2 carries all four.

## N1–N8 gate checklist (as originally drafted — superseded where it conflicts with the table above)

**N1 — alternative routes (6 named, ≥5 required).**

| # | route | outcome |
|---|---|---|
| R1 | The spread is a mass-scheme/mixed-convention artifact | **ATTEMPTED, closed** by the common-rescaling invariance theorem (#6020): dial stable to 2.4e-12 across six common scales |
| R2 | Quarks don't register the same C₃ carrier, so `r` isn't one observable | **OPEN → wall W1**; `SPECIES_BRIDGE_MINIMUM_DECOMPOSITION_..._2026-06-13` explicitly excludes across-fermion-type alignment |
| R3 | The framework supplies `r` at another scale; SM Yukawa running is flavour-*dependent* | **OPEN → wall W2**; #6020 covers QCD running only |
| R4 | A universal rule with sector-dependent registered input | **OUT OF SCOPE by construction** — see N5 |
| R5 | A different C₃ coordinatization makes the dials agree | ~~ATTEMPTED, closed~~ — **superseded by review: this is now wall W4.** The original closure argument (that changing the `√m` coordinate breaks the leptonic `Q = 2/3` that motivates the lane) is motivational, not a closure |
| R6 | `r` isn't a count, so "between rungs" proves nothing | **PARTIAL → wall W3**; T1 does not depend on it |

*(Missing from this table, added by review: the route "this is already recorded in the repo." It is — see the review table, item 1.)*

**N2 — wall independence.** All pairs independent. Wall count is now **4**, of which T1 carries 3 and T2 carries 4.

**N3 — hidden-wall scan.** Hits on "registered" → wall W2. Hit on "standard" for QCD flavour-universality → explicitly supplied condition in #6020. PDG inputs annotated as external comparators. **Review found one hidden wall this scan missed: the mass-to-dial dictionary, now W4.**

**N4 — residual matching.** `ACPHILAMBDA_RECORD_OUTCOME_ORBIT_OCCUPANCY_NON_SUPPLY_NO_GO_NOTE_2026-07-04` attacks the residual "the axioms do not select `F_C` vs `F_R`"; this note's residual is "no single numerical `r` fits three sectors." **Residuals differ → citation dropped as a witness**, retained only as complementary context. Witness count after dropping: 0. *Review verified this against the cited note: correct.*

**N5 — rhetoric audit.** "No sector-blind supply" tested at resolutions *value* / *rule* / *input*. Verified only at **value**; the broader phrase does not appear in the note. Runner Part C gives the existence proof that rule-level supplies survive — *and, after review, additionally checks that the surviving rule genuinely needs three distinct inputs, so it is not a disguised single value.*

**N6 — partial-closure path scan.** A convention-level reframe *does* dissolve this wall: declaring the grain to be per-sector registered data rather than axiom content. Reported rather than suppressed; the note does **not** say "a new axiom is required." Primitive registry check run — the three approved primitives each declare they supply no dimensionless quantity, weighting rule, or probability rule, so no "no retained primitive supplies this" language is used.

**N7 — steelman (this is why it's demoted).** *"Your comparison presupposes the quark sectors register the same C₃ object as the leptons — precisely the across-fermion-type alignment the species bridge note lists as an open residual. Grant that, and you still assume the framework's `r` is the low-scale registered one; your own invariance theorem only covers QCD, and SM Yukawa contributions to `γ_m` are flavour-dependent, so a substrate-scale `r` could be universal and run apart. You have excluded a value nobody claimed, at a scale you did not establish, for an observable you did not show is the same across sectors."* **Review extends the steelman: "…and computed through a mass-to-dial dictionary your own repo records as non-retained."** That is why this ships bounded-with-named-walls.

**N8 — cross-cycle echo.** Structurally similar walls here have been retired by **reframe** (moving a residual from physics to convention). That mechanism applies and is named in the note as the route the claim does not reach.

**Verdict: demoted, then PASS at the narrowed scope — with novelty reduced to quantification and a fourth wall added in review.**

## Non-claims

Not a no_go. Asserts no route is structurally closed. Derives no `r`, `Q`, `δ`,
mass, scale, mixing angle, probability rule, species map, or sector weight —
every number is an external comparator. **The observation is not new**; only the
error bars, the exact minimax, the K-independence, and the common-scale
comparators are. **T1 is not dictionary-free** — it escapes W3 and rides W4.
Makes no claim about *why* the sectors differ, or that Admissibility supplies
the difference. Does not reconcile the four landed notes that still carry the
superseded `0.597`/`0.773` values. No axiom, primitive, registry entry, or audit
verdict added, edited, retired, or predicted; status authority is the
independent audit lane alone.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EEnGMUdwTHWaaAH43odZqx)_

### DIFF
--- a/docs/SECTOR_INDEPENDENT_DIAL_EXCLUSION_WITH_NAMED_WALLS_BOUNDED_THEOREM_NOTE_2026-08-07.md
+++ b/docs/SECTOR_INDEPENDENT_DIAL_EXCLUSION_WITH_NAMED_WALLS_BOUNDED_THEOREM_NOTE_2026-08-07.md
@@ -1,4 +1,4 @@
-# No Sector-Independent Value of `r` Agrees With the Registered Dials, Subject to Three Named Walls (Bounded Theorem)
+# No Sector-Independent Value of `r` Agrees With the Common-Scale Dial Comparators, Subject to Four Named Walls (Bounded Theorem)
 
 **Date:** 2026-08-07
 **Type:** bounded_theorem
@@ -11,6 +11,15 @@ set, predict, or apply an audit outcome, and edits no registry.
 **Cached runner output:**
 [`logs/runner-cache/frontier_sector_independent_dial_exclusion_bounded_2026_08_07.txt`](../logs/runner-cache/frontier_sector_independent_dial_exclusion_bounded_2026_08_07.txt)
 
+> **Unlanded dependency — read before anything else.** Every number in T1 and
+> T2 is consumed from
+> `SECTOR_DIAL_SCALE_INVARIANCE_AND_COMMON_SCALE_COMPARATOR_BOUNDED_THEOREM_NOTE_2026-08-07.md`,
+> which is **in flight and unaudited** and is **not present in this repository
+> yet**. The link to it below **does not resolve on this branch**, so the
+> citation graph cannot seed that edge and the comparators cannot be checked
+> against their source here. This note must not be landed or audited before that
+> companion note lands. If its comparators move, T1 and T2 must be re-run.
+
 ## This is not a no_go
 
 It was drafted as one — "no sector-blind supply of the occupancy grain is
@@ -20,9 +29,12 @@ convincing steelman using either. Per the gate's own instruction, the claim is
 demoted rather than shipped in its original framing.
 
 What ships is the narrow, honest version: a **bounded exclusion of one named
-class of candidate supplies, with three walls stated as load-bearing
+class of candidate supplies, with four walls stated as load-bearing
 conditions rather than as caveats.** The full N1–N8 checklist is in the PR
 body.
+
+A fourth wall (W4) and the prior-art section below were added in review; the
+PR body's N1 route table and its "three walls" count predate them.
 
 ## Claim
 
@@ -30,14 +42,20 @@ Let a *candidate sector-independent supply* be any axiom, primitive, or
 theorem whose output is **one numerical value** `r*` of the C₃ dial
 coordinate, the same for every fermion sector.
 
-**T1 (exact).** No such `r*` exists within 5σ of all three registered
-common-scale dials. The three 5σ intervals are pairwise disjoint and their
-triple intersection is empty. The best possible single value — the `r*`
-minimising the largest pull over the three sectors — still sits **149.5σ**
-from at least one sector, because the charged-lepton dial is pinned to
-`1.0 × 10⁻⁵`.
+**T1 (exact).** No such `r*` exists within 5σ of all three common-scale dial
+comparators. The three 5σ intervals are pairwise disjoint and their triple
+intersection is empty. The **exact** minimax — the `r*` minimising the largest
+pull over the three sectors, computed by envelope-breakpoint enumeration, not
+by a grid scan — still sits **149.4807σ** from at least one sector. So T1 does
+not depend on the choice of `K`: it holds for every `K < 149.48`.
 
-Registered inputs (comparators, frozen as exact rationals in the runner):
+Two honesty notes on that σ figure. It is large mainly because the
+charged-lepton dial is pinned to `1.0 × 10⁻⁵`, three orders of magnitude
+tighter than the quark dials; the *physical* separation at the minimax is
+`Δr ≈ 0.33`, not something 149 times any physical width. And a σ-distance is
+only as good as the σ — see "Robustness" below.
+
+Comparator inputs (frozen as exact rationals in the runner):
 
 ```text
 charged lepton   r = 0.499990767 ± 0.0000102     (0.9σ from exactly 1/2)
@@ -52,7 +70,7 @@ atoms has singlet weight `w = 1/n`, hence `r = (n−1)/2` — the ladder
 from its nearest rung, up-type 76.7σ. So no uniform count of atoms, of any
 `n`, reaches either quark sector.
 
-## The three walls
+## The four walls
 
 These are load-bearing conditions, not caveats. Any one of them failing voids
 the corresponding part of the claim.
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Closing. This does not clear the bar for a claim row.

After review, its novelty reduced to putting error bars on an observation already recorded qualitatively — including a rough minimax — in `FOURTH_AXIOM_RG_SCALE_DYNAMICS_SCOPING_2026-06-05` §2.1, a `Type: meta` non-claim note. What remained rested on four open walls (carrier universality, registration scale, an unadopted energy dictionary, and a non-retained mass-to-dial dictionary), and its most useful contribution — naming that fourth wall — applies to the comparator values themselves, so it has been moved to #6020 where it constrains live numbers.

Also for the record, two things in this PR were wrong, not merely loose:

- The 149.5σ minimax was computed by a **grid scan**, which gives an *upper* bound on the minimax — the wrong direction for an exclusion claim. The exact envelope-breakpoint value is 149.4807σ, so the digit quoted happened to be right while the argument was not sound.
- Three of the runner's checks were `check(..., True, …)` fiat assertions inflating the headline count.

What genuinely survives from this work is one sentence, now in #6020: the corrected comparators sit *further apart* than the values they replace, so the correction widens the known sector spread rather than narrowing it.

Superseded by #6020.

---

PR #6041, opened 2026-08-09
feat(emergent-geometry): derive the three-parity law that decides which splits can carry a charge

jonathonreilly:
## What this responds to

Cycles 742-745 established that the size-sixteen frontier of the cut object is
reached by exactly one reading (four), that the symmetry group of the incidence
is the certified order-384 group, and that the complete census at sixteen is
132 = 12 x 11. Every one of those results came out of an anchored search that
splits a target size across the four quarters of the 192 pieces and then lifts
each split. Which splits the search must lift has been decided empirically per
run, by the search's own licensing test.

This package derives that licensing test in closed form and shows it is a law
of the object, not a property of the search.

## What changed

One source note, one paired runner, one cached output, one cold output, one
receipt.

The runner rebuilds the incidence from scratch (15800 cuttings, 192 pieces, 24
pieces per cutting, 1975 cuttings per piece, pivot rank 88, 18 readings, six of
them charges) and then measures, for each reading, which named blocks of pieces
the reading fixes the parity of.

**The law.** The answer is the same five blocks for every one of the 18
readings: the size, both halves, and quarters two and three. Quarters zero and
one are never fixed. Each half parity is the sum of the two quarter parities it
covers, so the five collapse to three independent conditions. Licensing a
reading is exactly three parities.

**Two independent confirmations, not one measurement restated.** Row reduction
over the field with two elements on the 88 pivot cuttings produces 104
independent piece sets that no cutting can see, each then checked invisible
directly against all 15800 cuttings rather than only the 88 it was built from.
A block is left free exactly when one of those sets meets it in an odd number
of pieces, which happens for quarter zero and quarter one and no other block.
Separately, the size parity a reading fixes is recovered as the parity of how
many cuttings the reading marks, which the odd cuttings-per-piece count already
forces. So two of the five determinations are derived rather than read off.

**The three classes.** Those parities sort the 18 readings into exactly three
classes, and all six charge readings land in the single class where all three
are even. Physically: every carrier of a charge reading has even size and meets
quarters two and three in an even number of pieces. Exactly one class demands an
odd size and it holds no charge. Against the search's own licensing test the
three-parity rule agrees on all 191250 pairs of a split and a reading up to size
twenty, with no disagreement.

**Closed forms, with the anchor priced.** For a charge at size two k the
licensed splits are the sum of the first k plus one squares (5, 14, 30, 55, 91,
140, 204, 285, 385, 506 at sizes two through twenty); those holding a piece in
the anchor quarter are the sum of the first k squares (1, 5, 14, 30, 55, 91,
140, 204, 285, 385). The anchored count at one size equals the licensed count
one size below, so drawing every subset through the anchor costs exactly one
size step. The splits a charge licenses that miss the anchor number k plus one
squared: 4, 9, 16, 25, 36, 49, 64, 81, 100, 121. The even-size odd-quarter class
licenses one step below a charge and anchoring there is free; the odd-size class
licenses twice a tetrahedral number and nothing at any even size.

**The counts are rejected against the wrong parities.** A closed form measured
against the machinery that produced it can hold by construction, so the runner
builds the count for all four parity pairs on quarters two and three and asks
which reproduces the charge count. Only the even pair does. Both closed forms
are also matched against a direct enumeration that never consults the incidence.

## Runner numbers

`TOTAL: PASS=24 FAIL=0`, under 20 s, peak memory well inside budget, output
under the character ceiling. No solver: integer and field-with-two-elements
arithmetic on a finite explicit object.

## Boundary

The note states its own limits. Licensing bounds the search's work, not the
answer: a licensed split can be empty of carriers, and at the sizes searched in
earlier cycles many are. The 104 invisible sets are one basis, chosen by the row
reduction, and the free-block conclusion is basis-independent because a block
met evenly by every basis member is met evenly by every combination; the sizes
quoted for them describe the basis built, not a bound on how small an invisible
set can be. Counts are measured through size twenty. No physical reading of the
parity classes is claimed.

## Ordering

This package is self-contained. It rebuilds the system from the incidence alone
and uses no group theory, no census, and no search result, so it does not depend
on any other in-flight package and can be reviewed in any order relative to
them. Earlier-cycle artifacts are named in backticks rather than linked, and the
sole citation edge is to the minimal axioms note.

## Audit

Audit status is left unset; the independent audit lane owns grading.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_CARRIER_PARITY_LAW_CYCLE746_NOTE_2026-08-08.md
+++ b/docs/PHYSICAL_CELL_CUTTING_CARRIER_PARITY_LAW_CYCLE746_NOTE_2026-08-08.md
@@ -1,0 +1,146 @@
+# The parity law that decides which splits can carry a charge — Cycle 746
+
+Date: 2026-08-08
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_carrier_parity_law_cycle746_2026_08_08.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+Carrying a charge is exactly three parity conditions. The 192 pieces of the
+cut object sit in two halves and four quarters, and for each of the 18
+readings the runner reads off the incidence which of those blocks the reading
+fixes the parity of. The answer is the same five blocks for every reading: the
+size, both halves, and quarters two and three. Quarters zero and one are never
+fixed, and each half parity is the sum of the two quarter parities it covers,
+so the five collapse to three. Those three sort the 18 readings into exactly
+three classes, and all six charge readings land in the single class where the
+three are even: every carrier of a charge has even size and meets quarters two
+and three in an even number of pieces. Each class then has its own count of
+licensed splits in closed form, and demanding a piece in the anchor quarter
+costs exactly one size step.
+
+## The rebuilt system
+
+The runner rebuilds the incidence table of the cutting system from scratch:
+15800 distinct cuttings on 192 pieces, each cutting using 24 pieces, each
+piece used in exactly 1975 cuttings. The incidence has pivot rank 88 and
+carries 18 readings, six of them charges. A set of pieces carries a reading
+when, on every cutting, the parity of how many of its pieces that cutting uses
+reproduces the reading.
+
+## Which blocks a reading can fix
+
+For a named block of pieces and a reading, either every carrier of that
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop landed with bounded claims.

Provenance: exact submitted head `03815bd45668bd1f2f256ac29762f36ed53bac36`; submitted merge-base `9b8207504c80d3056d025458f7e0cbd90c55b72f`; reconstruction review base `b02f50a9cfb8ca57c2dbe7026d06487947d22331`; final landed main `0e7b8819931620350ff8155160e8b0c3014d4563`.

The submitted result was not landable unchanged. Review repaired two substantive errors: `odd-ctl` is inconsistent and cannot define a third carrier class, so the exact result is 17 realizable supplied targets in classes 15 + 2; and the submitted chunk loop covered only 7,900 of 15,800 first endpoints. The landing also added a non-importing opposite-pivot checker, hostile controls, current proof/import/trace/N1-N8 surfaces, exact packet-helper registration, and refreshed the fail-closed Cycle 737→738→739→741→745 evidence chain.

Final scope: an exact finite GF(2) necessary block-parity licensing theorem for one supplied 15,800×192 incidence table. It does not prove carrier existence, sufficiency, a physical charge identification, another target family, a multi-cell result, or a continuum result.

Checks: Cycle 746 primary `TOTAL: PASS=31 FAIL=0`; independent `TOTAL: PASS=15 FAIL=0`; full audit pipeline PASS; strict audit lint zero errors; changed-audit-evidence `checked=1 failures=0`; 32 fingerprint tests PASS; portable-link, native-language, py_compile, diff, and manifest gates PASS.

Review disposition: PASS WITH BOUNDED CLAIMS. This review applies no audit verdict; the new bounded-theorem row still requires independent audit and dependency-chain ratification.

PR #6042, opened 2026-08-09
feat(emergent-geometry): bracket the least carrier size for the flip partner of a charge

jonathonreilly:
## What this responds to

An earlier cycle measured the least number of pieces that can carry the charge
called four on the cut object: sixteen. Each of the eight basic readings has a
flip partner, the reading marking exactly the cuttings it leaves unmarked, and
the pair sits in a four-element subgroup of the reading group. This cycle asks
the same least-size question of the flip partner of four and answers it as a
bracket: at least eighteen, at most twenty.

## The ceiling is a theorem the runner proves outright

The all-marked reading, which marks every one of the 15800 cuttings, needs
exactly eight pieces. The counting floor is immediate — each piece meets 1975
cuttings, so eight pieces are needed to cover 15800 marks — and the converse is
sharp: eight pieces meet 1975 times eight, exactly 15800, cuttings with
multiplicity, so an eight-piece carrier meets every cutting exactly once, which
says no cutting uses two of its pieces. Reading that relation off the incidence
(each piece stands in it to exactly 33 others) and enumerating its eight-element
cliques gives exactly 192 sets, one per piece, and all 192 verify directly. Each
piece lies in exactly 8 of them, and two of them share 0, 1, 2 or 4 pieces and
never 3, with counts 15072, 1920, 960, 384 over every pair.

Adding one of those to a sixteen-piece carrier of four gives a carrier of the
flip partner. Each of the eleven anchored sixteen-piece carriers of four meets
each of the 192 in 0, 1 or 2 pieces (counts 1216, 384, 512; five distinct
profiles; 128 met with multiplicity per carrier, as the 8-per-piece fact
requires). Largest overlap 2, so the smallest such sum has 16 + 8 - 4 = 20
pieces. All 512 of them have exactly 20 pieces and every one recomputes to the
flip partner against the incidence.

The same overlap cap rules out both short routes to eighteen on their own terms:
an eighteen-piece sum from one of the eleven would need an overlap of 3, and a
sum with a ten-piece carrier of the all-marked reading would need such a carrier
to exist, which the sweep at ten does not find.

## The floor is a search, licensed to the whole system

Every sweep here is anchored through one fixed piece. The runner checks the
licence in place rather than assuming it: for each of the eight basic readings
separately, the symmetries of the incidence table fixing that reading number
exactly 48, the full cube group, and they carry any piece to any other. So an
anchored sweep that comes back empty at a size is empty for the whole system at
that size.

The anchored search finds no carrier of the flip partner at any even size from
two up to sixteen. Odd sizes need none: both readings force an even total. Two
things make that emptiness informative rather than blind. At sizes eight and ten
the same machinery is asked for the all-marked reading and returns exactly the 8
eight-piece carriers through the anchor and nothing at ten — a positive control.
At sixteen the two questions are asked in one pass over the same 204 licensed
cells and the same 2004 splits: four comes back with eleven, the flip partner
with none. Every carrier recorded anywhere in the sweep, 19 in all, is rechecked
directly against the incidence with no failures.

## Eighteen, reported with its gap

Eighteen is the only even size between the emptiness at sixteen and the carriers
at twenty, so the bracket turns on it, and this package does not overstate what
was measured there.

An earlier exploratory pass covered 228 of the 285 anchored licensed cells before
its own row guard stopped it, and the verdict line it printed tested only whether
the run finished and abandoned nothing — never whether any join had been refused.
That pass is not evidence and is not part of this package.

The landed sweep meets all 285 anchored licensed cells, an independent recount of
them agreeing with what the sweep met, and processes 4796 splits: 4770 searched,
26 refused, 0 carriers. Each refusal is a join whose intermediate table would
want more rows than the guard allows; the guard sits at 30 million and the
refused joins asked for between 31 and 21766 million, the largest more than 700
times the guard. The refusals sit in 11 of the 285 cells. The note states this as
a complete pass over every licensed cell with 26 of 4796 splits refused, which is
bounded evidence and not an emptiness certificate.

## What is not claimed

The bracket is {18, 20}, not a value. The ceiling is exact for sums of one of the
eleven anchored carriers with an eight-piece carrier of the all-marked reading,
and the overlap cap of two is measured on those eleven; the note does not claim
every twenty-piece carrier arises that way. The reachable subgroup from four and
the all-marked reading has four elements, so nothing here bounds six, seven, or
their flip partners. Counting arguments were tried against eighteen and did not
settle it: a carrier can meet an eight-piece carrier in at most four pieces, but
the resulting total over the 192 is consistent with an eighteen-piece carrier
existing, and the 192 sets are not a balanced pair design, so the second-moment
step is unavailable. No physical reading of the readings, the carriers or the
bracket is claimed.

## Runner

34 sequential gates, `TOTAL: PASS=34 FAIL=0`, 406 s, peak memory under 2500 MB,
stdout under 5500 characters. Class-A integer and two-element-field arithmetic on
a finite explicit object, no solver; the cell complex, the cuttings, the readings
and the block bookkeeping are rebuilt from scratch in the runner and every number
in the note appears in its own output.

Constitutional effect: none. No axiom, Admissibility rule, primitive, policy or
audit status changes, and no import or assumption is added. The note links only
its paired runner and the minimal-axioms note; earlier-cycle artifacts whose
packages are in flight are named in backticks and carry no citation edge.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_FLIP_PARTNER_CARRIER_BRACKET_CYCLE747_NOTE_2026-08-08.md
+++ b/docs/PHYSICAL_CELL_CUTTING_FLIP_PARTNER_CARRIER_BRACKET_CYCLE747_NOTE_2026-08-08.md
@@ -1,0 +1,221 @@
+# How few pieces can carry the flip partner of a charge — Cycle 747
+
+Date: 2026-08-08
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_flip_partner_carrier_bracket_cycle747_2026_08_08.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+Each of the eight basic readings of the cut object has a flip partner: the
+reading that marks exactly the cuttings it leaves unmarked. An earlier cycle
+measured that the charge called four needs sixteen pieces to carry it. This
+cycle asks the same question of the flip partner of four, and brackets the
+answer between eighteen and twenty.
+
+The ceiling comes from a small theorem the runner proves outright. The
+all-marked reading, which marks every one of the 15800 cuttings, needs exactly
+eight pieces, and its eight-piece carriers are exactly the 192 sets of eight
+pieces that no cutting uses twice — one such set for every piece of the cut
+object. Adding one of those to a sixteen-piece carrier of four gives a carrier
+of the flip partner, and because the eleven anchored sixteen-piece carriers of
+four never share more than two pieces with one of them, the smallest sum built
+that way has twenty pieces. There are 512 distinct such sets and every one of
+them checks back against the incidence.
+
+The floor comes from search. Every search here is anchored, asking only for
+carriers through one fixed piece, and that costs nothing: the runner checks that
+the symmetries fixing a reading carry any piece to any other, so an empty
+anchored sweep is an empty sweep for the whole system. The anchored search finds
+no carrier of the flip partner at any even size from two up to sixteen, and at
+sixteen it is asked of exactly the cells and splits that deliver the eleven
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop PASS and landed to main at 92a7dbeeed819d61ce32d370ec80940b71fc90cb after focused repair and confirmation review. Science boundary: exact lower bound 18 and explicit upper bound 20; size 18 remains open because the licensed search refused 26 oversized cells, so no unsupported emptiness claim is made. Repairs include full predecessor traversal, the correctly named symmetry action, current Cycle 745/746 dependency receipts, independent non-importing checker, fail-closed hostile controls, compact canonical evidence, N1-N8/proof surfaces, and regenerated combined citation-graph acknowledgment. Final primary 43 PASS / 0 FAIL; independent 11 PASS / 0 FAIL; one final 18-stage pipeline PASS; strict audit lint and changed-evidence readiness PASS; caches fresh; generated audit/status outputs stripped. No audit verdict was applied.

PR #6043, opened 2026-08-09
feat(emergent-geometry): count the smallest carriers of a charge across the whole system

jonathonreilly:
## What this responds to

Cycle 747 bracketed the least carrier size for the flip partner of the charge called four between eighteen and twenty, using an **anchored** input: the eleven sixteen-piece carriers of four through one fixed piece, and the overlap cap of two measured against that anchored handful. Every conclusion in it therefore inherited an anchored scope. This package removes that scope.

## What changed

The symmetries of the incidence table close into a group of 384 which is transitive on the 192 pieces and permutes the 192 eight-piece carriers of the all-marked reading among themselves. Carrying the anchored eleven around by that group produces **132** sixteen-piece carriers of four, each rechecked against the incidence directly rather than trusted from the group action, and every one of the 192 pieces sits on exactly **eleven** of them.

Eleven is the same count the anchored sweep returned, and the anchored sweep at sixteen was complete. So every piece stands to the 132 exactly as the anchor does, and **the census is complete, not a sample**: any sixteen-piece carrier of four contains some piece, a group element carries that piece to the anchor, and the image is one of the eleven.

Three results follow.

**The group is not transitive on the smallest carriers.** The 132 split into six families of sizes **12, 12, 12, 24, 24, 48**. Transitivity on pieces does not descend to transitivity on the carriers, which is a structural fact about the object rather than a bookkeeping artifact.

**The overlap pattern is a proper coarsening of the family structure.** Sorting the 132 by how they meet the 192 eight-piece carriers gives only **five** profiles, of sizes 12, 12, 24, 36 and 48. The profile is constant on a family, so five profiles from six families means one profile covers two families; the sizes identify it as the profile of size 36, covering a family of 12 and a family of 24. The overlap pattern separates, but not completely.

**The ceiling at twenty becomes a system statement.** Across all 25344 pairs the overlap is 0 in 14592 cases, 1 in 4608 and 2 in 6144, and never more, so the cap of two is a property of the system. The 6144 pairs at the cap give **6144 distinct** twenty-piece carriers of the flip partner, one per pair with no collisions, each of the 192 pieces on 640 of them, in twenty families of 192 and six of 384.

Finally, a derivation rather than a search: were an eighteen-piece carrier of the flip partner to exist and meet an eight-piece carrier of the all-marked reading in k pieces, their sum would carry four at 26 less twice k pieces. For k at least 6 that is under sixteen, which no carrier of four achieves; for k equal to 5 it is exactly sixteen, so the sum is a census member meeting that carrier in 3 pieces, breaking the cap of 2. So **k is at most 4** — a shape constraint any construction or refutation at eighteen must respect, now stated system-wide rather than anchored.

## Runner

36 gates, `TOTAL: PASS=36 FAIL=0`, 107 s, output 5435 characters. The runner rebuilds the cell complex, the cuttings, the readings and the block bookkeeping from scratch; it does not read cycle-747 results in. Arithmetic identities cross-check every count in place: 132 x 16 = 192 x 11, 14592 + 4608 + 6144 = 25344 = 132 x 192, 132 x 128 = 16896 total meetings, 6144 x 20 = 192 x 640, 20 x 192 + 6 x 384 = 6144, and the six family sizes sum to 132.

## Boundary

The note states plainly that the group of 384 is the one recorded in this runner and does not claim it is the full symmetry group of the table; that the 6144 counted here are the twenty-piece carriers that arise as such sums, with no claim that every twenty-piece carrier does; that the bound of 4 at eighteen is conditional on such a carrier existing, and the bracket between eighteen and twenty is unchanged; and that the families are counted, not classified — no invariant separating the two families sharing the size-36 profile is named here, and finding one is open.

Earlier-cycle artifacts are backticked rather than linked because their packages are in flight. The note carries exactly two citation links: its paired runner and the axiom registry.

## Downstream

The natural next unit is the complete least-carrier-size table for the eight basic readings, where the group law four + six = seven gives a triangle of subadditivity bounds once any two entries are known.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_CENSUS_FAMILIES_CYCLE748_NOTE_2026-08-08.md
+++ b/docs/PHYSICAL_CELL_CUTTING_CENSUS_FAMILIES_CYCLE748_NOTE_2026-08-08.md
@@ -1,0 +1,182 @@
+# The smallest carriers of a charge, counted and sorted into families — Cycle 748
+
+Date: 2026-08-08
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_census_families_cycle748_2026_08_08.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+An earlier cycle measured that the charge called four needs sixteen pieces to
+carry it, and found eleven sixteen-piece carriers through one fixed anchor
+piece. That was an anchored count: it said what passes through one piece, not
+what the system contains. This cycle answers the whole-system question.
+
+The symmetries of the incidence table close into a group of 384 which is
+transitive on the 192 pieces and permutes the 192 eight-piece carriers of the
+all-marked reading among themselves. Carrying the anchored eleven around by
+that group produces 132 sixteen-piece carriers of four, each checked back
+against the incidence directly, and every one of the 192 pieces sits on exactly
+eleven of them. Eleven is the same count the anchored sweep returned, so every
+piece stands to the 132 exactly as the anchor does, and the 132 are all of
+them: the census is complete, not a sample.
+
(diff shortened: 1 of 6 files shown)


jonathonreilly:
review-loop PASS and landed directly to main at `9a572a2b92ed63765229e0990231f53ca8a451c9`.

Review repaired the stale 7,900/15,800 pair scan, added fail-first/nonzero receipts and exact Cycle746/Cycle747 dependency binding, introduced an import-independent opposite-pivot checker, corrected the transitivity boundary to the 48 geometric maps plus `b0`,`b1`, added proof/trace/N1-N8 and packet-helper surfaces, and removed the tracked cold output. Final evidence: primary 46 PASS / 0 FAIL; independent 17 PASS / 0 FAIL; sole final repository pipeline PASS through stage 18b; citation topology acknowledged. The exact finite result is the complete 132-carrier weight-16 census and stated group families; weight 18 remains unresolved, with only the conditional intersection bound proved. Independent audit remains unset.

jonathonreilly:
Closing because the reviewed content is now contained in main at 9a572a2b92ed63765229e0990231f53ca8a451c9.

PR #6044, opened 2026-08-09
feat(emergent-geometry): read a carrier's family from its own pieces

jonathonreilly:
## What this responds to

The in-flight cycle 748 package ([#6043](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6043)) established that the 132 sixteen-piece carriers of the charge called four are the complete census across the system, and that the group of 384 symmetries of the incidence table cuts them into six families of sizes 12, 12, 12, 24, 24 and 48. Its own boundary recorded an open question, quoted verbatim from that note:

> This note names no invariant that tells the two families sharing a profile of size 36 apart, and finding one is open.

This package answers it, and answers it in a stronger form than the question asked.

## The result

A family is, on its face, a fact about a carrier's *place* in the system: which other carriers the symmetries take it to. This package shows the family is also a fact about the **sixteen pieces themselves**.

Take a smallest carrier of four and throw away everything except the incidence table restricted to its sixteen pieces. For each pair of its pieces, count the cuttings that use both. The sorted list of those counts is different for every one of the six families. No two share it. So a smallest carrier of four already carries its family label in itself, and the group is not needed to read it off.

The separating list did not arrive alone. Four coarser readings of the same sixteen pieces are measured beside it, and the five nest strictly, each telling apart everything the one before it does and more:

| reading of the sixteen pieces | groups | families held together |
| --- | --- | --- |
| how often it meets the marked cuttings | 1 | 12+12+12+24+24+48 |
| pairs of its pieces no cutting shares | 2 | 12+12+12+24 \| 24+48 |
| the total of its pair counts | 4 | 12 \| 12 \| 12+24 \| 24+48 |
| its meetings with the eight-piece carriers | 5 | 12 \| 12 \| 12+24 \| 24 \| 48 |
| the whole list of its pair counts | 6 | 12 \| 12 \| 12 \| 24 \| 24 \| 48 |

The fourth row is the reading cycle 748 used, and the row where its two families sit together. The first four rows are what makes the fifth a measured result rather than a definition: the same construction applied four other ways lands short at one, two, four and five groups. The runner gates the whole sequence 1, 2, 4, 5, 6, checks the nesting group by group, and requires the last reading to give one group per family while each of the other four holds at least two families together.

Two of these readings are the same quantity wearing different clothes: summing the pair counts over pairs of pieces is summing the squares of the numbers of pieces each cutting contributes, with a fixed diagonal removed. They land in different places because the spread keeps information the count of never-sharing pairs throws away.

## Two further laws from the same computation

**The marked cuttings are met identically by all of them.** Four marks 5664 of the 15800 cuttings. Every one of the 132 carriers meets 2832 of those marked cuttings once and the other 2832 three times, and meets none of them twice or more than three times. Not merely a coarse reading; a constant one.

**Each carrier comes with a pairing of its own sixteen pieces.** Within a smallest carrier of four, the largest pair count is reached on exactly 8 pairs, and those 8 pairs cover all 16 pieces. The largest value is 433 on 60 of the carriers and 666 on the other 72, and 60 and 72 make 132.

**Overlaps take five values.** Two of the 132 share 0, 1, 2, 4 or 8 pieces and never 3, 5, 6 or 7, with counts 4926, 960, 1440, 960 and 360 — mirroring the same gap at 3 that the eight-piece carriers of the all-marked reading show.

The pairing claim is the one that looked too clean, so it ships with a control rather than an assertion. Sixteen pieces taken evenly spaced through the table's own ordering, in four different spacings, give 768 control sets. The first version of the control gate asserted none of them would have the pairing property; that gate **failed on measurement** — 1 control set does. Rather than drop the control or quietly weaken the sentence, the control was widened from one spacing to four and the gate restated as what is actually measured: 1 of 768, against 132 of 132 for the carriers. Every sentence about the pairing in the note and the receipt carries that qualifier, and the note's boundary says plainly that a control is not a proof of rarity.

The note's boundary also records that the separating reading is measured to separate and not proved to, that the five readings are five that were tried and a finer sixth is not ruled out, that the pairing is exhibited and not explained, and that nothing here moves the bracket on an eighteen-piece carrier of the flip partner.

## Runner

`scripts/physical_cell_cutting_family_separator_cycle749_2026_08_08.py` — rebuilds the cell complex, the cuttings, the readings and the block bookkeeping from scratch and gates every quantity in place. `TOTAL: PASS=37 FAIL=0`, exit 0, 95 s, peak memory under 2500 MB, 5837 characters of output. Re-run from the committed file reproduces the landed cold output byte for byte.

## Files

Five science files plus a separate `audit-infra:` commit that acknowledges the new note in the citation graph manifest (one node, one edge, `out_degree` 1 to the minimal-axioms note). No ledger row, no status, no policy change. This package changes no axiom, no Admissibility rule, no primitive, and adds no import and no assumption.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_FAMILY_SEPARATOR_CYCLE749_NOTE_2026-08-08.md
+++ b/docs/PHYSICAL_CELL_CUTTING_FAMILY_SEPARATOR_CYCLE749_NOTE_2026-08-08.md
@@ -1,0 +1,138 @@
+# A smallest carrier of a charge carries its own family label — Cycle 749
+
+Date: 2026-08-08
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_family_separator_cycle749_2026_08_08.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a cutting
+and 1975 cuttings through each piece. The charge called four needs sixteen
+pieces to carry it, and the system holds exactly 132 such carriers, every
+piece on eleven of them. The symmetries of the incidence table close into a
+group of 384 which does not carry every one of the 132 onto every other: they
+fall into six families, of sizes 12, 12, 12, 24, 24 and 48.
+
+A family is a fact about a carrier's place in the system — which other
+carriers the symmetries take it to. The question this cycle asks is whether
+the family is also a fact about the sixteen pieces themselves, readable
+without looking at the rest of the system. It is.
+
+Take a smallest carrier of four and throw away everything except the
+incidence table restricted to its sixteen pieces. For each pair of its pieces,
+count the cuttings that use both. The list of those counts, sorted, is
+different for every one of the six families. No two families share it. So a
+smallest carrier of four already carries its family label in itself, and the
+group is not needed to read it off.
+
+## Four readings that do not separate, and one that does
+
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop PASS and landed directly to main at 405ee12804a1646b34ff5bb5f7c35589178050bf. Repaired the half-row pair scan, full marked-intersection histogram, exact 1-of-768 control gate, and fail-closed exit behavior; added an opposite-pivot row-streaming independent checker plus current Cycle 748 dependency receipts and No-Go boundary record. Corrected primary: 42/42 PASS. Independent checker: 23/23 PASS. Load-bearing pair-count mutation: nonzero exit with 3 failed invariants. Canonical caches fresh. Sole full repository pipeline PASS. Audit verdict remains unset; no audit verdict was applied.

jonathonreilly:
Exact landed main SHA: 405ee1280482f7676d1625d7298735ad8c4163ad.

PR #6045, opened 2026-08-09
feat(emergent-geometry): a distance inside a smallest carrier of a charge

jonathonreilly:
## What this responds to

The in-flight cycle 749 package ([#6044](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6044)) showed that a smallest carrier of the charge called four carries its own family label, readable from the incidence table restricted to its sixteen pieces alone. Among the five readings it measured was one it recorded without explanation: the count of pairs of a carrier's pieces that no cutting shares takes exactly two values, 32 and 24, and nothing else.

This package asks what those never-sharing pairs *are*, and the answer turns a count into a shape, and then the shape into a distance.

## The result

Take a smallest carrier of four, throw away everything except the incidence table restricted to its own sixteen pieces, and join two pieces when no cutting among the 15800 uses both. The result is not an assortment. It is one of exactly two shapes and never anything else:

| shape of the never-sharing joins | carriers | joins | families |
| --- | --- | --- | --- |
| the corners and edges of a four-cube | 60 | 32 | 12, 12, 12, 24 |
| two separate three-cubes | 72 | 24 | 24, 48 |

Sixty and seventy-two make 132, and no carrier gives a third answer. The two join counts are cycle 749's two unexplained values, now named: a four-cube has 32 edges and two three-cubes have 24 between them. The shape is constant on each family, and the split it makes is exactly the split that reading made.

**The shape then carries a distance, and the incidence table gives the distance back.** Walk from one piece to another along the joins, count the steps, then look up how many cuttings the two pieces share:

| steps apart in the four-cube | cuttings the pair shares |
| --- | --- |
| 1 | 0 |
| 2 | 170 to 184 |
| 3 | 240 to 250 |
| 4, the far corner | 433, every time |

Twelve counts occur in all, and each of the twelve belongs to one number of steps and no other. **The number of steps is a function of the count.** Counting, with no coordinates anywhere in the computation, produces a distance on sixteen pieces.

Two things follow from that with no further computation. The distance a pair receives cannot depend on which relabelling of the shape onto the four-cube the search happens to return, since the count is a property of the pair alone and already fixes the distance. And the metric is recovered, not assumed: the bands do not meet, so a single integer count settles how far apart two pieces are.

The other 72 carriers behave the same way inside each of their two cubes: one step gives 0, two steps give 170 to 174, and the far corner of a three-cube gives 250 every time. Pairs taken across the two cubes give 202 to 666, meeting none of the inside values, and the largest count anywhere in such a carrier is always a crossing pair, in 576 of 576 cases.

**The tautology.** The first version of this package gated "one step gives 0" as a result. It is not one: one step in the shape *is* a pair sharing no cutting, by the definition of the joining, so that row is true by construction and would hold for any object whatever. The gate was rewritten to say so in its own text, the note says so twice, and the receipt records it as `one_step_is_true_by_construction`. What is claimed is only what the other three rows show.

**The relabelling objection.** If the search returned some other labelling of the same shape, would the bands survive? A scratch check found every four-cube carrier admits many relabellings, none of which disagree about any pair's distance. That check turned out to be unnecessary once the stronger statement was gated: because each count belongs to one distance, relabelling-independence is a consequence, not an extra assumption. The runner gates the stronger statement.

**The obvious wrong identification, measured and refused.** The pieces are cells on five of the object's own 16 corners, so the natural guess is that this four-cube is the ambient four-cube in disguise. A carrier's cells do cover all 16 corners, and joined pieces do meet in exactly 2 corners, a shared edge. But 2016 farther pairs meet in 2 corners as well. Meeting in an edge is therefore not what makes a join, and the package states plainly that the shape reported is not the ambient four-cube relabelled.

**Wrong-value rejector.** The relabelling search is handed a four-cube with two of its joins moved elsewhere, keeping both the total join count and the number of joins at every piece, and refuses it; it also refuses the two three-cubes against the four-cube in both directions.

**Control.** Sixteen pieces taken evenly spaced through the table's own ordering, in four spacings, give 768 control sets, of which **0** produce either shape, against 132 of 132 for the carriers. The note records that a control is evidence and not a proof of rarity.

## Runner

`scripts/physical_cell_cutting_carrier_cube_metric_cycle750_2026_08_09.py` rebuilds the cell complex, the cuttings, the readings and the block bookkeeping from scratch and gates every quantity in place. `TOTAL: PASS=39 FAIL=0`, exit 0, 96 s, peak memory under 2500 MB, 5876 characters of output. Re-run from the committed file reproduces the landed cold output byte for byte.

## Files

Five science files plus a separate `audit-infra:` commit that acknowledges the new note in the citation graph manifest (one node, one edge, `out_degree` 1 to the minimal-axioms note). No ledger row, no status, no policy change. This package changes no axiom, no Admissibility rule, no primitive, and adds no import and no assumption.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_CARRIER_CUBE_METRIC_CYCLE750_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_CARRIER_CUBE_METRIC_CYCLE750_NOTE_2026-08-09.md
@@ -1,0 +1,136 @@
+# A distance appears inside a smallest carrier of a charge — Cycle 750
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_carrier_cube_metric_cycle750_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a cutting
+and 1975 cuttings through each piece. The charge called four is marked on 5664
+of the cuttings, needs sixteen pieces to carry it, and the system holds
+exactly 132 such smallest carriers, each piece lying on 11 of them.
+
+Take one of those carriers, throw away everything except the incidence table
+restricted to its own sixteen pieces, and join two pieces when no cutting in
+the whole system uses both. What comes back is not an assortment. It is one of
+exactly two shapes and never anything else:
+
+| shape of the never-sharing joins | carriers | joins | families |
+| --- | --- | --- | --- |
+| the corners and edges of a four-cube | 60 | 32 | 12, 12, 12, 24 |
+| two separate three-cubes | 72 | 24 | 24, 48 |
+
+Sixty and seventy-two make 132, and no carrier gives a third answer. The two
+join counts, 32 and 24, are the two values an earlier cycle recorded without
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop PASS; content landed directly to main at 68b41b4ae2399f985860e8d75c3f40993773d423. The review repaired the half-row predecessor scan, added fail-closed exits and current Cycle 749 receipt binding, added an independent row-streamed / constructive-bit-label checker, narrowed the ambient-coordinate overclaim, and landed current caches, receipts, N1-N8/N5 evidence, helper mappings, and regenerated graph manifest. Primary PASS 42/42; independent PASS 14/14; hostile source mutations failed nonzero; full repository pipeline and changed-evidence readiness passed. No audit verdict was applied.

PR #6046, opened 2026-08-09
feat(emergent-geometry): the shape inside a smallest carrier is the object's own distance

jonathonreilly:
## What this responds to

The [cycle 750 package](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6045)
found that a smallest carrier of the charge called four has a definite shape: join two
of its sixteen pieces whenever no cutting of the unit four-cube uses both, and you get
the corners and edges of a four-cube on 60 of the 132 carriers and two separate
three-cubes on the other 72. That result threw the rest of the system away before
looking. Sixteen pieces held in isolation can be made to look like almost anything, so
the shape was open to the charge of being an artefact of the isolation.

This cycle puts the rest of the system back and asks whether the shape survives.

## The result

Join every one of the 192 pieces to every other that shares none of the 15800 cuttings
with it. That is one object, not 132 little ones:

| property of the whole never-sharing object | value |
| --- | --- |
| joins at every piece | 33 |
| joins in all | 3168 |
| symmetries of the table carrying joins to joins | 384 of 384 |
| pairs at one, two, three steps | 3168, 12576, 2592 |
| pairs where walking outward and multiplying the join table agree | 18336 of 18336 |

Now restore a carrier into it. Every pair that the carrier's own shape places one, two
or three steps apart stands exactly that far in the whole object:

| pairs inside a carrier | count | forced by construction? |
| --- | --- | --- |
| within three steps in the shape, same in the object | 10752 of 10752 | rows one and two only |
| of those, at three steps | 2496 | no |
| far corners of a four-cube, four steps in the shape | 480 | folds to three |

The shape is the distance the object already had. The single place a carrier bends is
the eight far corners of each four-cube, which have nowhere to be, since nothing in the
object stands more than three steps from anything else.

Two further readings. The two three-cubes of a split carrier interleave rather than
stand apart: 4032 pairs across them at two steps and 576 at three. And sixteen pieces
taken evenly spaced through the table's own ordering fail the isometry in 730 of 768
control sets, against 0 failing pairs for the carriers.

**Two of the three rows are tautologies, and the note says so in its own voice.** For an
induced subgraph, agreement at one step and at two steps cannot fail: a join in the shape
is the same relation restricted, and a two-path inside the carrier caps the ambient
distance at two while a non-join in the shape is a non-join in the object. The note states
that "anyone reading 10752 of 10752 as one measurement is reading it wrong; the
measurement is 2496." Only the three-step row had freedom to fail, by routing through a
piece the carrier does not contain.

**The far-corner fold is forced too, and is reported rather than hidden.** It is stated as
a consequence of the object having no fourth step. Leaving it out would make the isometry
look total when it is not.

**The surround is measured, not assumed to be interesting.** If the whole object were
evenly built, a carrier inheriting a four-cube from it would be unremarkable. It is not
evenly built: the number of pieces joined to both ends of a pair takes 9 different shapes
at one step and 10 at three. The four-cube is produced by carrying the charge.

**The relabelling search cannot have arranged the answer.** It reads the join table alone
and never how many cuttings a pair shares, and it is handed a four-cube with two joins
moved that preserves both the join count and the count of joins at every corner, which it
refuses.

**The distance is cross-checked by an independent method.** Walking outward from each
piece and taking boolean powers of the join table are computed separately and compared on
every pair.

**A limit on the preceding cycle is reported against interest.** Inside a carrier the
number of shared cuttings settles the number of steps. Across the whole object it does
not: of the 47 counts that occur, 44 settle the step and three do not, at 202, 212 and
250, on 1632 pairs. The count-to-distance law is a carrier phenomenon, and this note says
so rather than extending it.

**A control is not a proof of rarity.** That 730 of 768 evenly spaced sets are shortened
from outside is evidence that isometric placement tracks the charge, not a theorem that no
other sixteen pieces can be placed isometrically.

## Runner

```
TOTAL: PASS=43 FAIL=0
```

Under 100 s, peak memory under 2500 MB, output under 6000 characters.

## Files

- `docs/PHYSICAL_CELL_CUTTING_OBJECT_DISTANCE_CYCLE751_NOTE_2026-08-09.md` — source note
- `scripts/physical_cell_cutting_object_distance_cycle751_2026_08_09.py` — paired runner
- `logs/runner-cache/physical_cell_cutting_object_distance_cycle751_2026_08_09.txt` — cache

and no ledger row.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_OBJECT_DISTANCE_CYCLE751_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_OBJECT_DISTANCE_CYCLE751_NOTE_2026-08-09.md
@@ -1,0 +1,145 @@
+# The shape inside a smallest carrier is the object's own distance — Cycle 751
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_object_distance_cycle751_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## Headline
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a cutting
+and 1975 cuttings through each piece. The charge called four is marked on 5664
+of the cuttings, needs sixteen pieces to carry it, and the system holds
+exactly 132 such smallest carriers, each piece lying on 11 of them.
+
+A preceding cycle joined two pieces of one carrier whenever no cutting used
+both, threw the rest of the system away, and got one of exactly two shapes:
+the corners and edges of a four-cube on 60 of the 132, two separate
+three-cubes on the other 72. Throwing the rest of the system away is the
+weakness of that result. Sixteen pieces looked at alone can be made to look
+like anything; the question is whether the shape survives putting the rest of
+the system back.
+
+It does, and that is this cycle's claim. Join every one of the 192 pieces to
+every other piece that shares no cutting with it. The result is a single
+object, the same 33 joins at every piece, 3168 joins in all, carried to itself
+by all 384 symmetries of the table, and with no two pieces standing more than
+three steps apart:
+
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop completed on frozen head 9b4ac00ad55c2d84ae0d05300cc26acff48df3fe and landed the repaired content directly to main at ed4bbf49acc0cb9943f3e85ee483ec6ca36af6d0.

The submitted total-isometry headline was narrowed to the exact bounded theorem: all 10,752 within-component carrier pair occurrences through intrinsic distance three are preserved in the 192-vertex ambient never-sharing graph, including all 2,496 nontrivial distance-three occurrences. All 480 Q4 antipodes shorten from four to three, and split-carrier cross-component pairs occur at ambient distances two and three, so full isometry is explicitly not claimed.

Review repairs also restored the full 15,800-row scan, bound current Cycle749 and Cycle750 primary and independent receipts, added fail-first and nonzero-exit receipts, added a structurally independent streamed-pair/Floyd checker, landed N1-N8 and N5 evidence, removed the raw cold output, and regenerated the citation topology manifest. Primary PASS 49/49; independent checker PASS 16/16; sole full repository pipeline, strict lint, and changed-science evidence readiness all passed. Audit status remains unset for the independent audit lane.

PR #6047, opened 2026-08-09
feat(emergent-geometry): the carrier shape is common, the shared-cutting counts carry the charge

jonathonreilly:
## What this responds to

Two cycles found that a smallest carrier of the charge called four has a shape. The
[cycle 750 package](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6045)
joined two of a carrier's sixteen pieces whenever no cutting of the unit four-cube uses
both and got the corners and edges of a four-cube on 60 of the 132 carriers. The
[cycle 751 package](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6046)
put the other 176 pieces back and showed that shape is the distance the whole object
already had.

Neither asked the flat question that comes next: **is the shape rare?** If the object is
full of four-cubes, then finding one inside a carrier says nothing about the charge, and
both preceding results need narrowing.

This cycle asks, and the answer goes against the two cycles before it.

## The shape is common

Join every one of the 192 pieces to every other sharing none of the 15800 cuttings, then
count every sixteen pieces joined exactly like the corners of a four-cube.

| four-cubes among the never-sharing joins | count |
| --- | --- |
| through one piece | 4978, the same number at all 192 |
| in all | 59736 |
| carrying none of the eight readings | 59676 |
| carrying the charge called four | 60 |
| generating symmetries carrying every shape to a shape | 50 of 50 |

The four-cube is available at every piece, thousands of times over, and almost all of
them read as nothing. The shape is not the mark of the charge.

## What is the mark is a count

There is a single number attached to a shape with no reading anywhere in it: take the 120
pairs of pieces, count for each pair how many of the 15800 cuttings use both, and add the
120 counts. Nothing in that recipe consults a mark or a charge.

Rank all 59736 shapes by it and the 60 carriers fall out at the bottom, with a clear step:

| shapes ranked by the cuttings their 120 pairs share | total |
| --- | --- |
| every four-cube carrier of four | 19800 and below |
| every one of the other 59676 shapes | 20338 and above |

The same holds locally, not just in aggregate: at all 192 of the 192 pieces, the 5
lightest of the 4978 shapes through that piece are the 5 carriers through it, and the
sixth is at least 538 heavier.

Two weaker tests are reported alongside to show that not every plausible criterion does
this work. Asking that pairs three steps apart share fewer cuttings than pairs four steps
apart lets 1488 shapes through; asking that all eight far corners sit at 433 lets 672
through.

## Why there is a floor underneath the ranking

Part of this is derived rather than measured, which is what makes it more than a census.

A sixteen-piece set meets the 15800 cuttings 31600 times, since each piece lies on 1975
of them. Carrying the charge called four means meeting each of the 5664 marked cuttings
an odd number of times and each of the rest an even number. The ranked total is the sum
over cuttings of the pairs each cutting contributes, so it falls as the 31600 meetings
spread evenly, and the parity requirement is exactly what stops them. The cheapest spread
consistent with it costs

```
15800 + 5664 / 2 = 18632
```

shared cuttings at the very least, for any carrier of this reading whatsoever. The
carriers come within 1008 of that floor, sitting at 19640, and the 1008 is accounted for
exactly: 252 cuttings the carrier misses altogether and 252 it meets four times. Over all
132 carriers the total takes just 4 values, least 19640 and most 24216.

**The result was independently replicated blind, on a different algorithm.** A second
worker was given the object and the question but not the method or the answer, and built
the shapes by gluing Cartesian products upward rather than by relabelling — a four-cycle
into a three-cube into a four-cube, with a proof-carrying check at each step. It returned
4978, 59736 and 60, agreeing on every count. It also rejected three near-misses that a
sloppy enumerator would have swallowed (a twisted prism, a four-cube with two joins
moved, and a doubled complete bipartite shape), cross-checked the total against the
identity that sixteen times 59736 equals the sum of the per-piece counts, and ran a
random-restart completeness probe that found nothing absent in nearly two and a half
million hits.

**The separating test is not claimed to be the only description that fits, and the note
says so.** A one-sided threshold — asking only that no pair two steps apart shares more
than 184 cuttings — selects the same 60, as does a bound on the width of the two-step
band. The virtue of the form reported in the note is that it names no value, not that it
is unique. The note states this rather than presenting the band separation as the sole
criterion.

**The metric test does not tell four from the other readings, and the note leads with
that.** Of the 59736 shapes, 59676 carry no reading at all and 60 carry four. Within this
population, carrying a reading and carrying four are the same condition, so what the
ranking demonstrably separates is shapes that carry something from shapes that carry
nothing. Nothing here shows it would discriminate four from a different reading in a
population holding both.

**The floor is derived but is not shown to be attained.** 18632 is a lower bound. Nothing
here shows 19640 is forced, or that 18632 is out of reach for some carrier this search
did not produce. The note separates the derived half from the measured half explicitly.

**The gate that carries the headline discriminated in practice.** An earlier draft cut
the ranking at the wrong count and the gate failed, which is direct evidence it is not
satisfied by construction.

**This narrows the two preceding cycles, in their favour only where the evidence
supports it.** What survives from them is that the shape is real and that it sits
isometrically in the object. What does not survive is any reading in which the shape
detects the charge.

## Runner

```
TOTAL: PASS=44 FAIL=0
```

Under 300 s, peak memory under 2500 MB, output under 6000 characters.

## Files

- `docs/PHYSICAL_CELL_CUTTING_SHAPE_CENSUS_LEAST_SHARING_CYCLE752_NOTE_2026-08-09.md` — source note
- `scripts/physical_cell_cutting_shape_census_least_sharing_cycle752_2026_08_09.py` — paired runner
- `logs/runner-cache/physical_cell_cutting_shape_census_least_sharing_cycle752_2026_08_09.txt` — cache

and no ledger row.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SHAPE_CENSUS_LEAST_SHARING_CYCLE752_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SHAPE_CENSUS_LEAST_SHARING_CYCLE752_NOTE_2026-08-09.md
@@ -1,0 +1,147 @@
+# The shape is common; the counts of shared cuttings carry the charge — Cycle 752
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_shape_census_least_sharing_cycle752_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a cutting
+and 1975 cuttings through each piece. The charge called four is marked on 5664
+of the cuttings, needs sixteen pieces to carry it, and the system holds 132
+such smallest carriers, each piece lying on 11 of them.
+
+An earlier cycle joined two pieces of one carrier whenever no cutting used
+both and found the corners and edges of a four-cube on 60 of those 132; the
+cycle after it put the rest of the 192 pieces back and found that shape to be the
+distance the whole object already had. Neither asked the flat question that
+comes next. Is the four-cube shape rare?
+
+It is not, and this note reports that against the interest of the two cycles
+before it.
+
+## The shape is common
+
+Join every one of the 192 pieces to every other sharing none of the 15800
+cuttings, which gives 33 joins at each piece, and then count every sixteen
+pieces joined exactly like the corners of a four-cube.
+
+| four-cubes among the never-sharing joins | count |
+| --- | --- |
+| through one piece | 4978, the same number at all 192 |
+| in all | 59736 |
+| carrying none of the eight readings | 59676 |
+| carrying the charge called four | 60 |
+
+All 50 generating symmetries of the table carry every one of the 59736 shapes
+to a shape. The four-cube is therefore not the mark of the charge. It is
+available at every piece in the object, thousands of times over, and 59676 of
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop disposition: PASS after repair; durable content landed directly on main at 8b79924363823264f138d8ecc273d4e305fbd8e4. The review fixed the skipped half-row census, added fail-closed/nonzero behavior and fail-first receipts, bound the Cycle750/751 dependencies, added an independent Cycle752 checker, refreshed canonical caches, and narrowed the note to the proved supplied-object boundary (Q4 shape alone is insufficient; restricted pair-total separation), without causal charge/commonness or physical/multicell overclaim. Verification: sole final 18-stage review pipeline PASS; strict lint and changed-evidence checks PASS; primary 48/48 and independent 19/19; hostile expected-count mutations failed nonzero in both runners. Audit remains independent.

PR #6048, opened 2026-08-09
feat(emergent-geometry): derive the cycle-752 ranking as a spread about a forced mean

jonathonreilly:
## What this responds to

The [cycle 752 package](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6047)
ranked all 59736 four-cube-shaped sixteen-piece sets by a single number — for each of the
120 pairs of pieces, how many of the 15800 cuttings use both, summed — and reported that
the 60 sets carrying the charge called four come out lightest, above a derived floor of
18632. It reported the ranking as a measurement and gave no account of why ranking by that
number should track the charge at all.

This cycle was opened to attack that headline, not to decorate it. The suspicion was that
the ranking is a disguised count of the reading support: a carrier of four is odd on 5664
cuttings, and if the total carries a term of that size against a separating step of 538,
then the recipe "consults no mark" only in form. The measurement went the other way, and
in going the other way it produced the account the preceding cycle lacked.

## The ranked total is a spread about a mean the system fixes in advance

Each of the 192 pieces lies on 1975 of the 15800 cuttings, so **whichever** sixteen pieces
are taken — carrier or not, four-cube-shaped or not — they meet the cuttings 31600 times,
and 31600 is exactly twice 15800. The average number of the sixteen met per cutting is two,
and no choice of pieces can move it.

Writing `m` for the number met on a given cutting, the ranked total is the sum of
`m(m-1)/2`, and expanding about two gives

```
ranked total  =  15800  +  (1/2) x sum over cuttings of (m - 2)^2
```

for every sixteen-piece set whatsoever. This is an identity with no condition on how large
`m` may get. What the runner measures is the premise and the arithmetic, not the identity:
it checks on all 59736 shapes that the multiplicities sum to 15800 cuttings and weigh 31600,
and that the total computed from the multiplicity profile agrees with the total computed
from the shared-cutting matrix. Given the premise the identity is algebra, which is the
point of it. So the quantity cycle 752 ranked by is a variance: light means flat, and
the ranking is a measure of how far a set's meetings stray from an average the system
imposed before any set was chosen.

## The floor gains an attainment profile

Carrying a reading marked on `S` cuttings forces an odd number of meetings on each of them,
so the squared departure is at least one there, and

```
ranked total  >=  15800 + S/2
```

which for four is the 18632 cycle 752 derived. The identity says more than the bound:
equality holds **precisely** when the carrier meets each of the 5664 marked cuttings once or
three times and each of the other 10136 exactly twice, and the mean then forces the split
into 2832 met once and 2832 met three times. The floor is no longer just a number to sit
above; it names one profile. The lightest carrier found here misses that profile by exactly
252 cuttings it does not meet at all and 252 it meets four times, which is the whole of its
1008 excess.

## The attack, reported against interest

The hypothesis this cycle was built to test is refuted, and refuted in the direction
opposite to the one that would have hurt cycle 752.

| shapes ranked by cuttings met an odd number of times, largest first | position |
| --- | --- |
| the 60 carriers of four | 57973rd to 58032nd of 59736 |
| shapes strictly above them | 57972 |

Far from being extreme in reading support, the carriers are near the bottom of that ranking,
so the odd-count term works against their position rather than for it. What does separate
them is the count of cuttings met exactly twice — the cuttings sitting exactly on the mean:

| shapes ranked by cuttings met exactly twice, largest first | count |
| --- | --- |
| the 60 heaviest are exactly the 60 carriers | 60 of 60 |
| least carrier value | 9616 |
| next shape down the list | 8688 |

with no shape tying across that boundary — 36 shapes lie strictly above 9616 and 24 sit
exactly at it, all 60 of them carriers. That is a sharper statement than cycle 752's, and
it names what the ranking is sensitive to.

## Where the natural first identity does not hold

If no cutting is met more than four times, eliminating the thrice-met count between the two
sum rules gives a linear form in the odd count and the twice-met count alone. That form
holds on exactly 17544 of the 59736 shapes — and the runner gates that this set is
*identical* to the set of shapes no cutting meets more than four times, rather than merely
the same size. The condition is not idle: the largest multiplicity reached anywhere is 8,
and 42192 shapes pass four. The variance identity above is the one that holds unconditionally.

**This cycle is the adversarial review of the preceding one, run as a probe before any
artifact was written.** The probe was specified to test my own anticipated deflation of
cycle 752's headline, with the anticipated numbers written into the spec so a disagreement
would be visible. The measurement disagreed. The note leads with that, in the section
"What does not pick them out is the reading", and states plainly that the refutation is a
refutation of one hypothesis about this ranking, not a demonstration that no count of
reading support could separate these shapes.

**The equal-support control this probe wanted is not available in this population.** No
non-carrier shape has the carriers' odd count, so the cleanest test — hold the reading
support fixed and ask whether the twice-met term still separates — could not be run here.
The note does not claim it was.

**The gates discriminate rather than restate.** The ranked total is computed on every shape
by two independent routes, one from the multiplicity profile and one from the shared-cutting
matrix, and gated to agree; the floor is built from the reading support and the cutting count
alone rather than pinned; the identity is tested in exact integer form; and the linear-form
gate checks set identity, not cardinality, so a coincidence of counts would fail it.

**What is derived and what is measured are separated in the note's own Boundary.** The
identity and the floor with its equality condition are derived and hold for every
sixteen-piece set in the system. Everything about the 59736 shapes is measured over that
population and is not claimed beyond it. In particular **18632 is still not shown to be
attained**, nothing here shows 19640 is forced, and nothing here derives 9616.

**The reading-blindness caveat from cycle 752 survives unchanged and is repeated.** Within
these 59736 shapes, carrying a reading and carrying four coincide, so what both rankings
demonstrably separate is sets that carry something from sets that carry nothing. Neither is
shown to tell four from a different reading.

## Runner

```
TOTAL: PASS=42 FAIL=0
```

Under 900 s, peak memory under 2500 MB, output under 6000 characters.

## Files

- `docs/PHYSICAL_CELL_CUTTING_SHARED_COUNT_VARIANCE_LAW_CYCLE753_NOTE_2026-08-09.md` — source note
- `scripts/physical_cell_cutting_shared_count_variance_law_cycle753_2026_08_09.py` — paired runner
- `logs/runner-cache/physical_cell_cutting_shared_count_variance_law_cycle753_2026_08_09.txt` — cache

and no ledger row.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SHARED_COUNT_VARIANCE_LAW_CYCLE753_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SHARED_COUNT_VARIANCE_LAW_CYCLE753_NOTE_2026-08-09.md
@@ -1,0 +1,219 @@
+# The ranked total is a spread about a forced mean — Cycle 753
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_shared_count_variance_law_cycle753_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a cutting
+and 1975 cuttings through each piece. The charge called four is marked on 5664
+of the cuttings and needs sixteen pieces to carry it.
+
+A preceding cycle joined two pieces whenever no cutting used both, counted
+every sixteen pieces joined exactly like the corners of a four-cube, found
+59736 of them with 4978 through every single piece, and ranked them by one
+number: take each pair of pieces, count how many cuttings use both, and add
+those counts over all pairs. The 60 shapes that carry the charge came out
+lightest. That cycle reported the ranking as a measurement and a floor
+underneath it as a derivation, with no account of why the ranking works.
+
+This cycle supplies the account, and it is not the account the preceding cycle
+would have guessed. The ranked total turns out to be a spread about a mean the
+system fixes in advance, and the statistic that picks the carriers out is not
+the size of their reading support.
+
+## The mean is fixed before any shape is chosen
+
+For a set of sixteen pieces write `m` for the number of its pieces lying on a
+given cutting. Each piece lies on 1975 cuttings, so the sixteen pieces meet
+the 15800 cuttings 31600 times in total, and
+
+```
+31600 = 2 x 15800
+```
+
+Whichever sixteen pieces are taken, and whether or not they carry anything,
+the average number of them met per cutting is exactly two. Nothing about the
+choice can move it.
+
+## The ranked total is the squared departure from that mean
+
+Every cutting met `m` times contributes `m(m-1)/2` such pairs, so the ranked
+total is the sum of `m(m-1)/2` over the 15800 cuttings. Expanding about two,
+
+```
+sum (m-2)^2  =  sum m^2  -  4 x 31600  +  4 x 15800
+```
+
+and the ranked total is `(sum m^2 - 31600) / 2`, so for every sixteen-piece
+set whatsoever,
+
+```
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop completed and landed directly on main at `35635c28d237ff9ceae3633b3bb6544dc31d5ee2`.

Scientific disposition: the forced-mean identity, parity floor, minimum-carrier profile, and finite twice-met ranking are correct. Review found and repaired one substantive overclaim: odd-count equality at 5664 also identifies exactly the same 60 Q4 carriers, so twice-met is not the unique classifier. The retained negative is narrower: descending odd-count rank does not reproduce the low-total extremum (57,972 above / 60 equal / 1,704 below).

Repairs include fail-first input-bound receipts, current Cycle752 primary/helper contracts, a structurally independent row-blocked checker, canonical cache envelopes, N1-N8 boundary documentation, helper registration, and removal of raw cold stdout.

Verification: primary `PASS=44 FAIL=0`; independent helper `PASS=13 FAIL=0`; hostile primary/helper boundary mutations each exited nonzero; committed-tree changed-evidence passed; sole 18-stage repository pipeline passed, including audit lint with no errors, acknowledged graph delta, repo invariants, and skill axiom baselines. No audit verdict was applied.

PR #6049, opened 2026-08-09
feat(emergent-geometry): bound cutting invariants by an 88-dimensional shadow

jonathonreilly:
## What this responds to

Cycles 748 through 753 separated the sixteen-piece carriers of the charge called four from
the wider population of four-cube-shaped sixteens by numbers built out of how often pairs
of pieces share cuttings: a ranked total, a multiplicity profile, a spread about a forced
mean of two. Every one of those numbers is a function of a single object — the vector of
multiplicities a piece-set induces on the 15800 cuttings. This cycle asks what that object
can see, and finds a hard ceiling on the whole family.

## Two results against interest, first

**The spectral floor this cycle was opened to derive does not exist.** The plan was to
write the ranked total as a quadratic form and read a floor off the smallest eigenvalue of
the Gram matrix of the incidence. Measured exactly over the rationals, that matrix is
singular: its rank is 88, not 192, so it is positive semi-definite with smallest eigenvalue
zero at multiplicity 104. The bound the plan would have produced returns 15800 and nothing
more. Nothing here explains 19640.

**The rank argument that would have settled the two-cover question does not hold either.**
"No sixteen pieces meet every one of the 15800 cuttings exactly twice" would follow at once
if the incidence had full column rank, since the equation has an obvious rational solution.
It does not have full column rank, so that equation has a solution space of affine dimension
104 and the question stays open. What this cycle supplies in its place is a sharp per-piece
test, below.

## The cuttings see the pieces through an 88-dimensional shadow

The incidence table is 15800 by 192, every column carrying 1975 ones and every row 24. Its
exact rational rank is 88. So there is a 104-dimensional space of piece-weightings that
every one of the 15800 cuttings totals to zero — call it the blind space. All 104 basis
vectors are checked against every cutting in exact integers, not floating point.

The blind space is not an artefact of duplication: all 192 columns are distinct, the largest
number of cuttings two different pieces share is 1266 against 1975, and no two pieces and no
three pieces are dependent.

**Every blind vector is balanced, and this is derived rather than observed.** The 15800 rows
sum to 1975 times the all-ones vector, because every column carries 1975 ones. So the
all-ones vector lies in the span of the rows, and the blind space, being that span's
orthogonal complement, is orthogonal to it. A blind vector with entries in minus one, zero
and one is then exactly an **exchange**: two disjoint equal-size sets of pieces that every
one of the 15800 cuttings meets equally often. The basis found here has entries in exactly
that range, with supports 8, 12, 14, 16, 18 and 20.

## The smallest thing the cuttings cannot see is four pieces for four

| exchange size | verdict | how it is settled |
| --- | --- | --- |
| one for one | none | all 192 columns distinct |
| two for two | none | exact determinants over all 18336 pairs |
| three for three | none | exact determinants over all 1161280 triples |
| four for four | **exists** | pieces 4, 5, 10, 11 against 1, 3, 7, 9 |

Two independent routes settle the negative rows, and the stronger one is the determinant
route. For every pair and every triple the runner forms the small Gram matrix of
shared-cutting counts and evaluates its determinant in exact integers — the 3-by-3 case
by the explicit cofactor expansion, largest term about 7.7e9, far inside int64. None
vanishes. That rules out not merely an exchange but **any** linear dependence among three
or fewer columns, whatever the coefficients, which is strictly stronger than the exchange
claim and is what actually carries the minimality.

The second route is an independent sift. If two disjoint sets of pieces have the same column
sum then every weighting of the cuttings gives them the same total; weighting cutting `c` by
`c`, `c` squared and `c` cubed turns that into a three-number signature a genuine exchange is
obliged to share. **The sweep found no signature shared by two different sets at either size,
so no candidate ever reached the exact confirmation step and none was run.** The count of
exact confirmations is therefore zero by absence of candidates, not by anything having been
checked, and both the PR and the note say so rather than presenting a zero as evidence. What
shows the sift is not simply letting everything through is a separate gate: the known
four-for-four exchange **does** share its signature across the two halves under the same
weighting. The integer range of the weighted sums is gated too (three times the largest cube
sum against 2 to the 62), so the signature arithmetic is exact rather than merely large.

The four-for-four exchange is met equally often by every one of the 15800 cuttings, checked
directly against the incidence rather than inferred from the signature, and its orbit under
the 384 symmetries has 96 members. Those 384 symmetries fix the table of shared-cutting
counts and carry the blind space into itself; that check is done in float32 under an explicit
gated bound (192 x 1975 x 1 = 379200, under 2 to the 24) which makes every partial sum an
exactly representable integer, so the speed costs no exactness.

## The blindness is realized on the census, not just in principle

The 132 sixteen-piece carriers of four fall into six families of sizes 12, 12, 12, 24, 24
and 48, as cycle 748 reported. They give only **108 distinct multiplicity vectors**. The 24
coinciding pairs are the interesting part:

- each coincidence is a pair, never a larger group
- the two members are **wholly disjoint** — symmetric difference 32, no shared piece
- both members lie in the **same family**
- the difference of their indicators is a blind vector, a sixteen-for-sixteen exchange

So there are 24 pairs of disjoint sixteen-piece sets, both carrying the charge called four,
that no cutting in the system can tell apart. Across the wider population the same collapse
appears: the 59736 four-cube-shaped sets give 53632 distinct multiplicity vectors, 6104
fewer than there are sets.

**The consequence for the preceding cycles is stated plainly and is not comfortable.** The
cycle-752 ranked total and the cycle-753 spread are functions of the multiplicity vector, so
they are constant on these coincidences. What those cycles separate is classes of sets, not
sets. That is a ceiling on the entire family of invariants built from shared cutting counts,
and it is derived here rather than guessed.

## Over the two-element field

The rank is again 88, by rows and by columns alike. So the piece-sets carrying no reading at
all form a space of dimension 104, the reachable readings number 2 to the 88, and each one
is carried by exactly 2 to the 104 piece-sets — a partition of all 2 to the 192 piece-sets.
All eight named readings are reachable.

## The floor gains a per-piece test

Cycle 753 derived that the ranked total is 15800 plus half the squared spread about a mean
of two, so 15800 is an absolute floor. This cycle localizes the excess. For a sixteen-piece
set, write the shortfall of a piece as the number of cuttings it shares with the other
fifteen, less 1975. Counting incidences two ways gives

```
sum of the per-piece shortfalls  =  2 x (ranked total - 15800)
```

so the excess is exactly half the total shortfall, and **a sixteen-piece set sits at the
absolute floor if and only if every one of its pieces shares exactly 1975 cuttings with the
other fifteen.** That is an equivalence, argued both directions, and it is the sharp test the
failed rank argument was supposed to supply. Both sides are computed on all 132 carriers by
two routes that share no machinery — one through the shared-cutting table, one directly from
the incidence — and gated to agree. None of the 132 passes the test; the least total remains
19640.

**This cycle refutes its own design, and the refutation is the lead.** The programme was a
spectral floor; four scratch probes were run before any artifact was written, and the first
one killed it by measuring the rank. Both the note and this body lead with the vacuous bound
and with the failure of the rank route to the two-cover question, not with the positive
results.

**The zero counts in the sweep are reported as absence of candidates, not as evidence.** My
line-by-line review of the runner changed how this claim is presented. An earlier draft of
this body implied the sweep verified colliding candidates; it verified none, because none
arose. The minimality is carried by the exact determinants over all 18336 pairs and all
1161280 triples — a statement about arbitrary coefficients, stronger than the exchange claim
— with the sift as an independent second route whose empty result is corroborating rather
than load-bearing, and with the four-for-four tie gate showing the sift is not vacuous.

**Where a gate is implied by its premise, both artifacts say so.** Two of the 45 gates are
weaker than they look. Once two carriers are measured to share a multiplicity vector, their
agreement on the ranked total follows by algebra, and the ranked total takes only 4 distinct
values across the 132 carriers, so agreement on it is weak evidence on its own; what is
genuinely measured is the premise plus the agreement of two separately computed numbers.
Likewise the shortfall identity follows from the symmetry of the shared-cutting table, so
what discriminates there is the agreement of the two independent routes, not the identity.
Both disclosures are in the note, not only here.

**The minimality claim is bounded exactly where the sweep is exact.** Settled with certainty:
no linear dependence among three or fewer columns for any integer coefficients, and no
two-for-two and no three-for-three exchange. A general integer relation on four or more
columns with coefficients outside minus one, zero and one was not swept, and the note says so.

**What is derived and what is measured stay separated.** Derived: the balance of every blind
vector; the localization identity and its equivalence; the soundness of the sift; the count
of piece-sets per reachable reading. Measured on this object: the rank 88, the support
histogram, the coincidence counts, the census numbers. The rank is a measurement — nothing
here derives 88 from the symmetry, and the note does not pretend otherwise.

**The standing negatives from earlier cycles are repeated unchanged.** 18632 is still not
shown to be attained, 19640 is still not shown to be forced, and within this population
carrying a reading and carrying four coincide, so nothing here tells four from a different
reading.

## Runner

Rebuilds the cell complex, the cuttings, the readings and the block bookkeeping from scratch
and gates every quantity in place. Class-A: exact rational, exact integer and two-element-field
arithmetic on a finite explicit object, no solver. The Gram matrix is built in int64 and gated
against a shared-cutting table computed through a different arithmetic route, so the two
agree or the run fails.

```
TOTAL: PASS=45 FAIL=0
```

Under 300 s, peak memory under 2500 MB, output under 6000 characters. Re-run independently
after review; output byte-identical, sha256 `f4309f572ca5cfd26e84c8f0bddb1d097d5d5e91b5b892756104192d13db304b`,
empty stderr.

## Files

- `docs/PHYSICAL_CELL_CUTTING_SHADOW_RANK_UNSEEN_SWAP_CYCLE754_NOTE_2026-08-09.md` — source note
- `scripts/physical_cell_cutting_shadow_rank_unseen_swap_cycle754_2026_08_09.py` — paired runner
- `logs/runner-cache/physical_cell_cutting_shadow_rank_unseen_swap_cycle754_2026_08_09.txt` — cache

The manifest acknowledgement is a separate `audit-infra:` commit with no science content and
no ledger row.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SHADOW_RANK_UNSEEN_SWAP_CYCLE754_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SHADOW_RANK_UNSEEN_SWAP_CYCLE754_NOTE_2026-08-09.md
@@ -1,0 +1,232 @@
+# The cuttings see the pieces through an 88-dimensional shadow — Cycle 754
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_shadow_rank_unseen_swap_cycle754_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings and the block bookkeeping from scratch and
+gates each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The unit four-cube cuts into least-volume pieces at the adjacency cost floor
+in 15800 ways, and those cuttings use 192 pieces between them, 24 to a
+cutting and 1975 cuttings through each piece. Several preceding cycles ranked
+and separated sixteen-piece sets by numbers built from how often pairs of
+pieces share cuttings. Every one of those numbers is a function of a single
+object: the vector recording, for each of the 15800 cuttings, how many of the
+sixteen pieces it meets. This cycle asks what that object can and cannot
+distinguish, and the answer puts a ceiling on the whole family of such
+numbers.
+
+## Two results against interest, first
+
+**The spectral floor this cycle was opened to derive does not exist.** The
+plan was to write the ranked total as a quadratic form in the indicator of a
+piece-set and read a floor off the smallest eigenvalue of the incidence
+table's Gram matrix. Measured exactly over the rationals, that matrix is
+singular. Its rank is 88, not 192, so it is a positive semi-definite matrix
+whose smallest eigenvalue is zero with multiplicity 104, and the bound the
+plan would have produced returns 15800 and nothing beyond it. Nothing here
+accounts for 19640.
+
+**The rank argument for the two-cover question does not hold either.** The
+statement "no sixteen pieces meet every one of the 15800 cuttings exactly
+twice" would follow at once if the incidence table had full column rank,
+since that equation has an evident rational solution. It does not have full
+column rank, so the solution space has affine dimension 104 and the question
+stays live. What this cycle supplies in its place is a sharp per-piece test,
+given below.
+
+## The shadow, and the space the cuttings cannot see
+
+The incidence table is 15800 by 192, every column carrying 1975 ones and
+every row 24. Its exact rational rank is 88. So there is a 104-dimensional
+space of piece-weightings that every one of the 15800 cuttings totals to
+zero. Call it the blind space; 88 and 104 together make 192.
+
+The blind space is not an artefact of repetition. All 192 columns are
+distinct, the largest number of cuttings two different pieces share is 1266
+against the 1975 each of them lies on, and no two pieces and no three pieces
+are linearly dependent.
+
+A basis for the blind space is produced by exact rational elimination and
+then checked against the incidence table in exact integers, all 104 vectors,
+so the dimension does not rest on floating-point arithmetic. Its entries lie
+in minus one, zero and one, with supports 8, 12, 14, 16, 18 and 20 occurring
+38, 30, 14, 13, 3 and 6 times.
+
+## Every blind vector is balanced, and that is derived
+
+The 15800 rows of the table sum to 1975 times the all-ones vector, because
+every column carries 1975 ones. So the all-ones vector lies in the span of
+the rows, and the blind space, being the orthogonal complement of that span,
+is orthogonal to it. Every blind vector therefore sums to zero. This is an
+argument, not an observation; the runner gates the premise and the
+conclusion separately.
+
+A blind vector whose entries lie in minus one, zero and one is then exactly
+an **exchange**: two disjoint sets of pieces, necessarily of equal size, that
+every one of the 15800 cuttings meets equally often.
+
+The same fact is what defeats the spectral plan above, and it is worth
+stating in one place. Subtracting its average from the indicator of a
+sixteen-piece set leaves a balanced vector, and the ranked total is 15800 plus
+half the Gram form evaluated on it. A useful floor therefore needs the Gram
+form to be bounded away from zero **on balanced vectors specifically**. But
+the blind space is balanced, so it sits inside exactly the subspace the bound
+must avoid, and the smallest value there is zero. The bound returns the floor
+that was already in hand and adds nothing to it.
+
+## The smallest exchange no cutting sees is four pieces for four
+
+| exchange size | verdict | how it is settled |
+| --- | --- | --- |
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop PASS; landed the repaired science directly on main at 52d4a30218558b3ec33f034451aad91e24e5158f.

Retained finite result: exact rational and GF(2) rank 88 with kernel dimension 104; a basis-dependent exact kernel profile; no equal-cardinality exchange through three-for-three; an explicit four-for-four witness with orbit 96; 132 minimum carriers collapsing to 108 multiplicity vectors; and 59,736 induced-Q4 sets collapsing to 53,632 vectors.

Review repairs: the determinant argument now excludes only dependencies supported on at most three columns, while the complete signature sweeps carry the two-for-two and three-for-three exchange exclusions; the spectral result is narrowed to failure of the named minimum-eigenvalue route; rational rank deficiency does not decide 0/1 two-cover feasibility; the local equivalence explicitly consumes the Cycle 753 spread identity and distinguishes the universal 15,800 baseline from the four-reading floor 18,632; arbitrary pairwise invariants are not claimed to factor through multiplicity.

Evidence: primary cache PASS 48/48 in 244.18 s; independent helper PASS 15/15 in 150.36 s with live predecessor replay, SymPy exact nullspace, modular rank/signature routes, exact collision confirmation, and hostile contract checks. Sole final audit pipeline passed, strict lint had no errors, and changed-evidence readiness returned true. The raw cold transcript was removed; canonical source/input-bound receipts and caches landed. No audit verdict was applied; independent audit authority remains unset.

PR #6050, opened 2026-08-09
feat(cutting-system): the space the cuttings cannot see is not a symmetry object

jonathonreilly:
## What this responds to

The preceding cycle measured the exact rational rank of the piece sharing table of the 15800 least-volume cuttings of the unit four-cube as 88, and said in as many words that it did not derive it. The obvious place to look for a derivation was the group of 384 symmetries of the whole system: if the 104-dimensional space no cutting can see were a sum of whole same-type blocks of that group, its dimension would be fixed by the group's trace counts alone and 88 would follow as the complement. This cycle tests that and reports the answer, which is no.

## The result

**The blind space is not a symmetry object.** A subspace of a permutation module is a sum of whole same-type blocks exactly when the averaged product of its trace count with its complement's is zero. Here that averaged product is **21**. Twenty-one same-type parts, counted with multiplicity, are shared between the seen space and the blind space, so the splitting under the group is coarser than the seen/blind splitting and the rank 88 is not fixed by the group's trace counts. This is not a claim that the group forbids a 104-dimensional blind space, which plainly exists and is carried into itself by every symmetry; it is that the group does not single this one out.

**The same 21, by a second route sharing no machinery.** The maps commuting with every symmetry span 104 dimensions, and the blind-plus-seen splitting splits that algebra into four blocks of dimensions 33, 29, 21, 21. A commuting map carries the blind space into itself exactly when its blind-to-seen block vanishes, so exactly **83 of the 104** dimensions preserve it. In coordinates that condition reads: sharing table, times the map, times the transpose of the blind basis is zero. Elimination over a prime field on the flattened residuals of the 104 pair-orbit matrices returns rank 21 — a floor on the rational rank, hence a ceiling on the preserving dimension, meeting the character count from the other side. One route averages products of trace counts over 384 group elements; the other does elimination on residual matrices in a different arithmetic. Either alone could be a coding error and look clean; both landing on 21 is the check that neither is.

**A correction this cycle makes to its own first reading.** Of the 104 pair-orbit matrices, only 2 carry the blind space into itself on their own, and both are permutation matrices already in the group of 384; the other 102 miss by as much as 12738 in a single entry. The first reading was that 102 failures out of 104 means the preserving space is tiny. That is wrong — it confuses the failure count of a badly adapted basis with the dimension of a subspace. The true dimension is 83. The correction is set down in the note rather than quietly dropped.

**Cheap trace reading.** With the blind basis in reduced row echelon form, the trace of a symmetry is the sum over the 104 rows of the single basis entry standing where the row's own pivot column was moved to — 104 lookups per group element, no projector formed. The runner gates that this one-entry-per-row sum agrees with the formed trace on all 384 symmetries.

**Pair-orbit count 104, three ways**, the sharpest tying it to group structure: one orbit on the 192 pieces, and a piece stabiliser of order 2 whose non-identity element fixes 16 pieces, so twice the pair-orbit count is 192 plus 16.

**One four-for-four exchange orbit** has 192 signed images, every one blind against all 15800 cuttings, spanning **60** of the 104 blind dimensions and leaving **44**. That span is carried into itself by every symmetry and holds no constant weighting. What reaches the remaining 44 is open, and whether a four-for-four exchange exists outside this orbit was not swept, so no class-level statement about least exchanges is made.

## Runner

`TOTAL: PASS=32 FAIL=0`, under 100 s, under 2500 MB, output under 6000 characters. The group is rebuilt from scratch and its order 384 gated; the sharing table is built in exact integers with the constant diagonal 1975 and the largest two-piece share 1266 gated; the rank 88 comes from exact rational elimination, not a numerical threshold; the symmetry check runs in exact integers and again in a bounded second arithmetic and the two agree; every averaged product is gated to come out a whole number with remainder zero, which a wrong trace count would not give.

## Boundary

The rank 88 stays measured and not derived, and this cycle adds that the group's trace counts do not derive it either. Where a gate follows from its premise — the rebuild of 33 from 13, 6, 6 and 8, and of 104 from 29, 21, 21 and 33 — that is disclosed in the note, with the measured content named as the whole-number remainders. Whether every blind vector supported on eight pieces lies in the single orbit of 192 was not swept.

## Files

Note, paired runner, cached output, cold-run record and receipt, plus a separate `audit-infra:` commit adding one node and one edge to the citation graph manifest. No axiom, no framework Admissibility rule, no primitive, no policy and no audit status is changed, and no import or assumption is added.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_BLIND_SPACE_SYMMETRY_CYCLE755_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_BLIND_SPACE_SYMMETRY_CYCLE755_NOTE_2026-08-09.md
@@ -1,0 +1,250 @@
+# The space the cuttings cannot see is not built from whole symmetry blocks — Cycle 755
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_blind_space_symmetry_cycle755_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the cuttings, the readings, the piece sharing table and the group of
+symmetries from scratch and gates each quantity in place. Constitutional
+effect: none. This package changes no axiom, no framework Admissibility rule,
+no primitive, no policy, and no audit status, and it adds no import and no
+assumption to [MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The object is the unit four-cube on sixteen corners, cut into least-volume
+pieces at the adjacency cost floor. There are 15800 such cuttings. Between
+them they draw on 192 pieces, 24 pieces to a cutting, and each piece lies on
+1975 of the cuttings. From that, the piece sharing table: the 192-by-192
+integer table whose entry at a pair of pieces counts the cuttings using both,
+its diagonal constant at 1975. Call a weighting of the 192 pieces seen if it
+lies in the image of that table, and blind if the table sends it to zero —
+blind meaning that every one of the 15800 cuttings, totalled piece by piece
+against the weighting, comes out zero.
+
+The preceding cycle measured the exact rational rank of the sharing table as
+88, so the blind space has dimension 104 and the two add back to 192. That
+cycle recorded the rank as a measurement and said in as many words that it did
+not derive it. This cycle asks the next question: does the group of 384
+symmetries of the whole system — the piece permutations that carry cuttings to
+cuttings — single the blind space out on its own? Is the blind space a
+symmetry object? The answer is no, and the obstruction is measured twice, by
+two routes sharing no machinery, both returning 21.
+
+## Two results against interest, first
+
+**The blind space is not a symmetry object, so the route this cycle was opened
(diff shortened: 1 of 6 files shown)


jonathonreilly:
Review-loop disposition: PASS WITH BOUNDED CLAIMS; landed directly on main at `eee6ab5874e2fc207db5526dc82d9f71ae550c7c`.

The submitted finite arithmetic survived independent reconstruction, but the claim boundary required a substantive correction: the 104-dimensional cutting kernel is invariant under all 384 symmetries, so “not a symmetry object” was overbroad. The landed theorem says the invariant kernel is not a sum of complete isotypic components. Exact character inner products are seen/seen 29, seen/blind 21, blind/blind 33; the 21 is `sum m_seen(lambda)m_blind(lambda)`, not 21 distinct irreducible types. A direct ordered-pair-orbital reconstruction independently gives residual rank 21 over two primes, hence a commutant-preserving dimension of 83. The signed orbit of the supplied least four-for-four exchange has 192 vectors and exact/modular span rank 60, leaving 44 dimensions open. No universal symmetry/incidence rank no-go, classification of all support-eight vectors, or generator for the remaining 44 is claimed.

Repairs landed with a new independent helper, current Cycle 754 primary/helper receipt contracts, source/input-bound canonical receipts and caches, hostile overlap/boundary/predecessor mutations, fail-closed nonzero exits, the committed N1-N8 and five-line N5 evidence, helper registry wiring, raw-output removal, and a current citation-graph manifest. The exact-final pipeline passed; strict lint had no errors; changed-audit-evidence checked 1 row with 0 failures. The new bounded-theorem candidate remains `unaudited`; the independent audit lane owns any verdict or effective-status change.

PR #6051, opened 2026-08-09
feat(emergent-geometry): cycle 756 blind space is the span of exact-cover differences

jonathonreilly:
## What this responds to

Cycle 755 measured the rank of the cutting-system incidence table as 88 and did
not derive it, then showed that the group of 384 symmetries does not pick that
rank out on its own: the seen space and the space the cuttings cannot see share
same-type parts, so the group's trace counts are too coarse. That cycle named
three next places to look, and the third was to find 88 in the structure of the
incidence map itself rather than in the trace counts of the group.

This cycle looks there, and finds the whole blind space sitting inside a family
of objects the incidence table names by itself.

## The result

The unit four-cube on sixteen corners, cut into least-volume pieces at the
adjacency cost floor, has 15800 cuttings drawing on 192 pieces, 24 pieces to a
cutting, each piece on 1975 cuttings. Exactly 192 sets of eight pieces are used
at most once by any cutting, and each of those meets every one of the 15800
cuttings exactly once. Each is an exact cover of the cutting system.

**Every difference of two exact covers is blind**, with no symmetry argument
used: both send every cutting to one, so the difference sends every cutting to
zero.

**Those differences are the whole blind space, and this is exact.** The
argument is two-sided and uses only bounded-arithmetic ranks, each of which is
a floor for the exact rank:

- the incidence table has bounded rank 88, so its exact rank is at least 88, so
  the blind space has dimension at most 104;
- the 192 exact covers have bounded rank 105, so their exact rank is at least
  105, so their differences span at least 104 dimensions, all of them blind.

At most 104 and at least 104 leaves exactly 104, and forces both bounded ranks
to have been exact. Two different bounded arithmetics are run and agree on both
numbers. So the space the cuttings cannot see is precisely the span of the
differences of the eight-piece exact covers, and the rank 88 is the 192
coordinates less those 104. No floating-point rank, no threshold, and no value
carried in from the preceding cycle.

**No one class of exchange is the source.** Two exact covers share 0, 1, 2 or 4
pieces, never 3, with counts 15072, 1920, 960 and 384 over 18336 pairs, and the
differences have supports 16, 14, 12 and 8. Each of the four sharing classes,
taken alone, already spans all 104 dimensions. This sharpens cycle 755's
negative result at the class level: that cycle measured one orbit of least
exchanges falling short of the whole blind space and was careful to make no
class-level claim; the class-level truth is that the class spans everything and
what fell short was one orbit inside it.

**The smallest blind support has exactly the size of an exact cover.** Every
cutting meets the support of a blind weighting in 0 or in at least 2 pieces.
That is necessary and not sufficient, and necessity is the direction used here:
a support failing it carries no blind weighting. So each member of a blind
support must have all 1975 of its cuttings met by the others. For each of the four kinds of piece, those 1975
cuttings are not covered by any 2 other pieces nor by any 3, and are covered by
4, which kills sizes 2, 3 and 4. Sizes 5, 6 and 7 are killed by a complete
sweep: the 48 proper cube symmetries carry the table to itself and leave four
kinds of piece, 48 of each, so anchoring on one piece of each kind is without
loss; every valid support contains a least cover of its anchor's cuttings, so
enumerating all least covers and extending each to the target size builds every
candidate. That is 3306820 candidates, of which 27824 pass a necessary count
bound, and none is blind. And 8 is reached, by the 384 least exchanges. So no
blind weighting of the cutting table touches fewer than 8 pieces.

The two results fit together: the blind space is generated by differences of
eight-piece exact covers, and 8 is the smallest support any blind weighting can
have, so the exact covers sit at the floor rather than being one convenient
family of generators among many.

## Runner

`TOTAL: PASS=15 FAIL=0`, 3278 characters of output, 221 s, 341 MB peak. Fifteen
gates, contiguous, each measuring in place. What the gates defend:

- the cell complex, pieces and cuttings are rebuilt from scratch and the two
  independent counts of the incidences are gated against each other;
- the exact covers are found as eight-piece sets no cutting uses twice, then
  separately gated to meet every cutting exactly once, so the exact-cover
  property is measured rather than assumed;
- both ranks are measured twice, in two different bounded arithmetics, and the
  gate requires the two to agree before the inequality argument runs;
- the covering test used by the sweep is gated to discriminate, on both sides:
  it returns yes on the eight pieces a least exchange touches and no on the
  eight pieces of an exact cover, so its three zeros are not vacuous;
- each of the 48 proper cube symmetries is gated to carry the table to itself
  by a piece relabelling paired with a cutting relabelling, which is what
  licenses anchoring the sweep on four pieces instead of 192;
- the sweep is gated on both the count it built and the count that survived the
  necessary bound, both non-zero, so a sweep that silently built nothing cannot
  pass.

## Boundary

The rank 88 is relocated, not derived: it is now identified with a structural
quantity, the 192 coordinates less the span of the exact-cover differences,
rather than left as a bare elimination result. What remains open is why the
exact covers span 105. The sweep settles that no blind weighting has support
smaller than 8; it does not settle which supports of size 8 carry blind
weightings beyond the 384 least exchanges, which would need least covers of
size 7, and the note declines to claim it.

## Files

- `docs/PHYSICAL_CELL_CUTTING_BLIND_SPACE_CARRIER_SPAN_CYCLE756_NOTE_2026-08-09.md`
- `scripts/physical_cell_cutting_blind_space_carrier_span_cycle756_2026_08_09.py`
- `logs/runner-cache/physical_cell_cutting_blind_space_carrier_span_cycle756_2026_08_09.txt`
- `outputs/physical_cell_cutting_blind_space_carrier_span_cycle756_2026_08_09_cold_2026-08-09.txt`
- `outputs/physical_cell_cutting_blind_space_carrier_span_cycle756_2026_08_09_receipt_2026-08-09.json`

plus a separate `audit-infra:` commit adding the new note to the citation-graph
manifest.

## Downstream order

This is a follow-on to the cycle 755 package and should land after it. Its only
citation edge is to the minimal axioms; every other prior artifact is referred
to in prose without a link.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_BLIND_SPACE_CARRIER_SPAN_CYCLE756_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_BLIND_SPACE_CARRIER_SPAN_CYCLE756_NOTE_2026-08-09.md
@@ -1,0 +1,216 @@
+# The space the cuttings cannot see is the span of the differences of their eight-piece exact covers — Cycle 756
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_blind_space_carrier_span_cycle756_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the least-volume pieces, the cuttings at the adjacency cost floor,
+the piece sharing table, the eight-piece carriers, the span of their
+differences and a complete sweep of the small supports a blind weighting could
+have, gating each quantity in place. Constitutional effect: none. This package
+changes no axiom, no framework Admissibility rule, no primitive, no policy,
+and no audit status, and it adds no import and no assumption to
+[MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The object is the unit four-cube on sixteen corners, cut into least-volume
+pieces at the adjacency cost floor. There are 15800 such cuttings. Between
+them they draw on 192 pieces, 24 pieces to a cutting, and each piece lies on
+1975 of the cuttings. Write the incidence table of that system: 15800 rows,
+one per cutting, 192 columns, one per piece, a one where the cutting uses the
+piece. Call a weighting of the 192 pieces blind when this table sends it to
+zero, so that every one of the 15800 cuttings, totalled piece by piece against
+the weighting, comes out zero.
+
+The preceding cycle measured the rank of the table as 88, so the blind space
+has dimension 104 and the two add back to 192, and then asked whether the
+group of 384 symmetries fixes 88 on its own. It does not: the seen and the
+blind space share same-type parts, so the trace counts of the group are too
+coarse to pick the blind space out. That cycle left the rank as a
+measurement and named the next place to look — the structure of the incidence
+map itself rather than the trace counts of the group.
+
+This cycle looks there, and finds the whole blind space sitting inside a
+family of objects the incidence table names by itself.
+
+## The result
+
+There are exactly 192 sets of eight pieces that no cutting uses twice. Each
+one of them meets every single one of the 15800 cuttings exactly once: each is
+an exact cover of the cutting system. Call them carriers. Two immediate
+consequences.
+
+**Every carrier difference is blind.** If two carriers both send every cutting
+to one, their difference sends every cutting to zero. So the differences of
+the 192 carriers all lie in the blind space, and nothing about symmetry is
+used to say so.
+
+**Those differences are the whole blind space, and this is exact.** The 192
+carriers, as weightings of the 192 pieces, span 105 dimensions. Their
+differences therefore span 104, one fewer, because no carrier lies in the span
+of the differences — a carrier sends every cutting to one, a difference sends
+every cutting to zero. And 104 is the dimension of the blind space. So the two
+coincide:
+
+> the space the cuttings cannot see is exactly the span of the differences of
+> the eight-piece exact covers of the cutting system, and the rank 88 is the
+> 192 coordinates less those 104.
+
+## The two-sided count that makes it exact
+
+The two numbers 105 and 88 are each measured in a bounded arithmetic, and a
+bounded rank is never larger than the exact rank. That inequality points the
+same way both times, which is what makes the argument close without any exact
+rational elimination at all:
+
+- the table has bounded rank 88, so its exact rank is at least 88, so the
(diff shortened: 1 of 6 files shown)


JONbridgerapps:
Review-loop disposition: **PASS; repaired content landed on `main`.**

Scope/provenance:

- Original PR head reviewed: `1a266ab800c10b96fdc65cd2b70b7dccd74ae725`.
- Exact repaired confirmation state: `c711a51c95617f572fcbd7ad8220586937f509b7` (source replay `80b81b4e8318348de8105ab5e2d4a92a174fadb4`, iteration 1 `abc3d988286629dcc9c76856a8d95a19dba685fc`, iteration 2 `c711a51c95617f572fcbd7ad8220586937f509b7`).
- One combined independent reviewer, GPT-5.6-Sol/max, covered every required lens. Initial result was `FIX_THEN_PROCEED` with C1-C12; the same reviewer independently rechecked all interacting fixes and returned `FINAL PASS`.
- Landing replay on then-current `origin/main`: `1d393f7a90`, `6b7556770d`, `473a540b93`; contained landing SHA `473a540b938ac4dcd685bbe4dbe1f9d896437a13`.
- Cycle 758 overlapped only the generated citation-graph manifest. The serialized graph build regenerated the combined acknowledgment (5,536 nodes / 15,738 edges), found the auto-merged node count stale by one, and the mechanical manifest correction landed and was containment-verified at `5f081b997f5eb682082a373e9c49a944bf80e14e`.

Narrow repairs closed C1-C12: fresh-main preservation; governed proposal metadata/machine trace; one 900-second timeout contract; fail-closed exits and mutations; removal of unbound raw/PASS artifacts; explicit supplied-fixture and Q-to-R scope; exact 48-element group naming; withdrawal of all minimum/lower-bound/route-foreclosure/exclusivity claims; implementation-disjoint exact helper and bound receipt; exact inverse/collision/face gates; portable RSS and complete-output accounting; and parity in both packet-helper registries. The landed result retains only the positive 384 support-eight witnesses and the exact characteristic-zero rank/span theorem.

Recovery-only mathematical record (not a landed claim): the reviewer also independently reconstructed the broader finite computation—768 maximal cliques, maximum clique size eight, 192 size-eight carriers, exhaustive support sweep through seven, and agreement of all four overlap-class ranks at 104. That broader minimum-support result was deliberately withdrawn from every landed surface rather than asserting a lower bound/no-go without its full governance packet.

Validation evidence:

- Two complete 18-stage pipelines passed before final confirmation.
- Strict audit lint passed with zero errors (baseline warnings/notices only); changed-audit evidence reported `checked=1 failures=0 control_failures=0`.
- Primary `10/10` and independent helper `12/12`; four primary mutations and one helper mutation each failed visibly and exited nonzero.
- Both runner caches were fresh; helper receipt bindings and hostile checks passed; the parity unit test and both helper consumers agreed exactly.
- `py_compile`, vocabulary (`0`), JSON/link/no-go/diff checks, exact graph extraction, and post-landing repo invariants passed. Validation-only generated outputs were excluded from the landed diff.

This review did **not** apply or imply an audit verdict. The bounded theorem remains proposed and requires the independent post-landing audit lane.

PR #6054, opened 2026-08-09
cycle 758: the 104 blind directions are the cover table rank less one

jonathonreilly:
## What this responds to

Three cycles have measured the blind space of the least-volume cuttings of the unit four-cube at 104 dimensions and tried to account for that number through the symmetries of the cube. Cycle 755 found the group's trace counts too coarse to fix it, cycle 756 found that each sharing class alone already spans all 104, and cycle 757 found that no symmetry orbit reaches across it (orbit difference ranks 23, 47, 23, 35, 29, all short of 104). The instrument was wrong. This cycle puts the question to the incidence structure instead, and reduces 104 to a single rank.

## What is new

The 192 eight-piece exact covers give a cover-by-piece table that is 8-regular on both sides. Measured exactly over the rationals:

- rank of the cover-by-piece table: **105**
- rank of its 191 cover differences: **104**
- nullity of the transposed table: **87**

Two derivations in the note carry the reduction, and they use **opposite regularities**.

**Rows bound the blind space from above.** The blind space is the image, under the transposed table, of the coefficient vectors that sum to zero. Row 8-regularity forces every kernel vector of that transposed table to sum to zero over the covers: adding the 192 piece entries of an annihilated combination gives 8 times the sum of its coefficients, so that sum vanishes. Hence the blind dimension is 191 less the nullity 87, which is 104, the rank less one. The runner gates the same fact in dual form by stacking one all-ones cover row onto the transposed table and finding the rank unmoved at 105.

**Columns pin it from below.** All 192 rows of the table add to 8 at every piece, so the all-ones piece vector lies in the row space; the cutting table sends that vector to the single value 24 at every one of the 15800 cuttings, so it is not blind. The row space is therefore the blind space plus that one direction, and 105 is 104 plus 1.

So **"why 104" is now "why is the rank of the cover table 105"** — one number about one table, in place of a dimension count about a 192-dimensional space.

## The rational spectrum, settled in full

Both sharing tables also carry rank 105, so the question can be put to a symmetric table. Three facts make a finite scan a complete test of which eigenvalues are rational: the cover-side table is an integer table times its transpose, so nothing is below 0; its largest row sum 64 bounds it above; and a rational eigenvalue of an integer symmetric table is an algebraic integer, hence a whole number. A fourth makes the scan cheap: rank over the prime field 1000003 never exceeds rank over the rationals, so that field rules candidates out and exact integer elimination confirms every survivor.

| eigenvalue | 0 | 2 | 4 | 8 | 10 | 12 | 16 | 20 | 24 | 64 |
|---|---|---|---|---|---|---|---|---|---|---|
| multiplicity | 87 | 8 | 8 | 3 | 8 | 6 | 2 | 10 | 3 | 1 |

That is 136 of the 192 eigenvalues counted with multiplicity; the other **56 are not rational**. Two totals confirm the list independently: the trace 1536 splits as 592 from the whole numbers plus 944, and the trace of the square 36096 splits as 12352 plus 23744, both leftovers strictly positive as eigenvalues that are not rational require. I re-derived 1536 and 592 and 36096 and 12352 by hand from cycle 757's landed sharing counts before reading the runner's numbers, and all four agree.

The piece-side table returns the same whole-number spectrum. That much is forced, since a table times its transpose and the transpose times the table carry the one nonzero spectrum and both are square of the same size here. Its interest is the contrast with cycle 757, which proved by a permutation-similarity argument that the two tables are **not** related by any relabelling. They carry one spectrum without being permutation-similar: **the spectrum is blind to exactly the asymmetry the sharing counts detect.**

## Where this leaves the question

Rank is 192 less the multiplicity of 0, and that multiplicity is 87 exactly, so the 56 eigenvalues that are not rational are all nonzero and play no part in the rank. The whole of the 104 now rests on one question: **why is 0 an eigenvalue of the sharing table exactly 87 times?**

## Runner

`scripts/physical_cell_cutting_cover_table_rank_cycle758_2026_08_09.py` — `TOTAL: PASS=15 FAIL=0`, 46 s, 186 MB, 3721 characters of output, deterministic across two runs. Its construction slice is byte-identical to cycle 757's runner, so the object is the same object.

Gates: C0-C2 rebuild the object and the 8-regularity; C3-C4 place the all-ones piece vector in the row space and show it is not blind; C5-C7 give the three ranks and the nullity, with C6 the dual form; C8 the three agreeing ranks; C9-C10 the two spectra; C11 the two totals; C12-C13 the controls; C14 the resource bounds.

The scan has both controls. **Positive:** the sixteen corners of the four-cube joined when they differ in one coordinate, whose spectrum (4 once, 2 four times, 0 six times, -2 four times, -4 once) was fixed before any measurement — the scan recovers it. **Negative:** the same table with 1 added at one symmetric pair; the gate requires only that the whole-number spectrum change and the count drop, hard-coding no answer, so a scan returning a canned result fails it.

## Boundary

Stated in the note: rank 105 is measured here, not derived; the two derivations are prose arguments over gated inputs, not machine-checked proofs; the cospectrality of the two sharing tables is forced and carries no new information on its own; the 56 eigenvalues that are not rational are unidentified; the scan settles the rational part and says nothing about the rest; and nothing here bears on any physical claim.

Adds no import and no assumption to `MINIMAL_AXIOMS_2026-06-29.md`. The note's one citation edge is to that file; every prior-cycle artifact is referred to in prose or backticks.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_COVER_TABLE_RANK_CYCLE758_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_COVER_TABLE_RANK_CYCLE758_NOTE_2026-08-09.md
@@ -1,0 +1,256 @@
+# Why the cuttings are blind in 104 directions: it is the rank of the cover table, less the all-ones direction — Cycle 758
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: computational identities of the finite cutting system
+
+Claim type: computational identities
+
+Runner:
+
+- [paired rebuild-and-gate runner](../scripts/physical_cell_cutting_cover_table_rank_cycle758_2026_08_09.py)
+
+Scope: computational identities of the finite cutting system. Every number
+below is machine-checked by the paired runner, which rebuilds the cell
+complex, the least-volume pieces, the cuttings at the adjacency cost floor,
+the eight-piece exact covers, the cover-by-piece table, its exact rational
+rank and nullity, the exact ranks of the two products of that table with its
+transpose, and the exact whole-number part of the spectrum of those products,
+gating each quantity in place. Two of the gates are controls whose job is to
+show the spectrum scan is neither vacuous nor canned. Constitutional effect:
+none. This package changes no axiom, no framework Admissibility rule, no
+primitive, no policy, and no audit status, and it adds no import and no
+assumption to [MINIMAL_AXIOMS_2026-06-29.md](MINIMAL_AXIOMS_2026-06-29.md).
+
+## What this answers
+
+The object is the unit four-cube on sixteen corners, cut into least-volume
+pieces at the adjacency cost floor. There are 15800 such cuttings; between
+them they draw on 192 pieces, 24 pieces to a cutting, and each piece lies on
+1975 cuttings. A vector on the pieces is *blind* to the cuttings when every
+cutting sums it to zero. Two cycles back the blind space was measured at
+exactly 104 dimensions and identified with the span of the differences of the
+192 eight-piece exact covers; since then, three separate attempts to explain
+that 104 through the symmetries of the cube have each fallen short, the last
+of them finding that no symmetry orbit reaches across it at all. The
+preceding cycle left the question open in its own words: why the covers
+differ inside exactly 104 dimensions.
+
+This cycle answers it, in the sense that a question is answered when it is
+reduced to a smaller one. **104 is not an independent number.** It is the
+rank of the 192 by 192 cover-by-piece table, less one, and the one is the
+all-ones direction on the pieces. Both steps are derived below from gated
+inputs, not measured; what is measured is the rank, 105, and every input the
+derivations use. The question "why 104" is now the question "why 105", asked
+of a 192 by 192 table instead of a 15800 by 192 one.
+
+The second half of the cycle takes the sharing table apart spectrally, since
+that is where a rank of 105 would have to come from. The whole-number part of
+the spectrum is settled exactly: ten eigenvalues with exact multiplicities,
+accounting for 136 of the 192, with the remaining 56 not rational.
+
+## The blind space is the cover table applied to the zero-sum covers
+
+Write M for the cover-by-piece table: 192 rows, one per exact cover, 192
+columns, one per piece, a one where the cover uses the piece. The runner
+gates that M is 8-regular on both sides, row sums and column sums alike (C2).
+
+**Derived here, from gated inputs.** Every cover meets every cutting exactly
+once (C1), so the difference of two covers is summed to zero by every cutting:
+the differences are blind. A general element of their span is a combination
+of covers whose coefficients add to zero, so the span is exactly the image of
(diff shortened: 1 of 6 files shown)


JONbridgerapps:
Review-loop provenance — landed as a narrow self-contained salvage.

Disposition: PASS WITH BOUNDED CLAIMS after 2 reviewer iterations using the same GPT-5.6-Sol reviewer at maximum reasoning.

Iteration 1: FAIL with 9 consolidated findings. The blockers were the unsupported equality between the full cutting-blind kernel and the 104-dimensional cover-difference image, plus invalid evidence/input-contract handling. Additional findings covered unlanded-predecessor and broad spectral/no-go language, hidden supplied finite-object choices, missing machine-status/trace/import fields, the factor-of-eight C3 prose error, timeout/RSS mismatch, overstated trace confirmation, and an ungated rounded inverse.

Fixes: retained only the finite exact-algebra core B=M^T(H) subset ker(A), dim(B)=104, row(M)=B direct-sum the all-ones direction, rank(M)=rank(S)=rank(N)=105, and the exact rational Gram spectrum for the supplied labelled construction. Full-kernel equality, predecessor authority, non-permutation claims, and broad spectral/no-go claims were removed. Supplied choices and the primitive-registry result are explicit. The runner now inventories zero repository-science reads, gates exact two-sided inverse identities, states the C3 factor correctly, and uses a consistent 600-second portable resource contract.

Iteration 2 focused confirmation: PASS. All 9 findings CLOSED; no new findings. Lens outcomes: Code PASS; Physics BOUNDED/PASS; Proof CLOSED for the narrowed target; Imports DISCLOSED; Nature BOUNDED; Labeling PASS; Governance PASS; audit compatibility source-PASS; No-Go and Methodology not applicable.

Independent and mechanical checks:
- paired runner: TOTAL PASS=15 FAIL=0 (38.9 s targeted cache run; 41.18 s reviewer run)
- independent modular ranks at 101, 1009, 10007: rank(M,D,S,N)=(105,104,105,105)
- independent exact characteristic polynomial reproduced all rational factors and a degree-56 irrational remainder
- mutations rejected a changed inverse, changed C3 incidence, and timeout boundary violations
- exact incidence checks: A M^T all ones, A D^T zero, both M sums 8, and A times all-ones equal 24
- vocabulary lint, Python compilation, portable-link scan, and diff checks passed
- one full audit pipeline passed; strict lint exited 0 with baseline warnings/notices only
- temporary canonical cache passed changed-evidence readiness with forensic_evidence_ready=true
- all generated validation outputs and the cache were stripped afterward; only the required citation manifest landed
- citation topology: +1 node, +0 edges, with no existing node changed

Landed commits on current main:
- 2aff990514 — source salvage
- 3ed1371f36 — review fixes
- 37d77613a4 — citation-topology acknowledgment and landing tip

Remote containment verified: 37d77613a4372c8d733aab240a7d094ceb22942f is contained in origin/main, and the reviewed note, runner, and manifest object hashes match exactly.

No audit verdict was authored or applied. The new bounded-theorem row remains for the independent audit lane.

PR #6055, opened 2026-08-09
feat(cell-cutting): the smallest cover exchanges generate the blind lattice over the whole numbers

jonathonreilly:
## What this responds to

The previous cycle measured the cover-by-piece table of the finite cutting system: rank 105, with the differences of exact covers spanning a 104-dimensional space no cutting can see. Spanning over the rationals is a weak statement. It leaves open whether an integer labelling of the 192 pieces splits into an integer part the cuttings measure and an integer part they cannot see, and whether the two ranks survive at every prime. This package answers both over the whole numbers.

## What the runner measures

**The smallest exchange is a four-for-four swap.** Two exact covers share 0, 1, 2 or 4 pieces, in counts 157, 20, 10, 4 per cover. Largest sharing gives smallest difference, so the smallest cover difference moves 8 pieces. There are 384 such pairs, entries in -1, 0, 1, and the cutting table sends every one of them to zero.

**They generate, not merely span.** The 384 smallest exchanges carry a 104 by 104 minor of determinant one. Every invariant factor is one, so their row lattice is saturated: every integer vector the cuttings cannot see is an integer combination of smallest exchanges. What the cuttings cannot see is generated over the whole numbers by the smallest thing they cannot see. A single cover together with its 191 differences to the rest already gives an integer basis of the same lattice.

**Both ranks hold in every characteristic.** An 88 by 88 minor of determinant one in the cutting table and a 105 by 105 minor of determinant one in the cover table force rank 88 and rank 105 at every prime, with torsion-free quotients of rank 104 and 87. Direct prime-field measurement over 13 primes agrees, and a fixed control drops to 0, 1 and 3 where the witness predicts it should.

**The counterweight, reported in full.** The measured lattice and the blind lattice are orthogonal complements over the rationals and each is saturated, yet they do not glue: the index is 2^114 x 3 x 5^30, agreeing across three matrices of three different sizes and from both choices of basis. The quotient group has 42 cyclic parts of orders 2:12 10:8 20:2 40:2 80:8 160:8 320:1 960:1, exponent 960, which is 5 times the 192 pieces. So an integer piece labelling does not split into integer parts, and 960 is the smallest whole number that forces it to.

## Runner

`scripts/physical_cell_cutting_blind_lattice_generation_cycle759_2026_08_09.py`, 19 gates, TOTAL: PASS=19 FAIL=0, 11.1 s, peak memory about 315 MB, stdout 4572 characters. Re-run by hand after the cache was written; the fresh cold output is byte-identical to the cached one.

Four gates carry controls, so a report of "one" is information rather than a default. A determinant of 144 is recovered from a diagonal whose invariant factors are not all one; the invariant-factor routine returns the true chain for handed-in diagonals whose chains differ from their entries; the gluing measurement returns index 2 for a two-dimensional case known by hand; and the prime-field routine is checked against a second scalar implementation.

## Boundary

The gluing index and the group behind it are measured, not derived. Nothing here says the 384 smallest exchanges are the only vectors of support 8 in the blind lattice, nor that no blind vector has smaller support. The object is a finite combinatorial one and every statement about it is a count. This package changes no axiom, no primitive, no framework rule and no policy, and it adds no import.

## Next

Why the sharing counts take the values they do, and how the 88 measured directions distribute across the 192 pieces.


JONbridgerapps:
Review-loop result: **PASS** after one combined review pass and one focused repair/confirmation iteration.

Reviewer provenance: GPT-5.6-Sol, maximum reasoning. One combined reviewer covered CodeRunner, claim/scope/assumption/physics/equation/symbolic/numerical/dependency/artifact/methodology/governance/primitive/no-go/vocabulary/labeling/audit-compatibility/generated-artifact/self-containment lenses; the same session performed focused confirmation on the repaired five-file delta.

Eight blocking findings were resolved:

1. bound C0 to the complete reconstruction, exact inverse, exact action, and collision/boundary invariants;
2. replaced aggregate C3 division with rowwise sharing-profile checks;
3. replaced modulo-prime row selections with four unit-minor-certified integer bases;
4. narrowed negative splitting rhetoric to positive finite quotient/index/exponent identities;
5. made the packet self-contained with explicit inputs, primitive-registry N/A, exact target, proof ledger, and finite boundary;
6. aligned C18 with the 300-second contract and portable RSS units;
7. corrected generator-family versus basis wording in C9/C10; and
8. added canonical discoverability, retained only the canonical cache, and acknowledged citation topology deterministically.

Validation: primary runner and source-pinned cache `PASS=19 FAIL=0`; `py_compile`; independent 192-row sharing census, SymPy rank/unit-minor/determinant checks, separate local p-adic Smith computation, and 676 random/edge arithmetic comparisons; 8/8 load-bearing mutation families rejected; 34 relevant existing tests passed; full pipeline through 18b passed on confirmation base `5f081b997f`; strict lint had zero errors; changed-audit evidence checked one row with zero failures; vocabulary, portable-link, diff, graph, and generated-output-stripping gates passed. The generated row is `bounded_theorem`, `unaudited`, with `deps=[]`; no audit verdict was issued or applied.

Durable content landed directly to `main` at `07610d4b977820eb7a0f5c97123300a8acc81bbf`. Remote containment and the five landed paths were verified before this comment.

PR #6061, opened 2026-08-09
[review-loop] audit-infra: repair dependency-policy helper registry epoch gate

JONbridgerapps:
## Outcome

Repairs the live dependency-policy epoch invariant that prevents a clean full
audit pipeline on `main` from reaching/restoring science fingerprints.

`main`'s controlled manifest records an old raw hash for
`build_citation_graph.py`. Subsequent additions to the claim-scoped
`EXPLICIT_PACKET_HELPER_RUNNER_PATHS` registry changed that raw hash, so the
full pipeline fails closed when restoration constructs a fingerprint. A no-op
scan can appear healthy because every committed live science row is currently
unaudited and does not need an epoch comparison.

## Reviewed design

- Keep every dependency-policy source governed.
- For `build_citation_graph.py`, exclude only the contents of the one
  module-level, single-target, strictly literal `{str: [str, ...]}` helper
  registry assignment.
- Bind the assignment's exact byte position into a domain-separated normalized
  digest, so moving it across executable code is still a policy change.
- Locate the span through the Python AST with byte offsets that correctly model
  LF, CRLF, and lone-CR newlines.
- Fail closed on missing, duplicated, executable, malformed, or otherwise
  rebound registry forms.
- Continue binding each registered helper's path, exact source bytes, and
  declared inputs into the affected claim's own `science_fingerprint_v2`.
- Use one `dependency_policy_source_sha256()` producer for both comparison and
  controlled-manifest refresh.

The controlled manifest advances to `dependency_policy_v3`.

## Review-loop corrections

Independent review initially returned `FIX_THEN_PROCEED` for two P1 hash
collisions:

1. LF-only line accounting could consume a governed byte after a lone CR.
2. Deleting the registry span without a location marker allowed relocation of
   the assignment to preserve the normalized hash while changing execution
   order.

Both were fixed in `4357c44991`, with exact mixed/all-CR and relocation
regressions. Independent re-review returned **PASS with no remaining findings**.

## Final landing surface

Exactly five files:

- controlled dependency-policy manifest;
- fingerprint implementation;
- fingerprint and pipeline fixture tests;
- active-review-queue resolution record.

The original branch's generated ledger, queue, status, residue, and restoration
outputs were removed. No audit verdict or retained-grade decision is authored.

## Verification

- Full `run_pipeline.sh`: exit 0 through all 18 stages and checkpoint 18b;
  stage 7 scanned 4,784 rows with 0 hard resets and completed restoration.
- Fingerprint suite: 32/32 pass.
- Direct controlled epoch comparison: pass.
- Changed-audit-evidence control: `checked=0 failures=0 control_failures=0`.
- `py_compile` and `git diff --check`: pass.
- Broad audit-pipeline unit suite: 484/491 pass; all seven failures reproduce on
  unmodified pre-PR `main` (legacy framework-epoch short circuit / audit-runner
  lock baseline), with no PR-specific regression.

The restore-policy redesign discussed on the earlier working branch remains
deferred and is not part of this PR.


jonathonreilly:
Review-loop landed the independently approved five-file delta on main as 39c74017b870c27c804e3992f2a11e90336476b2. Generated ledger/status outputs were stripped; no audit verdicts were applied. Closing rather than merging because the reviewed content is already present on main.

PR #6060, opened 2026-08-09
science(cell-cutting): certify finite orbit-table rank witnesses

jonathonreilly:
## Salvaged scope

This revision retains the independently reproduced finite construction and
withdraws the former physical/covariance interpretation. It is self-contained
on current `main` and consumes zero authority from Cycle 762, PR #6059, or any
other unlanded stack.

## Exact finite result

The runner rebuilds the labelled unit four-cube object from scratch:

- 2,672 unit-determinant candidate simplices, with 400 at adjacency-cost floor 6;
- 15,800 certified cuttings on 192 supported pieces;
- 192 eight-piece covers;
- the 384-element coordinate-relabeling action; and
- 96 cover-piece cell orbits.

Every cell-orbit indicator decomposes into 48 four-cycles and therefore has
exact rank 144. The reconstructed cover table is a union of four orbit
indicators, has row and column sum 8, and has exact integer rank 105. The
explicit orbit union `(6, 11, 28, 39)` has the same binary row-and-column shape
and exact integer rank 144. These two matrices discharge the bounded positive
existence target.

## Scope and imports

The theorem ranges over this declared finite incidence object. Physical and
multicell interpretations are separate targets. The note inventories every
model and computational choice, records an empty primitive-premise dependency
set, and includes an acyclic proof-obligation graph. The one-prime sampled-rank
distribution from the submitted version has been removed; modular values that
remain are explicitly labelled diagnostics, while both advertised witness
ranks are checked by integer elimination.

The commutant statement is also narrowed to its computed content: exactly two
of 104 canonical orbital basis matrices individually preserve each row space.
It makes no assertion about the dimension of the preserving subspace of
arbitrary linear combinations.

## Runner and review repair

Primary runner:
`scripts/physical_cell_cutting_finite_orbit_ranks_cycle763_2026_08_09.py`

The repaired source declares a 600-second audit budget, uses platform-correct
resident-memory units with synthetic Linux/Darwin controls, and exits nonzero
whenever any gate fails. A live run gives `TOTAL: PASS=47 FAIL=0`. A deliberate
G0 predicate mutation gives `TOTAL: PASS=46 FAIL=1` and process exit 1.

The combined review independently reproduced the finite enumeration, group
orders, exact primary ranks, cycle decomposition, rank-144 witness, and
citation topology. The renamed note and runner now carry only the bounded
finite theorem and its diagnostics.

PR #6064, opened 2026-08-10
feat(cell-cutting): exact ceiling 144 and floor 48 from the cutting cell's symmetry

jonathonreilly:
## What this responds to

Cycle 763 ([#6060](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6060)) asked whether the cutting cell's own symmetry picks out the cover table's dimension of 105, and answered no: covariant tables of that shape run over a wide band and 105 sits low in it. But the band was only sampled — 1500 of the 3.3 million four-subsets — so nothing said whether the largest dimension seen, 144, was the true ceiling or an artefact of sampling. The hand bound derived there fell short of 144 and could not settle it.

This PR replaces the sample with a derivation.

## What changed

One note, one paired runner, one cache, plus a separate `audit-infra:` commit adding the new note to the citation manifest (one node, out-degree 1).

**The decomposition.** The algebra of maps commuting with the symmetry on the 192 piece directions has dimension 104, one basis element per orbit on ordered pairs of pieces. Its centre is the kernel of 10816 commutator conditions: a modular selection of 84 independent rows caps it at dimension 20, and 20 integer vectors each verified against all 10816 conditions in both multiplication orders floor it at 20. So the centre has dimension exactly 20. One central element separates all twenty parts — its matrix on the centre has 20 distinct integer eigenvalues each of nullity one, found by an exact integer scan over the interval its own row sums allow. The parts have dimensions 1, 1, 1, 1, 3, 3, 3, 3, 4, 4, 8, 8, 8, 8, 12, 12, 24, 24, 32, 32.

**The theorem.** A table from covers to pieces that commutes with the symmetry acts on each part as one small matrix tensored with an identity. Its rank is therefore a sum of twenty tiny ranks, each at most `min(m, mc)`. Hence for every covariant table

```
rank  at most   sum of d * min(m, mc)      = 144
blind at least  sum of d * max(0, m - mc)  =  48
```

**The ceiling is attained**, not merely bounded: an explicit equivariant integer matrix built from all 96 cell orbits has exact rational rank 144, computed over the rationals with no modulus anywhere. So 144 is the true greatest available to any covariant table, and 48 the true least blind space. Cycle 763's sampled largest becomes a derived and attained value, and a rank question on a 192 by 192 table becomes twenty rank questions on matrices of size at most 6 by 4.

**What the cover table does with the room.** Its own rank is 105 and its blind space 87 — strictly inside the derived interval, 39 below the ceiling and 39 above the floor. Twelve parts sit exactly at the forced value; eight exceed it, with excesses 2, 4, 6, 8, 6, 6, 6 and 1 summing to the whole 39.

Six of those eight have more cover-side room than piece-side room and **two have less**. So the natural reading — that the cover side offers room the table declines to use — is wrong as a general statement, and the note says so.

**Three candidates for the 39, all refuted.** The even coordinate permutations form a subgroup of order 192 and index 2 splitting both the pieces and the covers into classes of 96 and 96. That class label is itself one of the twenty parts, and it is blind for a reason the run exhibits: every cover splits its eight pieces 4 and 4 between the classes, the same holds dually on the piece side, and the single non-identity element fixing a cover keeps the class. On that same element the flip-count parity and its product with the permutation sign both act by -1, which forces those two alternative labels blind as well. The aggregate incidence carries nothing by arithmetic alone. And two geometric labels read off the pieces each agree with the class label on exactly half of the 192 pieces; the determinant label is genuinely covariant but tracks the product character, matching on all 73728 element-piece pairs with 0 misses under source ordering and failing on exactly half of them under re-sorted ordering.

## Why the modular step is sound

Every per-part quantity is a rank computed modulo one prime, where the modular value can differ from the exact one in a single direction only. Six sums pin the whole table to independently computed exact totals: sum of dimensions 192, sum of m squared 104, sum of m times mc 96, sum of mc squared 120, sum of d times mc 192, sum of blind 87. Equality of every sum therefore forces every individual per-part entry to be exact. Three of those totals — 96, 120, 192 — come from orbit counts that were never used to build the decomposition, so it is cross-checked against objects outside itself.

## Runner

`physical_cell_cutting_symmetry_ceiling_cycle764_2026_08_09.py`, 28 gates, `TOTAL: PASS=28 FAIL=0`, about 165 seconds, 125 MB peak resident, 3431 characters of stdout. Every gate is exact integer arithmetic; no floating point enters any gate. Re-run independently before landing; every science line is byte-identical to the cache, the two runs differing only in the wall-time and resident-memory line.

Controls carried inside the run: 40 equivariance checks on the orbit matrices; the cover table rebuilt entry by entry as a sum of whole cell orbits; 7 products of orbit matrices recomputed directly against the structure constants; an integer overflow guard on those products; three products on the full 192 square pinning the abstract central matrix to the concrete action; and the witness rank computed over the rationals.

## Honest boundary

- Two identities in the run hold by arithmetic whatever the part table says — ceiling plus floor is 192, and ceiling minus rank equals blind minus floor. The note names them as bookkeeping, not evidence. The content is that the measured 105 is **strictly** below 144 and the measured 87 **strictly** above 48, and that the ceiling is attained.
- Likewise, that every part with cover-side multiplicity 0 has no excess is arithmetic; it is reported as a check that the cover table annihilates those five parts, not as support.
- The weights inside the central element and the witness are arbitrary deterministic genericity devices with no meaning. Each is gated by a wrong-value rejector — twenty distinct integer eigenvalues each of nullity one, and an exact rational rank of exactly 144 — so a non-generic choice would fail its gate rather than pass quietly.
- The symmetry used is the full symmetry of the four-cube, larger than the proper cubic rotations the admissibility axiom names. A ceiling from a larger group is still a ceiling, but the smaller group would give a weaker one; nothing here should be read as a statement about the axiom's own covariance, and the note says this explicitly.
- **The 39 is not explained.** Three candidate labels are refuted; the residual stands.

## Order

No dependency on [#6060](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6060) or any other open PR — this note cites only its paired runner and the axioms file, and the runner recomputes the object from scratch. It can land in any order.

## What it opens

The residual is now a finite and small question: name the eight small matrices whose rank drops make up the 39. The same reduction also makes an exact census over all 3321960 covariant tables of the cover table's shape tractable, replacing cycle 763's 1500-point sample with the full distribution.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SYMMETRY_CEILING_CYCLE764_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SYMMETRY_CEILING_CYCLE764_NOTE_2026-08-09.md
@@ -1,0 +1,209 @@
+# The cell symmetry caps the cover table's rank at 144 and floors its blind space at 48 — Cycle 764
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: derived ceiling and floor with a measured, unaccounted residual
+
+Claim type: bounded_theorem
+
+Runner:
+
+- [`physical_cell_cutting_symmetry_ceiling_cycle764_2026_08_09.py`](../scripts/physical_cell_cutting_symmetry_ceiling_cycle764_2026_08_09.py)
+
+Axioms:
+
+- [`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md)
+
+Constitutional effect: none. This note changes no axiom, primitive, registry,
+policy, audit verdict, effective status, or framework claim.
+
+## What this responds to
+
+Cycle 763 asked whether the cell's own symmetry picks out the cover table's rank
+of 105, and answered no: tables built to be covariant under that symmetry run
+over a wide band and 105 sits low in it. But the band itself was only sampled.
+The largest dimension seen was 144 and the bound derived by hand fell short of
+it, so nothing said whether 144 was the true ceiling or an artefact of sampling.
+
+This cycle replaces the sample with a derivation. The symmetry alone fixes an
+exact ceiling on the rank of any covariant table, an exact floor on its blind
+space, and the ceiling is shown to be attained.
+
+## The object
+
+The unit four-cube on sixteen corners, cut into least-volume pieces at the
+adjacency-cost floor: 2672 candidate pieces of determinant one, 400 of them at
+cost floor 6, 15800 cuttings of 24 pieces, 192 pieces actually used and 192
+eight-piece covers.
+
+The symmetry permutes the four coordinates and flips any of them: 384 maps,
+closed over all 147456 products, acting transitively on the 192 pieces and on
+the 192 covers. It has 104 orbits on ordered pairs of pieces, 120 on ordered
+pairs of covers and 96 on the cells of the cover-by-piece square. By exact
+rational arithmetic the cutting table has rank 88 with kernel 104, and the cover
+table has rank 105 with kernel 87.
+
+## Twenty parts
+
+The algebra of maps commuting with the symmetry on the 192 piece directions has
+dimension 104, one basis element per orbit on ordered pairs. Its centre is the
+kernel of the 10816 conditions saying that a combination commutes with every
+basis element. A modular selection of 84 independent rows caps that centre at
+dimension 20, and the 20 integer vectors recovered are each verified against all
+10816 conditions in both multiplication orders, which floors it at 20. So the
+centre has dimension exactly 20.
+
+One central element separates all twenty. Its matrix on the centre has 20
+distinct integer eigenvalues, each of nullity one, found by an exact integer scan
+over the interval its own row sums allow, bound 6009; three direct products on
+the full 192 square confirm that this matrix really is multiplication by that
+element. Its eigenspaces on the 192 piece directions are the twenty parts, of
+dimensions
+
+    1, 1, 1, 1, 3, 3, 3, 3, 4, 4, 8, 8, 8, 8, 12, 12, 24, 24, 32, 32
+
+Each part carries three integers: a degree d, a piece-side multiplicity m and a
+cover-side multiplicity mc. Six sums certify the whole table at once:
+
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
## Review-loop provenance

- Reviewer seat: **GPT-5.6-Sol / max reasoning** (single combined all-lens reviewer; same-session focused confirmation)
- Review iterations: **2**
- Original reviewed head: `96e1c1f55ba4b58a751db153c081de06a746fcad`
- Findings: **9 total, 9 resolved, 0 skipped** — 1 critical, 5 high, 2 medium, 1 low (`R6064-001` through `R6064-009`)
- Final focused verdict: **PASS**; claim-strength disposition **BOUNDED**

Narrow repairs added a source-declared/reproducible timeout, corrected the piece-to-cover map direction, supplied the imports and seven-obligation rational/mod-p proof bridge, narrowed negative rhetoric to positive finite comparisons, bound construction controls and exact diagnostics into gates, removed the false axiom dependency, made gate failures exit nonzero, and scoped C27 as a resource-envelope check. The canonical cache is SHA-bound, fresh, and records `TOTAL: PASS=28 FAIL=0`. No audit verdict or effective-status mutation was authored or applied.

Validation passed: independent determinant/enumeration checks; independent two-prime rank checks and exact part sums; combined 28-family mutation (`PASS=0 FAIL=28`, exit 1); Python compilation; vocabulary fix/lint; forensic trigger checks; portable links; citation graph reproduction; full audit pipeline; strict audit lint (pre-existing notices only); changed-evidence against the integration base and current `origin/main`; generated/audit-surface stripping; and diff cleanliness.

The reviewed note, runner, and cache landed byte-for-byte. Current-main integration regenerated only the required citation-manifest topology acknowledgment (`5542` nodes, `15741` edges; new node out-degree `0`). The prescribed first landing loop exhausted fail-closed on two sequential manifest-only conflicts with no push; the consolidated generated-only integration then passed the exact loop.

Landed directly to `main` as `e9472c2522e35941e57aa8b8beb1ec2bc40c0f4f`; explicit refetch and ancestor containment verification passed. Independent post-landing audit remains required before any retained-grade status.

PR #6068, opened 2026-08-10
feat(cell-cutting): cycle 765 twenty small matrices carry every covariant table, and the 39 resolves into eight named rank drops

jonathonreilly:
## What this responds to

Cycle 764 derived a ceiling of 144 on the rank of any table covariant under the
cell's own symmetry and a floor of 48 on its blind space, and showed the ceiling
attained by an exact rational witness. The cover table sits at rank 105 with blind
space 87, leaving 39 of the available rank unused. Cycle 764 ended by naming the
next question: which small matrices lose that rank, and by how much.

This PR answers it and then measures the entire band the ceiling bounds.

## The reduction

The twenty parts of cycle 764 are eigenspaces of one central element on the 192
piece directions, each carrying a degree d, a piece-side multiplicity m and a
cover-side multiplicity mc. A covariant table therefore acts on each part as the
identity on the degree-d space times a single mc by m matrix, so its rank is the
sum of d times the rank of that small matrix.

The runner builds those matrices rather than assuming them: restricted to a part,
a deterministic small-integer combination of the commuting basis has a minimal
polynomial of degree at most m, whose roots are found by scanning every residue of
the prime with nested evaluation, and a root whose eigenspace has dimension exactly
d yields a pure-tensor basis. All 20 of 20 parts pass all five construction
conditions; five of the twenty matrices are empty because those parts have no
cover-side copy.

The correspondence runs both ways: the 96 cell-orbit coefficients stack to a matrix
of width 96 whose rank is 96, so a covariant table is exactly one list of twenty
small matrices and every list is a covariant table. The reduction matches the
direct rank on 62 of 62 tables and returns 105 with blind space 87 on the cover
table itself.

## Where the 39 goes

Eight small matrices are rank-deficient. As degree, piece-side multiplicity,
cover-side multiplicity, rank and rank forgone:

    2/2/2/1/2   4/2/3/1/4   6/2/3/1/6   8/4/6/3/8
    6/4/3/2/6   6/2/3/1/6   6/4/3/2/6   1/1/1/0/1

The drops sum to exactly 39, and these are precisely cycle 764's eight excess
parts reached by an independent route. On all twenty parts the blind space equals
d times m minus the rank, so the cover table's blind space of 87 is now accounted
for part by part rather than in aggregate.

## The band

With the reduction in hand the runner censuses every four-subset of the 96 cell
orbits, all 3321960 of them rather than a sample: 79 distinct ranks, least 24,
greatest 144 meeting the derived ceiling, with 511872 attaining it. At or below the
cover table's rank of 105 sit 106536 four-subsets, of which 24768 equal 105 and
81768 fall below, with 72 at the least. Far more four-subsets reach the ceiling
than sit at or below 105, so the cover table is low in the band rather than
typical of it.

Three points are pinned by exact rational arithmetic with no modulus: 0/2/12/14 has
rank 24, 0/1/2/7 has rank 144, and the cover table's own 0/1/2/3 has rank 105, each
agreeing with its census value. Independently, all twenty small matrices are rebuilt
from scratch at a second prime and reproduce the census rank on 200000 of 200000
sampled four-subsets.

## Runner

`physical_cell_cutting_small_matrix_census_cycle765_2026_08_09.py`, 41 gates,
`TOTAL: PASS=41 FAIL=0`, re-run by hand before landing. Every gate number is an
exact integer computation; no floating point enters any gate. The run finishes well
inside its budget of 900 seconds of wall time and 2500 MB of peak resident memory.

Discriminating controls: the stacked coefficient matrix having full rank 96; the
reduction matched against the direct rank on 62 of 62 tables; a closed-form blind
count of 1272384 matched against the brute count from the census; the whole
construction repeated at a second prime; exact rational ranks at three named
four-subsets; and the requirement that every part's construction come out at
dimension exactly d, so a non-generic choice fails its gate rather than passing
quietly.

## Boundary, stated in the note

Census ranks are modular and can only fall below the true rank. The greatest, 144,
meets the derived ceiling, so the maximum is exact; the least, 24, is confirmed by
an exact rational rank, so the minimum is exact. The counts at or below 105 are
upper bounds and 511872 is a lower bound, and the 79 distinct ranks is a count of
distinct modular values, not a bound in either direction.

Two lines are named as bookkeeping rather than evidence: the part table rebuilt at
the prime cycle 764 itself used, and the at-or-below line, which partitions a census
checked elsewhere rather than testing it.

The cover table's 105 remains measured, not derived. What changes is the shape of
the residual: the 39 is no longer one unexplained number but eight named drops on
eight named parts, and the census says a covariant table at or below 105 is
uncommon while one at the ceiling is not. The symmetry used is the full symmetry of
the four-cube, larger than the proper cubic rotations the admissibility axiom names,
so nothing here is a statement about the axiom's own covariance.

The next step this opens: 0/1/2/7 reaches the ceiling while the cover table's
0/1/2/3 gives 105, sharing three of their four cell orbits and differing in one.
Naming the linear relations that the swap creates or destroys is the concrete
handle the earlier cycles lacked.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SMALL_MATRIX_CENSUS_CYCLE765_NOTE_2026-08-09.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SMALL_MATRIX_CENSUS_CYCLE765_NOTE_2026-08-09.md
@@ -1,0 +1,197 @@
+# Twenty small matrices carry every covariant table, and the 39 resolves into eight named rank drops — Cycle 765
+
+Date: 2026-08-09
+
+Authority: none
+
+Audit: unset.
+
+Status: derived reduction, complete census of the covariant band, and the residual named part by part
+
+Claim type: bounded_theorem
+
+Runner:
+
+- [`physical_cell_cutting_small_matrix_census_cycle765_2026_08_09.py`](../scripts/physical_cell_cutting_small_matrix_census_cycle765_2026_08_09.py)
+
+Axioms:
+
+- [`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md)
+
+Constitutional effect: none. This note changes no axiom, primitive, registry,
+policy, audit verdict, effective status, or framework claim.
+
+## What this responds to
+
+Cycle 764 derived a ceiling of 144 on the rank of any table covariant under the
+cell's own symmetry, a floor of 48 on its blind space, and showed the ceiling
+attained by an exact rational witness. The cover table sits at rank 105 with
+blind space 87, so 39 of the available rank goes unused. Cycle 764 could say
+which twenty parts the room is distributed over, but not where the 39 goes, and
+it ended by naming that as a finite and small question: which small matrices lose
+rank, and by how much.
+
+This cycle answers that question, and then measures the whole band the ceiling
+bounds.
+
+## The object
+
+The unit four-cube on sixteen corners, cut into least-volume pieces at the
+adjacency-cost floor: 2672 candidate pieces of determinant one, 400 of them at
+cost floor 6, 15800 cuttings of 24 pieces, 192 pieces actually used and 192
+eight-piece covers.
+
+The symmetry permutes the four coordinates and flips any of them: 384 maps,
+closed over all 147456 products, acting transitively on the 192 pieces and on the
+192 covers. It has 104 orbits on ordered pairs of pieces, 120 on ordered pairs of
+covers and 96 on the cells of the cover-by-piece square. By exact rational
+arithmetic the cutting table has rank 88 with kernel 104, and the cover table has
+rank 105 with kernel 87.
+
+## The reduction
+
+The twenty parts of cycle 764 are the eigenspaces of one central element acting
+on the 192 piece directions. Each part carries a degree d, a piece-side
+multiplicity m and a cover-side multiplicity mc. Written out, the piece side is a
+sum over parts of a degree-d space times an m-dimensional multiplicity space, and
+the cover side is the same sum with mc in place of m. A covariant table therefore
+acts on each part as the identity on the degree-d space times a single mc by m
+matrix. Its rank is the sum over parts of d times the rank of that small matrix,
+and its blind space the sum of d times m minus that rank.
+
+The runner builds those matrices rather than assuming them. Restricted to a part,
+the commuting basis acts through m by m matrices; a deterministic small-integer
+combination of them has a minimal polynomial of degree at most m, whose roots are
+found by scanning every residue of the prime 1000003 with nested evaluation. A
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Review-loop provenance (GPT-5.6-Sol/max; one combined reviewer, resumed for focused confirmation)

Original PR head: `05b7ec8416c90fd709d5d7aa1aaf37366429affb`
Original merge base against the landed history: `323d7fc32d77598f74ea6cd4d30c38dda0fe5070`

Review iterations:
1. Combined all-lens review: `FINAL VERDICT: FIX_THEN_PROCEED (C1-C12)`. The repair made the finite cell-cutting packet self-contained from closed/unlanded ancestry, restored fail-closed runner preflight and exact mutation pins, bound the rational rank/span and finite census claims to independently checked evidence, corrected status/trace/import/proof-DAG surfaces, removed unsupported broader claim language, and added the canonical harness/citation-graph acknowledgment.
2. Same-reviewer focused confirmation: `FINAL VERDICT: FIX_THEN_PROCEED (C13)`. C13 found residual explanatory symmetry non-forcing wording in the runner source. The narrow fix rewrote only the docstring/headings to positive invariant-rank/drop statements and regenerated the canonical cache.
3. Same-reviewer resumed confirmation on frozen candidate `d2cbf3c21ac17324cef64062f92dfe9f3ca1a969` over `e9472c2522e35941e57aa8b8beb1ec2bc40c0f4f`: `FINAL VERDICT: PASS`; C13 closed, 0 new and 0 remaining actionable findings. Code/runner, physics boundary, proof obligations, imports/support, hostile retention, repo governance, and audit compatibility passed; the negative-claim and labeling gates were not applicable after the wording repair.

Validation and landing gates:
- live runner and canonical cache: `TOTAL: PASS=43 FAIL=0`, exit 0, cache fresh and SHA/input-bound;
- independent rank/decomposition/census recomputation plus hostile mutations of the fail-closed gate families;
- `py_compile`, vocabulary lint (0 violations), portable-link/forbidden-verdict scans, and merge-base `git diff --check`;
- full audit pipeline PASS after integration with current main; strict lint: 0 errors (pre-existing warnings/notices only); changed-evidence: `checked=1 failures=0 control_failures=0`;
- current-tree serialized graph regeneration: 5,544 nodes / 15,743 edges; validation-only generated audit/effective-status outputs stripped before landing;
- no audit verdict, audit-status promotion, or effective-status claim was authored or applied. Independent audit remains the ratification lane.

The frozen four-commit repair was integrated onto current main with only the generated citation-manifest overlap, regenerated before push, and landed through the exact fail-closed retry/containment loop on attempt 1.

Landed main SHA: `795e851254e689a66fa9e3fe619823835d4d8661`
Remote containment: verified. The landed five-path content is byte-identical to the validated candidate.

Disposition: PASS; durable bounded finite-theorem content landed. Closing this superseded PR and deleting its unique head branch.

PR #6069, opened 2026-08-10
[physics-loop][review-loop] inter-site-gate blockIS2 — Cycle 972: bounded theorem (the covariant neighbour-dependence law)

JONbridgerapps:
## Result

Characterizes the bounded covariant neighbour-dependence law in the landed Cycle-719 basis-state substrate. The declared family is exactly 20 distinct length-zero/one words on a target-centred seven-site nearest-neighbour star. Six incoming CNOT words are witnesses with exact law `y = x XOR n_d`; the other 14 words are target-neighbour-independent.

The witness family is covariant under `Z^3 semidirect O_cubic^+`: 24 proper rotations and six translation generators transport every witness with zero failures. There is one induced word-law class and two fixed-input state-resolved row orbits. Uniform target-input averaging gives 0/20 marginal-dependent words by the exact XOR permutation identity.

## Artifacts

- [bounded theorem note](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/physics-loop/inter-site-gate-blockIS2-20260809/docs/COVARIANT_DEPENDENCE_LAW_CYCLE972_BOUNDED_THEOREM_NOTE_2026-08-09.md) (includes trace gate, status certificate, exact family/caps/horizon, and remaining boundary)
- [primary runner](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/physics-loop/inter-site-gate-blockIS2-20260809/scripts/frontier_cycle972_covariant_dependence_law_2026_08_09.py)
- [independent refuter](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/physics-loop/inter-site-gate-blockIS2-20260809/scripts/frontier_cycle972_law_independent_check_2026_08_09.py)
- Pinned caches under `logs/runner-cache/`; machine-readable receipts under `outputs/`

## Verification

- Primary: `TOTAL: PASS=5 FAIL=0` (20 words; 6 witnesses; 0/768 XOR failures; 0/768 control failures; 0/2,560 landed-coordinate bridge failures; 0/61,440 rotation failures; 0/15,360 translation failures)
- Independent checker: `TOTAL: PASS=6 FAIL=0`, including an active XOR-to-XNOR corruption probe
- `python3 -m py_compile` on both runners
- `scripts/vocab_lint.py --fix`: 0 violations
- `audit_lint.py --strict`: 0 errors
- `git diff --check`: clean
[...]

PR #6065, opened 2026-08-10
[physics-loop][review-loop][bounded-support] Admissibility global measure / menu-kernel type separation

jonathonreilly:
## Result

This PR is now based on `main`; its parent (#6063) is already landed.

At one fixed `M_2(C)` site and nearest-neighbor condition, the current Admissibility probability measure on the full possibility space is not automatically the normalized effect grade consumed by the parent binary/ternary frame-lift theorem.

1. Raw singleton identification fails exactly: disjoint `z` and `x` binary projective menus would each have mass one, forcing mass two inside a probability space of mass one.
2. Normalized finite restriction is not universal: an explicit full-support atomless Gaussian family allowed by the current wording makes every finite menu a zero-measure event, hence `0/0`.
3. Where finite restriction is defined, an exact five-atom witness assigns the shared ternary-menu effect `25/142` in one context and `2/11` in another, differing by `-9/1562`.

The broad claim that the four axioms cannot derive Born probabilities is explicitly rejected. Apparatus/menu-conditioned measure, barycenter/evaluation, Record event-algebra, contact-compiler, and owner-approved axiom routes remain live.

## Conditional composition

The sufficient construction in the note is now deliberately an **abstract menu kernel and effect grade**. Composed with the already-landed low-arity frame theorem, it conditionally supplies a unique local density matrix and trace grade. Interpreting that kernel as a physical Record outcome still requires a separate bridge showing that the menu/context label is encoded in Record content, because current Record readout is content-only.

No canonical axiom file is edited. The construction is not claimed necessary, minimal, adopted, or derived from current Record dynamics.

## Review-loop repairs

- Corrected the machine-readable claim type to `bounded_theorem`.
- Removed three accidental, non-load-bearing citation-graph edges; dependencies are now exactly the parent theorem and `minimal_axioms`.
- Narrowed the physical claim to the proved abstract kernel and made the content-only Record bridge explicit.
- Replaced noncanonical proof-registry strength/status labels with governed enums.

## Verification

- Direct exact runner/cache: `TOTAL: PASS=20 FAIL=0`, including all five forensic N5 resolution lines.
- Independent symbolic/manual checks: both menu sums and ranks, Gaussian normalization in `R^8`, seven distinct centers, `Z=509/200`, `25/142`, `2/11`, and `-9/1562`.
- Full 18-stage audit/integration pipeline on current `main` `ef97fa9c10a6aedfd366c475f42e90f0c1fd36b7`: pass through enforced repository invariants and skill-baseline freshness.
[...]

PR #6063, opened 2026-08-10
[physics-loop] Born block 1 — binary/ternary scaled menus force trace form

jonathonreilly:
## Result

At one `M_2(C)` site, a menu-independent grading on scaled rank-one effects
and scalar coins, normalized on every binary and ternary nonzero resolution of
the identity, has a unique density-matrix trace form. No four-outcome menu is
used.

The new mechanism is a one-ancilla lift: every orthonormal basis of `C^3`
compresses onto `C^2` as either a ternary scaled-projector menu or a binary
menu plus the zero effect. The resulting nonnegative weight-one frame function
is quadratic by the explicitly named dimension-three frame theorem; zero
ancilla expectation then deletes the ancilla block.

## Exact status

- This closes the landed “prove ternary scaled-projector sufficiency or find a
  rogue” frontier at the repository's existing named-theorem granularity.
- It advances, but does not close, the campaign's deliberately stronger
  self-contained/no-literature-import target. A native no-regularity proof of
  the special compression-invariant frame lemma remains open.
- It does not derive physical Born probabilities from the four axioms. The
  remaining physical bridge has two exposed clauses: distribution-to-effect-
  grade identification/functionality, and universal binary/ternary scaled-menu
  eligibility/normalization.
- No canonical axiom, primitive, registry, policy, queue, or audit verdict is
  edited.

## Hypothetical axiom-side fallback

The source maps—but does not recommend or adopt—the following sufficient
wording:

> For each site and nearest-neighbor condition, the Admissibility distribution
> extends to a menu-independent grading `w` on the one-site scaled-projector
> domain. Every binary and ternary nonzero scaled-projector resolution of the
> identity is an eligible local menu and its grades sum to one.

N1–N8 is committed in the source note. Its disposition is PASS only for the
narrow factual statement that these clauses are sufficient and are not
literally stated in the current axiom memo; it is FAIL / DO NOT SHIP for “a new
axiom is required” or any claim that constructive Record routes are exhausted.

## Landing surfaces

- [theorem note](docs/BORN_FORM_FROM_BINARY_TERNARY_SCALED_PROJECTOR_FRAME_LIFT_BOUNDED_THEOREM_NOTE_2026-08-09.md)
- [exact runner](scripts/born_form_binary_ternary_scaled_projector_frame_lift_2026_08_09.py)
- [SHA/input-bound cache](logs/runner-cache/born_form_binary_ternary_scaled_projector_frame_lift_2026_08_09.txt)
[...]

PR #6070, opened 2026-08-10
[physics-loop] Fixed-condition Gaussian density-operator / Record-decoder compiler

jonathonreilly:
## What this block advances

This is TOE campaign Block 4, stacked on #6067. It takes Block 3's explicit full-support Gaussian Admissibility law and constructs the whole fixed-condition mathematical compiler rather than merely postulating its decoder:

- `E[A A^dagger] = C^2 + 2I`, hence the normalized second moment is a positive trace-one density-operator candidate;
- `Phi(Re Tr(A-C))` is uniform at fixed `C`, so cumulative effect weights define an explicit Borel program-relative content decoder for every supplied finite ordered resolution;
- at `C=P_z`, the two exact hostile ternary menus have weights `(3/10,19/50,8/25)` and `(3/10,7/20,7/20)`, so their shared effect descends to `3/10` inside the compiler.

The positive result is bounded to the displayed Gaussian family, one fixed mathematical condition, and supplied ordered programs. It does not identify the extracted density operator as the physical state, encode a physical apparatus, establish one preparation across program-dependent neighbor conditions, cause Records, or derive realized-history/frequency semantics.

## Exact residuals

The source's N1-N8 pass collapses the remaining local walls to three:

1. physical density-operator/compiler selection;
2. atom splitting or a stochastic resource outside the atomless Gaussian completion;
3. physical preparation/program registration and quotient, including cross-condition equivalence and occurrence.

An isotropic extractor family preserves the displayed compiler while changing the shared-effect grade from `3/10` to `2/7`, an exact `1/70` bounded selection control. A covariant singleton-support law cannot deterministically realize a half-half target, isolating the splitting interface. Neither is a global no-go.

No canonical axiom or approved primitive is edited. The checkpoint-zero TOE percentages remain unchanged because no physical or autonomous current-axiom obligation is retired.

## Evidence

- primary runner/cache: `TOTAL: PASS=29 FAIL=0`, SHA/input-fresh, 4,655 stdout characters;
- independent SymPy route: both density operators, both exact menu sums and weight vectors, lambda-one vectors, and the `1/70` gap agree;
- 18 process-local load-bearing mutation probes reach their intended named failure families;
- vocabulary lint: zero findings;
- strict audit lint: zero errors;
- graph/link invariants: zero link or class-F violations; intended graph delta acknowledged at 5,450 nodes / 15,502 edges, new note out-degree five;
- changed-audit-evidence: 25 changed stacked paths, zero failures and zero control failures;
- Python compile, YAML parse, local-link check, cached-runner freshness, and cold diff: pass.

The full audit pipeline passes through stale-audit invalidation and then reproduces the current-main dependency-policy epoch mismatch owned by #6061: the manifest pins `build_citation_graph.py` at `20698263cbf6ad82569f15ceee0642525f18c9c093d329be1149253217d33415`, while the current file hashes to `c58180b642cbc7714a270c54258ab4a6c3588a83df54d573170152f7a1a4ceb1`. This branch does not copy or modify that policy repair. Pipeline cleanup restored 1,048 tracked generated changes and removed 816 verified ledger shards.

## Stack and collision boundary

Hard landing order: #6063, then #6065, then #6067, then this PR. The exact base is #6067 head `bcd2edf4dad188554e9341f412639916efc8a953`.

The final open-PR sweep found two later adjacent submissions:

- #6068 is a four-file cell-cutting small-matrix/rank census and changes no axiom, primitive, Admissibility measure, Record decoder, effect quotient, or compiler-selection surface.
- #6069 classifies the finite basis-state length-zero/one word family and finds six covariant incoming-CNOT/XOR witnesses. Its own boundary explicitly leaves the continuous `M_2(C)` Admissibility law open, so it is a next-block contact seed rather than a collision with this compiler.

Independent audit remains required before any effective status change. `review-loop` was not invoked in authoring or self-reviewing this block.


jonathonreilly:
Review-loop disposition: PASS after current-main salvage. The submitted stack base (#6067) was closed without merge, so I replayed only the canonical note/runner/cache, removed the retired Record scalar/finite-additivity premise, and retyped the program-dependent quantile fibers as mathematical event partitions rather than direct content-only Record readout. The Gaussian claims were independently rederived; a separate 500,000-sample check confirmed the matrix moment and intrinsic CDF uniformizer, while omitted-centering and variance mutations failed. The repaired primary runner passes 30/0. The exact candidate completed all 18 pipeline stages, strict lint with zero errors, changed-audit-evidence 1/1 with zero failures/controls, and acknowledged graph topology at 5,523 nodes / 15,706 edges. Landed directly to main at 5ad583dee2a66fae31fab70ca5913f639a68a012. Immediate child #6071 was retargeted to main before branch cleanup; this PR is now superseded by the reviewed main content.

PR #6097, opened 2026-08-10
science(gravity): prove additive-zero contact ambiguity

jonathonreilly:
## Scope

Stacks exactly on Block 24 delivery head `298d50332e` / PR #6095.

This bounded axiom-interface packet proves the exact normalized-family gauge

```text
S_i(g) -> S_i(g) + F(g)
```

for a common geometry-dependent shift. Every normalized configuration
probability is unchanged, while `log Z` and its geometry Hessian shift by
`-F` and `-F''`. A common quadratic realizes any symmetric local Hessian; a
Hermitian six-mode tensor is the corresponding symmetric form on twelve real
coordinates.

For each reconstructed Block-23 source matrix `M_s`, the packet constructs the
explicit fifteen-edge shift `D_s=-U M_s U^dagger`. It cancels the complete
six-mode tensor, annihilates the inherited four gauge columns, and enters at
first source order.

## Evidence

- primary source/input-pinned cache: `PASS=13 FAIL=0` in `14.35 s`
- stdout: 3,701 characters
- maximum quotient cancellation residual: `2.037e-15`
- maximum inherited-gauge residual: `4.580e-16`
- maximum rescaled first-order residual: `4.945e-17`
- 13 isolated runner-copy mutations each fail exactly one intended named check
- citation graph: 5,471 nodes / 15,620 edges; new node out-degree three
- direct vocabulary, strict lint, premise/model-family, Python/YAML,
  whitespace, cache, canonical-nonmutation, and graph/link/class-F gates pass
- exact stack seeds 4,809 rows and classifies 3,687 runners before reproducing
  only the inherited stage-7 dependency-policy epoch mismatch
- controlled five-file current-main overlay passes all 18 stages and finalizes
  checkpoint `b7f009c19958`
- changed-evidence readiness: Block 25 `1/1`, full stack `25/25`, zero failures
  and zero control failures

## Boundaries and axiom consequence

The explicit completion is local to one flat chart and supplied momentum
direction. It does not select a local covariant action, prove cellwise gluing
or global integrability, define one joint source coordinate, or establish
continuous-zone or Lorentzian nonlinear closure.

The exact missing interface is a registered action representative: the
physical law must fix an allowed local covariant counterterm/equivalence
class, action unit, and source/constraint transformations before geometry
variation. Common shifts are physically equivalent only if their complete
geometry responses are equivalent or derived to vanish. Normalized Record
probabilities alone cannot supply this authority.

This is not an action-selection, locality, covariance, gravity, or axiom
[...]

PR #6096, opened 2026-08-10
science(gravity): bound proper-length source contact

jonathonreilly:
Summary:
- reconstruct the Block-23 six-mode carrier inside the affine metric tangent
- derive the proper-length source contact on all fifteen actual edge classes
- prove the retained 2/3/4-edge supports cannot cancel rank-six source coefficients
- prove coordinate-only source seagulls cancel the geometry connection at full stationarity
- construct a common six-edge algebraic escape, preserving wider local-action and dynamical-source routes
- sharpen the unadopted joint source/geometry law interface without editing canonical axioms

Verification:
- primary source/input-pinned runner: PASS=16 FAIL=0
- fresh-process mutation harness: PASS=16 FAIL=0, one intended failed check per mutation
- citation graph: 5,470 nodes / 15,617 edges; new out-degree 4 with 12 helper runners
- direct vocabulary, lint, premise, normalization, cache, YAML, link, class-F, and invariant gates pass
- exact stack reproduces only the inherited dependency-policy epoch mismatch after 4,808 ledger rows and 3,686 runners
- five-file current-main overlay passes all 18 stages at checkpoint 6252de92588d
- changed evidence: 1/1 Block 24 and 24/24 stacked claims ready

Boundary:
This is a support-confined contact and coordinate-rewrite boundary, not a gravity no-go. It does not select a wider local source law, extend the stationary identity to the partial-stationary reacted roots, prove nonuniform or continuous-momentum closure, establish Lorentzian nonlinear stability, move the fixed TOE rubric, edit canonical axioms, or apply an audit verdict. No review-loop was used.

PR #6098, opened 2026-08-10
science(gravity): anchor null-record action representative

jonathonreilly:
## Result

Block 25 showed that normalized probabilities leave an arbitrary geometry-dependent common action shift. This block proves the minimal exact conditional repair: for any finite strictly positive registered family with distinguished null configuration,

`A_x(g) = -log[pi_x(g)/pi_null(g)]`

is the unique dimensionless matter/source action representative satisfying `A_null(g)=0`.

The theorem reconstructs the normalized family, removes every common shift, fixes relative first/second geometry derivatives, and composes additively for factorized independent families. The retained nonuniform coframe cut family realizes the anchor exactly across all 256 binary configurations tested. Block-25 common completions are therefore reclassified as new joint interactions, not anchored normalization equivalences.

## Boundary

This does not derive or adopt physical gravity. Current Record wording supplies `I(empty)=0` only for readout; it does not register the positive joint family, identify readout with action, fix the action unit, or select the separate source-independent pure-geometry action/causal update. Local/projective consistency, full stationary nonuniform Ward response, continuous momentum, and Lorentzian nonlinear stability remain open. Canonical axioms and fixed TOE percentages are unchanged.

## Verification

- primary source/input-pinned runner: `PASS=14 FAIL=0` in `14.4 s`
- 14 isolated fresh-process mutations: each fails exactly one intended check
- citation graph: `5,472` nodes / `15,624` edges; out-degree 4; 14 helpers
- direct vocabulary, premise, model, syntax/YAML, cache, strict-lint, claim-typing, link/invariant, and canonical-nonmutation gates pass
- exact stack reaches only the inherited stage-7 policy-epoch mismatch after `4,810` rows / `3,688` runners
- controlled five-file current-main overlay passes all 18 stages at checkpoint `016787d4c504`
- changed evidence: Block 26 `1/1`; full stack `26/26`; zero failures/control failures

Stacked on PR #6097. Complementary proper-length PR #6096 is not included and requires explicit downstream reconciliation. Independent audit remains required. `review-loop` was not used.


JONbridgerapps:
Consolidated to the archive: ledger id pr-6098 (archive/ledger/98/pr-6098.json). Carrier: #6100 restates and strengthens the uniqueness (total null-relative action fixed; only the geometry/matter split is gauge); ancestry of #7753. Branch retained; science line in the ledger entry.

PR #6099, opened 2026-08-10
science(gravity): compose null anchor with RN sources

jonathonreilly:
## Result

This block composes the Block 26 null-record action anchor with finite positive Radon-Nikodym source interventions.

- Proves the exact null-relative action cocycle and exponential-normalizer cancellation.
- Separates positive Fisher source-scale selection from orientation and the dimensionful action quantum.
- Exhausts all 256 retained cut configurations and shows the retained topological source has exactly zero coframe contact across 1,024 geometry-path checks.
- Constructs a cubic-invariant anchored interaction that cancels all three supplied Block 23 contact matrices below 2.037e-15 while changing probability odds, proving there is no algebraic gravity no-go.
- Narrows the live repair to a physically selected local covariant geometry-dependent RN score plus a separate source-independent pure-geometry action or causal update.

## Boundaries

The constructive cancellation is target-tailored and nonlocal in geometry. This does not claim a physical gravity law, canonical axiom closure, continuum or Lorentzian closure, a TOE percentage move, or an audit verdict. Independent audit remains required.

## Verification

- primary cache: PASS=17 FAIL=0
- isolated mutations: 17/17 each fail exactly one named check
- citation graph: 5,473 nodes / 15,629 edges; out-degree 5; 15 resolved helpers
- exact stack: only inherited stage-7 dependency-policy epoch mismatch after 4,811 rows / 3,689 runners
- controlled five-file current-main overlay: all 18 stages pass; checkpoint 87d3c253d997
- changed evidence: 1/1 block claim and 27/27 stacked claims clean

Stacked exactly on Block 26 delivery head 8281f0a0ea. Complementary proper-length sibling PR #6096 requires later explicit reconciliation.


JONbridgerapps:
Consolidated to the archive: ledger id pr-6099 (archive/ledger/99/pr-6099.json). Carrier: #6100 supersedes the tailored nonlocal cancellation with a shared covariant null-anchored source-linear RN contact; ancestry of #7753. Branch retained; science line in the ledger entry.

PR #6117, opened 2026-08-11
feat(cell-cutting): complement spaces and partner pairs (cycle 768)

jonathonreilly:
## What this responds to

Cycle 767 measured that a set of cell orbits and its complement have equal rank. This cycle asks what stands behind the equality and gets two exact statements, both proved from the structure of the orbit tables rather than read off a computation.

## Theorem A — complements share the space, not just its size

Every cell-orbit table is a zero-one matrix with exactly two entries in each row and each column, and the ninety-six tables sum to the all-ones matrix. So the table of any set S of orbits has all row and column sums equal to twice the size of S. If a vector is blind to that table, then twice-the-size times its coordinate sum vanishes, hence the sum vanishes, hence the all-ones matrix annihilates it, hence it is blind to the complement's table as well. The argument is symmetric and runs equally on transposes, so the two tables have the *same* blind space and the *same* seen space, not merely spaces of equal dimension. Equality of rank is a corollary.

The runner checks this at sixteen sets spread across the whole range of sizes, exactly over the integers, with a wrong-value rejector; the two tables differ at every one of the sixteen.

## Theorem B — the ninety-six orbits are one cover's partner pairs

The symmetry group has order 384 and is transitive on the 192 covers, so a single cover's stabiliser has order two, and its non-identity element moves every piece. Every cell orbit meets that cover's row in exactly one partner pair, which gives a bijection from the ninety-six cell orbits onto the ninety-six partner pairs of pieces. Under it, the four incidence orbits are exactly the labels carried by that one cover's own eight blocks, two blocks to a label.

That turns a global orbit count into a local statement about a single cover, and it names the four incidence orbits geometrically for the first time in this lane.

## Two complete censuses

- All 4560 pairs of orbits: thirteen distinct ranks from 48 to 144, with 960 pairs at the ceiling and 1104 at or below the four-orbit rank of 105.
- All 142880 triples: eighteen distinct ranks from 64 to 144, with 60960 at the ceiling.

Ranks in both censuses are computed at a prime, so a rank can only be under-reported: the counts at the ceiling are lower bounds and the counts at or below 105 are upper bounds. Twenty-five pairs and fifteen triples were re-ranked on the full 192 by 192 tables as a check, and the whole pair census was repeated at a second prime with nothing off.

By Theorem A each census carries across to sizes 94 and 93 unchanged. Together with the size-one spectrum derived at cycle 763 and the complete four-subset census of cycle 765, the four smallest and the four largest set sizes now all have complete spectra.

## Two predictions failed, and are reported as measured

The design predicted a least pair rank of 72 and a least triple rank of 114. The measurements are 48 and 64. Both predictions were taken out of the runner rather than weakened, and the note reports the measured values.

## Runner

`TOTAL: PASS=68 FAIL=0`, 5889 characters of output, about 240 s wall, about 190 MB peak resident. The new size-one gate was checked for discrimination on a perturbed copy of the object: swapping a single label in one cover's row makes the cycle structure irregular and pushes two orbit ranks above the ceiling, and the gate fails.

## Boundary

The set sizes between the four smallest and the four largest are untouched, and the number of subsets there is far past what a complete census can reach. Why the rank ceiling and the blind floor should add to the piece count is not derived here.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_COMPLEMENT_SPACES_CYCLE768_NOTE_2026-08-11.md
+++ b/docs/PHYSICAL_CELL_CUTTING_COMPLEMENT_SPACES_CYCLE768_NOTE_2026-08-11.md
@@ -1,0 +1,271 @@
+# Complementary sets of cell orbits are blind to the same directions, and the 96 cell orbits are the partner pairs of a single cover — Cycle 768
+
+Date: 2026-08-11
+
+Authority: none
+
+Audit: unset.
+
+Status: two derived structure theorems on the cell-orbit lattice, with the
+complete rank spectrum at the four smallest set sizes and, by the first
+theorem, at the four largest
+
+Claim type: bounded_theorem
+
+Runner:
+
+- [`physical_cell_cutting_complement_spaces_cycle768_2026_08_11.py`](../scripts/physical_cell_cutting_complement_spaces_cycle768_2026_08_11.py)
+
+Axioms:
+
+- [`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md)
+
+Constitutional effect: none. This note changes no axiom, primitive, registry,
+policy, audit verdict, effective status, or framework claim.
+
+## 1. What this responds to
+
+Cycles 764 to 766 mapped the rank band of the cover table: a ceiling of 144, a
+floor of 48 on the blind space, a measured rank of 105 with an 87-dimensional
+blind space, and therefore an excess of 39 that is pure cancellation, born on
+pairs of cell orbits rather than on any single one. Cycle 767 added that a set
+of cell orbits and its complement carry equal rank, and cut the 96 orbits into
+25 corner strata.
+
+Two questions were left standing. First, whether that equality of rank is an
+accident of counting — two subspaces that happen to have the same dimension —
+or an equality of the subspaces themselves. Second, what the 96 cell orbits
+actually are as objects, rather than as indices into a table.
+
+Both are answered here, and neither answer needed a new axiom or a new input.
+The runner prints every check as one of a list of computational identities,
+with no floating point entering any gate.
+
+## 2. The setting
+
+The object is the one inherited from cycle 764 and unchanged since. Inside the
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
`/review-loop` completed and the reviewed science was landed directly to `main` at `7ec6a3865a4b23af21f1561a19e2946e17d0c960`.

Review disposition:

- Original hostile review: `FIX_THEN_PROCEED (C1-C18)`.
- Repair confirmation: the bounded finite mathematics passed; three remaining conformance gaps were identified and fixed (governed trace enums, a landed exact-rank/kernel mutation family, and a landed drop-block-identity mutation family).
- Same GPT-5.6-Sol/max focused re-review of only the three iteration-2 files: `FINAL CONFIRMATION: PASS`; no new blocker.

Science boundary retained:

- self-contained complement-kernel/image theorem over `Q`;
- all-cover stabilizer partner-pair theorem;
- explicitly modular exhaustive pair/triple census over `F_1000003`, with the pair vector independently replicated over `F_1000033`;
- physical/Record/continuum/multicell bridges and a rational full-census interpretation remain open;
- no no-go conclusion and no imported scientific dependency (`deps=[]`).

Evidence and validation:

- content-pinned live runner: source/cache SHA `74380ce5092163ec039fdb7aa9e53e0a6c3be69539c0cee4734a470db45509db`, exit 0, 192.02 seconds under the declared/resolved/cache timeout of 900 seconds;
- sequential C0-C68, `TOTAL: PASS=69 FAIL=0`, with `13/13` scratch mutation families rejecting;
- independent raw finite checks reproduced all 15,800 cuttings, 96 two-regular orbit tables summing to `J`, all 192 order-two cover stabilizers and partner fibres, and the pinned pair/triple vector invariants;
- full audit-compatibility pipeline reached stage 18b PASS;
- strict audit lint: zero errors; changed-science evidence readiness: 1 checked, 0 failures, 0 control failures; vocabulary lint and `git diff --check`: clean;
- generated audit/ledger/effective-status outputs were stripped. No audit verdict or status promotion was authored or landed; the new bounded-theorem row remains for the independent audit lane.

Landing containment was exactly the note, runner, source-bound cache, canonical harness row, and generated citation-graph manifest acknowledgment. Remote containment was verified after push.

PR #6119, opened 2026-08-11
feat(cell-cutting): fixed-point counts force the cell-orbit count and a positive blind floor (cycle 769)

jonathonreilly:
## What this responds to

Cycle 764 landed the twenty-row part table of the four-cube cell object and read two
numbers off it: a covariant-rank ceiling of 144 and a blind-dimension floor of 48. Cycle
768 established that the group acts freely on the 36864 cover-by-piece cells and used
that freeness to match the 96 cell orbits with one cover's 96 partner pairs. Both cycles
had the freeness and the two endpoint numbers as measured facts. This cycle derives them
from counting.

## The statement

Let a group act transitively on a set X whose point stabiliser has order two, generated
by a flip s, and let Y be any second set the group acts on. Every orbit on X x Y has a
representative whose first entry is a fixed base point, so orbits on X x Y are exactly
the orbits of that two-element stabiliser on Y. A two-element group has orbits of size
one at its fixed points and size two elsewhere, so that number is
(|Y| + fix_Y(s)) / 2. Y need not be transitive.

The cell object has 192 pieces and 192 covers, each in a single orbit, so both point
stabilisers have order 384 / 192 = 2. Applying the statement four ways turns the three
landed orbit counts into four fixed-point numbers:

| sum over parts | value | what it forces |
| --- | --- | --- |
| m dot m | 104 | the piece flip leaves 16 pieces alone |
| mc dot mc | 120 | the cover flip leaves 48 covers alone |
| m dot mc | 96 | the cover flip leaves no piece alone |
| m dot mc | 96 | the piece flip leaves no cover alone |

All four are measured directly on the permutations and come out 16, 48, 0, 0.

## Three consequences

**The free action is derived.** A zero says the cover flip lies in no piece stabiliser,
so the two stabilisers meet in the identity alone and no non-identity element fixes a
cell. Hence 36864 / 384 = 96 = 384 / (2 x 2). Cycle 768 had this as measurement.

**The blind floor is positive for a reason.** If every part had equal piece and cover
multiplicities, the first and third sums would agree, that is 104 would equal 96. The
difference, the sum of m times (m minus mc), is 8, so at least one part carries more
piece copies than cover copies; ten do. The floor is the dimension-weighted sum of those
positive differences, so it cannot vanish.

**Ceiling plus floor is the piece count identically.** min(x, y) + max(0, x - y) = x for
any two numbers, so summing against the part dimensions gives ceiling plus floor equal
to the sum of d times m, which is 192. That 144 + 48 = 192 needed no measurement. The
runner checks the identity on each of the twenty rows separately, not merely on the
totals.

## Runner

35 gates, TOTAL: PASS=35 FAIL=0, 107 s, 126 MB, 3736 characters of output. It is
standalone: it rebuilds the 2672 candidate pieces, the 400 at adjacency cost 6, the
15800 cuttings of 24, the 192 pieces used, the 192 covers and the group of order 384
from nothing, and takes no landed number as an input.

Discrimination: each of the three orbit counts is obtained three independent ways
(direct enumeration, a walk from two generators alone, and fixed-point averaging over
the group) before being compared to the fixed-point prediction. Twelve perturbed copies
of the action, six on each side, each break at least three of the four counts while the
true action breaks none. Eight altered fixed-point tuples and forty altered part-table
rows are rejected.

## Boundary

The counting statement needs the point stabiliser to have order exactly two, which holds
here only because the group has order 384 and is transitive on 192 points; a larger
stabiliser would break the statement as written, and the runner gates the order. The
part table is built at one prime, mitigated because its three multiplicity sums are
gated against prime-free orbit counts. The excess 39 = 144 - 105 is untouched: this
cycle fixes the band and its two endpoints, not where inside the band the cover table
sits.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_FLIP_COUNTS_CYCLE769_NOTE_2026-08-11.md
+++ b/docs/PHYSICAL_CELL_CUTTING_FLIP_COUNTS_CYCLE769_NOTE_2026-08-11.md
@@ -1,0 +1,259 @@
+# Four fixed-point counts force the cell-orbit count and a positive blind floor — Cycle 769
+
+Date: 2026-08-11
+
+Authority: none
+
+Audit: unset.
+
+Status: Four fixed-point counts, one per stabiliser and target set, reproduce all three orbit counts of the
+cell object; the zero among them forces the group to act freely on the 36864 cells, which cycle 768 took as
+measured input, and the gap between 104 and 96 forces the blind floor positive and makes ceiling plus floor
+equal 192.
+
+Claim type: bounded_theorem
+
+Runner:
+
+- [`physical_cell_cutting_flip_counts_cycle769_2026_08_11.py`](../scripts/physical_cell_cutting_flip_counts_cycle769_2026_08_11.py)
+
+Axioms:
+
+- [`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md)
+
+Constitutional effect: none. This note changes no axiom, primitive, registry,
+policy, audit verdict, effective status, or framework claim.
+
+## 1. What this responds to
+
+The object is the unit four-cube with its 16 corners. Among the corner simplices of
+unit absolute determinant there are 2672 candidates; the adjacency cost floor is 6 and
+400 candidates attain it; those 400 admit 15800 cuttings of 24 pieces each; 192 of the
+pieces actually occur, and there are 192 covers, a cover being 8 pieces that are
+pairwise never in a common cutting and that together meet every cutting exactly once.
+Each piece lies in 1975 cuttings and 8 times 1975 is 15800. The symmetry group is the
+group of coordinate permutations together with independent flips of the axes; it has
+order 384 and is transitive both on the 192 pieces and on the 192 covers.
+
+Three cycles left three loose ends here.
+
+Cycle 764 landed a ceiling of 144 and a blind floor of 48 and never said why the two
+add to the number of pieces. Cycle 768 built the 20-row part table from which those two
+numbers are read, and used the fact that the group acts freely on the cover-by-piece
+cells; it took that freeness as an input rather than producing it. Cycle 768 also found,
+separately, that one cell orbit on its own has rank 144 and blind dimension 48.
+
+This cycle supplies the missing arguments from four counts that had never been measured:
+how many pieces and how many covers are left in place by the non-trivial element of one
+piece stabiliser, and the same two counts for one cover stabiliser.
+
+## 2. The counting statement and its proof, in words
+
+Let a finite group act transitively on a set X in such a way that the subgroup fixing a
+chosen point has exactly 2 elements, the identity and one further element s, which is
+therefore its own inverse. Let Y be any other set the same group acts on. Then the
+number of orbits the group has on the set of ordered pairs, first entry from X and
+second from Y, is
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
## Review-loop landing record

**Disposition:** PASS after two review iterations; reviewed science landed on `main` at `cdb355be2f1373ca97a6d7245c419e1402979d65`.

**Reviewer provenance**

- Combined reviewer: GPT-5.6-Sol, maximum reasoning.
- Persistent reviewer session: `01a00f82-3750-7771-ad6b-978b18776acf`.
- Original PR merge base: `ea0968c7`; original PR contained two unique commits and no stack.
- Reviewed correction head: `9a6b3488199760807804ac0cf2ca51d9a55968b1`.
- Iteration 1: `FIX_THEN_PROCEED`, 11 findings (7 blocking, 2 major, 2 minor).
- Iteration 2 focused confirmation: all 11 findings resolved, no new findings, final `VERDICT: PASS`.

**Narrow science repairs**

- Replaced campaign naming with the domain-explicit four-cube fixed-point/orbit/floor identity.
- Scoped the theorem to exact finite geometry and group action, with the full numerical part table explicitly bounded to `F_1000003`; `F_1000033` is only the one-matrix comparator.
- Added the exact target, O1-O9 obligation graph, boundary cases, and explicit open characteristic-zero full-table lift.
- Added a rational exact geometry gate over all 15,168 co-occurring simplex pairs compressed to 59 checked symmetry classes, including the sample-grid false-negative witness with exact margin `1/11`.
- Added direct corner-set fixed-set recomputation plus fail-closed mutations for geometry, action, freeness, modular table, and floor identity.
- Added `AUDIT_TIMEOUT_SEC = 300`, regenerated the canonical cache, and removed the unused axiom dependency and author-owned audit wording.

**Independent science confirmation**

- Live runner: `TOTAL: PASS=41 FAIL=0`; cache fresh with declared/resolved timeout 300.
- Separate reconstruction reproduced 2,672 unit simplices, 400 cost-floor candidates, 15,800 cuttings, 192 pieces, 192 covers, 15,168 co-pairs, and 59 pair orbits; an independent LP check found no positive-margin selected representative.
- Fixed sets `(16,0,0,48)` reproduce orbit counts `(104,120,96)` and cell freeness.
- Literal table arithmetic reproduced dimensions 192/192, multiplicity products 104/120/96, gap 8, ten positive rows, ceiling 144, and floor 48.
- Independent exact-rational rank of the single C30 matrix is 144 with nullity 48; no full-table characteristic-zero lift is claimed.

**Landing evidence**

- Full landing-tree audit pipeline: PASS; graph delta acknowledged at 5,548 nodes and 15,746 edges.
- Strict audit lint: exit 0, no errors (existing global warnings/notices only).
- Changed-evidence check: one row checked, no failures or control failures, forensic evidence ready.
- Vocabulary lint, Python compile, cache identity, and `git diff --check`: PASS.
- Manifest-only integration conflicts were resolved solely by current-tree regeneration; the three reviewed science blobs on `main` exactly match reviewed head `9a6b348819`.
- `cdb355be2f1373ca97a6d7245c419e1402979d65` is durably contained in `origin/main`.

**Audit boundary**

The proposed claim ID is `four_cube_cutting_fixed_point_orbit_floor_note_2026-08-11` and still requires independent post-landing audit. Review-loop applied no audit verdict, no `audited_clean` status, and no effective-status change.

PR #6121, opened 2026-08-11
[physics-loop][review-loop] inter-site-gate blockIS8 — Cycle 986: bounded theorem (patch-uniformity induction)

JONbridgerapps:
## Result

Closes the star-gluing induction at the declared finite-patch scope. For every finite `n >= 2` and every ordered distinct target tuple rooted at an adjacent P2x pair, with each later closed star meeting prior support, the same relative 23-program chart is uniform on the finite support union under the separately declared chart-role record convention. This does not claim one infinite-lattice allocation or simultaneous global schedule.

## Evidence

- Primary reconstructs all 24 overlap offsets: adjacent/axial-two/diagonal-two = 6/6/12.
- Mechanical extensions pass for `k=2 -> 3` (support 17) and `k=3 -> 4` (support 20), including multiply shared site cocycles.
- Refutation checker independently derives route, overlap, base, case, and induction records; rejects 25/25 mutations and accepts coherent adverse outcomes through outcome-neutral bookkeeping.
- Cycle-719 top module, full scripts tree, and 52-file transitive closure are disclosed and pinned.
- Chart-role labeling is split into a separate `meta` note; checker evidence is explicitly review-only.

## Controls

Primary 4/4; checker 5/5; full audit pipeline passes; strict lint has no errors beyond 8 existing warnings; changed-evidence readiness reports `checked=1 failures=0 control_failures=0`; citation topology and worktree hygiene pass.

Review-loop recommendation: PASS WITH BOUNDED CLAIMS.


jonathonreilly:
Review-loop disposition: **close without landing; source branch preserved**.

The exact diagonal overlap exposes a claim-level blocker. For centres `t=(0,0,0)` and `s=(1,1,0)`,

```text
S(t) ∩ S(s) = {(0,1,0), (1,0,0)}.
```

Both shared sites are controls; neither distinct target/output site `t` or `s` belongs to the overlap. Nevertheless, `normalized_overlap_table()` supplies each chart with an extra anonymous target bit and compares two tables of the form `x XOR (c1 AND c2)`. Equality holds only after forgetting that the two `x` variables are different global output components. It is therefore equality of a declared role template, not equality of coordinate-bound function restrictions on the common domain. The stated “ordinary compatible-chart” gluing conclusion does not follow.

The independent checker reconstructs the same role-erasing definition in `overlap_table()`, so its agreement and mutation campaign validate bookkeeping but do not test this missing domain condition.

Making the overlap map honest removes the diagonal shared Boolean row and leaves only:

- coordinate labels and global pair/path geometry, which agree by construction; and
- copying the same declared 23-row target-indexed template to each target.

Those residual facts are definitional or already supplied by current-main Cycle 983, which owns the translated 23-row two-centre result and the exhaustive 24-offset geometry census (`6/6/12`). I therefore found no distinct promotable scientific salvage that could be reached by a narrow review repair.

There is also a current-main freshness failure: the primary pins pre-#6247 minimal-axiom SHA/blob `53175250…` / `2f5fdd26…`, while current main has `93af34cf…` / `bc23300b…`; direct execution reports `D_CONTROLS FAIL`. Refreshing those generated bindings would be mechanical but would not repair the semantic blocker.

No audit verdict was applied. The source branch is intentionally retained for research history.

PR #6127, opened 2026-08-11
feat(cell-cutting): two kernels and a closed-form colour on the 96 orbit tables (cycle 772)

jonathonreilly:
## What this responds to

The preceding cycle proved that the 384 signed coordinate maps act freely on the
36864 piece-cover pairs, so those pairs fall into 96 orbits of size 384, each of
which reads as a 192 by 192 zero-one table over covers by pieces, every one of
rank 144 with a 48-dimensional kernel. That left the family of 96 tables
undescribed: 96 subspaces of dimension 48, with no statement of how many are
distinct, how they sit inside one another, or what distinguishes them.

## What is new

**The 96 tables carry exactly two kernels.** Reducing all 96 kernels to
canonical form leaves exactly 2 distinct subspaces, each shared by exactly 48
tables, and the same split appears at both fixed primes. The two meet in 12
dimensions and join to 84. The meet is identified: it is the sign-anti-invariant
space of the 16 pure flips, which has rank 12, lies in every one of the 96
kernels, and dies on all 96 tables. The invariant space of the same 16 flips
also has rank 12 but meets every kernel in 0 and survives every table, so the
sign is what is load-bearing, not the dimension.

**Each kernel is a submodule for the whole group of 384.** That forces a stack
rule, and the runner measures it: all 48 tables of one colour stack to rank 144,
the rank of every single table, while one table of each colour stacks to
180 = 192 - 12. All 2256 same-colour pairs stay at 144 and all 2304 cross-colour
pairs reach 180.

**The colour has a closed form in the geometry of the cell.** Each cover has
exactly one non-identity map fixing it, and that map is a pure flip along a
single axis, so every cover carries an axis, 4 values of 48 covers each. Each
piece has a unique pair of axes attaining the max of min(k, 5 - k) over the four
axes, so every piece carries a pair, 6 values of 32 pieces each. The colour of an
orbit is decided by whether the cover axis lies in the piece pair, and this
agrees with the algebraic classing on all 36864 pairs, up to one global choice of
which class is called which. Both labels are equivariant under all 384 maps.

The two colour sums A and B are then computed directly. They add to the all-ones
matrix, both are 96-regular, and both have rank 4, factoring through a 4 by 6
pattern of axis in pair with 12 ones of 24, row sums 3 and column sums 2. Exactly
16 of the 384 maps hold every row and column block, and they are exactly the 16
pure flips.

**Reading the incidence through the same lens.** The cover-by-piece incidence is
the entrywise sum of exactly 4 of the 96 orbits, 3 of one colour and 1 of the
other. Their stack has rank 180 while their sum has rank 105, and 192 - 105 = 87.
Every cover holds 8 pieces of which 6 carry its axis, and every piece lies in 8
covers of which 6 carry the axis, so the incidence splits canonically into a
6-regular part and a 2-regular part with no orbit enumeration at all.

## Runner

31 gates, all passing, of which 7 are labelled rejectors or honest negatives and
3 more carry a control that fails where the real object passes. Everything is
exact over the integers and two fixed primes; no floating point enters any gate
and no constant is fitted. The runner rebuilds the 16 corners, the pieces, the
cuttings, the covers, the group, the orbits and the kernels from nothing, and
reads no cached data. 19 s, peak resident 176 MB.

The controls are the discriminating part. Twenty copies of one kernel with two
piece coordinates swapped all keep dimension 48 yet match neither class, so the
collapse to 2 is a property of the subspaces and not of their shape. Dropping one
basis vector leaves 5 of 329 generator images outside the span, so the submodule
property is sharp. An index-order split of the 96 reproduces the all-ones sum and
the 96-regularity but has rank 133 and fails to kill 48 kernel vectors of each
class, so regularity alone carries none of the content. Shifting either
geometric label by one axis disagrees on 24576 of the 36864 pairs. Two honest
negatives are recorded: the adjacent-corner count and the piece fixer weight
give 1 and 3 classes, short of the 6 needed, and the shared-corner profile takes
25 values of which 21 carry both colours, so it cannot fix the colour.

## Boundary

The rank drop from 144 on a single table to 105 on the incidence sum is
localised to the sum and is not explained here; this note offers no mechanism for
it. All ranks are measured at 2 fixed primes and the annihilation statements are
over the integers; nothing here is measured in any other characteristic.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_KERNEL_COLOUR_CLOSED_FORM_CYCLE772_NOTE_2026-08-11.md
+++ b/docs/PHYSICAL_CELL_CUTTING_KERNEL_COLOUR_CLOSED_FORM_CYCLE772_NOTE_2026-08-11.md
@@ -1,0 +1,177 @@
+# Physical cell cutting: the 96 orbit tables carry exactly two kernels, and the colour is a closed form (cycle 772)
+
+Date: 2026-08-11
+Authority: none
+Audit: unset.
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## 1. Status
+
+The unit four-cube has 2672 five-corner subsets of unit determinant. 400 of them sit at
+the adjacency cost floor 6, and those 400 admit 15800 exact cuttings of the cell into 24
+pieces. Exactly 192 distinct pieces occur, each in 1975 of the cuttings, and 24 x 15800 =
+379200 = 192 x 1975. There are also 192 eight-piece covers, each meeting every cutting
+exactly once, so 8 x 1975 = 15800. The group of 384 signed coordinate maps acts by 384
+distinct bijections on the pieces and on the covers, transitively on each, and freely on
+the 36864 pairs made of one piece and one cover, so those pairs fall into 96 orbits of
+size 384. Read as a zero-one table over covers by pieces, each orbit is a 192 by 192
+matrix, and those 96 matrices are the subject here.
+
+Three things are proved. First, the 96 tables carry exactly 2 kernels: the kernel of each
+table has dimension 48, and reducing all 96 of them to canonical form leaves exactly 2
+distinct subspaces, each shared by 48 tables, at each of the two primes 1000003 and
+1000033. Second, each of the 2 kernels is a submodule for the whole group of 384, which
+forces a stack rule: all 48 tables of one colour stack to rank 144, while a table of each
+colour stacks to 180. Third, the colour has a closed form in the geometry of the cell:
+each cover carries one axis, each piece carries a pair of axes, and the colour of an orbit
+is decided by whether the cover axis lies in the piece pair, on all 36864 pairs.
+
+The runner
+[physical_cell_cutting_kernel_colour_closed_form_cycle772_2026_08_11.py](../scripts/physical_cell_cutting_kernel_colour_closed_form_cycle772_2026_08_11.py)
+rebuilds every one of those objects from nothing: the 16 corners, the pieces, the
+cuttings, the covers, the group, the orbits and the kernels are all constructed in the
+run, and no cached data is read. It reports 31 gates, all passing, of which 7 are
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
`/review-loop` completed and the durable science has landed directly on `main` at `bd1ab6834bf77adf432f4b32210506569e4b6a4d`.

Review provenance:

- Frozen PR head: `1cf14afec45db4864d024eee72feeecdae9f4a41`
- Narrow review repair: `0aa9788df83ddde42da085b26ac8b46e447adf27`
- Landing commit: `bd1ab6834bf77adf432f4b32210506569e4b6a4d`
- Sole combined reviewer: GPT-5.6-Sol, maximum reasoning, persistent session `01a00fd5-9b04-7b70-b1f8-4c9cec6425df`
- Two science review iterations plus a narrow identity-only closure correcting an orchestrator SHA transcription; 10 unique findings, all 10 resolved, 0 open
- Final disposition: `PASS WITH BOUNDED CLAIMS`; audit compatibility `PASS`

The repair replaced campaign/cycle framing with a canonical bounded theorem note, supplied the exact 4-cube separation geometry, disclosed field and timeout/cache boundaries, removed the false dependency and overbroad no-go rhetoric, and added load-bearing mutation gates. Independent checks reconstructed all 15,168 co-pairs and their exact integer separators, verified the explicit overlapping-pair barycentrics, checked the determinant-2 characteristic-zero certificate, and confirmed the rank-changing mutations.

Integrated validation on fresh `main` passed: live runner 38/38, committed cache SHA and timeout exact, Python compilation, full 18-stage audit pipeline, strict audit lint with zero errors, `git diff --check`, and changed-science evidence readiness (`checked=1`, `failures=0`). Validation-only ledger/effective-status outputs were stripped; only the current-tree citation manifest acknowledgment was retained.

New claim `four_cube_cutting_kernel_colour_closed_form_note_2026-08-11` remains `bounded_theorem`, `unaudited`, with `deps: []`. It now requires the independent audit lane; no audit verdict was authored or applied by review-loop.

PR #6129, opened 2026-08-11
feat(cell-cutting): sign-pattern blocks split every rank (cycle 773)

jonathonreilly:
## What this responds to

Cycles 771 and 772 measured that the cover incidence of the cell cutting family has rank 105 with a 39-dimensional shortfall against the 144 of a single orbit table, and that the family carries a closed-form two-colour label. Both left the 39 as a number with no location. This cycle gives the coordinates in which it sits.

## The mechanism

The sixteen sign patterns of the four axes cut the 192 piece coordinates into sixteen blocks of dimension twelve. The identity this cycle runs on: when a matrix over the pieces has its kernel held by the sixteen pure flips, its rank is the sum of its sixteen per-block ranks. That recomposition is the discriminating statistic, and three rejectors show what is not. Constancy by weight is not sufficient: a 48-row coordinate slice is perfectly constant at rank twelve in every block, recomposes to 192, and has actual rank 48. Rank alone is not sufficient: swapping two pieces of the incidence leaves the rank at 105 but breaks the hold of fifteen of the sixteen flips, and the recomposition returns 121. And the weaker "pair meets the pattern" rule gives nullity 144 rather than the measured 48 on every one of the ninety-six tables.

## Results

- A single orbit table splits `[12, 12, 10, 6, 0]` recomposing to 144, and this is derived rather than measured: each of its 48 eight-cycles is held by exactly the four flips of one axis pair, that label is equivariant under all 384 maps with 0 of 18432 checks failing, and a cycle survives into a pattern exactly when its pair sits inside it. No matrix is reduced anywhere in that derivation.
- The cover incidence splits `[9, 9, 6, 6, 0]` to 105, with per-block kernel `[3, 3, 6, 6, 12]` summing to 87, at both primes.
- Splitting the incidence by the two labels the object already carries gives kernel 78 recomposing to 114, and kernel 48 recomposing to 144. The second part is a single orbit table, so its profile is exactly the derived one.
- The shortfall is located: `[3, 3, 4, 0, 0]` by weight, weighting to 3 + 12 + 24 = 39, and identically zero in both blocks of weight above two. Every one of the 39 lost dimensions sits in a block of weight at most two.
- The kernel of the incidence splits cleanly by mechanism. Twelve dimensions are the all-axes block, killed by each of the four parts separately. The other 75 are killed by the sum and by no single part, with not one near miss.

## Honest negatives, stated rather than softened

- At characteristic two the sixteen block bases coincide and stack to rank twelve, so the block route says nothing there. The value 144 at that characteristic is reported as a separate direct measurement.
- The block profile cannot separate the two distinct kernel classes among the ninety-six tables: both give `[0, 0, 2, 6, 12]`.
- Summing the four orbit tables that come next in index order produces a zero-one table, 8-regular both ways, none of whose 192 rows is a cover, which nonetheless reaches the same rank 105, the same drop `[3, 3, 4, 0, 0]`, the same weighting 39, and the same kernel dimension 87. Keeping three parts and swapping only the fourth does move both readings, to rank 93 and drop `[3, 3, 4, 3, 0]`, so the comparison is not vacuous. What separates the two sums is the kernel as a subspace and not as a number: both have dimension 87 and they meet in only 33. Every reading this note extracts from the block split is a dimension count, and dimension counts are exactly what the two sums share. This is stated in the note's boundary so that no claim here can be read as a fingerprint of the cover incidence.

## Runner

The paired runner rebuilds the cell, the 2672 unit-determinant subsets, the 400 at the adjacency floor, the 15800 cuttings, the 192 pieces, the 192 covers, the 384 maps, the 96 orbits and the incidence from the sixteen corners upward. It reads no cached data and no input file, works exactly over the integers and two fixed primes, and prints one line per gate.

`TOTAL: PASS=28 FAIL=0`, exit 0, elapsed 13 s, peak resident well inside budget. Re-run independently before committing; the cached body matches.

No axiom and no import is added. The note stands on the minimal axioms alone; the two companion stems in flight are named in backticks only, and nothing here reads them, cites a value from them, or depends on them.

The second commit is the citation-graph manifest acknowledgement required by the link guard: one new node with out-degree one, no other derived file touched.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_SIGN_PATTERN_BLOCKS_CYCLE773_NOTE_2026-08-11.md
+++ b/docs/PHYSICAL_CELL_CUTTING_SIGN_PATTERN_BLOCKS_CYCLE773_NOTE_2026-08-11.md
@@ -1,0 +1,160 @@
+# Physical cell cutting: the sign-pattern blocks split every rank, and locate the 39 (cycle 773)
+
+Date: 2026-08-11
+Authority: none
+Audit: unset.
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## 1. What is measured
+
+The object is the cell cutting family of the unit four-cube: its 16 corners, the 2672
+five-corner subsets of unit determinant, the 400 of those at the adjacency cost floor 6,
+the 15800 cuttings of the cell by 24 such pieces, the 192 pieces that actually occur (each
+in 1975 cuttings, so 379200 piece slots counted either way), and the 192 covers of 8
+pieces each that meet every cutting exactly once. The paired runner
+[scripts/physical_cell_cutting_sign_pattern_blocks_cycle773_2026_08_11.py](../scripts/physical_cell_cutting_sign_pattern_blocks_cycle773_2026_08_11.py)
+builds every one of those objects in-run from the corners upward; it reads no cached data
+and no input file, works exactly over the integers and two fixed primes, and prints one
+line per gate. Every one of those gate lines passes; 3 of them are labelled rejectors and
+3 are labelled honest negatives. The honest negatives carry the sharpest content in this
+note and are stated in section 8, not softened away.
+
+## 2. Sixteen blocks of dimension twelve
+
+The 384 signed coordinate maps act freely on the 36864 pairs made of one piece and one
+cover, giving 96 orbits of 384, and each orbit read as a zero-one table over covers by
+pieces is one of the 96 tables below. Inside that group sit the 16 pure flips, one per
+subset of the 4 axes. They form a subgroup, they act freely on the 192 pieces, and they
+leave 12 orbits of 16 pieces each.
+
+Every sign pattern of the 4 axes is a subset s, and its character sends the flip f to plus
+one when s and f overlap evenly and to minus one when they overlap oddly. Fix one piece per
+flip orbit; the vector carrying the character value of s at f on the image of that piece
+under f, and nothing elsewhere, is one basis vector, so the block of s has dimension 12.
+Then 16 times 12 is 192, the whole piece count, and the 16 blocks together stack to rank
+192, so they fill the piece coordinates rather than merely fitting inside them. The block
+of the empty pattern is the flip-invariant space and the block of the all-axes pattern is
+the space where every flip acts by its sign.
+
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
`/review-loop` completed; durable science was salvaged and landed directly on `main`.

- PR: #6129, original head `5f9c0e6d299449bc0831f5469abc5ec198e784ba`
- Frozen merge base: `ea0968c71ad46c39c6dacb39f88a18780363b71f`
- Reviewed preparation commit: `fb17161cd685cdb9b4949e8f1dd5f58afb8d681a`
- Reviewed fix commit: `d32fa36f9a50c391e61e2e10f547709b125de123`
- Fresh-main integration and landed SHA: `53fab8f45f73cc5bd68ff56e73ad5c78058e49d9`
- Reviewer: GPT-5.6-Sol, maximum reasoning, same session `01a00ff0-0bea-72c3-a2d5-a2815471cfaf`
- Findings: 8 historical findings resolved (1 critical, 6 high, 1 medium); 0 new and 0 open. Final terminal: `open=0 critical=0 high=0 medium=0 low=0 status=PASS`.

The landed claim is bounded, self-contained finite combinatorics: an exact rational certificate for the stated unit-four-cube simplex dissections, plus rank/profile/kernel facts over explicitly named finite fields. The repair does not claim a face-to-face triangulation, physical interpretation, universal pruning conclusion, or retained/Nature-grade status. The false D24 mechanism was replaced by actual kernel-intersection dimensions `[12,12,12,20]` over both declared large fields; the quotient extras are `[0,0,0,8]`.

Validation included live runner `PASS=29 FAIL=0`, independent exact-rational rank checks, independent D24 recomputation, all 29 gates failing closed under combined mutation, authenticated runner cache, full audit pipeline, strict lint, changed-evidence readiness, no-go non-trigger, generated-output stripping, and proactive current-tree citation-manifest regeneration. A fresh-main harness conflict was resolved by preserving every current-main row byte-for-byte and adding exactly the reviewed Cycle-773 row; the same reviewer reconfirmed that integrated surface at terminal zero.

No audit verdict, effective-status output, reviewer transcript, methodology/process change, or generated audit/status surface was landed. The new bounded claim remains for the independent audit lane.

PR #6136, opened 2026-08-11
feat(cell-cutting): force the cover incidence from the symmetry group up to the centre (cycle 776)

jonathonreilly:
## What this responds to

Cycles 771 to 775 established a genuinely uncomfortable fact: the cover incidence of the unit four-cube cell sits inside a family of **3321960** four-table sums, and no linear instrument this framework has separates it from the rest. The family shares the rank, the kernel, the regularity and the whole block-rank profile; the finest image label of cycle 775 still left **30720** candidates. The honest description of the object was "one of millions of look-alikes".

## What changed

The look-alikes are look-alikes only to linear instruments. This cycle imposes instead a condition that is part of the object's own definition — **every row of the member must be a cover of the cell**, not merely some 8-element set of pieces — and exactly **48** members survive.

The 48 is a group index rather than a measurement, and the note derives it:

- each of the 96 orbit tables is a single orbit of the diagonal action of the group of **384** signed coordinate maps, so every family member is carried to itself by every group element acting on both sides at once;
- therefore a member whose rows are covers induces a bijection of the 192 covers that **commutes with the whole group**;
- the group is transitive on the 192 covers with a point stabiliser of order **2** (a single-axis sign flip), so the commuting bijections are the right multiplications by the normalizer of that stabiliser — one per coset;
- the normalizer has order **96 = 6 x 16**: conjugating an axis flip by a signed coordinate map gives the flip of the permuted axis, so a group element commutes exactly when its coordinate permutation fixes that axis (6 of the 24), and all 16 sign flips commute because the sign group is abelian;
- the index is 96 over 2 = **48**.

Requiring in addition that the induced cover relabelling be an **actual symmetry of the cell** rather than an abstract bijection means the right multiplication must also be a left multiplication, which is the **centre — order 2**, the identity and the all-axes sign flip. The two survivors are the cover incidence and the cover incidence with its covers renamed by that flip: the same object twice.

## The load-bearing gate

The group-theoretic construction and an independent brute-force census over the family return **the same 48, symmetric difference 0** — element by element, not two agreeing counts. Without that gate the arithmetic agreement would be a coincidence; with it, the brute-force census has been identified as the coset space of the normalizer. Direct equivariance is separately re-verified against all 384 group elements at all 192 covers for each of the 48 members: **3538944 comparisons, 0 misses**.

## Rejectors

- Relabelling the incidence's covers by a group element and asking whether the result is still a sum of four tables succeeds for the **2** central elements and fails for **all 382 others**.
- The partner involution on the 96 tables gives 16 swap variants of the incidence. Of the four this runner tests, two carry the incidence's **entire** 16-pattern block-rank profile at both primes and are nevertheless outside the census — so a member can agree with the incidence on every rank the framework measures and still fail to have covers for rows. The rank profile could not have replaced the row condition.

## Honest negatives (stated in the note)

The census is 48 and not 2; only the symmetry requirement reaches 2, and the image-label filter reaches 4. The block-rank profile separates **none** of the 48 at either prime. 46 of the 48 realize their cover bijection by a map outside the group, so census members are not all symmetries in disguise. The residual pair differs by a renaming of the covers, so no rank or image instrument in this runner — all of them naming-blind — can reduce it further; that would need a naming-sensitive observable, which this note does not supply. The mirror condition on columns is not measured here.

## Runner

`scripts/physical_cell_cutting_normalizer_rigidity_cycle776_2026_08_11.py` — **26 gates, TOTAL: PASS=26 FAIL=0**. Builds everything in-run from the 16 corners upward (2672 unit-determinant five-corner subsets, 400 at the adjacency-cost floor 6, 15800 cuttings into 24 pieces, 192 pieces and 192 covers), reads no cached data and no input file, works exactly over the integers and two fixed primes, uses no floating point in any gate, and fits no constant.

Includes a separate `audit-infra:` commit acknowledging the new note in the citation graph manifest.


JONbridgerapps:
Review-loop landing provenance

- Original PR head reviewed: `48d7a3a0305cd1d7fc084cd4776782e773f63875`
- Corrected source reviewed: `d8c05ebe8a52b4298bf732dbc07af674ff28274c`
- Combined reviewer: GPT-5.6-Sol, maximum reasoning, session `01a01054-6c1e-7b81-9958-990ff8658345`
- Terminal review: R001-R008 closed; 0 open findings; same-session changed-file confirmation PASS
- Durable science retained: an explicitly supplied finite model; exact 4D dissection geometry; 192 cutting-family transversals; faithful group/coset/core conventions; the normalizer quotient and complete 48x48 product/action checks; distinct-prime certification
- Independent checks: direct D4 group/stabilizer/normalizer/quotient/centre enumeration; fail-closed mutations for duplicate fields, geometry bypass, deleted G23 operand, census corruption, and central-partner corruption
- Evidence: runner/cache `PASS=28 FAIL=0`; runner SHA-256 `d16517c0175a2fb45fb8633ed349737daed08574e6610950286bd391d603b7fc`; pipeline PASS; strict lint PASS; changed-evidence PASS; fresh-main replay vocabulary/compile/runner/strict/changed-evidence PASS
- Landed to `main`: `0c42c39fabc630f0057eb57a1bbaa4da75b68e38..c4acabe57b82cda211d6ac6e487031e2e834d8a5`
- Remote containment: note, runner, cache, and required citation-graph manifest only; generated audit/status surfaces were excluded

No audit verdict was applied. Independent audit remains required.

PR #6182, opened 2026-08-13
[physics-loop] toe-lphys aspeed: bounded_theorem / bounded-support

jonathonreilly:
## Summary

Among linear Wick maps `k4 = i a ω` of the Euclidean OS0 form `(k4^2+k^2)/4`, the extra matching that the Lorentzian null coincide with Euclidean equal coefficients holds iff `a^2 = 1`, so `a ∈ {+1, −1}` uniquely up to time-orientation. Neither the four axioms nor `c_t = c_s` names that matching. The note does not install `a = 1` and does not claim Lorentzian closure.

Time still has a primitive clock (record count). Relativistic time needs a Wick factor. This block pins the unique linear clock that **keeps the Euclidean speed**. The remaining wall is a *reason* for speed-preservation, not another listing of the `a` family.

## What changed

- Note: `docs/SPEED_PRESERVING_LINEAR_WICK_MAP_IS_UNIT_CLOCK_BOUNDED_THEOREM_NOTE_2026-08-13.md`
- Runner: `scripts/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.py`
- Cache: `logs/runner-cache/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.txt`

## Verification

```bash
python3 scripts/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.py
# TOTAL: PASS=28 FAIL=0
```

### DIFF
--- a/docs/SPEED_PRESERVING_LINEAR_WICK_MAP_IS_UNIT_CLOCK_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/SPEED_PRESERVING_LINEAR_WICK_MAP_IS_UNIT_CLOCK_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,478 @@
+---
+claim_id: speed_preserving_linear_wick_map_is_unit_clock_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "Among linear Wick maps k4=i a ω of the Euclidean OS0 form (k4^2+k^2)/4, the declared extra matching that the Lorentzian null a^2 ω^2=k^2 coincide with Euclidean equal coefficients holds if and only if a^2=1, hence a∈{+1,-1} uniquely up to time-orientation; neither the four axioms nor c_t=c_s names that matching, and this note does not install a=1 or claim Lorentzian closure."
+upstream_dependencies:
+  - minimal_axioms
+  - kinetic_isotropy_primitive
+runner: scripts/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.py
+---
+
+# Speed-Preserving Linear Wick Map Is The Unit Clock
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** exact algebra of the linear Wick family of the Euclidean OS0
+TT form, under a declared extra speed-preservation matching.
+**Audit-status authority:** independent audit lane only. This note authors no
+audit verdict and predicts none.
+**Primary runner:**
+[`scripts/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.py`](../scripts/speed_preserving_linear_wick_map_is_unit_clock_2026_08_13.py)
+
+## Result Up Front
+
+The approved kinetic-isotropy primitive
+[`KINETIC_ISOTROPY_PRIMITIVE_NOTE_2026-06-09.md`](KINETIC_ISOTROPY_PRIMITIVE_NOTE_2026-06-09.md)
+supplies one Euclidean kinetic-form equality,
+
+```text
+c_t = c_s,
+```
+
+equivalently the Osterwalder-Schrader OS0 kinetic normalization of the
+Euclidean regulator block `Z^3 x Z_tau`. In momentum space that is the
+declared Euclidean TT form
+
+```text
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **CLOSE without landing**. The remote source branch is intentionally preserved.

Provenance frozen before review:

- PR head: `14827386581b0f5d1ae71a1b2d78737889f47cae`
- source commit: `cef61e147d9669a86b05069475200c21dab01615`
- reviewed against current `origin/main`: `0dfd13c0a383e2ddde5660669bcff662be0e96d2`
- clean replay candidate: `dab431d1da2aedef81a9d0fa8275bcaae61fedf7`

The submitted algebra is exact: substituting `k4=i a omega` into the declared isotropic quadratic form gives a temporal coefficient proportional to `a^2`, and the separately defined predicate requiring that raw coefficient to equal the spatial coefficient is equivalent to `a^2=1`. The focused runner passed `28/28`, and `py_compile` and `git diff --check` passed.

That does not yield new retained science on current main. The landed theorem `linear_temporal_reparameterization_isotropic_quadratic_form_bounded_theorem_note_2026-08-13` already proves

```text
Q_b((a/b) omega,k) = Q_a(omega,k),
Omega = a omega,
Q_a = c(-Omega^2+|k|^2).
```

Consequently the proposed raw coefficient comparison is coordinate-dependent. Calling the `a^2=1` row “speed-preserving” defines `omega` to be the normalized coordinate; it does not distinguish a clock within the reparameterization orbit. The `a=1/2,1,2` witnesses are the same rows already carried by the landed theorem, and the `a=-1` extension adds only time-orientation reversal because the quadratic form depends on `a^2`. This fails the value gate as a one-step relabel/corollary rather than a load-bearing theorem.

Two current-framework inconsistencies would also block landing as written:

1. The kinetic-isotropy primitive supplies the ratio `c_t=c_s`, not the overall `1/4` normalization. Current main deliberately keeps the common coefficient `c` symbolic; `1/4` is at most a separately declared illustration.
2. The note says Record supplies an additive scalar `I` with `I(empty)=0`, and the PR body calls record count a primitive clock. The owner-approved premise update removed the named scalar, finite additivity, and a scalar value at absence. Current Record supplies no count or physical clock map.

Reviewer lenses: the symbolic implementation is deterministic and safe; there are no security, portability, or dependency-cycle findings. Governance/audit compatibility fails because the note carries retired Record semantics and duplicates the current canonical clock-reparameterization surface. No audit verdict was authored or applied.

No-Go Discipline (N1-N8):

- **N1 routes:** Euclidean isotropy alone, coordinate normalization, independent clock/readout calibration, Lorentzian reconstruction/dynamics, and time-orientation were examined separately. Only the coordinate-normalization route closes here; physical clock calibration and Lorentzian reconstruction remain live.
- **N2 independence:** a physical readout/calibration, Lorentzian dynamics, and a coordinate convention are different obligations. Closing the algebraic coordinate identity closes none of the physical routes.
- **N3 hidden conditions:** the claim fixes the `omega` coordinate, declares linear/rational maps, imports an overall `1/4`, and treats coefficient equality as “speed.” The first and last together are the selector being claimed, while the normalization is not primitive content.
- **N4 residual matching:** the kinetic primitive contributes only `c_t=c_s`; the landed reparameterization theorem contributes the exact quotient identity; current Record contributes neither scalar count nor clock. No source residual supplies a physical `omega` normalization.
- **N5 rhetoric:** this closure does not say speed preservation is impossible, that Lorentzian physics cannot close, or that `a=1` cannot be selected after new input. It says only that this PR does not add an invariant result beyond current main.
- **N6 partial closures:** an independently justified physical clock observable, calibrated frequency coordinate, reconstruction/dynamics interface, or coordinate-invariant discriminator could support a future theorem.
- **N7 steelman:** naming `a^2=1` as the unique equal-raw-coefficient row is mathematically correct and potentially convenient. It remains the direct consequence of the definition `speed_preserved(a) := (a*a==1)` after choosing normalized `omega`, so convenience does not clear the novelty gate.
- **N8 echo:** current main already includes the exact `a=1/2,1,2` coefficient table, `Omega=a omega`, the pairwise invertible rescaling law, and the boundary that no physical clock is selected. The proposal is a direct echo of that retained surface.

Because this is a no-land closure, review-loop correctly skips manifest regeneration and the full science pipeline. No tracked files were changed.

PR #6171, opened 2026-08-13
[physics-loop] toe-lphys block12 graph-separated-bits: bounded_theorem / bounded-support

jonathonreilly:
## Summary

On `Z^3` with nearest-neighbor adjacency, a one-site Admissibility law at an auxiliary site is a function of the 6-tuple on `N(x)`.

- `0 ∈ N(e1)`: adjacent placement does **not** force independence of the origin possibility.
- `0 ∉ N(2e1)` and `0 ∉ N(3e1)`: distance ≥ 2 excludes the origin from the condition domain.
- Spacing-3 sites `x_k=3k e1` have pairwise-disjoint neighbor supports, each disjoint from `N(0)`. Spacing 2 is not disjoint-support: `N(0) ∩ N(2e1)={e1}`.
- The product of one-site laws on `{3e1,6e1,9e1}` is therefore not a function of the origin possibility. This is a **conditional** independence compiler: it assumes those one-site laws and assumes records form there.

Uniform binary margins and record formation are not derived. No axiom edit.

Campaign `toe-lphys-20260812` Block 12. Independent of unmerged Blocks 01–11. Stretch on the physical-bit-compiler residual.

## What changed

- Note: `docs/GRAPH_SEPARATED_AUXILIARY_RECORDS_INDEPENDENT_OF_DISTANT_LOCK_BOUNDED_THEOREM_NOTE_2026-08-13.md`
- Runner: `scripts/graph_separated_auxiliary_records_independent_of_distant_lock_2026_08_13.py`
- Cache: `logs/runner-cache/graph_separated_auxiliary_records_independent_of_distant_lock_2026_08_13.txt`
- Citation-graph carve-out: `docs/audit/data/citation_graph_manifest.json` (new node, out-degree 2, parents Aug 10 / minimal axioms)

## Verification

```bash
python3 scripts/graph_separated_auxiliary_records_independent_of_distant_lock_2026_08_13.py
# TOTAL: PASS=35 FAIL=0
```

Cache written through `scripts/runner_cache.py` `execute_and_write_cache(..., 120)` matching `AUDIT_TIMEOUT_SEC = 120`.

Mutation checks: Moore neighborhood fails blanket distance-2 origin exclusion (`0` is a Moore neighbor of `e1+e2`); spacing 2 fails `disjoint_supports`; a declared always-independent predicate contradicts `0 ∈ N(e1)`.

## Promotion value gate

[...]

PR #6166, opened 2026-08-13
[physics-loop] toe-lphys block10 formation-site: bounded_theorem / bounded-support

jonathonreilly:
Admissibility support forbids locking a zero-mass possibility. It does not pick formation site or rate. Campaign Block 10. No axiom edit. Runner PASS=11 FAIL=0. bounded-support.

PR #6183, opened 2026-08-13
[physics-loop] toe-lphys formsite: bounded_theorem / bounded-support

jonathonreilly:
## Summary

On Z^3, sites 3e1 and 6e1 have disjoint NN 6-tuples. A product of one-site laws is the same function of those tuples whether the sites are occupied or not. Histories H_form and H_empty share that law. A bit compiler needs H_form; the axioms do not select it. No claim that no compiler exists. No axiom edit. L_phys not adopted.

Independence of auxiliary sites (the Born-compiler geometry) is not occurrence. Fair independent *laws* can sit on empty sites. Quantum mechanics still needs a formation rule before those sites become record bits.

## Verification

```bash
python3 scripts/product_law_does_not_force_formation_at_auxiliary_sites_2026_08_13.py
# TOTAL: PASS=27 FAIL=0
```

### DIFF
--- a/docs/PRODUCT_LAW_DOES_NOT_FORCE_FORMATION_AT_AUXILIARY_SITES_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/PRODUCT_LAW_DOES_NOT_FORCE_FORMATION_AT_AUXILIARY_SITES_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,454 @@
+---
+claim_id: product_law_does_not_force_formation_at_auxiliary_sites_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "On Z^3, the sites 3e1 and 6e1 have disjoint nearest-neighbor 6-tuples, so a product of one-site laws on {0,1} is the same function of those 6-tuples whether or not records occupy the two sites; the histories H_form and H_empty share that product law and differ only in occupancy; a bit compiler that needs records at those sites requires H_form, but the axioms do not select H_form over H_empty."
+upstream_dependencies:
+  - minimal_axioms
+runner: scripts/product_law_does_not_force_formation_at_auxiliary_sites_2026_08_13.py
+---
+
+# Product Law Does Not Force Formation At Auxiliary Sites
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** exact cubic nearest-neighbor listings for `3e1` and `6e1`; product
+of one-site laws as a function of those disjoint 6-tuples; occupancy-tagged
+histories `H_form` and `H_empty` sharing the same content law; a bit compiler
+that needs records at those sites is not selected by the axioms.
+**Audit-status authority:** independent audit lane only. This note authors no audit verdict and predicts none.
+**Primary runner:**
+[`scripts/product_law_does_not_force_formation_at_auxiliary_sites_2026_08_13.py`](../scripts/product_law_does_not_force_formation_at_auxiliary_sites_2026_08_13.py)
+
+## Result Up Front
+
+A product of one-site Admissibility laws is a content law. Occupancy is a
+separate mark: whether a record is present at a named site. The two objects
+are not the same.
+
+The current Admissibility sentence in
+[`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md) is quoted only
+as a premise and is not edited:
+
+For each site, the probability distribution over the possibilities is determined by, and varies with, the nearest-neighbor conditions.
+
+The current Lattice sentence is likewise quoted only as a premise:
+
+Physical sites are the points of the cubic lattice `Z^3`, with nearest-neighbor
+adjacency, standard translations, and proper cubic rotations about each site.
+
+The Admissibility reading note is quoted only as a premise. The distribution
+concerns which possibility a forming record locks, conditional on formation at
+that site; it does not supply the formation site, probability, or rate.
+
+The current Record lock sentence is quoted only as a premise:
+
+When present, a record locks exactly one admissible local possibility.
+
+Five exact statements locate the split.
+
+1. **Disjoint supports.** The six-point sets `N(3e1)` and `N(6e1)` are
+   disjoint. Neither contains the origin. Neither site lies in its own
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **CLOSE without landing**. The remote source branch is intentionally preserved.

Frozen provenance:

- exact PR head: `3dd185fdb07b2e0c9026cd3d5faf00abfac0074d`
- source commit: `53dc69a6833e1546af3c36601335941272962913`
- reviewed against current `origin/main`: `0dfd13c0a383e2ddde5660669bcff662be0e96d2`
- clean source replay: `a1bfe2ec704ac12470648355f539454e5ee02750`

The focused runner reports `27 PASS / 0 FAIL`; `py_compile`, both current comparator runners, and `git diff --check` also pass. Those green checks do not validate the load-bearing claim.

## Findings

1. **The positive geometry and kernel result is already on main, in stronger form.** `CUBIC_NN_CONDITION_DOMAIN_SEPARATION_BOUNDED_THEOREM_NOTE_2026-08-13.md` (landed at `b4099890d9dfad2141deaf03cdaf58d7a19f7cb5`) already proves the exact `N(3e1), N(6e1), N(9e1)` disjoint-neighborhood geometry and direct-coordinate invariance of an explicitly supplied, nonconstant conditional product kernel. The submitted V2 search was frozen at the old base `c45dd5ab30` and therefore missed that landed result.

2. **The formation-site non-injectivity result is already on main.** `ADMISSIBILITY_SUPPORT_CONSTRAINS_CONTENT_NOT_FORMATION_SITE_BOUNDED_THEOREM_NOTE_2026-08-13.md` (landed at `692af0b54fec50e55140bb685905da242ebdf1f5`) already proves that one conditional content law has site-distinct and multiplicity-distinct formation extensions, while keeping physical history, formation process, and rate open. `H_form/H_empty` is the same type-separation idea with two extra occupancy tags.

3. **The claimed histories are defined, not constructed as framework histories.** At note lines 122–129, a “history” is stipulated to be only `(law, occupancy)` at two sites. The runner mirrors that at lines 130–149 with a three-field dataclass. It provides no full `Z^3` state/configuration of records, no supplied condition for a framework law, no law-admissibility check, no locked contents for the formed case, and no formation rule. At runner lines 369–376, “axiom compatibility” passes by finding the note sentence `Both are compatible` plus the held-fixed law equality. This is a self-confirming prose check, not a model construction.

4. **The conditional-law test is vacuous where novelty would have to live.** `one_site_law` accepts `eta` but ignores every one of its six entries; consequently `prod_form == prod_empty == prod_other_eta` is true by construction. Current main’s comparator deliberately uses nonconstant kernels and verifies both origin invariance and sensitivity to a used support coordinate, expressly preventing this older vacuous-test pattern.

5. **The note contains a false general statement.** Lines 181–184 say distinct sites share an open nearest neighbor iff `d<=2`. In the bipartite cubic graph, adjacent distinct sites (`d=1`) have disjoint open neighborhoods; the exact condition is the appropriate distance-two displacement. The needed `d=3` nonintersection conclusion is true, but the stated iff is false and current main already carries its correction.

Reviewer summary: Code/Runner `FAIL` on the load-bearing history/compatibility test; Physics `REJECT` as duplicate plus semantic bridge; Proof Obligations `FAIL` because no full-history countermodel is constructed; Imports `DISCLOSED` (binary menu, margins, and product factorization are declared); Nature Retention `REJECT`; Repo Governance `CLOSE`; Audit Compatibility `NOT APPLICABLE` because nothing lands. No audit verdict was authored or applied.

## Salvage pass

- neighbor geometry: duplicate of the stronger landed cubic-condition-domain theorem;
- product-kernel invariance: duplicate, with a weaker constant-kernel runner;
- formation-site/multiplicity separation: duplicate of the landed formation-extension fiber;
- full-history countermodel: potentially novel, but absent and not repairable without a new scientific construction.

There is therefore no novel, correct bounded-science slice to salvage from this PR.

## No-Go Discipline (N1–N8)

- **N1 routes:** exact cubic geometry, nonconstant conditional-kernel invariance, finite formation-extension non-injectivity, an explicit pair of full `Z^3` law-admissible countermodels, formation dynamics/kernel derivation, and a Record content/registration compiler were considered. The first three are ruled out as already landed; the last three remain live and were not executed here.
- **N2 independence:** novelty and full-history validity initially appear as separate walls, but for this package they collapse to one terminal obligation: construct a genuinely new full-law countermodel beyond the landed local extension fiber. The constant-kernel test and false distance statement are additional implementation defects, not independent physics walls.
- **N3 hidden conditions:** “A history is,” “Both are compatible,” the binary menu, fair product, product factorization, and the claim that occurrence is realized elsewhere were scanned. The menu/margins/product are declared toy inputs; full-lattice state, law domain, lock contents, and realization elsewhere are hidden/unconstructed.
- **N4 residual matching:** the landed formation-extension note exactly addresses content-law versus site/multiplicity; the landed cubic-condition-domain note exactly addresses spacing-3 domains and product-kernel invariance. The older general formation no-go concerns a broader process/rate residual and is not evidence that these two local dataclass objects are lawful histories.
- **N5 rhetoric:** per-element neighbor listings and product atoms are executed; per-site occupancy tags are executed only inside the toy dataclass; per-mode is not applicable; per-block only the declared local pair is checked; lattice-wide history compatibility is not executed. The close verdict is therefore narrowly “this PR does not construct the advertised histories or add novel science,” not “formation or a compiler is impossible.”
- **N6 partial closure:** a future source could explicitly construct two full `Z^3` states and one or two laws satisfying the complete current axiom contract, prove both are in the law domain, and show identical local conditional kernels with different formation at the selected sites. A derived formation kernel/dynamics is another live positive route. Neither requires an axiom edit merely to be investigated.
- **N7 steelman:** combining the two landed parents does support the useful explanatory corollary that a supplied local product kernel does not itself encode which sites are formed. That is correct, but it is already the parents’ joint boundary. To become new theorem science, the submission must add the missing full-history model or another invariant not present in those parents.
- **N8 echo:** this PR is a direct cross-cycle echo of #6166’s formation-extension fiber and #6171’s spacing-3 product-kernel theorem, both landed after the PR’s stale novelty snapshot. Their repairs specifically avoid the invalid physical-history upgrade and the constant-kernel test used here.

Because this is a no-land closure, review-loop correctly skips manifest regeneration and the full pipeline. No tracked files were changed by review.

PR #6185, opened 2026-08-13
[physics-loop] toe-lphys grain: bounded_theorem / bounded-support

jonathonreilly:
## Summary

Inside the July 16 declared record-influence class, the unique interior 2-cell stationary weight is w=1/2 (recomputed, not claimed new). The full-support law w=1/3 is allowed by Admissibility and is not that fixed point. The four axioms do not name the class, so they do not select w=1/2 via it. The unadopted dictionary only displays r=1/2 and r=1. Occupancy-grain counting is an extra fair-weight selector. No axiom edit. r=1/2 is not forced.

There is a named extra class that *would* pick a fair 2-cell weight. The axioms do not include that class. Fairness for Born bits or Koide grains is still a selector, not a theorem.

## Verification

```bash
python3 scripts/occupancy_grain_counting_class_is_extra_fair_weight_selector_2026_08_13.py
# TOTAL: PASS=27 FAIL=0
```

### DIFF
--- a/docs/OCCUPANCY_GRAIN_COUNTING_CLASS_IS_EXTRA_FAIR_WEIGHT_SELECTOR_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/OCCUPANCY_GRAIN_COUNTING_CLASS_IS_EXTRA_FAIR_WEIGHT_SELECTOR_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,481 @@
+---
+claim_id: occupancy_grain_counting_class_is_extra_fair_weight_selector_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "Within the July 16 declared record-influence class the unique interior stationary 2-cell weight is w=1/2, recomputed and not claimed new; the distinct full-support law w=1/3 is allowed by Admissibility and is not that stationary point; the four axioms do not name the class, so they do not select w=1/2 via it; the unadopted map r=(1-w)/(2w) only displays r=1/2 and r=1; occupancy-grain counting is an extra fair-weight selector."
+upstream_dependencies:
+  - minimal_axioms
+  - acphilambda_occupancy_grain_menu_counting_measure_dynamical_static_correspondence_bounded_theorem_note_2026-07-16
+runner: scripts/occupancy_grain_counting_class_is_extra_fair_weight_selector_2026_08_13.py
+---
+
+# Occupancy-Grain Counting Class Is An Extra Fair-Weight Selector
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** exact 2-cell weights `(w, 1-w)` with `0 < w < 1`; July 16
+record-influence uniqueness recomputed only; class not adopted; dictionary
+not adopted; `r = 1/2` not forced.
+**Audit-status authority:** independent audit lane only. This note authors no audit verdict and predicts none.
+**Primary runner:**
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **CLOSE WITHOUT LANDING; preserve the source branch.**

Provenance:

- submitted PR head: `1dcf39674639c787c7abbbf69ee96350279c2ac5`
- submitted source commit: `cbd9961cd3cd2ad50525a0a9e2db1b4e70a8c9a0`
- current-main review intake: `6f8046f265befb7fbb8c8e04bb7f5d43ea440a83`
- exact source replay on current main: `8dff846b8` (byte-identical note, runner, and cache inputs)

The submitted packet does not pass against current premises. Its runner closes at `PASS=26 FAIL=1`: `source-record-and-formation` still requires the retired sentence

`For any finite collection of pairwise-disjoint records, scalar readout I is additive, with I(empty)=0.`

After the Record reset landed in #6247, current Record explicitly contains no named scalar collection functional, no finite additivity, and no value at absence. The submitted cache's `27/0` result is therefore stale.

The mathematical arithmetic is correct but not a novel salvage:

- the July 16 direct parent already proves under its declared D1 class that every stationary weight is uniform on support, hence the unique interior two-cell point is `w=1/2`;
- that parent already executes power profiles, records the exact two-cell fixed set `{0,1/2,1}`, includes the numerical set `{1/3,1/2}`, and displays `w=1/2 -> r=1/2` and `w=1/3 -> r=1`;
- `T_{x^2}(1/3)=1/5` and `T_{x^3}(1/3)=1/9` are valid spot checks, but they are immediate instances of the parent's stronger uniform-on-support classification;
- the July 4 measure-binary and formation non-supply no-go notes already establish that the four axioms and approved primitives do not supply the physical occupancy-grain selector;
- the landed formation-weight dictionary theorem already supplies the post-reset Record boundary and a stronger explicit continuum of normalized alternatives, including the same `w/r` comparison points.

Removing the stale additivity premise would therefore leave only a current-wording restatement of those existing results. It would not meet the review-loop novelty/promotion gate for genuinely new bounded science.

No-Go Discipline N1-N8:

- **N1 route scan:** current Record content, Admissibility values, the conditional July 16 `T_f` class, physical matter-action statistics, a readout-partition theorem, formation dynamics, and a future approved primitive were kept distinct. None turns this duplicate packet into a new result; the latter four remain live constructive routes.
- **N2 wall audit:** “the axioms do not name `T_f`” and “the axioms do not select the occupancy grain” are the same existing non-supply residual here, not independent advances.
- **N3 hidden-wall scan:** the removed scalar/additivity premise was the hidden stale condition. No current Record collection scalar, energy codomain, formation probability, or physical grain may be imported.
- **N4 residual matching:** the July 16 parent owns the conditional stationarity classification; the July 4 no-supply notes own the open physical-grain residual; the current formation-weight theorem owns post-reset dictionary underdetermination. The submitted target adds no unmatched residual.
- **N5 rhetoric/resolution:** the valid surface is one declared two-cell update class and two rational test weights. It is not a lattice-wide selector theorem, a physical preparation law, or an unconditional derivation of `w=1/2` or `r=1/2`.
- **N6 partial-closure paths:** derive the actual matter-action measure, the physical readout partition, or formation dynamics; alternatively approve a narrow primitive. All remain open.
- **N7 hostile steelman:** the `1/5` and `1/9` calculations are explicit and exact. They are nevertheless spot instances of the already-landed parent theorem, so exactness does not create novelty.
- **N8 cross-cycle echo:** #6247 retired the scalar premise and the landed formation-weight theorem already migrated this lane to the correct current boundary. Reintroducing the old premise or duplicating that migration would regress the chain.

No full pipeline was run because no repaired landing candidate survived the science/novelty/No-Go review. No audit verdict was authored or applied. The branch is intentionally preserved for provenance.

PR #6165, opened 2026-08-13
[physics-loop] toe-lphys block05 plaquette-b1: bounded_theorem / bounded-support

jonathonreilly:
## Summary

`g_bare=1` forces Wilson `beta=6`. The exact one-plaquette Haar mean at that coupling is remainder-bounded strictly below `1/2`, so it is not the admitted B1 numeral `0.5934`. B1 is not retired. Named remaining path is the June 10 three-point `ln Z_L` / mass-gap interface.

Campaign `toe-lphys-20260812` Block 05. No axiom edit.

## Verification

```bash
python3 scripts/plaquette_b1_convention_independent_of_one_plaquette_mean_2026_08_13.py
# TOTAL: PASS=25 FAIL=0
```

Honest status: `bounded_theorem` / `bounded-support`.

PR #6189, opened 2026-08-13
[physics-loop] toe-lphys u1lnz: bounded_theorem / bounded-support

jonathonreilly:
## Summary

Remainder-controlled rational enclosures of `I_0(β)` at `β=1,2,3` are the partition function of 2D U(1) one-plaquette (and of a tree-gauge-fixed 2D disk factor). They are not 4D SU(3) `ln Z_L`, because `N_p(L=2)=96≠1` and the June 10 wrapping count `72` is unused. The numeral `0.5934` is not an input.

This is a finite certified object June 10 could consume as a *model*, not as 4D SU(3) `ln Z_L`. The B1 interface remains open.

## What changed

- Note: `docs/U1_TWO_D_ONE_PLAQUETTE_LN_Z_IS_NOT_FOUR_D_SU3_LN_Z_L_BOUNDED_THEOREM_NOTE_2026-08-13.md`
- Runner: `scripts/u1_two_d_one_plaquette_ln_z_is_not_four_d_su3_ln_z_l_2026_08_13.py`
- Cache: `logs/runner-cache/u1_two_d_one_plaquette_ln_z_is_not_four_d_su3_ln_z_l_2026_08_13.txt`

## Verification

```bash
python3 scripts/u1_two_d_one_plaquette_ln_z_is_not_four_d_su3_ln_z_l_2026_08_13.py
# TOTAL: PASS=25 FAIL=0
```

### DIFF
--- a/docs/U1_TWO_D_ONE_PLAQUETTE_LN_Z_IS_NOT_FOUR_D_SU3_LN_Z_L_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/U1_TWO_D_ONE_PLAQUETTE_LN_Z_IS_NOT_FOUR_D_SU3_LN_Z_L_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,464 @@
+---
+claim_id: u1_two_d_one_plaquette_ln_z_is_not_four_d_su3_ln_z_l_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "Remainder-controlled rational enclosures of I_0(β) at β=1,2,3 are the partition function of 2D U(1) one-plaquette (and of a tree-gauge-fixed 2D disk factor); they are not 4D SU(3) ln Z_L, because N_p(L=2)=96≠1 and the June 10 wrapping count 72 is unused."
+upstream_dependencies:
+  - minimal_axioms
+  - plaquette_value_derivation_program_specification_and_bracket_reduction_narrow_theorem_note_2026-06-10
+runner: scripts/u1_two_d_one_plaquette_ln_z_is_not_four_d_su3_ln_z_l_2026_08_13.py
+---
+
+# Certified 2D U(1) One-Plaquette I_0 Is Not 4D SU(3) ln Z_L
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** remainder-controlled `I_0` at three couplings for 2D U(1)
+one-plaquette, and the type split versus 4D SU(3) `ln Z_L`.
+**Audit-status authority:** independent audit lane only. This note authors no audit verdict and predicts none.
+**Primary runner:**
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **CLOSE / DO NOT LAND**. The source branch is being preserved.

### Frozen provenance

- PR head reviewed: `4e543c8da29ccc1c45529b62c1a03c017f8d029b`
- Science source commit: `b8d7fcb5a0bbb8c14030185ae557ee3dec5e53fe`
- Historical merge base: `c45dd5ab30b5a65df74621af789ae34cf6e4ae35`
- Current-main intake base: `6f8046f265befb7fbb8c8e04bb7f5d43ea440a83`
- Replayed review candidate: `8e738cc6280f193d13dbde39594002aad634b4c2`

### What checks out

The positive one-plaquette U(1) calculation is correct: normalized Haar integration gives
`Z_1(beta)=I_0(beta)=sum_k beta^(2k)/(4^k(k!)^2)`, and the submitted geometric-tail bound follows from the displayed term-ratio estimate. The changed runner reports `TOTAL: PASS=25 FAIL=0`; `python3 -m py_compile` and merge-base `diff --check` also pass.

Those green checks do not validate the proposed bridge to the 4D SU(3) target. The runner's decisive function `i0_enclosure_is_four_d_su3_ln_z_l()` computes the partial sum and discards it, then returns only whether the 4D lattice plaquette count equals one. Thus it encodes the desired conclusion as a count comparison and never evaluates or compares either partition function. `N_p=96 != 1` distinguishes the model definitions, but it does not by itself prove that two numerical partition-function values cannot coincide at an isolated coupling.

### Novelty and target value

The durable positive core is already present on current main. The May 10/May 17 U(1) plaquette material contains the exact `I_0` identity, the same series, and its leading coefficients. The new `N=8` rational enclosures at `beta=1,2,3` are elementary substitutions at points unrelated to the June 10 4D SU(3) certification target. Current main additionally contains the stronger, target-relevant #6165 result, which certifies separation of the SU(3) one-plaquette Haar mean from the declared B1 comparator while correctly leaving the correlated 4D problem open.

The claimed 2D open-disk factorization also depends on an open-disk topology, Wilson-action convention, normalized gauge-volume/Haar convention, and a tree-gauge change-of-variables/factorization theorem. Those premises are not supplied by the minimal axioms, and the PR provides no direct source for that step. Consequently, the declared `minimal_axioms` upstream edge is non-load-bearing.

### Mandatory No-Go review

The negative claim does not pass the N1-N8 discipline:

- N1 lists SU(3) one-plaquette integration, Monte Carlo, and mass-gap work as attacks, but these are alternative routes to the open 4D target, not distinct attacks on the narrow model-distinction claim; some are labeled `ATTEMPTED` although their own entries say they were not performed.
- N2 provides no pairwise wall-independence table.
- N3 omits the topology, action, measure normalization, and gauge-fixing/factorization assumptions above.
- N7 steelmans only a category-error slogan and does not test effective reductions or isolated numerical coincidence.
- N8 misses both the already-landed exact U(1) series and current #6165.

There is no narrow salvage that is both novel and target-useful: retaining only the rational table would duplicate existing exact U(1) calculus without advancing the SU(3) interface, while repairing the decisive bridge would require a different scientific question. I therefore did not run the expensive full pipeline for this no-land closure. No audit verdict was applied.

PR #6195, opened 2026-08-13
[physics-loop] toe-lphys occmin: bounded_theorem / bounded-support

jonathonreilly:
## Summary

On a two-site window Record allows all four occupancy maps. Shared μ plus I=1 cannot split o10 from o01. A capacity-1 token cannot represent o11. The occupancy map o is a smallest complete extra formation input. Display it; do not adopt it. Rate is not a separate target.

This is the formation exercise's second live route. Occurrence is not a reading of the odds and not a token. It is a bit-pattern on sites. The axioms do not select it.

## Verification

```bash
python3 scripts/occupancy_pattern_is_smallest_complete_formation_input_2026_08_13.py
# TOTAL: PASS=21 FAIL=0
```

### DIFF
--- a/docs/OCCUPANCY_PATTERN_IS_SMALLEST_COMPLETE_FORMATION_INPUT_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/OCCUPANCY_PATTERN_IS_SMALLEST_COMPLETE_FORMATION_INPUT_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,440 @@
+---
+claim_id: occupancy_pattern_is_smallest_complete_formation_input_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "On a two-site window W={x,y}, Record allows all four occupancy maps o:W→{0,1}. The shared one-site law μ with μ(A)=1/3, μ(B)=2/3 together with the unit-lock count I does not recover o, because o10 and o01 share μ and I=1. A capacity-1 token T∈{x,y,none} cannot represent the Record-allowed history o11. The occupancy map itself represents all four histories, is not a function of μ, and is not a value of I. Any complete extra object on this window must distinguish those four histories, so o is a smallest complete extra formation input. The axioms do not select o; display it; do not adopt it. Rate |o|/|W| is not a separate target."
+upstream_dependencies:
+  - minimal_axioms
+runner: scripts/occupancy_pattern_is_smallest_complete_formation_input_2026_08_13.py
+---
+
+# Occupancy Pattern Is The Smallest Complete Extra Formation Input
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** exact four occupancy histories on a two-site window; unit-lock
+count `I`; one shared one-site law `μ`; a capacity-1 token as an incomplete
+slice.
+**Audit-status authority:** independent audit lane only. This note authors no
+audit verdict and predicts none.
+**Primary runner:**
+[`scripts/occupancy_pattern_is_smallest_complete_formation_input_2026_08_13.py`](../scripts/occupancy_pattern_is_smallest_complete_formation_input_2026_08_13.py)
+
+## Result Up Front
+
+This is a missing-input type theorem. It names the smallest extra object that
+represents every Record-allowed occupancy on a declared two-site window. It
+does not select that object from the axioms and it does not edit an axiom
+sentence.
+
+The current Admissibility sentence in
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **close without landing; source branch preserved**.

Frozen provenance:

- PR: #6195
- submitted head: `5b1721df1e4726aa20b293c202768cbc122a80ad`
- scientific source commit: `e5241cd056dcf1b675d5cd6c2d0bfab06332c815`
- cache-only follow-up: `2cf9400017ae304f33f1972523d66b9fa6280795`
- exact current-main intake: `5ad583dee2a66fae31fab70ca5913f639a68a012`
- same-worker replay candidate: `426bb3f76482d355f8b553b1f06f5d49c082f75d`

The four-row enumeration is arithmetically correct: `{0,1}^{\{x,y\}}` has the rows `00,10,01,11`; a three-valued capacity-one token cannot encode `11`; and the occupancy count collides on `10` and `01`. Those are elementary set/cardinality facts.

They do not yield the claimed new formation theorem.

First, the result is already subsumed by the stronger current-main theorem `ADMISSIBILITY_SUPPORT_CONSTRAINS_CONTENT_NOT_FORMATION_SITE_BOUNDED_THEOREM_NOTE_2026-08-13.md` (landed through #6166). That theorem uses the same supplied `mu(A)=1/3, mu(B)=2/3`, explicitly enumerates the two-label formation-extension fiber, and distinguishes site sets `{x}`, `{y}`, and `{x,y}` and multiplicities one and two. It correctly concludes that content-law projection is non-injective in site and multiplicity while keeping a physical formation kernel, full-lattice history, process, clock, and rate open. #6195 adds only the locally empty row and the observation that a four-history code needs four labels.

Second, “the occupancy map is the smallest complete extra formation input” is circular at the claimed scientific level. Once the target is defined as representing the four occupancy histories, the occupancy history itself is automatically a four-valued complete code; any other injective four-valued code is equivalent up to relabeling. This is a cardinality tautology, not a derivation of an input, selector, formation law, or physical carrier.

Third, the note overreads the axiom surface. `Records form` is a global occurrence statement, but it does not by itself prove that all four patterns occur or are compatible on an arbitrary physical two-site window with one held-fixed `mu`. A compatibility theorem would need a full `Z^3` model satisfying the nearest-neighbor Admissibility law. The submission constructs no such model. The landed #6166 note avoided exactly this overread by using explicitly abstract labels and refusing a full-lattice claim.

Fourth, `|o|/|W|` is a spatial occupancy fraction, not a formation **rate**. A rate needs a time parameter or clocked counting/transition process. The landed formation-extension theorem explicitly keeps static multiplicity separate from rate. Saying “rate is not a separate target” because `|o|/|W|=I/2` collapses two different types.

There is also a current-premise failure. On exact current `main`, the focused runner returns `TOTAL: PASS=20 FAIL=1`, not cached `21/0`: it asserts that Record supplies a scalar `I`, finite additivity, and `I(empty)=0`. Post-#6247, the current axiom explicitly says none of those are Record content. A mathematical cardinality can be defined after an occupancy set is supplied, but it is not an axiom-supplied readout and cannot do formation-selection work.

No-Go Discipline (N1-N8):

- **N1 alternative routes:** a site-indexed formation kernel from neighbor conditions, a stochastic/quantum record-production process, a dynamics-plus-clock construction, and a separately supplied occupancy state all remain live. Merely naming the output bit pattern does not test them.
- **N2 wall independence:** complete representation of a static pattern, physical selection of that pattern, and temporal formation rate are independent obligations. Closing the first by an identity code closes neither of the others.
- **N3 hidden walls:** the argument assumes a physical two-site window, a common `mu`, compatibility of every pattern, an axiom-supplied count `I`, and identification of density with rate. None is earned by the finite table.
- **N4 residual match:** the repository residual is a physical site/process/probability/rate rule. A four-label encoding of already-specified occupancies does not reduce that residual.
- **N5 rhetoric:** “Record allows all four,” “smallest complete formation input,” “advances,” and “rate is not a separate target” overstate a coding tautology and blur static versus temporal quantities.
- **N6 partial closure:** #6166 already provides the honest positive finite closure: the content law has a nontrivial extension fiber in site and multiplicity. It names the exact remaining physical bridge without adopting an occupancy input.
- **N7 steelman:** an occupancy bitmap is indeed a natural lossless representation of a finite occupancy state. That software/data-structure fact is useful only after the state is supplied; it does not explain how records form or which bitmap is realized.
- **N8 cross-cycle echo:** current formation notes already separate conditional content, occupancy/state, site selection, and rate. This PR repeats the split with stale scalar semantics and a weaker object boundary.

Because no genuinely novel and correctly typed repaired candidate remains, the repo-native cheap-close path applies: no full pipeline was run, no audit verdict was written or applied, and no tracked change is retained. The remote source branch is intentionally left intact.

PR #6200, opened 2026-08-13
[physics-loop] toe-lphys realens: bounded_theorem / bounded-support

jonathonreilly:
## Summary

Unit-lock Record `I` is a nonnegative integer. Remainder-controlled Bessel sums give `0 < ⟨P⟩(2) < 8/9 < 1`, so the U(1) one-plaquette Haar mean is not an integer. The realized-state primitive forbids identifying that ensemble mean with a pointwise Record readout. June 10 remains open. `0.5934` is not an input.

An ensemble plaquette mean is the wrong type for Record `I`.

## What changed

- Note: `docs/REALIZED_STATE_PRIMITIVE_FORBIDS_ENSEMBLE_PLAQUETTE_AS_RECORD_I_BOUNDED_THEOREM_NOTE_2026-08-13.md`
- Runner: `scripts/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.py`
- Cache: `logs/runner-cache/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.txt`

## Verification

```bash
python3 scripts/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.py
# TOTAL: PASS=23 FAIL=0
```

### DIFF
--- a/docs/REALIZED_STATE_PRIMITIVE_FORBIDS_ENSEMBLE_PLAQUETTE_AS_RECORD_I_BOUNDED_THEOREM_NOTE_2026-08-13.md
+++ b/docs/REALIZED_STATE_PRIMITIVE_FORBIDS_ENSEMBLE_PLAQUETTE_AS_RECORD_I_BOUNDED_THEOREM_NOTE_2026-08-13.md
@@ -1,0 +1,405 @@
+---
+claim_id: realized_state_primitive_forbids_ensemble_plaquette_as_record_i_bounded_theorem_note_2026-08-13
+claim_type: bounded_theorem
+claim_scope: "Unit-lock Record readout I is a nonnegative integer domain count with I(empty)=0. The U(1) one-plaquette Haar mean <P>(2)=I_1(2)/I_0(2) obeys 0 < <P>(2) < 8/9 < 1 by remainder-controlled Bessel partial sums, so it is not an integer and is not a Record I. The realized-state primitive forbids identifying that ensemble mean with a pointwise Record readout. The note does not retire June 10, does not import 0.5934, and does not claim 4D <P>*."
+upstream_dependencies:
+  - minimal_axioms
+  - realized_state_primitive
+runner: scripts/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.py
+---
+
+# Realized-State Primitive Forbids Ensemble Plaquette Mean as Record I
+
+**Date:** 2026-08-13
+**Type:** bounded_theorem
+**Scope:** readout type split between unit-lock Record `I` and the U(1)
+one-plaquette Haar mean `<P>(2) = I_1(2)/I_0(2)`.
+**Audit-status authority:** independent audit lane only. This note authors no audit verdict and predicts none.
+**Primary runner:**
+[`scripts/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.py`](../scripts/realized_state_primitive_forbids_ensemble_plaquette_as_record_i_2026_08_13.py)
+
+## Result Up Front
+
+Unit-lock Record readout is a domain count. For a finite collection of
+pairwise-disjoint unit locks, the scalar `I` is a nonnegative integer, and
+`I(empty) = 0`. That is the Record sentence in
+[`MINIMAL_AXIOMS_2026-06-29.md`](MINIMAL_AXIOMS_2026-06-29.md).
+
+The U(1) one-plaquette ensemble mean
+
+```text
+<P>(beta) := I_1(beta) / I_0(beta)
+```
+
+is used only as a typed contrast. It is not 4D SU(3). Here
+
+```text
+I_n(beta) = sum_{k >= 0} 1/(k! (k+n)!) (beta/2)^{2k+n}.
+```
+
+At `beta = 2` the remainder-controlled partial sums give
+`0 < I_1(2)/I_0(2) < 8/9 < 1`. So `<P>(2)` is not an integer. Record `I`
+of any unit-lock pattern is an integer. They are unequal types.
+
+The realized-state primitive
+[`REALIZED_STATE_PRIMITIVE_NOTE_2026-06-11.md`](REALIZED_STATE_PRIMITIVE_NOTE_2026-06-11.md)
+supplies no averaging over alternatives and permits only pointwise
+evaluation at the realized state. `<P>` is an ensemble mean over Haar.
+The primitive forbids identifying `<P>` with a Record readout.
+
+This note does not retire June 10. It does not import 0.5934. It does not
+claim 4D `<P>*`. It does not say a later law-level table is impossible.
+
+## Machine Status And Trace
+
+```yaml
+actual_current_surface_status: bounded-support
+target_claim_type: bounded_theorem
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Review-loop disposition: **close without landing**. This is a source/science review only and applies no audit verdict.

### Frozen provenance

- current `origin/main`: `35635c28d237ff9ceae3633b3bb6544dc31d5ee2`
- exact PR head: `1e8b73813ec9833531379efc3da8c0af5b3839a7`
- science commit replayed for review: `c937994ef30d47c8ccfc4af04657a4cd3ec57330`
- historical merge-base: `c45dd5ab30b5a65df74621af789ae34cf6e4ae35`
- isolated full worktree; no unmerged PR was used as authority

### Blocking premise and runner findings

The fresh runner reports `PASS=23 FAIL=0`, but that green result is invalid because its current-premise guard is negation-insensitive. Current `docs/MINIMAL_AXIOMS_2026-06-29.md` says:

```text
Finite additivity, a named scalar collection functional I, and an assigned
value I(empty)=0 are not Record axiom content.
```

and records that the 2026-08-13 owner-approved revision removed all three. The runner nevertheless passes `source-record-i` because it searches for the token `I(empty)=0` anywhere in the file, including inside this explicit negation/removal paragraph. Its `canonical-nonmutation` check repeats the same error. Thus the cache and fresh run both show a false-positive governance gate, not current-premise support.

There is a second independent premise error. Even the historical Record sentence supplied only additivity over disjoint records and `I(empty)=0`; it never normalized every singleton record to `1`, required nonnegative values, or defined `I` as a domain count. The submitted function

```python
def record_I(unit_locks):
    return len(unit_locks)
```

therefore inserts the load-bearing integer/counting structure that the note claims to quote. Additivity alone permits arbitrary content-dependent singleton values in the codomain; it does not imply cardinality.

### Surviving mathematics and novelty

The exact Bessel calculation is sound: positive partial sums and the displayed tail majorant establish

```text
0 < I_1(2)/I_0(2) < 8/9 < 1.
```

That pure U(1), `beta=2` inequality does not rescue the Record theorem. With no supplied scalar Record `I` and no singleton normalization, there is no current framework object against which “non-integer” creates the claimed type obstruction. It also does not advance the actual June-10 4D SU(3), `beta=6` residual. Current `main` already contains the more target-relevant `DECLARED_PLAQUETTE_COMPARISON_VALUE_DISTINCT_FROM_ONE_PLAQUETTE_HAAR_MEAN_BOUNDED_THEOREM_NOTE_2026-08-13.md`, which gives a certified SU(3) one-plaquette separation at the declared `beta=6` point while preserving the correlated 4D route.

The realized-state primitive is also overstated. It says that the primitive itself supplies pointwise evaluation and **no averaging over alternatives**. That prevents using the primitive as the source of an ensemble average. It does not globally forbid a separate law, retained observable bridge, calibration map, or registered datum from mapping a record content to an ensemble statistic, nor does it make numerical equality between differently sourced quantities impossible. “Not supplied by this primitive” is valid; “the primitive forbids the identification” is not.

### No-Go Discipline (N1-N8)

**N1 — alternative routes.** Six materially distinct routes were checked.

1. **ATTEMPTED:** define a scalar record readout directly from current Record. This fails because current Record deliberately supplies no scalar functional, empty value, or additivity.
2. **ATTEMPTED:** recover a count from historical finite additivity. This fails because additivity plus `I(empty)=0` does not normalize singleton content to one and does not imply nonnegative integers.
3. **ATTEMPTED:** evaluate the plaquette observable pointwise on the realized gauge configuration. This is compatible with the realized-state primitive but produces a configuration-dependent value, not the Haar ensemble mean.
4. **ATTEMPTED:** take a spatial plaquette average within one realized configuration. This remains a pointwise functional of the whole realized state; relating it to a Haar/thermodynamic mean needs an additional ergodic or limit theorem, not a prohibition.
5. **ATTEMPTED:** supply an explicit observable/calibration bridge that maps record content to a law-level ensemble statistic. The primitive does not supply this bridge, but it does not forbid one; the route remains constructively open.
6. **ATTEMPTED:** execute the June-10 certified 4D finite-volume/bracket route. That route targets the correlated SU(3) thermodynamic object and remains open; the U(1) `beta=2` contrast neither closes nor obstructs it.

These alternatives defeat the submitted categorical “forbids” language. Only the narrower statement “the present primitive does not itself supply ensemble averaging” survives.

**N2 — wall independence.** The submission collapses four independent walls: `(W1)` existence of a scalar Record readout, `(W2)` singleton normalization/cardinality, `(W3)` identification of a plaquette observable with record content, and `(W4)` a bridge from pointwise realized-state data to an ensemble or thermodynamic statistic. Closing any one does not close the others. Current Record removes W1 entirely; historical additivity did not close W2; the realized-state primitive addresses only what it supplies at W4.

**N3 — hidden-wall scan.** The hidden premises are “unit lock means scalar value one,” “additive readout means cardinality,” “Record values are nonnegative integers,” and “no averaging supplied means every external average-to-readout bridge is forbidden.” None follows from the cited sources. The U(1) Haar measure and `beta=2` are declared contrast data and do not match the 4D target.

**N4 — residual matching.** Current minimal axioms directly contradict the cited scalar-`I` premise. The realized-state primitive supports only the no-averaging-supplied residual, not a universal identification ban. June 10 targets the correlated 4D SU(3) thermodynamic plaquette, whereas this calculation is a single U(1) plaquette at `beta=2`; the residuals do not match. The landed SU(3) one-plaquette separation is the relevant existing route-pruning authority and already preserves the 4D route honestly.

**N5 — rhetoric audit.** Per-element: the Bessel partial sums and rational bound are correct. Per-site/plaquette: one declared U(1) plaquette distribution is evaluated. Per-record: no current scalar readout exists and no singleton normalization is derived. Per-block: no observable bridge is constructed. Lattice-wide: no 4D configuration, finite-volume limit, or thermodynamic mean is evaluated. “Wrong type for Record `I`” is therefore unsupported; the safe statement is only “this primitive does not supply an ensemble average.”

**N6 — partial-closure paths.** An explicit retained readout/observable bridge, a spatial-average plus ergodic theorem, a registered calibration datum, or the June-10 certified 4D bracket can each create a legitimate typed relation without amending the minimal axioms. The primitive's no-averaging boundary is a supplier boundary, not a ban on these import-retirement paths.

**N7 — steelman.** A hostile reviewer should argue that an instrument can report a precomputed or law-defined ensemble statistic while the realized record merely stores that report; alternatively, a self-averaging theorem can identify a spatial observable of one realized configuration with an ensemble limit. Both are compatible with the primitive because they add an explicit law/bridge rather than pretending the primitive supplied averaging. This steelman is convincing, so the categorical no-go must not ship.

**N8 — cross-cycle echo.** The #6247 premise reset is the direct cross-cycle answer: scalar `I`, finite additivity, and `I(empty)=0` were removed, and consumers relying on them must be re-reviewed rather than matching their names inside the removal notice. Separately, the landed beta-6 one-plaquette comparison shows the established pattern for plaquette negatives: separate exact objects narrowly and leave correlated 4D/bridge routes open. This PR regresses both boundaries.

### Disposition

No repair is landed. Removing the obsolete Record ontology leaves a standalone elementary U(1) Bessel bound, while narrowing the primitive claim leaves only a restatement of its explicit “no averaging supplied” sentence. Neither is a novel advance on the live plaquette target. The PR is closed without merge; its source branch is deliberately preserved for provenance. No full science pipeline was run because there is no landing candidate.

PR #6350, opened 2026-08-14
feat(cycle783): cuttings are exact group tilings by two eight-element tiles; census assembles from the sign character and profile orbits

jonathonreilly:
## What this lands

One source note + one paired runner + one cached output (source-only triple), cycle 783 of the lane campaign.

**Result.** The 15800 cuttings of the physical cell are exactly the partitions of an order-192 group into 24 left translates of two eight-element tiles, and on that picture the label-sum census is symmetry-forced modulo a short list of measured constants.

The chain, each step recomputed from the corner coordinates upward inside the runner:

- The order-384 symmetry group of the cell acts faithfully on the 192 pieces, and the piece label is covariant under its sign character (checked on all 73728 element-piece pairs). The census symmetry under change of sign is therefore a derivation.
- The even-mask subgroup (order 192) relabels the 192 chambers freely and transitively, so pieces become eight-element subsets of the group. Two base pieces are singled out without reference to any target; their transported subsets are the two tiles. Every piece is a left translate of one of the two tiles: 96 distinct translates per tile, each from exactly 2 translating elements, with order-two stabilizers of even permutation sign, and the translate-to-piece map is a bijection.
- The label of a translate equals the permutation sign of the translating element, well-defined by the stabilizer fact, so the label sum of a cutting is a signed permutation sum. An independent recount over the abstract group data alone (no coordinates, no chambers, no simplices) recovers 15800 and the exact census, and the tiling-to-cutting map is a bijection matching label sums object by object.
- Tile shadows give 12 positions; profiles of position multiplicities satisfy a linear system with 125 nonnegative solutions, of which exactly 25 are realized, all of constant parity (measured; the mod-two kernel has dimension 6 of 12, so level-one linear conditions cannot force it). The realized profiles fall into 6 orbits; every profile has a sign-minus stabilizer, so each per-profile census is symmetric (derived). The three census entries assemble from the orbit data as computational identities, and the 240 extreme cuttings decompose into 7 orbits with sign-plus stabilizers as a direct cross-check.

**Boundary (honest).** The six per-profile counts, the per-profile extreme counts, parity constancy of realized profiles, and the two purity facts are measured, not derived; the note says so explicitly. The note adds no citation edges.

[...]

PR #6348, opened 2026-08-14
feat: diagonal-label odd count of every cutting is divisible by four (cycle 782)

jonathonreilly:
## What this lands

Cycle 782 in the physical-cell cutting lane. Each of the 192 used pieces carries a diagonal label (the start corner of its minimal naming, one of eight values). On the declared family of 15,800 continuously certified unit-four-cube cuttings, the note proves that every odd-diagonal count is divisible by four.

## Exact finite chain

- A shifted rational grid exhaustively enumerates candidate covers; independent determinant formulas, normalized volumes, and 15,168 pairwise integer-separating-hyperplane checks certify that every emitted candidate is a continuous exact cutting rather than only a sample-grid cover.
- Eight sign classes of chambers, one per label, each hold 24 of the 192 chambers.
- The exact local law has values 1, 4, 4, 6, 0; chamber-partition telescoping gives `n_w = 24 - 4a_w - 4c_w - 6q_w`.
- A transported label preserves weight parity and identifies the six-valued class with the selected half set.
- A 20-chamber certificate makes the odd-diagonal selected-half-set count even, closing the mod-four result.

## Runner and review evidence

The standard-library runner uses exact integer/rational arithmetic and declares a 120-second audit timeout. Its landed cache records 15 PASS / 0 FAIL, 2,271 stdout bytes, and the exact final line `TOTAL: PASS=15 FAIL=0`.

Independent review recomputed all determinants with a separate 24-term formula, matched the 15,800-cover set with an alternate exact-cover algorithm, independently checked every co-occurring simplex pair, and reproduced both censuses by direct unpacked counting.

Mutation checks recorded during review:

- determinant-sign corruption -> K1G FAIL;
- strict-for-weak separation corruption -> K1G FAIL;
- chamber-incidence deletion -> K1/K3/K12 FAIL;
- telescoping coefficient `6 -> 4` -> K4 FAIL;
- transported-label bit-flip deletion -> K5/K6 FAIL;
- complement-name reversal deletion -> K13 FAIL;
- theorem modulus `4 -> 8` -> K10 FAIL;
- odd-label-set corruption -> K10/K14 FAIL;
- census-target corruption -> K9 FAIL;
- landed certificate and local-law controls -> K11/K12 discriminate as stated.

## Boundary

[...]

PR #6346, opened 2026-08-14
feat(cell-cutting): size bound theorem for the cutting label sum

jonathonreilly:
## What this lands

Cycle 781 in the physical-cell cutting lane: the size bound on the cutting label sum, derived rather than measured.

Each of the 192 staircase pieces carries a handedness label (sign of the axis order times the start-corner parity sign). The label sum S over any of the 15800 cuttings was measured to take values in {-8,-4,0,4,8} in cycle 779 ([PR #6307](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6307)) and derived mod 4 in cycle 780 ([PR #6344](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6344)). This cycle derives the bound itself:

- a halving certificate on the 192 chambers (closed form, nonzero on exactly 48 chambers) reduces S to twice its half-set part;
- a local membership certificate pins the half-set count at exactly 12 pieces in every cutting;
- a complete enumeration finds exactly 24 + 24 pairwise-disjoint nine-piece families on the two half-set sides, each sealed by an uncovered witness chamber, so no cutting contains nine disjoint pieces from one side.

Together: S = 4(p - 6) with p between 4 and 8, hence |S| <= 8 and S in {-8,-4,0,4,8}. The census constraint and the mod-four law become corollaries of one local identity.

## Runner

`scripts/physical_cell_cutting_label_sum_size_bound_cycle781_2026_08_14.py`: 14 checks, TOTAL: PASS=14 FAIL=0, deterministic, ~13 s, well under all budgets. Includes two perturbation rejectors: negating the certificate on one chamber breaks the per-piece identity at exactly its 8 holders; dropping one half-set piece breaks the constant twelve at exactly the 1975 cuttings through it.

## Review done before opening

Own re-run PASS with byte-identical stdout; the closed-form halving certificate verified pointwise (0 mismatches on all 192 chambers) against an independent referee computation; two independent family enumerations agree as sets; mechanical ban and number-discipline checks clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6344, opened 2026-08-14
feat(cell-cutting): cycle 780 covers are chambers with derived mod-four label law

jonathonreilly:
## What this PR lands

A finite physical-cell cutting theorem: one note, one exact runner, and one content-pinned cache.

**Result 1 — the 192 covers are the chambers of a wall arrangement.** The 192 eight-piece non-co-occurrence covers are identified with the chambers cut out of the open four-cube by the twelve walls x_i = x_j and x_i + x_j = 1. Exact rational sample points verify chamber incidence, and every chamber meets every cutting exactly once across 3,033,600 chamber/cutting pairs. The reviewed runner now also certifies that all 15,168 co-occurring simplex pairs have exact separating hyperplanes and that each 24-simplex selection tiles the cube by volume.

**Result 2 — a local label formula.** The signed label of a piece is computed from any incident chamber plus the piece's corner parity at two slots of that chamber's order, across 1,536 incident pairs.

**Result 3 — the mod-four law.** Two integer chamber certificates telescope through the meets-once property to force both parity counts even on every cutting. Therefore the label sum S is divisible by four on all 15,800 cuttings.

**Boundary.** The size bound |S| <= 8 and the stronger corner-parity count modulo four remain enumerated facts, not structural derivations. The result is a bounded theorem about the explicitly rebuilt finite unit four-cube object; it makes no claim about another cell, a lattice-wide construction, or a physical observable.

**Runner.** The standard-library runner rebuilds the object from the sixteen corners and executes 15 exact gates. Its refreshed envelope is pinned to the reviewed source at the declared 300-second timeout and records `status: ok`, exit 0, and `TOTAL: PASS=15 FAIL=0`. The runner now exits nonzero whenever any gate fails.

**Review confirmation.** Scratch mutations were applied independently to every K1–K15 check family. Each mutation reached its intended `FAIL` gate and returned a nonzero process status. Independent recomputations also reproduced the determinant/cost census, the 192 path/chamber classes, the local formula, both parity certificates, their global totals, and the exact simplex-separation certificate.

**Dependency boundary.** This PR is self-contained. It does not depend on PR #6307, which remains open at review time; earlier census work supplies motivation only, and every load-bearing object and identity is reconstructed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6353, opened 2026-08-14
feat(physical-cell): label sum is affine in one family trace, cycle 784

jonathonreilly:
## What this lands

One source note + one paired runner + one cached output (source-only triple), cycle 784 of the lane campaign.

**Result.** The label sum of a cutting of the physical cell is an affine function of a single family count: there is a naming-defined half-family H of 96 pieces such that every one of the 15800 cuttings meets H in exactly 12 pieces, and the label sum is S = 4(p − 6), where p is the number of trace pieces in the label-positive half of H. Twelve pieces of information collapse to one number, and that number enters linearly.

The chain, each step recomputed from the corner coordinates upward inside the runner:

- **The family.** H = the pieces whose minimal walk naming uses the last axis within its first two steps — a condition on the naming alone, nothing chosen. Size 96 of 192; the label splits it 48/48; every one of the 12 positions meets H in exactly 8 pieces and the positive half in exactly 4.
- **The two laws.** Constant trace: |T∩H| = 12 for all 15800 cuttings, no exceptions. Affine law: S = 4(p − 6), no exceptions. Three corollaries are then derived, not measured: the label-positive count is 2p, p runs exactly over 4..8, and |S| = 8 exactly when p ∈ {4, 8}.
- **The mirror.** The reflection of the first axis is one of exactly 24 sign-minus elements of the order-384 cell symmetry group that preserve H and swap its halves; it negates every label. This forces the trace layer's p ↔ 12 − p symmetry, verified entry by entry on all 6152 distinct traces: the p = 5 and p = 7 layers agree completion-count-for-completion-count.
- **The trace layer.** 6152 distinct traces; completion counts distribute as {1: 2688, 2: 1536, 3: 816, 4: 384, 5: 96, 7: 168, 8: 96, 9: 48, 10: 320} with weighted sum 15800 — and a gap: no trace has exactly 6 completions. The 240 extreme cuttings (|S| = 8) are each the **unique** completion of their trace, so extremality is a property of the 12 family pieces alone.
- **The refutation (honest negative).** The natural next guess — completion count depends only on p, so the census factorizes over the trace layer — is **false**: each interior p-class carries 6 distinct completion counts, with an explicit witness pair at p = 6 (counts 1 vs 10). The factorization question is settled negatively; whatever governs the completion count is finer than p.
- **The rejector.** Swapping a single piece of H for one outside it breaks the constant trace on 3390 cuttings and the affine law on 1975 — the laws are properties of this family, not of any 96-piece set.

**Boundary (honest).** The constant trace, the affine law, the family spread, the histogram, and the mirror's H-preservation are measured, not derived; the note says so explicitly and names the open question it does not attack (which 4 of the 8 family pieces at each position carry the positive label). The note adds no citation edges.

**Runner.** 14 gates, `TOTAL: PASS=14 FAIL=0`, exit 0, ~15 s, well under budget, stdout 2606 chars. Reviewed line by line, re-run byte-identical, and checker-verified independently of the executor.

Sibling context (previous cycles of the same campaign): [cycle 780](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6344), [cycle 781](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6346), [cycle 782](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6348), [cycle 783](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6350).

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_TRACE_LAW_COMPLETION_CYCLE784_NOTE_2026-08-14.md
+++ b/docs/PHYSICAL_CELL_CUTTING_TRACE_LAW_COMPLETION_CYCLE784_NOTE_2026-08-14.md
@@ -1,0 +1,230 @@
+# Physical cell cutting: the label sum is an affine function of one family trace
+
+Date: 2026-08-14
+Authority: none.
+Audit: unset.
+Scope: exact finite counts and identities for the declared unit four-cube cell object; no
+physical, dynamical or lattice-wide identification is claimed.
+
+This note adds no citation edges. The sibling artifacts
+`PHYSICAL_CELL_CUTTING_LABEL_SUM_SIZE_BOUND_CYCLE781_NOTE_2026-08-14`,
+`PHYSICAL_CELL_CUTTING_DIAGONAL_PARITY_CYCLE782_NOTE_2026-08-14` and
+`PHYSICAL_CELL_CUTTING_GROUP_TILING_CENSUS_CYCLE783_NOTE_2026-08-14` are written in backticks
+for context only. Nothing below depends on them: the runner shipped with this note rebuilds the
+object from the corners and produces every number quoted here, and each number is printed on a
+line of its output.
+
+## The object
+
+The cell is the unit four-cube. Its 16 corners carry 2672 five-corner subsets of unit
+determinant. The adjacency cost of such a subset has floor 6, and 400 candidates attain the
+floor; those 400 are the kept pieces. An exact cover search over 625 generic sample points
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6353 (archive/ledger/53/pr-6353.json). Carrier: The chamber/position machinery it defines is recomputed from scratch inside every later runner in the lane (each note states it imports nothing and rebuilds from the corner list), and the position coo. Branch retained; science line in the ledger entry.

PR #6355, opened 2026-08-14
feat(cell): cycle 785 parity constancy from sign characters, small-class equal split

jonathonreilly:
## What this lands

One source note + one paired runner + one cached output (source-only triple), cycle 785 of the lane campaign.

**Headline result (derived).** Parity constancy of the realized profiles — a debt sentence cycle 783 left as "measured" — is now a theorem. Identify the eight even masks with three bits (the runner gates the identification as a bijection). Because a cutting partitions the eight masks at every axis ordering, each of the seven nontrivial sign characters must sum to zero across the classes present, so the profile's parity vector lies in the mod-2 image of an exact integer kernel for each character. The runner computes all seven kernels exactly (fraction-free row reduction with a saturation gate that makes the mod-2 image provably the true one), intersects them with the level-one incidence condition, and finds a 128-vector space in which **every single vector is parity-constant** — no exception, so parity constancy follows for all 15800 cuttings. Honesty clause, gated: the seven characters are collectively sufficient but individually redundant — dropping any one leaves the intersection unchanged, and the character conditions alone already imply the level-one condition. The rejector shows the level-one condition alone is far too weak: its 262144-vector space has 253952 parity-mixed members.

**Second result (derived).** The per-profile counts of the three small classes — 24, 20 and 16, also "measured, not derived" in cycle 783 — now follow from structure:

- **Base fact.** Every class offset image lies in exactly one of the three four-element mask subgroups containing the all-axes flip; the twenty-four classes split 8/8/8.
- **Coset blocks.** The 43 affine mask subsets per class (28 pairs, 14 four-element cosets, the whole group) give 1032 blocks; exact cover with at most one block per class yields 1224 block partitions, and that set **equals** the set of coset-pure cuttings, verified in both directions.
- **Class content.** The 1224 pure cuttings contain all of B (72), all of C (120), all of D (96), 144 of A's 480, 792 of U's 9368, and none of O's 5664.
- **Equal split.** All 384 cell-group elements carry cuttings to cuttings (verified exhaustively on all 15800), the induced profile action is transitive on each class, so the class totals split evenly: 72/3 = 24, 120/6 = 20, 96/6 = 16 — the measured counts, now forced.
- **Closed form for the smallest family.** Per subgroup, 512 six-class two-covers of the ordering set; exactly 12 admit a consistent colouring under the derived edge parity; each is connected, so two colourings each: 12 × 2 = 24, reproducing the family count.

**Boundary (honest).** The U and O per-profile counts (9368, 944) and A's remaining 336 stay measured; the block-partition/pure-cutting equality is verified, not derived from first principles; O's zero purity and U's 792 are unexplained; the parity theorem constrains parity only, not the counts themselves. The note says all of this explicitly and adds no citation edges.

**Runner.** 16 gates, `TOTAL: PASS=16 FAIL=0`, exit 0, ~18 s, well under budget, stdout 2476 chars. Reviewed line by line, re-run independently, and checker-verified separately from the executor.

Sibling context (previous cycles of the same campaign): [cycle 780](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6344), [cycle 781](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6346), [cycle 782](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6348), [cycle 783](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6350), [cycle 784](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/6353).

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_PARITY_AND_SMALL_CLASS_DERIVATION_CYCLE785_NOTE_2026-08-14.md
+++ b/docs/PHYSICAL_CELL_CUTTING_PARITY_AND_SMALL_CLASS_DERIVATION_CYCLE785_NOTE_2026-08-14.md
@@ -1,0 +1,229 @@
+# Physical cell cutting: parity constancy is a theorem of the sign characters, and the small class counts follow from coset blocks
+
+Date: 2026-08-14
+Authority: none.
+Audit: unset.
+
+Scope: the cuttings of the open unit four-cube cell, relabelled as arrangements of translates of
+24 tile classes inside an order-192 group. Two debts left open by the census note are paid here.
+The parity of a cutting's class counts is shown to lie in the intersection of the kernels of seven
+sign characters, and every vector of that intersection is parity-constant, so parity constancy of
+the realized profiles becomes a theorem. The per-profile counts 24, 20 and 16 of the three small
+profile classes are then obtained from an independent enumeration of coset blocks together with an
+equal-split argument that the symmetry group supplies.
+
+This note adds no citation edges. Sibling artifacts are named in backticks for context only, never
+as dependencies: `PHYSICAL_CELL_CUTTING_GROUP_TILING_CENSUS_CYCLE783_NOTE_2026-08-14.md`,
+`PHYSICAL_CELL_CUTTING_TRACE_LAW_COMPLETION_CYCLE784_NOTE_2026-08-14.md`.
+
+Paired runner:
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6355 (archive/ledger/55/pr-6355.json). Carrier: Not restated later - but the lane retargets at cycle 786 to the interface/boundary reading and never returns to the profile census, and the note itself records that the remaining large-class counts st. Branch retained; science line in the ledger entry.

PR #6359, opened 2026-08-15
feat(cell-cutting): interface transfer spectrum, exact factorization, evenness law (cycle 787)

jonathonreilly:
## What this PR lands

Source-only triple for cycle 787 of the lane, plus a separate `audit-infra:` commit acknowledging the new note in the citation-graph manifest (the sole sanctioned `docs/audit/data` change):

- `docs/PHYSICAL_CELL_CUTTING_INTERFACE_TRANSFER_SPECTRUM_CYCLE787_NOTE_2026-08-14.md`
- `scripts/physical_cell_cutting_interface_transfer_spectrum_cycle787_2026_08_14.py`
- `logs/runner-cache/physical_cell_cutting_interface_transfer_spectrum_cycle787_2026_08_14.txt`

## The science

Working on the unit four-cube cell: 2672 candidate five-corner pieces of unit determinant, 400 kept at the adjacency-cost floor 6, 15800 exact covers by 24 pieces, 192 used pieces, order-384 cell symmetry group, sixteen-letter facet alphabet with the 12-light/4-heavy split from the previous cycle. All checks in exact integer/rational arithmetic on geometry rebuilt from the corner list.

- **The transfer matrix.** For each axis, the 16 x 16 letter-pair matrix counts covers by the letters on the two opposite facet slots. The four axis matrices are **equal as matrices** (not merely conjugate), symmetric, and strictly positive (minimum entry 18); row sums reproduce the multiplicity census (12 rows at 862, 4 at 1364), trace 2000, and the entries take exactly 9 distinct values.
- **Commutation, derived.** Every facet-holding symmetry of the cell induces the **same** signed three-cube map on both boundary cubes — a mechanism verified with 0 misses on the non-swapping half and a full held-versus-predicted cross-check covering the swapping half. Hence the matrix commutes with the whole order-48 boundary letter action, forcing constancy on the 12 letter-pair orbits; the algebra of matrices commuting with all 48 letter maps has dimension 12, so the matrix is as symmetric as its size permits.
- **Exact spectrum, certified.** The characteristic polynomial factors over the integers as (x-50)^2 (x-42)^2 (x-10)^3 (x+54) (x^2-1090x+53968) (x^2-250x+7728)^3, certified by exact fraction-elimination determinant comparison at 17 points, kernel-dimension counts summing to 16, and a degree-8 squarefree annihilating polynomial whose six single-factor drops are each nonzero. The growth eigenvalue 545+sqrt(243057) is simple, bracketed strictly between 1038 and 1039, and dominates every other eigenvalue magnitude (all at most 214) — all in integer arithmetic.
- **The light/heavy class block.** Summing rows over the light and heavy letter classes gives a constant 2 x 2 block [[578,284],[852,512]] whose quadratic is exactly the leading factor above — the growth of the interface is carried by the class split found in the previous cycle. The positive eigenvector over-weights heavy letters relative to their multiplicity, proved by an exact integer inequality.
- **Closed strips counted two ways.** Traces of powers count closed cyclic strips of glued covers; the runner computes n = 1..6 from matrix powers and re-derives the same six numbers from a power-sum recurrence on the expanded factored polynomial — two computations sharing no arithmetic, agreeing exactly (first entries 2000 and 1233040; trace anchored to the independently computed diagonal sum).
- **Integer invariants.** The invariant-factor chain is 2 repeated eleven times, then 210, 420, 19320, 96600, 17594917200 — all even, with product equal to the determinant's absolute value; the determinant is negative, matching the single negative eigenvalue -54.
- **An evenness law with a sharp boundary.** Every entry of the matrix is even. A derived mechanism — the axis flip fixes no cover and trades sides with the identity letter map, giving a free pairing on every diagonal fiber — proves evenness for 208 of the 256 fibers. The 48 fibers with no free pairing are **exactly** the entry-36 orbit, and none of them admits any of the 96 holding maps acting with all-even orbits: the evenness of that one orbit is measured, not derived, and is named as the honest wall.

Honest boundary: the specific entry values and censuses, the axis-equality of the four matrices, the class-block values, the polynomial identity itself, and entry-36 evenness are complete measurements at the declared finite scope; the commutation mechanism, orbit constancy, diagonalisability, spectrum certification, eigenvector inequalities, strip-count identity, and the 208-fiber evenness mechanism are derived.

## Verification

- Runner: stdlib only, exact arithmetic, no file I/O; my own independent re-run exits 0 with `TOTAL: PASS=22 FAIL=0` and byte-identical output to the cache.
- Independent execution per written spec with executor self-check, followed by my own adversarial review: line-by-line read of note and runner, a fabrication hunt (matrix built from enumerated covers, spectrum certified against exact determinant evaluation rather than assumed, strip counts double-computed by disjoint methods, every gate discriminating), and a mechanical banned-language/number-discipline sweep with 0 flags. The review caught and fixed one misstatement — the minimal-polynomial degree (eight, not nine as my spec had it); gates were unaffected.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_INTERFACE_TRANSFER_SPECTRUM_CYCLE787_NOTE_2026-08-14.md
+++ b/docs/PHYSICAL_CELL_CUTTING_INTERFACE_TRANSFER_SPECTRUM_CYCLE787_NOTE_2026-08-14.md
@@ -1,0 +1,245 @@
+# Physical cell cutting: the interface transfer operator, its exact spectrum, and strip growth
+
+Date: 2026-08-14
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: test whether the growth eigenvalue and the positive eigenvector class weights have a canonical downstream consumer; none is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
+- `claim_type_reason: exact spectrum, integer invariants, strip counts and evenness certificates for the interface transfer operator of the declared unit four-cube object; no physical or lattice-wide identification`
+- `audit_required_before_effective_retained: true`
+- `bare_retained_allowed: false`
+
+## Scope and objects
+
+The declared finite object consists of the 16 corners of the unit four-cube, the 2672
+five-corner pieces of determinant one built on them, the 400 pieces surviving at the adjacency
+cost floor 6, the 15800 cuttings assembled from 24 such pieces each, the 192 pieces that occur
+in some cutting, and the 384 signed coordinate maps of the cell. These are finite-scope object
+choices, not imported physical primitives. There are no load-bearing literature, empirical,
+fitted, external-data, or repository-derived scientific inputs. The linked runner uses the
+standard library only, performs no file input or output, draws no random numbers, and carries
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6359 (archive/ledger/59/pr-6359.json). Carrier: T, its symmetry, trace 2000 and the 48 entries of value 36 are recomputed from the object in #6361, #6363, #6367, #6369 and #6370, each stating it imports nothing. The reduction of the evenness questi. Branch retained; science line in the ledger entry.

PR #6361, opened 2026-08-15
feat(cell-cutting): refute five evenness mechanism routes at the interface

jonathonreilly:
## What this PR lands

Source-only triple for cycle 788 of the lane, plus a separate `audit-infra:` commit acknowledging the new note in the citation-graph manifest (the sole sanctioned `docs/audit/data` change):

- `docs/PHYSICAL_CELL_CUTTING_INTERFACE_EVENNESS_ROUTES_CYCLE788_NOTE_2026-08-14.md`
- `scripts/physical_cell_cutting_interface_evenness_routes_cycle788_2026_08_14.py`
- `logs/runner-cache/physical_cell_cutting_interface_evenness_routes_cycle788_2026_08_14.txt`

## The science

Working on the unit four-cube cell: 2672 candidate five-corner pieces of unit determinant, 400 kept at the adjacency-cost floor 6, 15800 exact covers by 24 pieces, 192 used pieces, order-384 cell symmetry group, sixteen-letter facet alphabet, and the transfer matrix from the previous cycle — even in every entry, with the entry-36 orbit (48 fibers) as the named wall where evenness is measured, not derived. This cycle asks **what mechanism produces the evenness** and refutes five candidate routes, all in exact integer/GF(2) arithmetic on geometry rebuilt from the corner list.

- **Letters are not linear in the pieces.** Over GF(2), the sixteen letter indicator vectors on covers are compared against the span of the 192 used-piece columns (rank 88 with the all-ones vector adjoined) and against the span of the 24 tetra functionals per side (rank 10 each): **0 of 16 letters lie in either span, on either side**. No GF(2)-affine functional of the piece content computes a letter, so evenness cannot be a linear-algebra corollary at the letter level.
- **No symmetry pairs the wall.** For each of the 48 uncovered fibers, the subgroup of the 96 facet-holding maps fixing the letter pair has order **exactly 2** (census: all 48 at order 2), and in every fiber the nontrivial element acts with at least one odd cycle. Since any pairing subgroup would have to act with all-even orbits, **no subgroup of the cell symmetry group pairs any uncovered fiber** — a two-case theorem covering the wall completely.
- **Letter evenness is a corollary, not a mechanism.** With A the 16 x 24 letter-tetra incidence matrix (row sums 6, col sums 4), the letter-pair count matrix N satisfies **N = A^T T A exactly** (0 misses of 576 entries), where T is the piece-level transfer matrix. So N-evenness follows from T-evenness by transport and carries no independent information. A has GF(2) rank 10 with kernel dimension 6: a **six-dimensional invisible letter-functional quotient** that the piece level cannot see — named as the next surface. (N's value census puts 862 and 904 tied at 168 apiece; the note flags the tie as an unexplained coincidence.)
- **Frames determine letters but do not pair the wall.** Every cover meets each boundary cube in a six-piece producer set on each side (census: all 15800 covers at (6,6), sides disjoint by a five-corner argument). The induced boundary frame is **single-valued onto the letter**: 1024 side-0 classes and 6184 joint classes, 0 exceptions — frame determinacy holds, a finer carrier than the letter. But as an evenness mechanism it fails: 3368 of 6184 joint classes are odd, 856 of 1024 side-0 classes are odd, and **inside the wall all 48 fibers have odd classes under both the within-fiber joint-frame partition and the full eight-slot profile partition**.
- **Column parities.** All 192 used-piece columns have odd support over the 15800 covers, and all 24 tetra functionals have even support on both sides — certified structure the mechanism hunt must respect.
- **The exchange-distance law.** Any two covers differ in an even number of pieces: d = 48 - 2t where t is the shared count, taking exactly 19 values, the even numbers from 8 to 48 with 10 and 46 absent — two covers never share exactly 1 or exactly 19 pieces, and the least exchange swaps four pieces. Odd-degree counting on the exchange graph is refuted **at every distance-value rule (0 of 256 pass) and every cutoff below 48**; the cutoff-48 pass is the tautological complete graph and is guarded as such in the note.
- **Inside a representative fiber.** In the first entry-36 fiber (36 members), every member's nearest partner sits at distance 8, but the least-distance pairing is neither unique (tied minimizers exist) nor involutive — nearest-partner matching supplies no free pairing either.

Honest boundary: the censuses, ranks, class counts, and parities are complete measurements at the declared finite scope; the two-case stabilizer theorem, the N = A^T T A identity, frame determinacy, and the distance law are derived. The evenness of the entry-36 orbit itself remains measured, not derived — this cycle narrows the mechanism space by five refuted routes and names two open entrances: the six-dimensional invisible letter-functional quotient and frame determinacy as a finer carrier.

## Verification

- Runner: stdlib only, exact arithmetic, no file I/O; my own independent re-run exits 0 with `TOTAL: PASS=16 FAIL=0` and byte-identical output to the cache.
- Independent execution per written spec with executor self-check, followed by my own adversarial review: line-by-line read of note and runner, a fabrication hunt (the N = A^T T A check crosses two implementations sharing no arithmetic, the stabilizer walk is computed on the rebuilt group, the distance scan is complete over all values and cutoffs, every gate discriminating), and a mechanical banned-language/number-discipline sweep with 0 flags. The review caught and fixed two defects: one interior gate was tightened from an at-least-one condition to the exact count the note cites (forty-eight, both partitions), and the runner docstring's route count was corrected to match the note (five refuted routes, not six — the column parities and the distance law are certified structure, not refuted mechanisms).

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_INTERFACE_EVENNESS_ROUTES_CYCLE788_NOTE_2026-08-14.md
+++ b/docs/PHYSICAL_CELL_CUTTING_INTERFACE_EVENNESS_ROUTES_CYCLE788_NOTE_2026-08-14.md
@@ -1,0 +1,246 @@
+# Physical cell cutting: interface evenness routes, the incidence transport obstruction, boundary frames, and the exchange distance law
+
+Date: 2026-08-14
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: test whether the six-dimensional letter-functional quotient invisible to the tetra incidence carries the evenness mechanism; none is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
+- `claim_type_reason: refutation certificates for candidate evenness mechanisms, an exact incidence transport identity with a rank obstruction, boundary-frame determinacy, and the within-fiber exchange distance law for the declared unit four-cube object; no physical or lattice-wide identification`
+- `audit_required_before_effective_retained: true`
+- `bare_retained_allowed: false`
+
+## Inputs and scope
+
+The declared finite object is the one this lane has carried throughout: the 16 corners of the
+unit four-cube, the 2672 five-corner determinant-one pieces built on them, the 400 pieces
+surviving at the adjacency-cost floor 6, the 15800 cuttings of 24 pieces each that those 400
+assemble into, the 192 pieces that actually occur in a cutting, the 384 signed coordinate maps
+of the cell, and the slot-preserving subgroup of order ninety-six that fixes the two boundary
+three-cubes of axis zero as a pair. Each cutting dissects each boundary three-cube into 6 unit
+tetrahedra drawn from a support of 24, and exactly 16 such dissections occur, the same 16 on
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6361 (archive/ledger/61/pr-6361.json). Carrier: The wall it defines is attacked with strictly stronger tools by #6363 (free fold reduction, full-group rigidity), #6367, #6369 and finally #6370, which collapses all 48 fiber instances to a single can. Branch retained; science line in the ledger entry.

PR #6363, opened 2026-08-15
feat(cycle789): free-fold parity reduction on the interface wall fibers

jonathonreilly:
## What this PR lands

## The science

The object is the landed finite cell: the unit four-cube with its sixteen corners, the four hundred kept unit-determinant five-corner pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet on each of the eight interface slots, and the forty-eight transfer-matrix entries at thirty-six — the named wall, whose evenness is measured, not derived. The cell group has order three hundred eighty-four; the slot-preserving subgroup has order ninety-six.

- **Stabilizer rigidity in the whole cell group.** Every one of the forty-eight wall fibers is held setwise by exactly one nontrivial map out of all three hundred eighty-four — and that map is always side-swapping and always lies inside the slot-preserving subgroup of order ninety-six. The wall's symmetry is as small as it could be, and it is the same shape in every fiber.
- **The fold is free.** The fiber's one nontrivial symmetry acts on the sixteen corners in eight two-cycles with no fixed corner, fixes none of the four hundred kept pieces, and closes the one hundred ninety-two used pieces into ninety-six two-orbits.
- **The parity reduction.** Acting on its own fiber of thirty-six cuttings, the fold fixes fourteen and pairs the remaining twenty-two, in every fiber: thirty-six equals fourteen plus twice eleven. The six hundred seventy-two folded cuttings are all distinct, and each one splits into twelve swapped piece-pairs with no fixed piece. The wall question "why is thirty-six even" reduces to "why is fourteen even" — a strictly smaller object. The evenness of fourteen is itself measured, not derived; nothing in the note removes that gap.
- **No shortcut exists.** No fiber carries a map acting freely on its cuttings (which would have derived evenness outright), and on the sample fiber the fourteen folded cuttings are rigid: only two maps of the three hundred eighty-four hold them, and both act as the identity on them.
- **Two refutations, reported as refutations.** The joint frame classes give fourteen classes per fiber but no bijection with the fourteen folded cuttings (eight classes hold none). Forty-nine shared-count degree rules on the folded sets all fail in every fiber; the single all-passing rule is the complete graph, which is tautological and excluded.
- **The quotient sees the heavy letters, not the wall.** The letter-tetra incidence has a six-dimensional kernel over two elements whose letter types recover exactly the four heavy letters. All forty-eight folds hold the kernel, but the wall orbit's kernel signature is shared with the entry-fifty-two orbit — the quotient is blind to the wall.

The honest boundary: everything here is an exact finite computational identity on the declared object, and nothing wider. The reduction relocates the evenness question; it does not close it. The named next entrance is direct enumeration of fold-invariant cuttings in the quotient, where a fold-invariant cutting is a selection of twelve of the ninety-six used-piece two-orbits that tiles the cell.

## Verification

- Runner re-run by the orchestrator in a clean shell: exit 0, `TOTAL: PASS=14 FAIL=0`, all fourteen gates recomputing their censuses from the rebuilt geometry.
- Full note read plus line-by-line runner review (fabrication hunt over both halves; every gate checked to discriminate; all printing routed through the guarded emitter).
- Mechanical pre-review checker over note, runner source, and fresh stdout: zero flags (forbidden-string sweep, numeral containment, link inventory, line-length and stdout budgets).
- Adversarial probe on the quotient claim: an independent recomputation refuted the executor's side-suggestion that a multiset-valued kernel signature separates the wall orbit from the entry-fifty-two orbit — it does not, and the probe's exhaustion argument shows no kernel-type-valued statistic of any kind can separate the colliding orbits. The note's scoped blindness statement stands unchanged; the refutation strengthens it.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_INTERFACE_FREE_FOLD_REDUCTION_CYCLE789_NOTE_2026-08-14.md
+++ b/docs/PHYSICAL_CELL_CUTTING_INTERFACE_FREE_FOLD_REDUCTION_CYCLE789_NOTE_2026-08-14.md
@@ -1,0 +1,268 @@
+# Physical cell cutting: the free fold reduction, full-group rigidity at the interface wall, and the quotient that sees the heavy letters
+
+Date: 2026-08-14
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: enumerate the folded cuttings directly in the quotient of the cell by the free involution and test whether the evenness of the folded count is forced there; none is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
+- `claim_type_reason: exact stabilizer rigidity in the whole cell group, a free involution on the declared unit four-cube object, the parity reduction of the stubborn interface fibers, two refutation certificates, and the letter-functional quotient; no physical or lattice-wide identification`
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6363 (archive/ledger/63/pr-6363.json). Carrier: #6367 (cycle 791) takes the fold reduction one level down to a single fiber and determines the whole symmetry group of that cover instance; #6369 (cycle 792) puts every fiber into linear form; #6370 (. Branch retained; science line in the ledger entry.

PR #6369, opened 2026-08-15
feat(cell-cutting): fiber linearization law, every wall fiber cover problem is a coset minimum-weight problem (cycle 792)

jonathonreilly:
## What this PR lands

## The science

The object is the landed finite cell: the unit four-cube with its sixteen corners, the four hundred kept unit-determinant five-corner pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet on the eight interface slots, and the forty-eight transfer-matrix entries at thirty-six — the named wall. The prior cycle refuted, on one sample fiber, every permutation-symmetry carrier of the evenness of fourteen — abstract, geometric, and cell-group — and named the entrance: leave group actions and take the linear structure over the field with two elements. This cycle takes it, at all forty-eight fibers at once.

- **The linearization.** Each fiber's fourteen fold-held cuttings, written over the two-orbits of the four hundred kept pieces under that fiber's own fold, are weight-twelve vectors on exactly forty rows. The six-hundred-twenty-five-point incidence has rank thirty-two and kernel dimension eight at every one of the forty-eight fibers; the kernel is an even-weight code; and the thirteen differences of the solutions from the first span it entirely, so the affine hull of the fourteen is the full two-hundred-fifty-six-element solution coset of the all-ones system.
- **The exact characterization — the headline.** Inside that coset the minimum weight is twelve, the weight-twelve members are exactly the fourteen held cuttings, and the members that are genuine integer-sense exact covers — every one of the six hundred twenty-five sample points in exactly one selected row — are exactly the same fourteen. Since any exact cover must solve the all-ones system and the coset is the complete solution set, each forty-row instance has exactly fourteen exact covers: the count fourteen is a coset minimum-weight count.
- **The universality.** One and the same coset weight census at all forty-eight fibers — one distinct census, total two hundred fifty-six — while the forty-eight instances are pairwise distinct: forty-eight distinct fold-and-row-set pairs, forty-eight distinct solution systems, six hundred seventy-two held cuttings in the union with no cutting held twice, and within-fold row overlaps between twelve and twenty-two of the forty rows, never all forty.
- **The refuted translation.** A nonzero translation preserving a solution set must carry the first solution into the set, so the thirteen differences are the complete candidate list; all thirteen fail at every fiber — zero of six hundred twenty-four tests pass. The same refutation holds natively before any fiber is singled out: of the three hundred thirty-five candidate translations of the sample fold's three-hundred-thirty-six-cutting system, zero pass. With the prior cycle, the parity carrier of the evenness of fourteen is neither a permutation symmetry of the instance nor a translation of its solution set.

The honest boundary: everything here is an exact finite computational identity on the declared object, and nothing wider. The census itself, the rank and kernel dimension, and above all the per-fiber count fourteen stay measured, not derived. What changed is the shape of the wall question: "why is fourteen even" is now exactly "why is the minimum-weight count of this one census even" — one question rather than forty-eight, asked of a linear object of dimension eight rather than of a cover search over four hundred pieces. The named next entrance is whether the forty-eight kernel-and-coset pairs are equivalent under a relabelling of rows — the cell group cannot carry it, since it already fails to move one fiber's solutions — which would descend the parity question to a single kernel and its distinguished coset.

## Verification

- Runner re-run by the orchestrator in a clean shell: exit 0, `TOTAL: PASS=12 FAIL=0`, stdout identical line-for-line to the hand-back, every census recomputed from the rebuilt geometry.
- Full note read plus line-by-line runner review (fabrication hunt; the exact-cover test verified integer-sense point-by-point, not a parity proxy; the kernel verified a real full enumeration with a basis-independence gate; the translation candidate lists verified derived from the solutions with the completeness argument stated in the note; the construction verified rebuilt from the corner list before any gate runs).
- The coset-exhaustiveness inference (exact covers lie in the enumerated coset, so the scan is complete) re-derived independently by the orchestrator before the spec was written, and the universality anchors verified twice by independent probe runs before dispatch.
- Mechanical pre-review checker over note, runner source, and fresh stdout: zero flags (forbidden-string sweep, numeral containment, link inventory, line-length and stdout budgets).

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_FIBER_LINEARIZATION_LAW_CYCLE792_NOTE_2026-08-15.md
+++ b/docs/PHYSICAL_CELL_CUTTING_FIBER_LINEARIZATION_LAW_CYCLE792_NOTE_2026-08-15.md
@@ -1,0 +1,164 @@
+# The fiber linearization law: every wall fiber cover problem is a coset minimum-weight problem
+
+Date: 2026-08-15
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: derive the single fiber-independent weight census, or at least the parity of its minimum-weight count, from the kernel and its coset alone, beginning with whether the pairwise-distinct row systems carry one census because their kernel-and-coset pairs are equivalent under a relabelling of rows; none is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
+- `claim_type_reason: an exact determination, at every wall fiber of the declared finite cell, of the rank and kernel of the fiber's point-row incidence over the field with two elements, of the coset carrying its cover solutions, of the weight census of that coset, of the identification of the fold-held cuttings with the coset minimum-weight members and with its exact covers, and of the complete refutation of translation carriers at both the fiber and the native level; no physical or lattice-wide identification`
+- `audit_required_before_effective_retained: true`
+- `bare_retained_allowed: false`
+
+## Inputs and scope
+
+The declared finite object is the one this lane has carried throughout: the 16 corners of the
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Reviewed head: 48989a7a49 (feat 1e8e45ab7e + audit-infra 48989a7a49). Runner sha256: b3ad3e99d2b17c8f076ac699d7a4d3a2e2006f4a41898ffade0bf6cdd0922881. Orchestrator re-run in a clean shell: exit 0, TOTAL: PASS=12 FAIL=0, stdout line-identical to the reviewed cache.

JONbridgerapps:
Consolidated to the archive: ledger id pr-6369 (archive/ledger/69/pr-6369.json). Carrier: #6370 (cycle 793) strengthens this from per-fiber linear form to a single instance: it shows the kernel weight census, the kernel and coset column profiles, and the cover and pairwise-intersection inv. Branch retained; science line in the ledger entry.

PR #6374, opened 2026-08-15
feat: cycle 794 clean-exchange symmetry and the parity descent to two cuttings

jonathonreilly:
## What this PR lands

## The science

The object is the landed finite cell: the unit four-cube with its sixteen corners, the four hundred kept unit-determinant five-corner pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet on the eight interface slots, and the forty-eight transfer-matrix entries at thirty-six — the named wall. The prior cycle showed the forty-eight fiber systems are ONE instance in forty-eight labellings, carrying exactly one nontrivial involution. This cycle finds where that involution comes from and drives the parity of the fourteen held cuttings down to two of them.

- **The minimal exchanges are the light kernel vectors.** The pairs of held cuttings sharing ten of their twelve rows form an eleven-edge graph at every fiber, and the eleven edge differences take exactly four distinct values — precisely the four weight-four vectors of the fiber's own kernel. The unique pair sharing nine rows differs by the unique weight-six kernel vector.
- **Four equal-union exchanges, found independently of the fourteen.** Over the forty rows there are exactly four two-for-two exchanges — two disjoint row pairs with equal point union — and their supports are exactly the same four weight-four kernel vectors. The held set is closed under all four toggles (zero images land outside), with the inside count equal to twice the pair multiplicity for every difference.
- **One of them is the whole symmetry — the headline.** Exactly one of the four exchanges is clean: it breaks no held cutting it does not move. A lemma proved in the note carries cleanliness to coordinate level — both half-pairings extend to self-equivalences of the instance, verified by explicit image on the fourteen held cuttings, all two hundred fifty-six kernel vectors, and all two hundred fifty-six coset members. A complete backtracking search then returns exactly two self-equivalences — the identity and the clean toggle — so the one clean exchange accounts for the instance's entire symmetry.
- **The parity of fourteen descends to two cuttings.** For every nonzero kernel vector the count of held cuttings it pairs is even — derived, from fixed-point-freeness of adding a nonzero vector on a coset — and equals twice the pair multiplicity, checked over all two hundred fifty-five. The maximum multiplicity six is attained by exactly one vector; that dominant exchange pairs twelve of the fourteen and strands exactly two — the two it breaks, which share eight of their twelve rows. The wall question "why is fourteen even" is now "why are these two cuttings two".

The honest boundary: everything here is an exact finite computational identity on the declared object, and nothing wider; the counts and censuses stay measured, not derived. One designed anchor missed honestly: the spec pinned the exchanges' point-union cardinalities to the sample fiber's values, and the executor's run reported them fiber-dependent — the forty-eight fibers split eight, thirty-two, eight by cardinality class — failing that gate as designed. The orchestrator re-anchored the gate to the full measured census (deterministic across three independent runs, byte-identical stdout) with matching note edits; these cardinalities are labelling properties, and nothing else in the note depends on them. The named next entrance is the anatomy of the two exceptional cuttings: whether their weight-eight joining kernel vector is distinguished in the kernel, and whether any involution pairs them — a pairing with a named reason would carry the evenness of fourteen outright.

## Verification

- Runner re-run by the orchestrator in a clean shell before the re-anchor: stdout byte-identical to the executor's hand-back. After the re-anchor: exit zero, `TOTAL: PASS=10 FAIL=0`, byte-stable across consecutive runs, every census byte-identical to the pre-fix measurement.
- Full note read plus line-by-line runner review (fabrication hunt): every census recomputed from the rebuilt geometry; the kernel row-reduced from each fiber's own incidence; equal-union tests compare point bitmasks, never sizes; the search's feasibility screens live inside the entry point; witnesses built from the half-pairing and checked against independently computed frozen targets, never derived from them.
- The clean-exchange lemma re-derived independently by the orchestrator before the spec was written; the executor ran its own self-check before hand-back.
- Mechanical pre-review checker over note, runner source, and fresh stdout: zero flags (it caught one over-long docstring line, fixed before landing).

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_CLEAN_EXCHANGE_SYMMETRY_CYCLE794_NOTE_2026-08-15.md
+++ b/docs/PHYSICAL_CELL_CUTTING_CLEAN_EXCHANGE_SYMMETRY_CYCLE794_NOTE_2026-08-15.md
@@ -1,0 +1,208 @@
+# The clean exchange: four minimal exchanges, one of them the whole symmetry, and the parity of fourteen down to two cuttings
+
+Date: 2026-08-15
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: read the anatomy of the two held cuttings that the dominant exchange leaves unpaired — what distinguishes them from the twelve it pairs, whether the kernel vector joining them is itself structured, and whether a natively named involution pairs them, since such a pairing would carry the evenness of fourteen outright; none is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Reviewed head: `f5bf7a567c` (feat commit `2a7f8dcb3d`). Runner sha256: `0b9f0c43d15762054bac7abea5f4f97d52bb98a5629732db57d0b4e5fb7b5978`. Orchestrator re-ran the runner at this head in a clean shell: exit 0, `TOTAL: PASS=10 FAIL=0`, stdout byte-identical to the committed cache body.

JONbridgerapps:
Consolidated to the archive: ledger id pr-6374 (archive/ledger/74/pr-6374.json). Carrier: #6375 (cycle 795) takes exactly the two exceptional cuttings this note isolates and gives their full anatomy, and it strengthens the outcome into a rigidity result (a degree lemma forbidding ANY share. Branch retained; science line in the ledger entry.

PR #6370, opened 2026-08-15
feat: cycle 793 — the wall instance is one: every fiber system is a relabelling of a single coset problem

jonathonreilly:
## What this PR lands

One source note, one paired runner, one runner cache (source-only triple), plus the separate audit-infra commit acknowledging the new note in the citation-graph manifest.

## The science

The object is the landed finite cell: the unit four-cube with its sixteen corners, the four hundred kept unit-determinant five-corner pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet on the eight interface slots, and the forty-eight transfer-matrix entries at thirty-six — the named wall. The prior cycle put every fiber into linear form over the field with two elements and found one coset weight census at all forty-eight fibers, while the forty-eight labelled systems stayed pairwise distinct. This cycle answers the question that left open: the forty-eight systems are not forty-eight objects that happen to agree — they are ONE object in forty-eight labellings.

- **The further invariants are single-valued.** Beyond the coset weight census, four more invariants take exactly one distinct value over all forty-eight fibers: the weight census of the kernel itself (total two hundred fifty-six), both column profiles (every one of the forty coordinates is one on exactly one hundred twenty-eight kernel vectors and one hundred twenty-eight coset members), the per-coordinate cover profile of the fourteen held cuttings (total incidence one hundred sixty-eight, which is fourteen times twelve), and the pairwise support-intersection census over the ninety-one pairs.
- **The equivalence is real, not inferred — the headline.** A complete backtracking search finds, for the sample fiber against every one of the other forty-seven, an explicit bijection of the forty coordinates, each verified by explicit image: all two hundred fifty-six kernel vectors map onto the target's kernel as a set, the fourteen held cuttings onto the target's fourteen, and all two hundred fifty-six coset members onto the target's coset. Completeness is licensed by a lemma proved in the note: instance equivalence is exactly equivalence of the fourteen-block systems, because weight preservation carries the fourteen forward and the difference span (regated set-level at all forty-eight fibers) carries kernel and coset backward.
- **The count is two, for a checked reason.** Enumerating every equivalence gives exactly two per target, matching the sample instance's exactly two symmetries — the identity and one nontrivial involution — and composing that involution with either of a target's two relabellings yields the other at all forty-seven targets, so composition acts freely and transitively.
[...]

PR #6367, opened 2026-08-15
feat(cell-cutting): single-fiber cover rigidity, the induced symmetry group of the fourteen-solution system (cycle 791)

jonathonreilly:
## What this PR lands

One source note, one paired runner, one runner cache (source-only triple), plus the separate audit-infra commit acknowledging the new note in the citation-graph manifest.

## The science

The object is the landed finite cell: the unit four-cube with its sixteen corners, the four hundred kept unit-determinant five-corner pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet on the eight interface slots, and the forty-eight transfer-matrix entries at thirty-six — the named wall. The prior cycle derived the evenness of the global fold-held count by a free pairing, showed the pairing never descends to a single fiber, and named the entrance: hunt a free involution on a single fiber's cover problem, or refute it. This cycle refutes it — at three distinct levels.

- **The fiber as a cover instance.** The sample fiber's fourteen fold-held cuttings, written as twelve-row subsets of the two hundred kept-piece two-orbits, use exactly forty distinct rows; the ninety-one solution pairs obey the exact identity that the symmetric difference is twenty-four minus twice the meet; the forty rows fall into twenty classes of equal membership, whose within-class swaps form a kernel of order twenty-one million two hundred thirty-three thousand six hundred sixty-four, every one holding all fourteen solutions.
- **The abstract refutation — the headline.** A complete unseeded backtracking over all permutations of the fourteen solutions leaves exactly two induced maps: the identity and one involution with two two-cycles and ten fixed points. Free involutions number zero. The whole group of row maps preserving the system has order forty-two million four hundred sixty-seven thousand three hundred twenty-eight.
- **The geometric refutation.** The forty rows cut the six hundred twenty-five sample points into sixty-eight incidence patterns with no point uncovered; of the whole group's forty-two million row maps, exactly one is carried by a point permutation — the identity. The involution has no geometric realization at all.
- **The cell-group refutation.** Of the three hundred eighty-four cell maps, exactly two hold the fiber's fourteen-set setwise, and both induce the identity on it. The cell group is blind to even the one nontrivial abstract symmetry.
[...]

PR #6378, opened 2026-08-15
feat: cycle 796 wall homogeneity and the fold as entry stabilizer

jonathonreilly:
## What this PR lands

## The science

The object is the landed finite cell: the unit four-cube with sixteen corners, the four hundred kept pieces at adjacency-cost floor six, the fifteen thousand eight hundred exact twenty-four-piece cuttings, the sixteen-letter facet alphabet, and the forty-eight transfer-matrix entries at thirty-six — the named wall. Until now the wall has been a list of forty-eight per-fiber systems that happen to look alike, each with a fold found by search one fiber at a time. This cycle asks whether the cell's own symmetry group acts on the wall itself, and the answer reorganizes the whole picture.

- **Exactly ninety-six of the three hundred eighty-four cell maps act on the wall — precisely those fixing the first axis.** Tested map by map: every one of the ninety-six maps fixing the letter-carrying axis induces a well-defined map on the forty-eight entries, and all two hundred eighty-eight others fail. The induced maps compose in the order the maps are applied at all nine thousand two hundred sixteen ordered pairs; the reverse ordering holds at only one thousand nine hundred twenty, so the orientation is measured, not assumed.
- **The wall is homogeneous.** The action is transitive: the forty-eight entries form a single orbit, so every wall entry is carried to every other by a symmetry of the cell. The forty-eight per-fiber systems are not merely isomorphic — they are one object seen from forty-eight positions of the cell.
- **The fold is derived, no longer searched.** Ninety-six acting maps over a single orbit of forty-eight forces every entry stabilizer to have order exactly two, and the nontrivial stabilizer element equals the fold from the independent per-fiber search at all forty-eight entries. What an earlier cycle could only measure — each fiber held setwise by exactly two maps — now follows from the orbit count.
- **The six folds are a single conjugacy class, transported by the action.** The fold of an image entry is the conjugate of the source fold by the acting map at all four thousand six hundred eight map-entry pairs, with every inverse found by search rather than assumed; the six distinct folds form one conjugacy class inside the acting ninety-six, each serving exactly eight fibers.
- **The light, middle and heavy split is frame-stable yet moved by the action.** The eight, thirty-two, eight split of the entries by exchange union sizes (fifty, one hundred, one hundred seventy-five — all multiples of twenty-five) reproduces exactly, member for member, on a second generic point frame; yet searched maps carry a light entry to a middle one and to a heavy one. Since the wall is homogeneous, any quantity intrinsic to the abstract instance must be constant across it, so the split is a property of how the instance sits relative to the point frame and the letter labels — and the note names the letter level of the interface matrix as the next entrance.

Honest boundary: the exactly-ninety-six count, the class sizes, the letter-pair lists, and the union values stay measured, not derived; the derived content is the stabilizer order from the counting identity, the conjugation law verified on every pair rather than sampled, and the embedding-relativity argument from transitivity. Frame stability is established at one additional generic frame, not proven for all frames. The executor caught two of its own mistakes before hand-back: the source entry for the transport search must be selected as an ordered letter pair (each unordered pair occupies two wall entries), and an anchor table first written in the wrong representation produced a legitimate gate failure that was fixed as representation only. Transporters are reported as the first found in construction order, not as canonical choices.

## Verification

- Runner re-run by the orchestrator in a clean shell: exit zero, `TOTAL: PASS=10 FAIL=0`, stdout byte-identical to the executor's hand-back (itself byte-identical across the executor's two runs), and byte-identical again after the orchestrator strengthened one gate in review.
- Full note read plus line-by-line runner review (fabrication hunt): the fold table and the wall action are computed by separate identification logic, neither reading the other's result; the composition-orientation gate is discriminating rather than tautological; anchors are compared against recomputed values, never used to build them; the second-frame masks are genuinely rebuilt from the new offsets.
- Review edits by the orchestrator: one note sentence tightened, and the per-fiber anatomy values printed in the fold gate's detail lines are now pinned by that gate as well.
- Mechanical pre-review checker over note, runner source, and fresh stdout: zero flags.

### DIFF
--- a/docs/PHYSICAL_CELL_CUTTING_WALL_HOMOGENEITY_CYCLE796_NOTE_2026-08-15.md
+++ b/docs/PHYSICAL_CELL_CUTTING_WALL_HOMOGENEITY_CYCLE796_NOTE_2026-08-15.md
@@ -1,0 +1,193 @@
+# Physical cell cutting: wall homogeneity, the fold as entry stabilizer, and the frame-stable class split
+
+Date: 2026-08-15
+Authority: none
+Audit: unset
+Status: proposed_retained
+Claim type: bounded_theorem
+Constitutional effect: none.
+
+## Trace gate
+
+- `trace_class: frontier_discovery`
+- `target_claim_id: null`
+- `target_blocker_text: null`
+- `source_of_blocker_text: frontier_question`
+- `reachability_to_target: unknown_frontier`
+- `artifact_role: theorem`
+- `next_trace_action: read the letter level of the interface matrix directly — why the 48 entries of value 36 are the letter pairs they are, why the acting maps move an entry between the light, middle and heavy classes of union size while the class sizes stay put, and where the union values 50, 100 and 175 come from as multiples of 25; none of that is claimed here`
+
+## Status contract
+
+- `actual_current_surface_status: bounded-support`
+- `target_claim_type: bounded_theorem`
+- `trace_class: frontier_discovery`
+- `reachability_to_target: unknown_frontier`
+- `conditional_surface_status: null`
+- `hypothetical_axiom_status: null`
+- `admitted_observation_status: null`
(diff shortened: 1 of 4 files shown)


jonathonreilly:
Reviewed head: 6f90079236 (feat) + b876b72afe (audit-infra). Runner sha256: 2959e15c50fcc82cbcd8dd61e98b3a42ef30f59367e70db3c7a5b1defe903483. Orchestrator re-ran the runner in a clean shell after the review edits: exit zero, TOTAL: PASS=10 FAIL=0, stdout byte-identical to both executor runs.

JONbridgerapps:
Consolidated to the archive: ledger id pr-6378 (archive/ledger/78/pr-6378.json). Carrier: The fold structure (48 folds collapsing to 6 maps with serving census {8:6}, each holding 14 of 36) is recomputed independently in #6367, #6369, #6370 and #6374 - this note supplies the group-theoreti. Branch retained; science line in the ledger entry.

PR #6633, opened 2026-08-16
physics: complete the half-space pairing with the reflection intertwiner

JONbridgerapps:
## Block 119 — The Reflection Intertwiner Completion

Stacked on #6624 (Block 118). Fifth-file discipline: bounded_theorem note,
runner, cache, block ledger, refreshed citation manifest.

### Result

Block 118 named the reflection intertwiner as the object that could
complete the rank-one geometric-Hankel action pairing. This block decides
it, two-sidedly:

- **The selection wall.** Every displayed fixed carrier-natural candidate
  fails the exact proportionality condition `Theta x_k ∝ y_k` at every
  momentum and both fixtures: the Block 114 dressing (direct and
  OS-reflected), the overlap Hodge, the reality conjugation composed with
  complex conjugation, the Klein and parity operators, `-I`, and their
  displayed simple products — fourteen rows, with momentum-mapping
  candidates (Hodge, parity) tested against their correct target sectors.
  All failures already occur at the first residual entry.
- **The completion exists.** The swap completion
  `Theta = I + V(S_swap − I)(V^H V)^{-1} V^H` on the four-column span
  `V = [x, y, Rx, Ry]` of the action's stable boundary data and its
  reality images is involutive *structurally* (the swap squares to the
  identity on the span; identity off it), sends `x ↔ y` with `mu = 1`
  (real, positive — exactly what the PSD reduction requires), is
  per-sector reflection-real at the self-conjugate momenta, and satisfies
  the coupled identity `Theta_3 = R Theta_1 R` at the paired momenta —
  forced, because R's column permutation `(0↔2)(1↔3)` commutes with the
  swap `(0↔1)(2↔3)`.
- **The positive package.** The completed kernel is exactly
  `rho^{m+n} y y^H` — Hermitian PSD with inertia `(1,0,23)` per momentum
  and `(4,0,92)` globally on the displayed three super-cells (one genuine
  positive direction per momentum: the rank-one moment structure). The
  quotient transfer is `diag(rho_k^2)` with all four contractive values
  pinned in `(0,1)` by exact isolating intervals, and the geometric
  semigroup law `T^n = diag(rho_k^{2n})` holds exactly — resolving, on
  this carrier, the non-semigroup pathology Block 116 found for the
  chart windows. **This is the campaign's first positive and contractive
  finite OS package.**
- **Firewalls.** Half-space carrier only; the completion is built from
  the pairing's own boundary data (action-derived in that bounded sense)
[...]

PR #6647, opened 2026-08-16
physics: isolate the torus wrap defect and pin the half-space carrier

JONbridgerapps:
## Block 120 — The Torus Wrap Defect And The Half-Space Carrier

Stacked on #6633 (Block 119). Fifth-file discipline: bounded_theorem note,
runner, cache, block ledger, refreshed citation manifest.

### Result

Block 119 completed the half-space pairing; this block answers what
happens on the original closed carrier — and pins the OS package's
correct home:

- **The torus rejects the completion.** Dressing the literal antiperiodic
  `Z8` torus pairing with the Block 119 swap family leaves it
  non-Hermitian with residual rank four per momentum: no inertia.
- **The keystone split.** `K_torus = K_half + D` exactly, with the wrap
  operator `R_AP[n,j] = -U[n,0](I+M^2)^{-1}U[4,j+1]e/C_j`, and **all**
  dressed non-Hermiticity lies in `D` — the half-space part is exactly
  Hermitian after the swap dressing. (These two split claims are
  recomputed live in the runner, entrywise.)
- **The vanishing pencil.** The completed half-space window pencil is the
  zero polynomial — both window determinants and both cross coefficients
  vanish identically — so the vacuum sits exactly at one in the
  strongest sense, and the torus pencil value at one is exactly the wrap
  contribution, which is nonzero.
- **The finite-size law and the saturation.** The wrap obeys the exact
  projector decomposition `(I+T^N)^{-1} = P_s/(1+a^N) + P_u/(1+a^{-N})`
  with `T = M^2` and `a = rho_F^2 in (0,1)`: the stable channel's
  coefficient `c_N = a^N/(1+a^N)` decays geometrically in torus length,
  but the rank-one unstable-projector channel **saturates** at an
  N-independent limit — on any fixed torus, the antiperiodic boundary
  speaks through the growing mode. Only the genuine half-space removes
  it.
- **The corrected completion.** Subtracting the displayed defect
  restores `y y^H` exactly: Hermitian PSD, inertia `(1,0,3)` per
  momentum on the window, contractive quotient `beta = a^2 = rho_F^4`
  per double period — the same geometric semigroup `rho_F^{2n}` as
  Block 119's per-period `rho_F^2` (bookkeeping at doubled step, not a
  discrepancy; the reconciliation identity is certified).
- **The carrier verdict.** The finite OS package's correct carrier is
  the half-space / large-torus limit. W1 is a carrier-transfer wall for
  the displayed dressing family and subtraction only — not an OS no-go,
  and the positive half-space theorem stands untouched.

### Verification

- Runner: 8/8 PASS (~200 s); the split and Hermiticity-localization
  claims recomputed live entrywise; the wrap operator identified against
  the dressed defect exactly; projector law certified for N = 1, 2, 3;
  every interval pin re-verified by exact root counting. Mutation sweep:
  15/15 clean. N5 byte-identical.
- Independent cross-model verification (Claude fraction checker): torus
  failure, vanishing pencil (confirmed in a stronger form — the zero
  polynomial — while catching and correcting its own earlier
  reflected-arm artifact), the projector law, the unstable-saturation
  characterization, and the power reconciliation all CONFIRMED; the two
  entrywise split claims were budget-limited on the checker side and are
  covered by the runner's live recompute plus the solve lane's 792
  checks.
  epoch-debt gate (queue item
  `2026-08-08-dependency-policy-epoch-debt-helper-registry`); identical
  on the clean base — unrelated to this PR.

### TOE accounting

referral for any verdict.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block120-torus-wrap-defect-20260816/NO_GO_LEDGER.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block120-torus-wrap-defect-20260816/NO_GO_LEDGER.md
@@ -1,0 +1,37 @@
+# Block 120 No-Go Ledger
+
+Scope: one narrow carrier-transfer wall inside a split-and-restore
+packet that preserves the positive half-space theorem. `W1` says the
+Block 119 swap family does not complete the literal antiperiodic Z8
+torus pairing — proven by exact mechanism, not search: the dressed
+torus pairing is non-Hermitian with residual rank four per momentum
+(no inertia), and the torus kernel splits exactly as
+K_torus = K_half + D with the displayed antiperiodic wrap operator
+R_AP[n,j] = -U[n,0](I+M^2)^{-1}U[4,j+1]e/C_j, where D carries ALL of
+the dressed non-Hermiticity (the half part is exactly Hermitian). The
+completed half-space window pencil vanishes identically (zero
+polynomial — the vacuum sits exactly at one), so the torus pencil value
+at one is exactly the wrap contribution, which is nonzero. The wrap
+obeys the exact projector law (I+T^N)^{-1} = P_s/(1+a^N) +
+P_u/(1+a^{-N}) with T = M^2 and a = rho_F^2 in (0,1): the
+stable-channel coefficient c_N = a^N/(1+a^N) decays geometrically,
+while the rank-one unstable-projector channel saturates at an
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6647 (archive/ledger/47/pr-6647.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #6651, opened 2026-08-16
physics: couple the constraint quotient and prove the population wall

JONbridgerapps:
## Block 121 — The Constraint-Quotient Coupling And The Closed-Carrier Gauss Obstruction

Stacked on #6647 (Block 120). First block of the constraint-quotient
contract (Block 105 §12 item 4). Fifth-file discipline: bounded_theorem
note, runner, cache, block ledger, refreshed citation manifest.

### Result

Block 105 named the final gate: "only then couple the physical gravity
constraint quotient and ask whether a retained obligation can move."
This block executes the coupling and proves exactly which half moves:

- **The discrete Noether identity.** The routed U(1) current of the
  Dirac–Kähler action obeys
  `Delta_t^- J + Delta_x^- S = phibar [E_z, Q] psi` as an exact
  *off-shell* matrix identity at all 32 sites — including the
  antiperiodic seam (whose sign lives inside Q) — at both fixtures. The
  current is unique only up to a discrete curl, and the identity is
  provably routing-independent: two genuinely different routings both
  reproduce the commutator, and their difference is an exact
  divergence-free curl. On-shell the right side vanishes by row/column
  ideal factorization.
- **The d=2 constraint count.** The symmetric-perturbation constraint
  `C_k = (1, kappa_k)^T` has (rank, kernel, cokernel) = (1, 0, 1) at
  every spatial momentum: zero TT coordinates everywhere. At `k = 0` the
  divergence row degenerates but the trace row alone has trivial kernel,
  so the count survives with the active-constraint refinement.
- **The Gauss intertwiner.** Constraint preservation under
  `Gamma_{t+1} = Gamma_t - (Delta_t rho + B_4 S)` is exactly equivalent
  to the source continuity identity — which the DK current satisfies
  on-shell. The coupling is exact at the local/density level.
- **The population wall (W1).** The closed carrier's sourced quotient is
  unpopulated: `sum_z E_z = I` exactly, so the total U(1) charge form on
  the certified positive package is the norm `||v||^2`, while the closed
  `Z4` incidence image is exactly the zero-sum subspace — periodic Gauss
  solvability forces the zero state. This is the third independent
  appearance of the campaign's closed-carrier obstruction (after the
  torus wrap defect and the TT lane's Record-charge finding).
  Background-neutralized sources, open carriers, the half-space
  boundary-flux sector, and the physical stress tensor are all outside
  the wall and named live.
- **The honesty split.** The total-charge transfer commutation
  (`[T^n, Q_U(1)] = 0`) is vacuous — the U(1) generator is the identity
  on field space and cannot fail to commute — and is explicitly *not*
  claimed as content. The genuine conserved-coupling object is the
  charge density (`[E_z, Q]` has up to 16 nonzero entries per site); the
  Ward/transfer-covariance of the current bilinears on the OS quotient
  is the named next gate. The coupling is r-blind: the r-weighted TT
  block is empty in d=2, so no matter datum selects the d=3+1 family.

### Verification

- Runner: 8/8 PASS (~5 s); recomputes the routed currents in both
  conventions, the 128 fixture/site identity cases, the curl
  reconstruction, the seam sign structure, the ideal factorization, the
  mode-wise constraint count with the k=0 certificate, the intertwiner
  algebra, the zero-sum incidence image, and the vacuity split.
  Mutation sweep: 15/15 clean (one gate-independence defect caught and
  fixed in review). N5 byte-identical.
- Independent cross-model verification (Claude fraction checker): the
  divergence-commutator identity confirmed at 32/32 sites for two
  independent routings at both fixtures with the routing-independence
  argument; the k=0 refinement; the intertwiner equivalence; the
  decisive negative (norm total charge vs zero-sum Gauss image) — and
  the vacuity refutation of the worker's original charge-commutation
  claim, which reshaped this note's claims to the density level.
  epoch-debt gate (queue item
  `2026-08-08-dependency-policy-epoch-debt-helper-registry`); identical
  on the clean base — unrelated to this PR.

### TOE accounting

referral for any verdict.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block121-constraint-quotient-coupling-20260816/NO_GO_LEDGER.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block121-constraint-quotient-coupling-20260816/NO_GO_LEDGER.md
@@ -1,0 +1,40 @@
+# Block 121 No-Go Ledger
+
+Scope: one narrow source-population wall inside a coupling packet whose
+density-level side is positive. `W1` says the certified positive matter
+package cannot populate the sourced gravity constraint quotient on the
+displayed closed d=2 carrier — proven by exact mechanism: the closed-Z4
+incidence image is exactly the zero-sum subspace (rank 3, cokernel the
+constants), while sum_z E_z = I makes the total U(1) charge form the
+norm ||v||^2, strictly positive on every nonzero state of the certified
+positive package — so periodic Gauss solvability forces the zero state.
+The coupling itself is exact and positive at density level: the routed
+U(1) current obeys the off-shell divergence-commutator identity
+Delta_t^- J + Delta_x^- S = phibar[E_z,Q]psi at every site including
+the antiperiodic seam, independently of routing (two displayed routings
+differ by an exact divergence-free curl); the identity vanishes
+on-shell (row/column-ideal factorization); the d=2 constraint sector
+has zero TT coordinates at every momentum (trace row alone active at
+k=0); and the Gauss intertwiner makes constraint preservation exactly
+equivalent to source continuity. The total-charge transfer commutation
+is vacuous ([I,Q] = 0 by construction) and is NOT claimed as content;
+the charge density is genuinely nontrivial ([E_z,Q] nonzero, up to 16
+entries). The coupling is r-blind (the r-weighted TT block is empty in
+d=2). Narrow scope: the displayed carrier, fixtures, the U(1)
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6651 (archive/ledger/51/pr-6651.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #6705, opened 2026-08-17
physics: execute the sourced quotient on the stable balanced sector

JONbridgerapps:
## Block 124 — The Executed Sourced Quotient And The Stable Balanced Sector

Stacked on #6700 (Block 123). Fifth-file discipline: bounded_theorem
note, runner, cache, block ledger, refreshed citation manifest.

### Result

Block 123 proved the momentum constraint populatable; this block
executes the sourced quotient — the lane's first physical gravity
quotient computation:

- **The class quadric.** Vanishing expected total momentum on the
  certified 4-dim quotient is `|c_1|^2 + 2|c_2|^2 = |c_3|^2` with `c_0`
  free — cone/normalized/ray dimensions 7/6/5 (convention-invariant).
  The `p_2` sign convention is load-bearing off `c_2 = 0`: the two
  conventions' classes are different sets intersecting exactly there,
  so displayed claims restrict to the sector where they agree.
- **The Gauss solution.** On the closed `Z4` cycle,
  `Gamma = (0, -r_0, -r_0-r_1, r_3) + lambda(1,1,1,1)` under the pinned
  forward-difference orientation (the opposite orientation provably
  fails the displayed witness), unique up to the constant gauge mode.
- **The physical quotient.** Isomorphic to the matter class — gravity
  contributes zero independent dimensions after quotienting its gauge
  mode, the d=2 zero-TT count made manifest.
- **The stable balanced sector (the keystone).** The transfer preserves
  exactly `{c_2 = 0, |c_1| = |c_3|}`: the paired momenta share one
  characteristic polynomial whose stable root is uniquely pinned by
  `|rho| < 1` — the identification `rho_1 = rho_3` is forced, not
  chosen — while unbalanced states violate the class by exactly
  `2|c_2|^2(rho_even^4 - rho_odd^4)` per step. **This transfer-stable
  sector coincides with the convention-free sector**: the physical
  gravity quotient that ships is simultaneously convention-independent
  and dynamical.
- **The localization verdict.** Cosmetic, not structural: every
  admissible Hermitian localization satisfies the sum rule
  `sum_x D_x = P`, so the class, Gauss solvability, and the invariant
  sector depend only on the ordering-independent total; only the
  spatial profile varies (and the `E_x P E_x` ordering is outright
  inadmissible — it fails the sum rule).

### Verification

- Runner: 8/8 PASS (~58 s); recomputes the quadric under both
  conventions with their intersection, the symbolic Gauss solution with
[...]

PR #6718, opened 2026-08-17
physics: prove the spatial-dressing invisibility at the quotient

JONbridgerapps:
## Block 125 — The Spatial-Dressing Invisibility And The Wall Extension

Stacked on #6705 (Block 124). Fifth-file discipline: bounded_theorem
note, runner, cache, block ledger, refreshed citation manifest.

### Result

Block 124 named non-local dressings for the observable wall; this block
decides the spatial half — with the campaign's fifth verification catch
shipped in full transparency:

- **The invisibility lemma (one line of algebra).** At fixed momentum
  the spatial shift acts as the scalar phase `i^k`, which cancels under
  conjugation: the momentum-diagonal block of `U_x^w J U_x^{-w}` equals
  that of `J` for every `w`. The quotient sees only diagonal blocks —
  so it is structurally blind to spatial smearing.
- **The collapse and the wall extension.** The displayed eight-weight
  (plain + Theta-conjugated) family collapses to exactly two effective
  weight-sums; the descent kernel is precisely the zero-sum directions,
  whose momentum blocks vanish identically and compress to the zero
  observable. No member with nonzero quotient compression descends.
- **The admissible-observable spaces.** Exact dimensions 57 over
  `Q(i, rho_k)` (seven F-linear conditions on 64 entries) and 113 real
  with reflection-adjointness — the rank-one arithmetic `n² − (n−1)`,
  which also produced Block 119's intertwiner dimensions: same
  arithmetic, *different subspaces*, displayed without conflation via a
  witness operator. The wall is about the dressing class's reach, not
  an empty target.
- **The time direction is visible (the positive finding).** A
  transfer-conjugated translate has a provably different diagonal block
  — time-smeared dressings are the genuinely live route, and the
  "time-dependent counterterms are required" conclusion ships as a
  labeled inference, not a theorem.
- **The catch record.** The solve lane's original rank-eight,
  kernel-zero certificate was **refuted** in cross-model verification:
  the translates coincide, the true rank is two with a six-dimensional
  zero-observable kernel. The corrected certificate is recomputed live
  in the runner (including the explicit refutation computation), and
  the wall conclusion survived in a stronger, simpler form.

### Verification

- Runner: 8/8 PASS (~250 s); certifies the scalar-phase lemma
  symbolically and entrywise, the family factoring through weight-sums,
  the corrected rank/kernel, the zero-compression of the kernel, both
  dimension counts, the non-conflation witness, the time-visibility
  nonequality, and the explicit refutation record. Mutation sweep:
  15/15 clean. N5 byte-identical.
- Independent cross-model verification (Claude fraction checker): the
  refutation is the checker's own finding — its translate-coincidence
  test overturned the worker's central certificate while confirming the
  wall conclusion, the dimension counts (by its own derivation), and
  the coincidence explanation. What ships is the checker-corrected
  theorem.
  epoch-debt gate; identical on the clean base — unrelated to this PR.

### TOE accounting

referral for any verdict.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block125-spatial-dressing-invisibility-20260817/NO_GO_LEDGER.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block125-spatial-dressing-invisibility-20260817/NO_GO_LEDGER.md
@@ -1,0 +1,39 @@
+# Block 125 No-Go Ledger
+
+Scope: one narrow structural wall extending the observable-descent
+obstruction to the spatial-dressing class, shipped with an explicit
+verification catch record. `W1` says no spatial smearing plus Theta
+conjugation of the routed density yields a nonzero conserved
+observable on the certified quotient — proven by the invisibility
+lemma, one line of algebra: at fixed momentum the spatial shift acts
+as the scalar phase i^k, which cancels under conjugation, so the
+momentum-diagonal blocks of all spatial translates coincide and the
+displayed eight-weight family collapses to exactly two effective
+weight-sums; the descent kernel is exactly the zero-sum-weight
+directions, whose momentum blocks vanish identically and compress to
+the zero observable — no member with nonzero quotient compression
+descends. The admissible-observable spaces are large (exact dimensions
+57 over the root field and 113 real, the rank-one arithmetic
+n^2 - (n-1), explicitly not conflated with the Block 119 intertwiner
+spaces); the wall is about what the displayed dressing class can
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-6718 (archive/ledger/18/pr-6718.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #7204, opened 2026-08-22
physics: block 173 — the stationary kernel characterized (signed, domain-bounded, undefined at the recorded class)

JONbridgerapps:
## Block 173 — The Stationary-Kernel Characterization (mu* is exactly characterized, and it is NOT a law)

Stacked on Block 172 (the refinement census and the stationary kernel). Block 172 returned a **non-supply** verdict on the statistical bridge and its disjoint checker **discovered** that the record-extension stationarity system has a unique strictly-positive solution `mu*` at every point it measured. Block 172 named the **characterization of `mu*`** as this block's charter and performed none of it. This block performs it, and **this block's centerpiece is again the disjoint checker's work** — the boundary map and the `sigma = 0` non-existence are the note's climax, and the credit is carried throughout.

### THE HEADLINE — the characterization lands, and it closes the last theorem-shaped door

> **`mu*` is EXACTLY CHARACTERIZED — it is the normalized kernel of the record-influence matrix, it is genuinely transport-sensitive, and it is exactly `W9`-derivable, so it is NOT a shim — AND IT IS STILL NOT A LAW.** It is signed off-region, its positive region is a bounded and disconnected box of the same order as the committed fixture, and at the zero-shear carrier — the value the records write — **it does not exist at all**. The bridge cannot be restated as a theorem about `mu*`, and the axiom route stands on **non-supply, not measured falsity**.

Block 172's handoff put the stake as a fork: *if `mu*` is transport-sensitive and `W9`-derivable the bridge becomes a theorem ABOUT `mu*`; if `mu*` is a shim the axiom route stands.* **The answer is neither horn.**

### The closed form

`E[b,a] = P(b | base + (ell,a)) - P(b | base)` — the **record-influence matrix**. Then `M = P0 1^T + E`, the normalization row is redundant, and the stationarity system is exactly `E mu = 0` with `sum mu = 1`:

> **`mu*` is the normalized kernel of `E`**, equivalently `mu*_a = Z_a (D^-1 d0)_a / Z_0` in raw `W9` diagonals.

Verified at **14 configurations at both extents** by **four routes that must agree entry for entry** — a cofactor kernel, sympy's nullspace, the b172 checker's `5x4` augmented `linsolve`, and the raw `W9` form — with the checker's own independent route (column solves, nullspace, Sylvester minors) agreeing again. The b172 brackets reproduce, and the landed b171 identity `mu*(bench, s_t = 1/4) == mu*(holo_t, 0 + i/4)` is met by two separate dial routes.

**Two qualifications, both the checker's, both load-bearing.** `rank E = 3` is **MEASURED AND NOT PROVEN**, and it **fails in-lane**. And **existence is NOT free**: `det E = 0` identically is normalization-forced and content-free, but a kernel *vector* is not a stationary *kernel*. The correct criterion is **`P0` in the affine hull** of the extension profiles, with a displayed counterexample — four legitimate probability columns all equal to `(1/2,1/4,1/8,1/8)` with base `(1/4,1/4,1/4,1/4)`, where `rank(Aeq) = 1` against `rank(Aeq|beq) = 2` and `linsolve` returns **EmptySet**. Block 172's L3 is **partially demoted** and the demotion is disclosed.

### Transport — not a shim, and strictly finer than its own input

`mu*` moves on all five registered dials at both extents (`> 1/53, 1/9, 1/28, 1/7, 1/13` at 12x4). At 8x4 the one-dial closed forms make that a **theorem in the dial**: even nonconstant rational functions of degree **30/30** (mass), **30/30** (`s_x`), **64/64** (`sigma`) and **54/54** (`g_re`), each summing to 1 identically; `s_t` and `g_im` did not finish and nothing is claimed about them. And the null-bench control **does not come out null**: where the `W9` profile is exactly uniform `(1/4,1/4,1/4,1/4)`, `mu*` is exactly **x-period-2** — it resolves the L2 residual symmetry the profile cannot see. The parity law **extends to 12x4 on four dials and FAILS on `s_t` and `g_im`**, exactly the two dials carrying the nearest sign boundary.

### Not the Perron vector, not a limit — and the ladder's directions were backwards

O(1) separations from the Perron vector, the `W9` diagonal and both fork arms. Five exact pushforward steps drive the distance to the **Perron vector** from `> 1/392251` to `> 1/34086734520363315429692` while the distance to `mu*` stays at bracket `1/3` at **every** step. The ladder factors exactly, `defect = E_d (mu - mu*)` — and **the solve read both factors backwards**. Down the ladder `max|E_d|` **GROWS** through `643/10000000`, `11635/10000000`, `272011/10000000` while `|mu_gibbs - mu*|` **FALLS** through `4107256/10000000`, `1785217/10000000`, `1022984/10000000`. The solve is **quoted then corrected**; the structural conclusion survives — a product of two opposing factors, **no single geometric law**, with `e1`, `e2`, `e3` all changing sign.

### THE EXHIBITS — the domain map, and non-existence at the recorded class

At **five** points `herm(Q)` reads `(24,0,0)(n+,n-,n0)[b165]`, the `W9` record-slice block reads `(4,0,0)(n+,n-,n0)[b165]`, `P0` is strictly positive — **and `mu*` still has a negative entry**. `W9`'s positivity is a theorem carried by **Sylvester**; `mu*`'s is **carried by nothing**. The reconciliation with Block 172 is clean: its checker class carried **no dial arguments**, so all its measurements sit at the committed point, and this block rebuilds them all as strictly positive.

**THE BOUNDARY MAP** (rational scan at step `1/40`, then bisection to width `<= 7/81920`; the scan comes first because the positive set is disconnected):

| dial | committed | nearest boundary | margin |
|---|---|---|---|
| `s_t` | `0` | `3063/12800` | **`0.2393`** — below the committed sister value `s_x = 3/5` |
| `sigma` | `3/5` | `8583/10240` | **`0.2382`** — the committed `3/5` is already 71.6% of the way |
| mass | `1` | `2357/5120` | `0.5396` |
| `s_x` | `3/5` | `10679/8192` | `0.7036` |
| `g_im` | `1/4` | `5801/15360` | **`0.1277`** |

Not "negativity at extreme dials only", and not "the committed point is on the boundary" — **the third thing**: strictly inside on every dial, but the positive region is a **bounded box of the same order as the committed fixture itself**, and it is **disconnected** (four crossings on `s_t`, exact brackets; a further positive island on `g_im`).

**AND THE SEALING DISCOVERY.** At the zero-shear carrier — `b171 SIGMA_CLASSES[0]`, **the value the records write** — every extension profile equals the base profile, `E` **vanishes identically**, `rank E = 0`, **every probability vector is stationary**, and **`mu*` is not defined**. Off the committed point `mu*` does not merely go signed: **it stops existing uniquely, on an in-lane declared class.**

### The kernel laws

**K1** is a theorem **conditional on solvability** (the checker's qualifier). **K3 splits**: `23 of 23` forward at the committed point and `15 of 15` at each of three further configurations, against `12 of 12` backward failures, `15 of 72` grid failures (one **double-booked** by the landed b171 identity, disclosed) and 5 in-fence failures. **K2 fails** by an exact nonzero defect. `mu*` is **slot-dependent** in value, argmax and ordering — so the slot-invariant object is the characterization `ker E_ell` and **never a numeral**.

### The verdict, and what comes next

> **The axiom route stands on NON-SUPPLY, now sealed at every theorem-shaped door this program has opened**: additivity vacuous-or-forced (172), stationarity fork-defective (172), and `mu*` signed off-region, domain-bounded and undefined at the recorded class (173). Three doors, three closures, **none of them a refutation of the bridge**.

**THE AXIOM MEMO TO THE OWNER IS THE PROGRAM'S NEXT DOCUMENT.** The design fork and the axiom decision remain the owner's and are not taken here.

### Discipline

Nothing registered, adopted or proposed; eight imposed objects declared and measured; the inertia-convention banner first and gated. **The solve is quoted then corrected on four points**, all four corrections the checker's. **This block's own defects are named as its own**: the solve's **AST gate was under-scoped** — it parsed only its own file, not the library where the arithmetic lives — and the runner **repairs it** to cover every file it reads code from; and the solve's **pre-registration is not auditable** from the artefact set and rests on the author's word. Two extents are not a limit; no float anywhere, gated by a multi-file AST scan; `--deep` widens the boundary map to all six dials, adds the `s_x` symbolic leg and runs all three re-censuses. Zero axiom retirement, zero obligation retirement, no TOE percentage movement, retained-positive end-to-end theory count remains zero.

**Runner:** `scripts/admissibility_dirac_kahler_stationary_kernel_characterization_2026_08_21.py` — gates A-H, 15 mutations mapped one per gate, baseline 24 s (`--deep` 49 s).


JONbridgerapps:
Landing disclosure: two supervisor-side repairs between draft and land — (1) the draft's N5_FENCE was assembled as a parenthesized join-block; the standard single-line-literal adoption orphaned its continuation lines, repaired by deleting the orphan block (the b145/b157 lesson, parenthesized variant); (2) the runner's __main__ exception handler printed [FAIL] INTERNAL-EXCEPTION + TOTAL FAIL=1 but exited 0 (fail-open) — one line added (raise, the landed b172 pattern) so the self-check that catches toothless mutations fails closed. Baseline 8/8 and deep 8/8 re-verified after both; the full 15-mutation sweep re-run from scratch on the fixed runner: 15/15 clean.

PR #7317, opened 2026-08-22
physics: block 174 — the site-conditional law family (the readout principle isolated)

JONbridgerapps:
## Block 174 — The Site-Conditional Law Family (the framework supplies a family, not a law — and the missing input is now one named object)

Stacked on Block 173 (the stationary-kernel characterization, at its nsimplify-hygiene tip). Block 173 closed the `mu*` route and named this block's charter: the candidate site-conditional law was **genuinely undefined** — the axioms' "possibilities at a site" reads as a local **value menu** while the computable G-A object is a site **choice**, and the normalized `W9` profile is globally normalized and therefore not blanket-local. This block defines the candidate, derives it as far as the framework forces, measures it against the four bridge obligations, and reports **exactly where the derivation stops**.

It stops at one place, and that is the headline.

### THE HEADLINE — a family, not a law, and the gap is now one named object

> **The complex Gaussian supplies an AMPLITUDE `Z = pi^N / det Q`. Turning an amplitude into a positive weight is an ADDITIONAL READOUT MAP, and the framework does not supply one.** The candidate family `{|det Q|^-2, det(herm Q)^-1, |det Q|^-1, positive-Re variants}` passes and fails the four advertised binary obligations **identically** at the fixtures. Only an **auxiliary parity fingerprint** discriminates. So "derived, not chosen" weakens to **"selected within a family"**, and **THE MISSING PIECE IS AN INDEPENDENT PROBABILITY/READOUT PRINCIPLE — the in-framework Born-readout problem, now localized with exact fixtures.**

The solve claimed the modulus-squared shape was **derived** — "the complex measure forces it". **That claim is quoted and refuted**, and the refutation is this note's spine rather than its footnote.

### What the framework does supply

A **measured, exact, level-indexed, chart-covariant, asymptotically-blanket-local candidate family** for its own Admissibility distribution. The axioms' clause (`MINIMAL_AXIOMS_2026-06-29.md`, lines 60–61) names a local probability law and leaves its form and values as Open-Gates content; this block measures the framework's own candidate for that named-but-unfilled object. **The axioms file is quoted and never edited.**

### The four obligations, measured exactly

| obligation | status |
|---|---|
| covariance | **HOLDS EXACTLY** — `P_0 = P_2` entry for entry; `P_0 != P_1` is the landed L2 chart structure, and it is **NOT symmetry-protected** |
| normalization / support | **HOLDS BY CONSTRUCTION**, on the corrected domain **`m > 0`** |
| blanket-only dependence | **FAILS EXACTLY** at every finite width; asymptotically supported and **not proven** |
| nonconstant blanket variation | **HOLDS** |

Covariance is not carried by a hidden symmetry, and that is gated directly on `Q`: translation by 4 breaks the unrecorded `48x48` twin `Q` at **80 nonzero differences** (witness `-123/448`), all eight reflections break with counts `(340,352,356,352,312,352,344,352)`, and the twin-axis witness is `23/25`.

The domain correction is the sharp one. At all **14** sampled nonzero-mass dial points every menu arm has 24 strictly positive exact `LDL*` pivots — but **at `m = 0`, `herm(Q) = 0` identically**, the first pivot is exactly 0, and **the Gaussian diverges**, while every `det Q` stays nonzero and the algebraic normalization stays positive. Raw weight positivity is **weaker** than Gaussian well-formedness. A 15-point census is not a global positivity theorem, and it is not called one.

### The tail — blanket-only fails, then decays

With blankets matched the twin gap is exactly in `[4931/1e8, 1233/25e6]` against a blanket response in `[223949/1e8, 4479/2e6]`, **ratio exactly in `(45, 46)`**. The law is blanket-dominated and quasi-local and **is not exactly blanket-determined**. Then it decays:

| `L_x` | max twin gap (guide) | tail/response (guide) |
|---|---|---|
| 8 | `4.93e-5` | `2.20e-2` |
| 12 | `1.20e-6` | `5.37e-4` |
| 16 | `5.12e-8` | `2.29e-5` |

**Three widths are finite-width evidence and are never a limit.** Asymptotic blanket determination is **supported, not proven**, and the note says so wherever the asymptotic reading appears.

### Two closures and one refuted attribution

**Finite-size exact Form B is CLOSED AS UNSAT AT BOTH LANDED EXTENTS.** At 8x4 the 16 rows are the **entire** G-A history set (`N = T_phys - 2 = 2`): 0 of 16 exact matches, target LCMs of 34–42 digits none of which divides `N = 2`, and a pigeonhole that bites because `2 < 4`. **The pigeonhole wiring is corrected and the correction is gated**: it obstructs exactly when `N < |A|`, so it does **not** close 12x4, where `N = 4 = |A|` — the earlier universal leg mixed the site-class count with the trail alphabet. At 12x4 the closure is the **full 256-history census**: 0 of 256, every target LCM failing to divide `N = 4`, LCMs spanning 44–65 digits.

**The `(1,2)` attribution is REFUTED.** The spatial-2 coupling **exists pre-quotient**: on the committed cover, `Q_cov[(2,2),(3,0)] = 33/128`, recomputed from two landed factors `H = -55/128` and `d = 3i/5`, generated by `iHd`. And **`fx.quotient` is an antiperiodic fold, not a Schur elimination** — it keeps the second cover half as rows and folds the column halves by a sign, with no eliminated variable and no block inverse anywhere in it. What survives is the **chart reading**, stated as a reading: on the period-2 chart lattice a `(1,2)` site hop is a `(1,1)` chart hop, consistent with the measured `2Z` covariance — and it is **not** a spatial-locality theorem for the cover action.

### The law is level-indexed, and the menu semantics were wrong

`t_r = 2, 3, 4, 5` give **four pairwise-distinct exact laws** on the constant carrier; fixing `t_r = 2` hid it. The object is `P_{s, t_r}`. **The supervisor's counter-read is recorded and is UNTESTED**: the levels differ in distance to the pinned band, so this may be the same screened effect as the width tail — the named follow-up probe is whether it decays with that distance.

Menu value `0` is the **one-cell disconnection value at a FREE cell**, algebraically the G-A update — and it is **not** the landed two-level region pin, which differs in 8 entries with witness `Delta Q[8,9] = -63/640`. **The law is restricted to free levels.**

### Cross-model verification — and it earned its keep

**This is the first block of the lane checked by a second model, and it is disclosed as such.** The adversarial checker (codex 5.6-sol xhigh) ran an independent determinant route (`DomainMatrix` fraction-field elimination against Berkowitz) and an independent positivity route (exact unpivoted `LDL*`), plus the tail probe, both censuses and the law check. **It cut the frontier model's claims down five times**: modulus-squared forced, one site law, the menu-zero semantics, the positivity domain, and the quotient-generation attribution. Common-mode risk is **reduced and not eliminated** — both sides read the same landed chain, the same rational arithmetic and the same imposed fixtures.

### Runner

Eight gates, fifteen mutations mapped one per gate across A–H, no float anywhere. **Gate A is the interesting one**: the parent tip's own content is a repair of four `nsimplify` call sites in three committed scripts, so gate A reads those three files, asserts all four sites absent — they are — and pins the pre-hygiene tip as the stale parent. The two Block 173 artifact blobs are **identical at both commits**, so the hygiene leg is the binding one.

**And gate A found that the repair is partial, which it reports rather than smooths.** Counting `(text mentions, live call sites)`: the three repaired files still carry **4 live call sites** between them (none of them one of the four repaired), and the import chain carries **0 live call sites inside this runner's AST audit surface** but **2 below it**, in the landed Block 166 module the chain reaches transitively. No verdict here depends on any of them — every decision is an exact rational route inside the audited surface — but the corpus-wide hygiene item **stays live**.

Gate H additionally requires the **full** scope-key set and not a subset, which is stricter than the Block 173 fence and is what gives the drop mutations teeth. Baseline runs a disclosed sample (26 of 256 census rows, 6 of 15 dial points, `L_x = 8` and `12`); `--deep` runs all 256, all 15 and adds `L_x = 16`.

### Discipline

Nothing registered, adopted or proposed; eight imposed objects declared and measured; the inertia-convention banner first and gated on the **landed** collision instrument. The solve is **quoted then corrected on four points** and one further attribution is refuted, all five corrections the checker's. **This block's own defects are named as its own**: the width ladder `{4,6,8}` is **not runnable width-uniformly** on the xgraded carrier and the favourable stability at `L_x = 4` and `8` is a property of the tiling rule, disclosed as rule-dependent; the tail is three widths; the positivity domain rests on a 15-point census; and the AST surface is the runner chain and **not** the full transitive closure. Scope is one `T_cover`, the CM-SITE alphabet, the committed action class, **free levels only** and `m > 0` — not a continuum statement, not an OS no-go. **The semantic-bridge row is quoted verbatim and remains NOT YET EARNED**, and it cannot be earned before the readout principle exists. Zero axiom retirement, zero obligation retirement, no TOE percentage movement, retained-positive end-to-end theory count remains zero.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7317 (archive/ledger/17/pr-7317.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #7316, opened 2026-08-22
physics: encode NN Record programs and expose self-hosting gap

jonathonreilly:
## Summary

- Gives an exact opposite-neighbor M2 encoding that separates a supplied density preparation from a binary or ternary scaled-projector program under all 24 proper cubic rotations.
- Shows that the selected trace law reproduces the parent one-site prefix marginal, literal projective contents, pure endpoints, repeated-effect menus, and additive refinements.
- Adds an exact self-hosting check: the current Record rule outputs only scalar anti-Hermitian coefficients, while active carrier inputs require nonscalar coefficients. This is therefore a conditional capacity theorem, not physical Record-carrier closure.

## Scientific status

This is stacked on #7210. It is source-level bounded evidence only: no axiom or primitive change, no audit verdict, no formal obligation retired, and no TOE-percentage movement.

Still open are a same-law physical writer/genesis mechanism, physical calibration or trace-law selection, multi-site formation/history, and a Bell-capable composite carrier.

The note includes the N1-N8 no-go stress test, including the five required N5 resolution lines.

## Verification

- Block 32 primary runner: TOTAL: PASS=22 FAIL=0
- Canonical cache is runner-SHA and input-fingerprint pinned; cached stdout is below 6000 characters
- Both parent #7210 runners: TOTAL: PASS=12 FAIL=0
- Python compile, vocabulary lint, staged diff check, and audit-runner freshness checks pass
- Independent cold check: packaging PASS after narrowing the original carrier-closure overclaim

PR #7318, opened 2026-08-23
physics: block 175 — the pincer identity (both lanes compute the same object)

JONbridgerapps:
## Block 175 — The Pincer Identity (the two lanes compute the same object; the readout question becomes a two-object question in a joint cage)

Stacked on Block 174 (#7317). Block 174 isolated this lane's missing input and named it the **readout principle** — the framework forces the construction as far as the complex amplitude `Z = π^N / det Q` and no further. The owner then pointed at the parallel record-born-composition lane (#7316), which carries a **selected** downstream rule assigning `Tr(C E_j)` and names its own selection as not derived. Same shape of gap, one lane at the effect level and one at the amplitude level. This block asks whether the two gaps are the same gap by asking whether the two lanes' objects are the same object — with both predictions registered before the arithmetic was run. **Both confirmed, with zero defect.**

### The identity — the interface object is one object

The precondition is measured first and is not assumed. The W9 record-slice block at the read slice `t* = 5` occupies ambient rows and columns `(20,21,22,23)` and its **off-diagonal support is empty exactly** — all twelve entries expand to exactly `0`, not to a small number. So its normalized diagonal is a Hermitian, unit-trace, strictly positive **classical density `C` over the menu**, displayed with its exact common denominator, and `Tr(C E_j)` is a classical marginal read. That S-DIAG classicality is the reason the two lanes can meet here at all.

Then, with `E_j = |j⟩⟨j|` in CM-SITE order:

**`Tr(C E_j) − P_W9(j) = (0,0,0,0)`, entry for entry, exactly.**

The 4×4 block and the profile are built by **two independent routes** — the landed `b171 Env` route and a rebuild from the Block 174 `Fixture` through exact inversion and Hermitian projection — and the gate requires the two routes to agree before either is compared to anything. The reading: the Born-trace reading of this kernel's record slice **is** this lane's marginal profile. Two programs that started from opposite ends have been computing the same four numbers.

### The split — and it is this block's own witness to its scope

The formation-conditional `|det Q(σ_(2,0)=a)|^-2` law at the free record cell differs from the trace/marginal law **in all four entries**: four exactly nonzero rationals over a displayed common denominator `K`, signs `(+,−,−,+)`, sum exactly zero (forced by shared normalization, so it is gated as an arithmetic control rather than reported as a discovery), absolute brackets from `[49979/20000000, 31237/12500000]` to `[2367393/100000000, 1183697/50000000]`.

So the framework now carries **two distinguished exact probability objects** at one fixture — the Born-trace **marginal** and the **formation conditional** — differing by exactly computable amounts. This is not a discrepancy and not a defect in either lane; a marginal and a conditional are different objects. What is now exact is *how* they differ and *where* the framework declines to say which one the Admissibility clause names. **The marginal-versus-conditional distinction is the sharp surviving form of the readout question.** And it is the block's own counterexample to any wider reading of the identity: the conditional is generated by the same committed action and the identity fails on it.

### Additivity, split honestly

The other lane advertises refinement additivity, so it is measured here on both objects, on two declared three-outcome partitions.

- **The trace law is natively additive.** Union effects are measured to be orthogonal projectors resolving the identity; the defect is `(0,0,0)` on both partitions. It holds because a union effect is an operator sum and the trace is linear — for any state, any partition.
- **The determinant law's coarse additivity is a definitional pushforward.** The machinery defines only atomic weights, and **no primitive value-union pin exists** — measured three ways, not asserted: the landed field builder maps a cell to one scalar; two values at one dictionary key collapse to one, so that is not a simultaneous pin; and a set handed to a record key sympifies to a `FiniteSet`, not a number, so no `Q(σ ∈ U)` is formed. The pushforward `w(U) = Σ_{a∈U} w(a)` reproduces `Σ p_det(a)` at defect `(0,0,0)` on both partitions **by definition of the pushforward**, and **this does not show action-level merge closure.** The Block 172 nonclosure concerns definable multi-cell *conjunction* pins, not mutually exclusive value alternatives at one cell — scoped here, neither strengthened nor weakened.

The honest consequence: a selection principle leaning on additivity constrains the two lanes' objects **unequally**, and that asymmetry is itself part of the cage.

### The joint cage

Their open item — *"preparation-matrix calibration and trace-Law selection from the four axioms | not derived"* — and ours, the readout principle, are **one question with two exact object families**. Any candidate principle must now survive both fixture sets at once: their refinement additivity, 24-rotation covariant decoding, normalization, pure endpoints and prefix marginals at the effect level; this lane's exact rationality, chart sensitivity on the L2 period-2 fingerprint, screened locality with an exactly nonzero finite-width twin gap, and level indexing at the amplitude level. Plus the bar this block adds: it must say **which** of the marginal and the conditional the Admissibility clause names, or derive one from the other. Their `M_2(C)` carriers are dimension 2 where Gleason does not apply, and this lane's S-DIAG classicality is the same wall from the other side — **flagged as a structural resonance, not asserted as a theorem.**

### Scope, and the cross-lane discipline

An identification of the **marginal record-slice object only**, at **one fixture** — one extent, one carrier, one read slice, one record cell, one menu. **No selection principle is supplied**; the cage constrains a candidate and does not produce one. The Block 174 semantic-bridge row is unchanged and remains not yet earned.

The other lane's note is an **in-repo reference: quoted, never adopted, never edited, never treated as authority, never re-run.** The discipline is mechanical, not promised — that note appears nowhere in `AUDIT_INPUT_PATHS`, nothing in the runner reads it, and every number here is rebuilt from this lane's own Block 171–174 machinery on this lane's committed fixtures. Gate F checks quotation **integrity** only, never their correctness. That is the standing converge-don't-borrow directive applied literally.

### Verification and independence disclosure

- Runner: `scripts/admissibility_dirac_kahler_pincer_identity_cross_lane_2026_08_22.py` — baseline **7/8 PASS** at draft (gate H on note-at-final-path alone; 8/8 expected once the note lands), ~7 s against a 120 s budget, `--deep` clean (all four pinned determinants recomputed by both exact routes, agreeing), N5 fence byte-identical, **mutation sweep 15/15 clean** with each mutation failing exactly its own gate (the two gate-H mutations vacuous until the note lands, disclosed). 94 of 94 scope keys satisfied. **No floats** anywhere — 0 float literals and 0 forbidden references across the 5-file AST surface, exact rationals only.
- Negative controls that bite: an exact rational perturbation of one diagonal entry of `C` is measured beside the true reading and its defect is exactly nonzero, so `break_identity` has something real to break; and the stale authority pin is the Block 173 hygiene tip, a real ancestor of HEAD that **predates both Block 174 artifacts**, so `stale_parent_authority` looks for blobs at a commit where they do not exist.
- **Worker profile:** supervisor-designed probe, written to the owner's pointer at the #7316 lane, with both predictions registered before the arithmetic; execution by a codex 5.6-sol xhigh worker (the reduced state and its S-DIAG check, the four determinant pins, the exact deltas, both partition batteries); the **cross-model verification chain from Block 174 carried forward** onto the fixtures this block reads; Opus mechanical drafts of note and runner; supervisor review. Common-mode risk reduced and **not** eliminated — same landed chain, same SymPy rational arithmetic, same imposed fixtures.

### Status discipline

Zero axiom retirement, zero obligation retirement, no TOE movement. Nothing registered, adopted, or proposed — here or in the other lane. The axioms file is quoted and never edited. Every negative is non-supply within this formalism and never metaphysical necessity; every positive, the identity included, is candidacy within this formalism and never a claim about nature (CYCLE913 verbatim, both directions). Successor question: **select-or-relate the marginal and the conditional.**


JONbridgerapps:
Consolidated to the archive: ledger id pr-7318 (archive/ledger/18/pr-7318.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #7334, opened 2026-08-23
physics: block 179 — derive the source-functional grading boundary

jonathonreilly:
## Scope\n\nStacked after #7332. This block asks whether the complex Gaussian source functional itself supplies the missing probability/Record bridge.\n\n## Result\n\n- Derives the exact two-point function and permanent Wick tower from the normalized Gaussian functional.\n- Separates normalized vacuum 1, one-copy unnormalized vacuum Z_Q, and doubled partition-level weight |Z_Q|².\n- Repairs the vacuum accounting: Z_Q is already positive and transport-sensitive on all four exact fixtures.\n- Finds the full raw reflected source kernel non-Hermitian and its Hermitian part indefinite.\n- Confirms the inherited rank-two restriction is positive and transport-sensitive, while proving period-two symmetry does not select its 4/5, 3/5 weights.\n- Detects the Block 107 prose/runner orientation mismatch; the two forms are adjoints on generic G.\n- Lands a fresh N1–N8 bounded no-go packet and complete relevant PR/primitive inventory.\n\nThe correction is deliberately narrow: Z_Q repairs only the degree-zero leg. Higher sectors still require a physically justified Hermitian reflection/source convention and event-domain selector. A specific doubled |Z_Q|² choice would additionally require a sewing/reweighting rule.\n\n## Verification\n\n- Exact primary runner: 32 PASS / 0 FAIL\n- Cached stdout, runner SHA, and 17-input fingerprint match\n- Python compilation and staged-diff hygiene pass\n- Independent cold-check: SHIP\n- No audit or ledger surfaces changed\n\n## TOE accounting\n\nThis derives a substantial algebraic boundary and removes an incorrect vacuum obstruction, but retires no axiom or TOE obligation and makes no TOE percentage claim.

### DIFF
--- a/docs/ADMISSIBILITY_DIRAC_KAHLER_SOURCE_FUNCTIONAL_GRADING_BOUNDARY_BOUNDED_THEOREM_NOTE_2026-08-23.md
+++ b/docs/ADMISSIBILITY_DIRAC_KAHLER_SOURCE_FUNCTIONAL_GRADING_BOUNDARY_BOUNDED_THEOREM_NOTE_2026-08-23.md
@@ -1,0 +1,484 @@
+---
+claim_id: admissibility_dirac_kahler_source_functional_grading_boundary_bounded_theorem_note_2026-08-23
+final_path: docs/ADMISSIBILITY_DIRAC_KAHLER_SOURCE_FUNCTIONAL_GRADING_BOUNDARY_BOUNDED_THEOREM_NOTE_2026-08-23.md
+claim_type: bounded_theorem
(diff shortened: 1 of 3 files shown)


JONbridgerapps:
Cross-lane G1 review from the gravity/generator lane (full review at .claude/science/physics-loops/generator-program-20260821/G1_7334_review.md on our chain): PARTIAL DISCHARGE of our block-177 grading premise, and a genuine upgrade — your source-differentiation derivation lands the functional grading on the RAW covariance-side kernel at our exact fixtures (your runner imports our b177 module; zero translation), which answers our B2b 'which kernel' question and explains our -35233/38760 mismatch witness as stipulated-vs-derived. With your tower + our Sym^n lemma, the conditional symmetric-power theorem becomes ONE-prescription conditional (the Hermitianization of the raw non-Hermitian K — the same theta-breaking witnesses our shear-mirror block measured — plus the domain and your |Z_Q|^2 sewing rule), no longer two-premise. Your vacuum repair supports our vacuum-leg results. One corpus flag from our reviewer worth your eyes: a Block-107 prose/runner transpose mismatch (adjoints on generic G; harmless on real fixtures; should be fixed before complex ones). Reference-and-re-prove both ways; nothing binds your lane.

PR #7384, opened 2026-08-25
physics: classify the site symmetries and certify the invariant-sector quotient (block 189)

JONbridgerapps:
Block 189 of the toe-axiom-closure series. Stacked on #7359 (block 188).

## What this block does

Symmetry classification of the b188 site-glued action (T=8, X=4, m=9/20, c=5/13), all exact:

1. **Odd shifts break the grading.** [U_x, d_K] has 64 nonzero entries (rank 16) intra-slice; the mechanism is located exactly: grade-projector defects (16,32,16), and the glue split [U_x, H_s] = 0 while [U_x, D_s] = 48 — the η_x = (−1)^t staggering is the obstruction.
2. **The exact-symmetry criterion is reflection-evenness.** Exactly 32/32 reflection-even per-slice even-shift words commute with Q_s; all 224 others fail. All 32 transport the OS pairing at zero residual.
3. **The stabilizer is order 2.** {identity, global} out of the faithful 256-element family, by exhaustive sweep (the adversarial check's correction, carried as its own gated claim family): 2-of-32 reflection-even, 0-of-8 adjacent-pair words.
4. **Wilson twists localize with range vector (2,1,2,1,1,1,2,1)** — range-two at three slots, NOT nearest-bond (the check's correction, gated as the measured number; the word 'bond' does not appear).
5. **The first executable gauge-invariance instance on the OS core:** (W₁W₂)ᵀ K (W₁W₂) − K = 0 exactly, against two 32-entry rank-4 single-twist failures.
6. **The invariant-sector quotient.** The Z₂×Z₂ projector has rank 4; the quotient Gram is PD at inertia (4,0,0), det scale 1/16 — with the automatic-positivity disclosure gated (positivity follows from b188's core PD + invariance; not an independent discovery), and the checker's diag(1,−1,−1,1) staggered-sign Gram reconciled as convention (all four minors, det, density, inertia agree exactly).

**The interpretations fence (required by the check):** 'gauge quotient' and 'lapse' are READINGS — 8 unsupplied gravity structures enumerated; no constraint quotient is formed; the theorems are the symmetry/stabilizer/projector/positivity facts.

## Check

Cross-model adversarial check (codex 5.6-sol xhigh): CONFIRMED-WITH-CORRECTION, folded as content — stabilizer order 2 (own family), range-two twist supports, the interpretations fence.

## Verification

- Baseline: 8/8 families pass, exit 0, ~1s (exact rational DomainMatrix/QQ linear algebra — no tolerance anywhere).
[...]

PR #7580, opened 2026-08-25
physics: prove the width-family transfer refutation and the positive unit-cell monodromy (block 190)

JONbridgerapps:
Block 190 of the toe-axiom-closure series. Stacked on #7384 (block 189).

## What this block does

The transfer question of the site-glued OS class, answered on the width family (Z_T × Z_4, T = 8/12/16/20, wrap-edge antiperiodic sign — a **disclosed variant** of the landed T=8 block: the sign-placement fork is measured as a pair, nnz(Q−Qᵀ) = 144 at the landed placement — b188's own number, with b188's core minors reproduced digit-for-digit — vs 160 at the wrap edge, both reflection-covariant):

1. **The naive transfer is refuted on the cores themselves.** L₁ and L₂ asymmetric at every bulk core with six exact rational witnesses. The check's C2 correction is carried as a declared constant: the global commutator [τᵏ,Q] ≠ 0 (fork pair 224 nnz/28 slice pairs vs 208/26, common witness −65/576) is the *mechanism*; the core witnesses are the *proof*.
2. **The step operator is local.** V₂@T12 = V₂@T16 = V₄@T16 and V₁@T12 = V₁@T16 **entrywise** while the Grams differ at 56 entries — dynamics width/position-invariant, boundary layer in the metric. Factor pattern (2,2,4) everywhere; quadratics rigid; the odd quartic has near-seam/mirror/deep layers and the even quartic its own far-seam layer (47667825, 63213480, 101294706, −55889280, 39529825) — every layer width-locked. Mirror covariance q_j = (−1)ʲp₈₋ⱼ exact.
3. **The unit-cell monodromy is primitive, parity-independent, and positive.** W ≠ V² at 32 entries (exact witnesses); charpoly(W) = (22569375z² − 233631106z + 22569375)²(39529825z² − 109432706z + 39529825)² identical at all three deep cores; positivity **proven** (discriminants 2⁸·13·31·37·71·313²·1979 and 2⁸·3⁷·7·13·31·37·313², traces > 2·leading, reciprocal pairs, distinct trace ratios): four distinct real positive roots, 2cosh θ₁ = 233631106/22569375, 2cosh θ₂ = 109432706/39529825. The coefficient identity: second W-quadratic = a z² − c z + a from the even V-quartic's (a,b,c), b drops out.
4. **The commutant is computed exhaustively.** 2048 signed-monomial candidates: W-commutants exactly {I, S, U, S³}, Gram isometries exactly {I, U}. S (one-site shift, S² = U) grades the two scales into momenta — p ∈ {0,2} light, p ∈ {1,3} heavy — with the p=0/p=2 equality declared an additional isospectrality, not group-forced. Spatial reflection refuted with witness 16334218/7905965.
5. **The adversarial check's Hodge refutation (C3/C6/C8) is resolved, not disputed: a volume convention.** The landed shear_hodge at v=1 is displayed inline entrywise and gated against the import; the check's own two rebuilds are reproduced digit-for-digit here at v = 12/13 and unit shear; its forensic fingerprint is measured spectrally identical to the landed block. The check's P1 runs the other way on this construction (K_c exactly symmetric ⟹ K-only transposition is a no-op; the consistent transposition is a similarity — charpoly preserved).

**Fences:** dispersion, mass scales, the physical time step, and transfer positivity are READINGS — what is proven is the positivity of an 8×8 rational spectrum; the OS reconstruction that would license the words is not performed; no gravity structure is supplied. One fixture, four widths, no window, nothing about the infinite-width limit.

## Check

Cross-model adversarial check (codex gpt-5.6-sol, xhigh): C1/C4/C5/C7 CONFIRMED (Hodge-robust — they held under the checker's variant Hodge too), C2 CONFIRMED-WITH-CORRECTION (folded), P4 positivity proven exactly, P2 found the S-commutant (rebuilt and extended here by the exhaustive census). The C3/C6/C8 coefficient flags and P1 asymmetry resolved as above — every one of the checker's displayed numbers is reproduced digit-for-digit under its identified convention.

## Verification

- Baseline: 82/82 checks across families A–H, ~16s.
- Mutation battery: 58 mutations, one-family-each, verified twice in the live worktree (re-run in full after the pin refresh and the AUDIT_INPUT_PATHS literalization).
- Pins current at main b11811704e (nightly audit commit; sole change dispatch_shadow_state.json — axiom/registry blobs unchanged); parent = b189 tip 996e5166.
- nsimplify count 0 (gated); N5 fence byte-identical; runner cache ×2 stable; citation manifest regenerated; audit_lint --strict OK.

### DIFF
--- a/.claude/science/physics-loops/generator-program-20260821/b190_check_findings.md
+++ b/.claude/science/physics-loops/generator-program-20260821/b190_check_findings.md
@@ -1,0 +1,357 @@
+# Block 190 transfer-theorem package: independent adversarial findings
+
+Date: 2026-08-24
+
+## Scope and method
+
+This is a fresh rebuild from the dispatch equations. No repository runner,
+script, scratchpad draft, ledger, or audit-data file was read or imported.
+All matrix entries and polynomial computations use exact SymPy `Integer` and
+`Rational` arithmetic; `nsimplify` is never used. The only science-source reads
+were the permitted landed block-188 note and its named block-105 primary note,
+solely to identify/cross-check the displayed shear-Hodge convention.
+
+Status below defaults to **REFUTED** when a stated conjunction has any failed
+part. Exact witnesses and narrower surviving subclaims are recorded.
+
+## Hodge convention control
+
+**FAILED / dependent numerical claims unverifiable as stated.** The permitted
+block-105 note displays, at `q=5/13`, `s=12/13`,
+using zero-based corner indices `0,1,2,3`,
+
+```
+B_105 = diag(12/13,13/12,13/12,13/12),
+(B_105)12=(B_105)21=-5/12.
+```
+
+The literal unit-volume version of the dispatch's fallback characterization is
+
+```
+B_fallback = I_4,  (B_fallback)12=(B_fallback)21=-5/13,
+```
+
+with `P4 B_fallback P4^T` moving that shear to the `(0,3)` pair on the image
+half. Neither rebuild gives the claimed even-core polynomial. Their exact
+`T=12,t0=2` factorizations are respectively
+
+```
+B_105:
+(1975z^2+1953z-2365)(2365z^2+1953z-1975)
+(4746925z^4+6783420z^3+13155959z^2-6783420z+4746925)
+
+B_fallback:
+(475z^2+468z-565)(565z^2+468z-475)
+(272425z^4+393120z^3+755774z^2-393120z+272425).
+```
+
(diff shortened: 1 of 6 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-7580 (archive/ledger/80/pr-7580.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #7669, opened 2026-08-25
physics: prove the boundary-monodromy table and the volume-sensitivity package (block 191)

JONbridgerapps:
Block 191 of the toe-axiom-closure series. Stacked on #7580 (block 190).

## What this block does

The boundary and volume structure of the unit-cell monodromy, all exact:

1. **The boundary-monodromy table (T=20, all valid cores).** Deep rigidity at t0 = 2..5. The light factor (39529825 z² − 109432706 z + 39529825)² is **boundary-rigid at every core** — the light mode does not see the edge. At t0=1 one heavy copy is replaced by the **non-reciprocal boundary quadratic** 43033320714375 z² − 445467467014578 z + 48554286398375; at t0=6 the same quadratic exactly coefficient-reversed (mirror covariance); at t0=7 a second distinct far layer. The whole table is **positionally width-locked** — identical polynomials at T=16 and T=20.
2. **The validity rule, cleaned of its false signature.** Cores whose two-step pairing crosses into the image half are not bulk objects — but the adversarial check refuted the non-factoring signature (crossing cores factor (2,2,4)), and the drafter extended the refutation from below (t0 = T/2 factors completely into rational quadratics). The rule stands on the pairing's definition alone.
3. **The volume dial reshapes, not rescales.** At uniform v = 4/5 the deep spectrum keeps its palindromic-squared positive form — (31260675 z² − 302948719 z + 31260675)²(50327125 z² − 139773119 z + 50327125)² — with the two exact traces moving in **opposite directions** (ratio 2.7362708113 → 2.6427023041, the check's corrected values).
4. **The localized bump breaks reciprocity exactly, decaying into the bulk.** v = 4/5 on positive slices {3,4}: Ps-covariance exact; every irreducible factor non-palindromic at t0 = 1, 3, 5 (exact); the complete complex inventory gated (seven nonreal pairs across both bump positions, (2,1,4) around the 0.002–0.003 band); max root shifts 1.398 at the bump vs 0.0145 one core away. The near-edge response is a **hybridization** of boundary and bulk-heavy factors — pure boundary-mode dominance is refuted (the check's P1), and bump-position dependence is measured at {2,3} including an exact zero.

**Fence:** the package establishes **Hodge-volume spectral sensitivity**; "lapse physicality" and physical-edge-state language are READINGS — the OS reconstruction is not performed and no gravity structure is supplied.

## Check

Cross-model adversarial check (gpt-5.6-sol, xhigh, pinned): C1 confirmed exactly; C2 confirmed with the signature correction; C3 confirmed with corrected ratio arithmetic; C4 confirmed with the completed complex inventory; P1 refuted as stated (hybridization, folded as content); P2 position-dependence measured. Corrections #26–#32 continue b189's ledger (the drafter caught that the specified #21+ would collide with b189's landed #21–25).

## Verification

- Baseline: 59/59 across families A–F, measure() ~36s (five exact inverses shared).
- Mutation battery: 49 mutations, one-family-each, verified in the live worktree, all nonzero exit.
- nsimplify count 0 (gated); AUDIT_INPUT_PATHS literal; N5 fence byte-identical; cache ×2 stable; manifest regenerated; audit_lint --strict OK.
- Pins current at main b11811704e; parent = b190 tip e75ad9f4. The campaign anchor is committed with the block (gate C-1 reads its verdict section from the worktree).


JONbridgerapps:
Consolidated to the archive: ledger id pr-7669 (archive/ledger/69/pr-7669.json). Carrier: #7753 (open closure-lane tip; commit-ancestry verified behind/ahead=0). Branch retained; science line in the ledger entry.

PR #7726, opened 2026-08-25
physics: isolate L24 fermionic OS spin boundary

jonathonreilly:
## Result

Block 197 executes the panel-selected literal finite-L24 OS/GNS/CAR gate. The frozen periodic/uniform Block-192 fermionic reflected covariance is indefinite before any quotient: each actual radius-one Clifford block has rank 4 and inertia (2 positive, 20 null, 2 negative), with exact two-generator norm -678223072849/77463616656739800 = -7^14/77463616656739800. The full 16-component fiber is eight equivalent blocks, rank 32 with inertia (16,160,16).

Both adjacent planes, both half orientations, every Hermitian global degree sign, covariance transposition, and phase undo preserve the boundary. The older right-Schur Gram remains positive and is explicitly kept distinct.

The scalar antiperiodic-wrap plus transported-seam control is positive on both tested planes, but exact mode arithmetic shows pi/6 and pi/4 share a scalar twist only for L=24n, where that twist is periodic +1 rather than antiperiodic -1. This repairs another carrier, not the frozen one.

## Scope

This is partial route pruning, not a broad OS/history no-go. Parity-doubled two-step OS, a full D4 spin-structure rebuild, action-derived process tensors, open time, and centered-symbol gravity remain live. No quotient, channel, response, heldout, axiom update, obligation retirement, retained status, or TOE score movement is claimed.

The landed source note contains the full current N1-N8 packet. N1 and N7 fail for a broad negative; the next exact target is the action-derived even/odd Schur reduction and parity-doubled two-step state.

## Checks

- primary 7/7
- independent no-import 14/14
- 12/12 isolated mutation families rejected
- strict audit lint: 0 errors
- axiom premise purity: PASS
- forensic N5 readiness: PASS
- repo link/class-F/graph invariants: PASS
- citation graph delta: one node, five intended dependency edges, acknowledged

The full audit pipeline passed through effective-status computation and then repeated the inherited parent-stack dependency-policy epoch mismatch: build_citation_graph.py actual SHA c58180... versus tracked epoch 206982.... All generated audit surfaces were dropped; only the required citation manifest lands.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7726 (archive/ledger/26/pr-7726.json). Carrier: #7727 (even/odd Schur marginals inherit the indefiniteness with the closed-form -1/(m^2+s) law); #7733 (tournament exhaustion subsumes it); #7781 branch ancestry. Branch retained; science line in the ledger entry.

PR #7727, opened 2026-08-25
physics: isolate L24 even-odd OS boundary

jonathonreilly:
## Summary

- execute the literal Block 192 periodic `L=24`, `m=2/7` even/odd temporal Schur reductions at all nine frozen spatial radii
- derive both inherited parity-preserving coarse reflections and test the resulting degree-one and fermion-even Berezin forms
- package exact primary and independent certificates, mutation tests, narrow no-go boundaries, and the post-result route ranking

## Exact result

Both parity marginals factor as `A_s^(2) = Q_s ⊗ B_s`.  Their reflected forms factor as `K_s = H_s ⊗ G_s`, where `det(G_s) = -1/(m^2+s) < 0`.  A matching local `2×2` principal minor is therefore strictly negative at every frozen radius.

Every reduced form has exact inertia `(2 positive, 8 null, 2 negative)`; the eight-copy lift has `(16, 64, 16)`.  The negative witness is an even exterior norm and survives the ordinary null quotient.  PSD heredity also excludes any joint parity Gram that retains either literal Schur form as a diagonal principal compression.

This is a bounded route result only: the literal Schur-coordinate algebra with inherited reflections does not seed the proposed ordinary two-step OS quotient.  It does **not** exclude a genuinely different link/cell event algebra, an action-derived process tensor, open-time descent, a rebuilt spin structure, gravity, Record, Born, or TOE closure.

No axiom changed, no obligation retired, and no TOE percentage moved.

## Validation

- primary exact runner: `7/7 PASS`
- independent no-import runner: `19/19 PASS`
- primary mutations: `13/13` rejected
- independent mutations: `10/10` rejected
- Python compile checks passed
- strict audit lint: `3972` rows, `0` errors
- premise-purity checks passed for all four axiom premise documents
- direct forensic N5 readiness passed
- repository invariants passed: links, retained-grade policy, class-F policy, and acknowledged graph delta
- both evidence caches are fresh

## Next route

The independent post-result panel ranked an event-generated, rank-minimal finite-memory process tensor first.  Its first gate is to derive the action-cut boundary functional and a causal/PSD three-crossing process on the fixed Block 194 event operator system, with memory dimension frozen from the exact moment-Hankel rank before testing positivity.

Review-loop was not used.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block198-even-odd-two-step-os-20260825/APPROACH_REGISTRY.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block198-even-odd-two-step-os-20260825/APPROACH_REGISTRY.md
@@ -2,10 +2,9 @@
 
 | family | object/formulation | mechanism/invariant | terminal obligation | strength | status | concrete evidence | reopen condition |
 |---|---|---|---|---|---|---|---|
-| A | literal even/odd Schur algebra | coarse Berezin RP plus null quotient | unique parity-memory history | target-equivalent | active | disclosed exact radius-one pilot | independent exact execution at all frozen radii |
-| B | global process tensor | positive multi-time Choi functional | unique causal action-derived comb | target-equivalent | unexplored | Block195 route plus finite-circle memory warning | derive boundary/history functional from action |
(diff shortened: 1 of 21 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-7727 (archive/ledger/27/pr-7727.json). Carrier: #7733 (the 2x2x2 sign/algebra/trace tournament exhausts this family and restates the witnesses); #7781 branch ancestry. Branch retained; science line in the ledger entry.

PR #7731, opened 2026-08-26
physics: prove the volume-response mechanism numbers and the exact support cutoff (block 192)

JONbridgerapps:
Block 192 of the toe-axiom-closure series. Stacked on #7669 (block 191).

## What this block does

First-order volume-response mechanics of the width family's unit-cell monodromy, all exact:

1. **The method theorem.** dW = K_c⁻¹(dL₂ − dK_c·W) with dG = −G·dQ·G and the exact Hodge volume derivative dB = −E00 − (169/144)(E11+E22) + (65/144)(E12+E21) + E33 — no symbolic inversion. FD-verified: exact (residual 0) at the cutoff pairs; convergence-certified with exact rational bounds at the response pairs (disclosed as a certificate, not an equality).
2. **The ten rationals.** Per-factor first-order trace responses through CRT spectral projectors (congruence-gated, squarefree-insensitive), sum rules exact — every value confirmed digit-for-digit by the adversarial check.
3. **The response table, honestly quantified.** Heavy/boundary sign-locking with exact differences — the "<1%" quantifier was refuted by the check and appears nowhere (correction #33); the solve's "9:1" on-site dominance replaced by the exact ratio 37533905844768035289054578457791/4229425500383349914656444790625, bracketed in (8,9) (correction #41).
4. **The exact support cutoff.** Exactly 3 of 12 (bump, core) pairs are matrix-level zero at finite δ — ({1,2},5), ({2,3},5), ({4,5},1) — identical at first order, δ = 1/5 and 1/7 (the check found two; the drafter's sweep found the third). The support-form routing mechanism is REFUTED: dG is dense (3968/4096) at every cutoff pair, and the overlap signature is sufficient but not necessary (exact counterexample gated). The true mechanism — the intertwining identity and the parity-resolved window law — is block 193's package, held here as a named open leg.

**Fences:** one fixture; perturbation-response language is a READING; no gravity structure supplied.

## Check

Cross-model adversarial check (gpt-5.6-sol, xhigh, pinned): C1/C2/C4/P1 exactly confirmed; C3's quantifier refuted (folded as content); the checker discovered the t0=5 exact zero and the {4,5} support cutoff — both now central gated content. Corrections #33–#41 continue the ledger.

## Verification

- Baseline: 44/44 across families A–G, ~42s (one 64×64 δ=0 inverse shared; twelve heavy inverses total).
- Mutation battery: 43 mutations, one-family-each, verified on the exact final bytes twice (drafter CLI sweep + live-worktree sweep).
[...]

PR #7734, opened 2026-08-26
physics: prove the parity-resolved window law and reduce the intertwining mechanism (block 193)

JONbridgerapps:
Block 193 of the toe-axiom-closure series. Stacked on #7731 (block 192).

## What this block does

The exact locality law of the volume response, unified and mechanism-reduced:

1. **The unified window law.** A source breaks W-invariance at core t0 **iff** its support intersects W(t0) = [2⌊t0/2⌋+1, 2⌊t0/2⌋+3] — one formula; cores 2j and 2j+1 share a window, so the parity switch found by the adversarial check is emergent, not a case split. Censuses: 40 cells at T=16, 70 at T=20 — including the fixed-slice anchors s=0 and s=T/2−1 (absent from both solve and check), which are what make the floor structure visible.
2. **The mechanism by reduction — closes b192's named open leg.** R[a,b] = −u_bᵀ·dQ·G[:,θ_a] with u_b = Gᵀd_b; the union of the u_b slice-supports IS the window, exactly, at every valid core and both widths — the raw-unit law verified over 65,280 units with zero mismatches. This also repairs b192's refuted overlap signature exactly (the ({4,5}, t0=1) counterexample explained: dQ slices {4,5,6} ∩ W(1) = {1,2,3} = ∅).
3. **Honest proof status.** LAW_PROVED_FROM_RECURRENCE_CLAIMED = False — the law is exhaustively verified and mechanism-reduced, not recurrence-proven; the recurrence-theoretic proof is a named open leg.
4. **The exemption structure.** Survives 16 attack directions at odd cores (0 breaks), fails at even (8) — the checker's prose said 16, its code ran 10; the note claims 16 on the drafter's own measurement. Break-density fingerprints (64,64,64,64) odd vs (64,64,32,32) even; the dual low-end exemption exhaustive at even cores, failing at odd.
5. **The validity rule is load-bearing.** Invalid cores (t0+3 > T/2) were run and measured to disobey the law — b191's touch/cross rule is not a convention.
6. **Equivalence strengthened.** The 20-pair first-order table equals the finite-δ table entry-for-entry at δ = 1/5, 1/3, 2/5 — 60 rebuilds, not just matching zero sets.

## Check

Cross-model adversarial check (gpt-5.6-sol, xhigh, pinned): C1–C4/P1/P3 exact, C2 18/18, C3 surviving exhaustive attacks; P2's parity discovery (the even-core shifted window) folded as the block's central law. Corrections #42–#51 continue the ledger.

## Verification

- Baseline: 37/37 across families A–F, ~130s (one 64×64 + one 80×80 δ=0 inverse shared + 12 finite ones).
- Mutation battery: 36 mutations, one-family-each, verified in the live worktree, all nonzero exit.
- nsimplify count 0; AUDIT_INPUT_PATHS AST-verified literal (9 strings); N5 fence byte-identical; cache ×2 stable; manifest regenerated; audit_lint --strict OK.
- Pins current at main b11811704e; parent = b192 tip afb66fc4.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7734 (archive/ledger/34/pr-7734.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7742, opened 2026-08-26
physics: prove the finite heavy-sector positive-symmetrizer cone

JONbridgerapps:
## Scoped result

At `T = 16`, unit volume, the interior window `D = {2,3,4}`, and the two declared rational fixtures, the heavy `U = -1` compression has an exact rational positive-symmetrizer cone.

- The symmetric intertwiner system `Theta = Theta^T`, `Theta T_h = T_h^T Theta` has coefficient rank `4` on ten unknowns, hence dimension `6`.
- Exact positive-definite rational symmetrizers exist.
- The exact identity `T_h^2 - tau T_h + I = 0` with `tau > 2` proves that every positive-definite symmetrizer makes `Theta T_h` symmetric positive definite.
- The declared one-site-shift condition leaves a two-dimensional displayed family.
- The declared Block-197 completion restricts to `I_4` on the heavy sector, so it does not further cut that family.

The last two bullets cover exactly two named compatibility conditions. They do not classify all commuting symmetries or physical selector principles, and this PR does not claim that canonicalization is impossible.

`Theta` is not the OS Gram. This is finite exact algebra at two fixtures, not an OS reconstruction, generic `(m,c)` theorem, continuum limit, Hamiltonian, energy, gravity result, or axiom change.

## Repair applied

- Preserved the exact cone theorem, both fixtures, and the valid rank/positivity lemmas.
- Stopped applying primary-fixture forced relations at the second fixture.
- Made the second fixture's family, parameterization, minimal-polynomial, and positive-definiteness residuals load-bearing gates with adversarial mutations.
- Replaced the exhaustive symmetry/canonicalization language with the exact two-condition scope.
- Corrected the weak-inequality geometry to a closed rational box contained in the open positive-definite region.
- Removed an unrelated R3-G/Gleason campaign hunk from this PR; it was preserved in the campaign scratchpad for separate validation.
- Rewrote the note/status surface as `proposed_retained`, conditional on the unmerged stack and independent audit.

## Verification

- Baseline: `TOTAL: PASS=35 FAIL=0`; stdout `1,898` bytes.
- All 35 predicates remain active; the mutation sweep fails only the intended gate family. The three repaired second-fixture mutations exit nonzero and fail family G.
- Independent exact re-review re-derived the `4 -> dim 6` and `8 -> dim 2` ranks, the reciprocal positive spectrum, the cone proof, the symbolic completion identity, and both-fixture positivity witnesses: `FINAL VERDICT: PASS`.
- Runner cache SHA matches the repaired runner: `403a25cc23e6622ee75e4b88f6d113812f3df4d73445b85f4feadbfb4044946f`.
- Citation graph: seven intended dependency edges; manifest regenerated and acknowledged.
- `repo_invariants_check.py --check --enforce-links`: PASS.
- `check_changed_audit_evidence.py --base e784ffc1... --json`: PASS.
- `git diff --check` and `py_compile`: PASS.

The full audit pipeline was attempted. It stops at the dependency-policy epoch manifest check. The same `ScienceFingerprintError` reproduces on the untouched parent commit `e784ffc1`, so this is an inherited stacked-base condition, not introduced by this repair. Pipeline-generated audit outputs were stripped; only the citation manifest acknowledgment remains.

## Stack status

This PR is stacked on the unmerged spatial-embedding branch and is not independently landable ahead of its ancestors. Nothing in this repair merges or closes any PR.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7742 (archive/ledger/42/pr-7742.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7741, opened 2026-08-26
[physics-loop] Block 205 H1 Schur Record probability germ — bounded theorem

jonathonreilly:
## Result

- Reconstructs the action-derived H1 two-sector right-Schur boundary state on the same `C32` carrier as the fixed Block-194 eight-effect PVM.
- Exact zero-source positivity, analytic inversion, and a nonzero algebraic derivative prove a finite interval of positive, normalized, nonuniform eight-outcome laws.
- The derivative pattern is `(ell/8)*(+,-,+,-,-,+,-,+)`, with all four coarse-port derivatives and both global pointer-sign derivatives zero.
- The fixed nonidentity `M2` writer reproduces the same effects; port-conditioned binary probabilities and one port-pointer correlation vary.

## Boundary

This is a conditional one-shot H1 law, not a periodic CAR or full-Fock state. Nearest-neighbor `eta`, autonomous formation, permanent history, H2 generality, an explicit positivity endpoint, two-TT gravity completion, axiom amendment, retained status, obligation retirement, and TOE percentage movement remain open.

## Evidence

- Primary exact runner: `8/8`; mutations rejected: `30/30`.
- Independent reconstruction: `8/8`; mutations rejected: `24/24`.
- SHA/input-pinned caches: fresh.
- Vocabulary lint: zero findings.
- Strict audit lint: no errors.
- Repository invariants: `3974` rows, zero link/class-F violations, acknowledged one-node/five-edge graph delta.
- The full audit pipeline is not reexecuted because the stacked parent already localizes the inherited dependency-policy epoch mismatch and this block changes none of its governed policy sources.

The result commit is an 18-file clean delta. The stacked PR has 24 unique paths including preregistration; six immutable prereg packet files retain trailing-blank-line warnings so their registered blob hashes remain unchanged. No audit verdict is asserted.

PR #7743, opened 2026-08-26
[physics-loop] Block 206 port-free neighbor-phase M2 context boundary

jonathonreilly:
## Result

Block 206 proves a bounded positive H1 result beyond the Block-205 Schur
probability germ.  After summing all four coarse ports before the binary
pointer, the linear response vanishes exactly, while the first surviving
coefficient is cubic:

`a_1 = 0`, `a_3 = i C`, with `C > 0` exactly.

For the six signed nearest-neighbor source displacements, the normalized
cubic contrasts are

`kappa (sqrt(3), -sqrt(3), 2, -2, 0, 0)`, with `kappa > 0`.

Thus the tested H1 state has five distinct port-free binary `M2` laws, four
nonzero, with strict inherited Schur positivity on a finite germ.  The result
passes all 24 proper-cubic frames and translations.

## Context boundary

The exact inverse-Fourier source has 110 Laurent terms, 78 spatial support
pairs, and matter/geometry radius three in both the literal forward and the
actual reverse constructions.  A same-`q`, different-`p` collision proves
that phase `q` alone does not reconstruct the action source.

The required second TT shear transforms as `T2`.  The scalar six-neighbor
shell has no linear `T2` map, whereas the conditional Pauli-adjoint `M2`
shell has a two-dimensional Hom space.  Explicit rank-three odd-shell and
even-shell decoders are constructed.  The current action/axioms do not yet
select between them.

This is a constructive escape route, not a no-go: the next decisive test is
to factor the native action into radius-one pieces and ask whether the actual
six neighboring records select one decoder without collisions.

## Scope

- Stacked on draft PR #7741 (Block 205).
- No axiom, claim-registry, audit-verdict, or TOE-score change.
- H2 remains sealed until the H1 action-factorization gate passes.
- Gravity/source work remains on its independent PR stack.
- `review-loop` was not used.

## Verification

- Primary exact runner: 8/8; 29/29 mutations rejected.
- Independent sector-path, character, and Laurent reconstruction: 8/8;
  24/24 mutations rejected.
- Both content/input-pinned caches freshly regenerated.
- Python compilation, YAML parse, vocabulary lint, staged claim typing,
  changed-evidence checks, runner-pin gate, strict audit lint, repository
  invariants, and diff checks pass.
- Citation graph: 5,635 nodes / 16,213 edges; new node outdegree 5.

## Next campaign

Run the action-factorized actual-six-record decoder discriminator.  Only if
that gate passes should H2 open; if it fails, the exact fork is to reject this
H1 realization or justify a wider physical context before considering an
axiom update.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7743 (archive/ledger/43/pr-7743.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7744, opened 2026-08-26
physics: fence Block 200 transfer probes from continuum interpolation

JONbridgerapps:
Block 200, repaired and restacked on the corrected Block 199 branch (#7742).

## What remains

- Finite exact carrier measurements at `T = 12,16,20,24` for `(m,c,v) = (9/20,5/13,1)`.
- Deep-factor agreement at `T = 16,20,24`; `T = 12,t0 = 3` is the seam-mirror slot and has a different exact factor. Boundary-slot agreement holds at all four tested widths. No all-`T` theorem is claimed.
- Exact classifications at two shear endpoints: the lower point has positive reciprocal heavy/light pairs; the upper point has a negative reciprocal heavy pair and positive reciprocal light pair. The two points do not locate an edge.
- Six exact first-order response slice sums. Per-anchor values follow parity, are distinct on `s = 2..5`, and coincide at zero on `s = 1,6`.
- Seven exact positive volume samples. No interval or maximality claim.

## Interpolation repair

The `[4/4]` object `beta_fit = -2N/D` is now explicitly typed as an interpolation model. It fits ten exact carrier values, reproduces two withheld values, and agrees projectively at seven listed probes (fourteen distinct probe shears in total).

Exact Sturm arithmetic remains valid for that model: `D` has one bracket root, while `N`, `N-D`, and `N+D` have none. Thus the model has a pole and excludes the unit band on the punctured bracket. No carrier degree bound or symbolic identity is proved, so none of those between-probe conclusions is promoted to the carrier.

The note retains the exact counter-interpolant showing why finite values alone are insufficient. It now lands a complete N1–N8 discipline packet, accepts the symbolic-degree steelman, and expressly withdraws any broad carrier no-go.

## Verification

- Baseline runner: `PASS=33 FAIL=0`; all gates `A` through `G` pass.
- Mutation sweep: all 33 declared mutations fail only their assigned gate family.
- Cached stdout: 2.9 KB, fresh SHA/input fingerprint, with five substantive N5 resolution certificates.
- Independent exact re-review: PASS after the N1–N8 and parity wording repairs.
- Citation topology refreshed: one new node with five dependency edges.
- `repo_invariants_check.py --check --enforce-links`: PASS.
- Forensic N5 readiness: PASS.
- Changed-audit-evidence check: no failures (no audit verdict is applied by this source PR).
- Full audit pipeline was attempted and stops at stage 7 with `dependency-policy epoch manifest does not exactly match its governed sources`. The same failure was independently reproduced on the untouched stacked parent during #7742 repair, so generated pipeline byproducts are excluded and the condition is disclosed as inherited.

Nothing is registered or adopted; there is no axiom movement, gravity, continuum, dynamics, energy, or claim about Nature.

Parent pin: `725269c6057deed9b7ac1f72a315297a9f99f35a`.

### DIFF
--- a/.claude/science/physics-loops/generator-program-20260821/scout_batch2_findings.md
+++ b/.claude/science/physics-loops/generator-program-20260821/scout_batch2_findings.md
@@ -1,0 +1,115 @@
+# Adversarial check: scout batch 2 (S5 + S7)
+
+Status: **COMPLETE**
+
+Construction authority: landed b190 note
+`docs/ADMISSIBILITY_DIRAC_KAHLER_WIDTH_FAMILY_TRANSFER_MONODROMY_BOUNDED_THEOREM_NOTE_2026-08-25.md`
+and its landed exact runner
+`scripts/admissibility_dirac_kahler_width_family_transfer_monodromy_2026_08_25.py`.
+
+Pinned fixture: `m = 9/20`, `c = 5/13`, `T = 16`, deep core `t0 = 3`.
+All arithmetic in this check is exact over `QQ`; `nsimplify` is not used.
+
+## S5 — response profile
+
+**Verdict: the six quoted numbers are exact, but the literal "per-single-cell"
+wording is refuted by the landed b190 cell definition.** They are the response
+of the **sum of all four spatial cells at temporal anchor `s`**, together with
+their four image partners. A single b190 cell is indexed by both `(s,x)` and
+does not give the quoted number.
+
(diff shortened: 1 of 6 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-7744 (archive/ledger/44/pr-7744.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7750, opened 2026-08-27
physics: fence Block 210 finite alphabet and four-facet diagnostics

JONbridgerapps:
## Scientific scope

- Retains the exact finite refined-alphabet result: the single-context system has ranks (4,4), while the six-context system has ranks (8,9) with augmented pivot 12.
- Identifies six exact normalization-row relations and the resulting coefficient-rank ceiling 19. This excludes a unit-sum real twelve-weight on the tested six-context design; it does not claim an outcome-count threshold or exhaust richer alphabets.
- Completes all four dichotomic Leggett-Garg facets across four chains and three splits (48 exact evaluations). The tested values remain below the macrorealist bound.
- Verifies nonnegative normalized probability profiles for every profile used by the finite landscape and seven-q sweep.

## Fences

The result is finite and diagnostic: no all-q theorem, classical no-go, joint-law theorem, physical-measurement identification, or exhaustive alphabet theorem is claimed. The macrorealist route is parked only for these tested constructions.

## Verification

- Exact runner: 29/29
- Mutation isolation: 29/29 mapped families
- Independent reconstruction from raw Block 171 data: PASS
- Cache and five-file citation-manifest delta verified
- The full repository pipeline reaches the inherited stage-7 dependency-policy epoch mismatch; this PR does not change that governed source.

This PR remains open and stacked on the repaired Block 209 branch. It is not a merge request by the campaign agent.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block210-refined-alphabet-macrorealist-landscape-20260826/NO_GO_LEDGER.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block210-refined-alphabet-macrorealist-landscape-20260826/NO_GO_LEDGER.md
@@ -1,0 +1,64 @@
+# Block 210 No-Go Ledger
+
+Date: 2026-08-26
+Status: source-side `proposed_retained`; independent audit required
+Parent: `07f0613c8730de54cd50403c809f7102bc1534bf`
+
+## Exact negative boundary
+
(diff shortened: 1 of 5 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-7750 (archive/ledger/50/pr-7750.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7749, opened 2026-08-27
physics: fence Block 209 three-direction gluing geometry

JONbridgerapps:
## Scope

Block 209 is stacked on repaired Block 202 (#7747), parent `141b9e8da04319eb2f31c53389de6edd0cf723bf`. It proposes finite exact three-direction generator, cell-gluing, and exterior-algebra results for retention. It does not register or adopt a physical rule, metric field, spacetime, dynamics, gravity law, or continuum limit.

## Retained exact results

- The declared three generators are involutions and anticommute pairwise. The three tested all-even extents scalarize on every directed link with the derived eta pattern; `(4,3,2)` has exactly eight non-scalar `x`-wrap links.
- The three-origin-face literal system has ranks `(22,23)` and five compatibility relations. The six-face system has ranks `(32,33)` and sixteen relations.
- On the uniform positive-volume locus, `c^2+v^2-1=0` and `c^2 v=0` give only `(c,v)=(0,1)`.
- The exhibited reciprocal point `(3/5,12/25,3/4,4/5)` is solvable with four free complementary pairings, but its parameter-free degree-1 and degree-2 principal blocks have spectra `(-3/20,6/5,6/5)` and `(-5/4,15/4,15/4)`, so that point is not positive definite.
- The declared standard exterior-algebra candidate `D3(g,V)=diag(V,Vg^-1,EgE/V,1/V)` obeys the exact Jacobi/compound identities. Its origin-face restrictions satisfy the stated Schur identity.
- On the tested isotropic locus, exact Sylvester minors give positivity ranges `0 <= kappa < 1` and `0 <= kappa < 1/2` for the two product-sign classes.

## Repaired boundary

The draft all-face Schur reading was false. Origin faces have global degrees `(0,1,1,2)`, while opposite faces have `(1,2,2,3)`. At `(c_tx,c_ty,c_xy,V)=(1/5,1/7,1/9,2)`, the opposite-face residual matrices have exact nonzero-entry counts `(5,5,5)` and `(0,0)` defects `(1310/23159,2282/23159,2682/23159)`. The three 2-form couplings are hidden on origin faces but visible on opposite faces. No local regrading or transport map is assumed.

Accordingly, the note withdraws every-face, framework-forced/uniquely-selected lift, global flat-only, and physical-selection readings while preserving all exact finite results and the open alternatives: other nonuniform compatible branches and a separately derived local face transport.

## Verification

- baseline runner: `36/36` PASS; stdout 2719 bytes
[...]

PR #7752, opened 2026-08-27
physics: prove Block 211 six-face positivity classification

JONbridgerapps:
Block 211 — THE SIX-FACE POSITIVITY CLASSIFICATION. Stacked on block 210 (#7750); the scientific parent is block 209 (#7749), whose named open branch this block closes. b209 is COMPLETED, NOT CORRECTED: its rigidity statement was scoped "on the checked branches" with positivity explicitly unclassified over the nonuniform branches — that scoping held exactly, and this block classifies the branch it named open.

THE VARIETY. The six-face-compatible moduli variety on the nonsingular domain is EXACTLY the per-offset-isotropic family: per-offset common volumes and common shear magnitudes tied by v1(1 - g0^2) = v0 and g1^2 = 1 - v0 v1 (equivalently v0^2 = (1-g0^2)(1-g1^2)). NO non-isotropic branch exists — the derivation uses exactly two divisions, both by volumes, and the converse polynomial normal form kills all sixteen landed relations identically. Signs enter only through the per-offset products: 64 sign cells in 4 gauge classes. The uniform locus still forces exactly (c, v) = (0, 1). The coefficient matrix is CONSTANT with exactly the four duality columns zero — which is why one classification covers every moduli point at once.

THE BLOCKS. Solvability always leaves exactly the four duality parameters, ABSENT from every degree block at SYMBOLIC moduli in every sign class (b209's principal-submatrix argument, promoted from one point to the whole variety). The signed-triangle determinants are (1+g)^2(1-2g) at pi = +1 and (1-g)^2(1+2g) at pi = -1 — an exact class-negated echo of b209's honest-lift orientation split, noted and never interpreted.

THE CLASSIFICATION (the headline). D is PD-solvable at a compatible point IFF gamma0 < 1/2 when pi0 = +1 (any gamma0 < 1 when pi0 = -1) AND independently the same for (gamma1, pi1). EVERY gauge class carries a two-parameter CURVED positive definite family; grid census over 32 exact solves: 3/8, 3/8, 3/8, 8/8. THE NATURAL FLAT-ONLY EXTENSION IS REFUTED — correction #111, the supervising record's own announced conjecture (never landed; b209 declined to assert it), killed by explicit witness W1 (magnitudes 1/4, all-plus, v0 = 15/16 — all eight leading minors strictly positive) and carried as the headline rather than smoothed.

THE RECONCILIATION. W3 carries the landed magnitudes (3/5, 4/5) with exactly one shear sign flipped per offset and is positive definite: b209's exhibited indefinite point was killed by its SIGN PATTERN, not its curvature. The all-plus point is reproduced unchanged — still indefinite, with the landed spectra (-3/20, 6/5 x2), (-5/4, 15/4 x2), read from b209's own runner literals rather than retyped (the runner imports b209 and gates against its constants).

Landing chain receipts: baseline 31/31 PASS in the landing worktree; 31-mutation battery one-family-each, all nonzero-exit; runner cache stable x2; audit pipeline exit 1 (expected) with docs/audit and docs/publication restored; citation graph manifest +1 node; audit_lint --strict OK; current_main 76df4becc8 re-verified at push; parent pin e3a6c02acb (block 210 tip). Hygiene: zero nsimplify, zero float literals, ZERO float call sites by AST. Checks: the R1 classification at 43/43 plus an independent exact reconstruction at 21/21 (C1/C2 in full; the C6 qualifications are the note's six fences). Same disclosed registry-note divergence as the rest of the stack.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7752 (archive/ledger/52/pr-7752.json). Carrier: #7781 (open L24-lane terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7756, opened 2026-08-27
physics: construct the metric-weighted exterior kernel

JONbridgerapps:
First new science block after the repaired #7753 stack. This PR is intentionally left open for the review-loop landing process.

MAIN RESULT. Starting from the declared Block 209 carrier D3(g,V)=diag(V,Vg^-1,EgE/V,1/V), define Gamma_d=epsilon_d+epsilon_d^dagger using the D3 inner product and K=sum_d Gamma_d nabla_d. Exact exterior algebra gives

- Gamma(q)^2 = q^T g^-1 q I8,
- -K(k)^2 = sin(k)^T g^-1 sin(k) I8,

including the off-diagonal direction-mixing terms.

NEW SELECTOR. The degree-2-to-1 adjoint carries the exact factor det(g)/V^2. For positive g and V, full generalized-Clifford closure is therefore necessary and sufficient at V^2=det(g). Inside the supplied D3 candidate family, propagation fixes V to the metric volume rather than leaving it independent.

BRIDGES CHECKED EXACTLY.

- At g=I,V=1 an explicit integer intertwiner shows the 8-component exterior carrier is two copies of Block 209s 4x4 Cl(3,0) rule; all 24 parity/direction links scalarize to eta_d I8/2.
- The rational 2D point c=3/5,V=4/5 restricts exactly to the landed shear-Hodge form and carries its inverse-metric quadratic.
- A genuinely mixed 3D point c_tx=c_ty=c_xy=11/50, V=117/125 has positive D3, nonzero off-diagonal co-metric entries, exact Clifford closure, and six exact values across all 64 momenta of the 4^3 grid.
- Pure scalar axis reweighting cannot reproduce shear because it has no mixed momentum coefficient; matrix-valued link mixing is load-bearing.

VERIFICATION. Runner PASS 18/18 with source/input-bound cache. A separate exact reconstruction returned PASS and independently confirmed the volume selector, symbol, flat intertwiner/scalarization, both rational witnesses, and positivity.

SCOPE. Constant positive Euclidean cell only. This does not select D3 or g from the axioms, identify physical time, provide a Lorentzian/causal cone, handle variable cells, supply dynamics/gravity, or take a continuum limit. The next science target is local transport between neighboring D3(g_s,V_s) carriers.

### DIFF
--- a/.claude/science/physics-loops/toe-axiom-closure-block214-weighted-kernel-metric-symbol-20260827/NO_GO_LEDGER.md
+++ b/.claude/science/physics-loops/toe-axiom-closure-block214-weighted-kernel-metric-symbol-20260827/NO_GO_LEDGER.md
@@ -1,0 +1,39 @@
+# Block 214 no-go ledger — weighted exterior kernel
+
+## Rejected routes
+
+1. **Independent scalar weights on the three staggered axes.** Their squared
+   symbol is diagonal in the direction labels,
+   `sum_d w_d^2 sin^2(k_d)`, so it cannot reproduce any nonzero
+   `(g^-1)_de sin(k_d) sin(k_e)` term. A sheared metric needs matrix-valued
+   direction mixing, oblique hops, or an equivalent non-scalar carrier.
+
+2. **Treating the D3 parameter `V` as independent while demanding a single
+   metric quadratic form on every exterior degree.** The exact middle-degree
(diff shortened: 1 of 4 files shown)


JONbridgerapps:
Consolidated to the archive: ledger id pr-7756 (archive/ledger/56/pr-7756.json). Carrier: #7803 (open metric-dynamics terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7762, opened 2026-08-28
[physics-loop] toe-axiom-closure-block219: bounded_theorem / bounded-support

jonathonreilly:
## Result

Block 219 replaces the Block-218 nonlocal shared-four-bit mixed-front
benchmark with an exact size-independent bounded-local formation law before
permanent Record commitment.

The explicit transient bit

`tau_b = |b>_center tensor (uniform shell-weight-three state)`

lies in the rank-76 complement of the seven-qubit Record code.  Proper-cubic
rotations fix it and global complement swaps its two states.  The identical
displacement-two comparison graph has four components and GF(2) rank
`L^2 - 4` on every finite even transverse period.  Exact ranks are 12, 32 and
60 on periods 4, 6 and held-out 8, with exactly 16 lawful words.

The local syndrome instrument has three orthogonal readable outcomes:
equal-transient, unequal-transient and outside-lock.  This separation prevents
an existing Record or malformed endpoint from masquerading as a successful
zero syndrome.  Every typed overlap critical pair commutes.  The all-zero
branch is a valid local factorization but remains a globally postselected
transcript, with raw weights `2^-12`, `2^-32` and `2^-60`.

A stronger normalized voter/coalescence channel removes that postselection
before permanent write on the all-transient precursor sector.  Every selected
mismatch becomes 00 or 11 with equal
weight.  On each finite parity component the only absorbers are consensus,
and absorption is almost sure.  Component densities and their products are
martingales, so all 16 terminal planes retain weight `1/16`, including under
fair adaptive edge ordering that cannot choose the random local outcome.  A
mixed Record/transient fixed witness makes the domain restriction explicit.

## Boundary and TOE effect

The surviving wall is conflict-safe irreversible Record conversion.  A site
whose complete local comparison view is zero may still have a remote one
domain.  Premature zero commit conflicts with eventual one consensus with
exact probabilities `1/4`, `4/9` and `11/16` on periods 4, 6 and 8.  This
excludes memoryless local-even and finite-patience quiet guards only; returning
coverage waves, emergent leaders, revocable reservations, event seeds, larger
alphabets/radii and coherent or continuous-time protocols remain live.

Event authority, the physical rate, permanent commit, renewal,
infinite-volume stopping and complete readable history remain supplied or
open.  No axiom amendment, audit verdict, retained status, obligation
retirement or TOE percentage movement is claimed.  Lane scores therefore
remain Records `95/92/50`, causal time `76/72/41`, matter `95/96/75`,
gravity/source `70/45/29`, and Born/history `84/63/34`.

## Verification

- Primary: `38/38`; mutations `26/26` rejected.
- Structurally independent held-out checker: `33/33`; mutations `23/23`
  rejected; no primary import.
- The held checker enumerates all 65,536 all-transient states of one
  period-eight component and certifies fair-scheduler absorption.
- The executable numerical digests freeze carrier supports, branch labels and
  entries, footprints, stencil, symmetries and outside lock.
- The committed N1--N8 checklist returns `FAIL` for a broad finality no-go and
  demotes the negative to `partial-narrowing`; every route outside the narrow
  local-even/finite-quiet counterexamples remains live.
- The user-disabled `review-loop` was not used.

## Next campaign

Preregister a size-independent coverage/finality tournament on the same typed
transient carrier.  Test returning waves, revocable reservations and
randomized emergent-leader/echo protocols on periods 4 and 6 with period 8
held out.  Require strong safe diamonds for every Record-touching critical
peak and explicitly search disjoint same-component opposite commits.  Keep
the same-variable Cartan/Palatini--Regge action/Hessian bridge as the
independent gravity hedge.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7762 (archive/ledger/62/pr-7762.json). Carrier: #7803 (open metric-dynamics terminal; commit-ancestry verified). Branch retained; science line in the ledger entry.

PR #7772, opened 2026-08-28
[physics-loop] Block 224: backoff fairness quotient boundary

jonathonreilly:
Stacked science draft over Block 223. Corrects the hostile scheduler classification: zero absorption under weak finite-delay fairness, but no nonterminal recurrent class under action/support strong fairness on the exact six-row quotient. Local dynamics and full CP remain open.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7772 (archive/ledger/72/pr-7772.json). Carrier: #7773 (builds the seam controller on the Block-223/224 carrier and static anchors, inheriting the strong-fair scheduler assumption). Branch retained; science line in the ledger entry.

PR #7784, opened 2026-08-28
[physics-loop] Block232: J_r temporal-spatial coarse response

JONbridgerapps:
## Science result

The literal Block220/221 metric/source derivative is not defined on the current Block229/231 carrier: that carrier is pure gauge and omits coframe, matter, source, and their measures. A scalar coefficient tangent alone would be response bookkeeping, so that route remains BACKLOG.

This block executes the joint space-time pivot on the actual changing carrier. For `J=J_r`, `Q=JJ*`, and the complete fine co-scaled step `S_epsilon`, it proves

```text
D_epsilon = J*S_epsilon^2 J - (J*S_epsilon J)^2
          = J*S_epsilon(I-Q)S_epsilon J
          = [(I-Q)S_epsilon J]*[(I-Q)S_epsilon J] >= 0.
```

On the physical Peter--Weyl core the leading term is `Gamma=J*G_f(I-Q)G_fJ`. For retain-every-two it is a nonconstant generated coarse interaction. For arbitrary fixed `r,q,n`, conditioned fibers are `(r-1)`-wise Haar and the leading nonconstant multiplication response is the centered `r`-fold convolution with coefficient `(-1)^r(2^r-2)`.

The exact quadratic half-response is `R_epsilon=(1/2)partial_lambda^2 D_epsilon|_0=L_epsilon*L_epsilon`. For `r>=3`, its next core coefficient closes on the actual O(3) carrier as

```text
B*A_fB=gamma A_c^ind+u_(r,q)I,
mathcal K=2gamma A_c^ind+u_(r,q)I.
```

Thus `epsilon^3 lambda^2` contains a non-scalar coarse-kinetic response, with no first-order or coarse-word-dependent zero-order remainder at that bidegree.

For determinant words `y!=z`, let `d=d_H(y,z)` and `m=rd`. The first possible complete physical offdiagonal derivative is order `m`; the exact full-O(3) coefficient is a positive sum over residual fine histories after `Q` deletes exactly the `2^d` block-cylindrical histories. Its small-step coefficient is

```text
epsilon^m(c_det^(n))^m(2^m-2^d) product_(p in H)a_p + o(epsilon^m).
```

The complete minimal determinant response is now resummed exactly as two ordered bond-dimension-four cell contractions. With

```text
T_(b,sigma)[u,v]=1_(sigma=1 or v=b)t_det^[2v+2(1-u)v],
M_(b,0)=T_(b,0) tensor T_(1-b,1),
[...]

PR #7786, opened 2026-08-29
[physics-loop] derive arbitrary-width scalar-fused vector transfer

JONbridgerapps:
## Summary

- derives the exact selected q=1 vacuum-to-coarse-defining-vector response for every supplied fixed finite r>=2 on the actual original-link O(3) ladder
- proves every derivative below r vanishes, both Gram orientations retain only the empty/full cylindrical histories, and the normalized complement overlap is 3^(1-r)
- resums the full finite-step coefficient into a three-state transfer with a separate exact two-state endpoint subtraction
- recovers the reviewed r=2 and r=3 entries and the small-step coefficient (2^r-2)/3^(r-1)

## Honest boundary

This is conditional support on the supplied action, temporal multipliers, Haar normalization, ladder, and physical J_r,Q stack. It is one selected one-cell vector entry, not a full non-determinant kernel, selected physical action/time, continuum result, or gravity/metric/source/matter identification. No axiom or approved primitive changes.

## Verification

- primary suite: 16/16
- independent mode wired through the audit entrypoint: 11/11
- standalone Fraction/signed-frame helper: 12/12
- 11 load-bearing mutations each fail exactly one intended check
- py_compile, diff check, vocabulary lint, fresh runner cache, N1-N8/N5 evidence, strict audit lint, and citation manifest pass
- two fresh same-session reviewers independently reconstructed/refuted the proof and returned conditional/bounded PASS after all fixes

The full audit pipeline on this stacked base still stops at the inherited dependency-policy epoch-manifest debt already recorded for the parent stack; this PR does not edit that policy surface. Independent audit and dependency closure remain required.

## Stack

Base: `physics-loop/toe-connection-dynamics-block232-metric-source-coarse-response-20260828` (parent review PR #7784). This PR is for review only and is not being merged here.


JONbridgerapps:
Consolidated to the archive: ledger id pr-7786 (archive/ledger/86/pr-7786.json). Carrier: #7803 (open conndyn terminal; decided reconcile item 8; ancestry verified). Branch retained; science line in the ledger entry.

PR #7834, opened 2026-09-03
feat(matter): finite BKSF sign checks and superlattice marker census

jonathonreilly:
## Summary

Bounded theorem note + exact runner + cache: a three-dimensional fermionic point particle exists on `Z^3` with **one qubit per site**, **one translation- and rotation-covariant law**, and **ordinary (commuting) composition** of the sites. The fermion is not a site; it is a shaped configuration (a marked corner site whose six neighbours together carry odd parity) on a **superlattice role pattern** the law forms by itself (period (4,2,2), 48 equivalent sectors, eight fine sites per fermionic mode). A separate exact theorem shows the pattern is not optional in the Pauli stabilizer class: one site per repeating unit gives no 3D point charge at all.

- `docs/EMERGENT_3D_FERMION_ONE_QUBIT_PER_SITE_SUPERLATTICE_ROLE_PATTERN_EXISTENCE_BOUNDED_THEOREM_NOTE_2026-09-02.md` (363 lines)
- `scripts/emergent_3d_fermion_one_qubit_per_site_superlattice_role_pattern_existence_check_2026_09_02.py` (`TOTAL: PASS=24 FAIL=0`, 72 s, every line `[exact]`, no floating point, no external solver)
- `logs/runner-cache/emergent_3d_fermion_one_qubit_per_site_superlattice_role_pattern_existence_check_2026_09_02.txt`

## Theorems (all exact, finite clusters named in the note)

- **T1 existence**: Bravyi-Kitaev superfast encoding on the coarse sublattice `2Z^3`; relations R0-R4; `k = V-1` on open blocks; full vacuum stabilizer set gives `k = 3` on tori `3^3`, `4^3`, `3x3x4`, independent of `L` (genuine 3D Z2 gauge structure, not foliated); hop support 11 edge sites, radius 2, commuting with all faces (144000/144000 pairs); closed 3D circuits in the stabilizer group with phase +1; Levin-Wen T-junction `-1` over 10 geometries and direct exchange `-1` over 4; controls: bound pair `+1`, 3D toric-code charge `+1`, 2D toric-code epsilon `-1`.
- **T2 selection**: the symmetrised marker rule's zero set is exactly the 48 sectors (branch-on-role propagation, no SAT), on `4x4x4` and `8x4x4`; incommensurate tori `5x4x4`, `4x5x4`, `7x4x4` carry no zero-penalty configuration; `5x5x5` window minimal.
[...]

PR #7844, opened 2026-09-03
[review-loop salvaged and landed] finite BKSF pi-flux sign class

jonathonreilly:
## Summary

Bounded theorem note + exact runner + cache: the framework's landed **staggered (Kawamoto-Smit) kinetic form**, read on the coarse sublattice `2Z^3` of the emergent fermion of PR #7834, is **the all-minus face-stabilizer sector of that fermion's own Z2 gauge structure**. Transporting one encoded excitation once around a coarse face equals the face stabilizer exactly, phase included, so the staggered sign field is the sector with every `S_f = -1` and the plain field the sector with every `S_f = +1`. Under the law as written the two sectors are a free choice.

- `docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md` (338 lines)
- `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py` (`TOTAL: PASS=23 FAIL=0`, 1.3 s; symbolic, `Z[i]`, finite-field rank, `F2`/`Z4` symplectic; numerical items are cross-checks)
- `logs/runner-cache/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.txt`

## Theorems (coarse lattice `2Z^3`, one fermionic mode per coarse vertex, named finite blocks and tori)

- **T1** the KS-phased hopping on a `2x2x2` cell is a `Cl(6)` operator: `H(q) = sum_a [(1 + cos q_a) Xi_a + sin q_a Gamma_a]`, `H^2 = (6 + 2 sum cos q_a) I`, spectrum `+-sqrt(6 + 2 sum cos q_a)` each fourfold; exact torus spectra.
- **T2** an exact intertwiner with `Z[i]` entries, unique up to a phase, gives `U H(pi + p) U^dag = sum sin p_a (sigma_a x T) + sum (1 - cos p_a)(I x B_a)`: an exact spin(2) x taste(4) split whose taste-mixing term is a spin singlet.
- **T3** the KS sign is a pi-flux class (holonomy `-1` on every plaquette), not a local gauge choice, and is a function of the fine coordinates mod 4, not of the mod-2 role pattern.
- **T4** the proper cubic group acts on the eight zero modes as the genuine representation `2 A1 + 2 T1` (sharpening the landed `1 + 3 + 3 + 1` Hamming grading), factorising into individually projective spin (spinor) and taste factors with one shared non-coboundary cocycle.
- **T5** the chirality grading `epsilon` is `I x (+- T B_1 B_2 B_3)`, never the taste pseudoscalar `T` (consistent with the landed narrow theorem).
- **T6** face transport `=` face stabilizer exactly; the all-minus sector exists on every open block and on an `L1 x L2 x L3` torus iff every pairwise product `L_a L_b` is even (at most one odd period, the KS periodicity condition), verified against an independent `F2` pi-flux solve on twelve blocks and tori; in that sector the encoded hopping is in the KS local gauge class with an explicit witness and the same one-particle spectrum.
- **T7** many-body cross-check on the `4096`-dimensional `2x2x2` coarse cube: the `S_f = -1` sector is the even-parity Fock space and the encoded hopping's 128 levels match the free-fermion spectrum with the induced signs.

## Corollary and boundary

- The staggered kinetic form and the plain form are the same one-qubit-per-site law read in the two flux sectors of its own gauge structure; the law attaches no coefficient to any face term, so the sector is a free choice; a face term `-J sum S_f` with `J < 0` would select the staggered sector (none is supplied).
- The staggered gate's spin-statistics support residual is supplied by construction on the coarse lattice (the modes belong to an encoding whose excitations exchange with `-1`); whether that changes its status is an audit question.
- Coarse lattice only; sign classes and free-hopping spectra, not dynamics; nothing derived from the axioms; the encoded model has no odd-parity sector; no corner of the taste cube is labelled (the labelling no-go is cited as a pointer only). The mod-4 structure of the KS sign and the period-(4,2,2) role pattern are not shown to be the same object.

## Dependencies and context

- Cites `EMERGENT_3D_FERMION_ONE_QUBIT_PER_SITE_SUPERLATTICE_ROLE_PATTERN_EXISTENCE_BOUNDED_THEOREM_NOTE_2026-09-02.md` (PR #7834) by filename as a pointer; the landed staggered-gate, chirality and labelling notes are quoted verbatim, each re-verified on `origin/main`.
- Independent verification (scratch, not landed): the whole package was re-derived by a second agent with its own code before landing; two of its claims were corrected in the process (a wrong torus criterion in its prose, and a convention-dependent cocycle count now flagged as such).
- Part of the owner-authorised 24-hour matter-to-TOE campaign (2026-09-02/03).

## Review-loop disposition (2026-09-04)

The original physical/framework claim was rejected by adversarial review. Its durable finite core was rewritten as a self-contained bounded theorem, re-reviewed to PASS, and landed on main as commit a1a64d36bc856c3ae330786972d3f2a02b678c52 (following the paired #7858 salvage commit). Full pipeline, strict lint, and changed-evidence gates passed on the exact current-main train. No audit verdict was applied; independent audit remains separate. The complete review and fix-confirmation record is attached in the comments.

### DIFF
--- a/docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md
+++ b/docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md
@@ -1,0 +1,338 @@
+---
+claim_id: emergent_fermion_pi_flux_sector_staggered_kinetic_form
+claim_type: bounded_theorem
+claim_scope: "On the coarse cubic lattice 2Z^3, carrying one fermionic mode per coarse vertex in the Bravyi-Kitaev superfast encoding written on it, with the Kawamoto-Smit link signs eta_1 = 1, eta_2(x) = (-1)^{x_1}, eta_3(x) = (-1)^{x_1+x_2} of the landed staggered kinetic-form clause read on that lattice, and on the named finite blocks and tori only: (T1) with one 2x2x2 cell of coarse sites as the unit cell the hopping has Bloch Hamiltonian H(q) = sum_a [(1 + cos q_a) Xi_a + sin q_a Gamma_a], Gamma = (Y_1, Z_1 Y_2, Z_1 Z_2 Y_3) and Xi = (X_1, Z_1 X_2, Z_1 Z_2 X_3) a Cl(6) set of anticommuting hermitian involutions, hence H(q)^2 = (6 + 2 sum_a cos q_a) I and tr H(q) = 0 symbolically; the coarse torus L = 4 spectrum is exactly -2sqrt3 x4, -2sqrt2 x12, -2 x12, 0 x8 with its mirror over 64 modes, and L = 6 (216 modes, 0 zero modes) and L = 8 (512 modes, 8 zero modes) reproduce the same Bloch prediction eigenvalue by eigenvalue at 1e-9. (T2) The intertwiner U with U Gamma_a U^dag = +- sigma_a (x) T, T = diag(1,1,-1,-1), and U Xi_a U^dag = I (x) B_a, B = (XX, XY, XZ), exists with entries in Z[i] by Clifford averaging and satisfies U U^dag = 16 I; the 384x64 intertwining system has rank 63 over a prime field with a square root of -1, so each branch nullspace is exactly one-dimensional and U is unique up to a phase; U H(pi + p) U^dag = sum_a sin p_a (sigma_a (x) T) + sum_a (1 - cos p_a)(I (x) B_a) over 300 random p at 1e-12, the taste-mixing term carrying the identity on the spin factor. (T3) Every plaquette holonomy of the KS signs is -1: all 24 coordinate-parity-class and plane combinations, and all 192, 648 and 1536 plaquettes of the coarse tori L = 4, 6, 8; no site gauge carries the plain signs to them, by loop invariance and by exhaustion of all 2^8 sign patterns on the open 2x2x2 coarse block; the fine-mod-2 role pattern of a coarse cell takes exactly 1 value over all coarse sites while the KS sign varies, and the KS sign is a function of the fine coordinates 2v mod 4 over an 8^3 coarse block. (T4) All 24 proper cubic rotations lift to signed permutations with C_R H C_R^T = H on the L = 4 coarse torus, all 576 products close exactly at the 64x64 level and all 576 on the eight zero modes with 0 pairs carrying -1, a genuine representation of O; its characters (E, 8C3, 3C2, 6C4, 6C2') = (8, 2, 0, 4, 0) decompose as 2 A1 + 2 T1 against a character table checked orthonormal in the runner; in the B basis every lift factorises as a 2x2 spin factor times a 4x4 taste factor at singular-value ratio 9e-17, both factors carrying one and the same sign cocycle with 208 of the 576 pairs at -1 under the runner's determinant normalisation, and that cocycle is not a coboundary, so each factor is individually projective while the product is genuine. (T5) epsilon = Z_1 Z_2 Z_3 anticommutes with all six generators, hence with H(q) at every q, and U epsilon U^dag = I (x) (+- T B_1 B_2 B_3) on the two branches and never +- I (x) T. (T6) (B_j - B_i) = 2 on exactly the legal steps, so the encoded hop T_ji = (i/2) A_ji (B_j - B_i) acts there as i A_ji, and the ordered product of four encoded hops around a coarse face equals the face stabilizer S_f exactly with its Z4 phase on 6, 36, 81 and 192 faces of the open 2x2x2, the open 3x3x3 and the tori 3^3 and 4^3; every F2 relation among the S_f has ordered product exactly +I on all twelve blocks and tori tested, so the sector S_f = -1 for every f is consistent exactly when every relation has even support, which holds on every open block and on an L1xL2xL3 coarse torus exactly when every pairwise product L_a L_b is even, verified against an independent F2 solve for a pi-flux edge sign field over all 2424 faces of the twelve; the sign field eta' read off that sector with no input from KS has holonomy -1 on every plaquette and admits a gauge witness s(v) carrying it to the KS signs on the open 2x2x2, 3x3x3 and 4x4x4 blocks and the 4x4x4 coarse torus, with 4, 7, 32 and 24 sites at s = -1, and the one-particle spectra of eta' and of KS agree there at 1e-9. (T7) On the open 2x2x2 coarse cube with 12 edge qubits and dimension 4096, prod_i B_i = +I, the six face stabilizers have F2 rank 5, the joint S_f = -1 eigenspace is one-dimensional on each of the 128 B-configurations giving 128 = 2^(V-1) states, and the encoded hopping restricted there matches the even-parity many-body spectrum of sum_ij eta'_ij c_i^dag c_j on all 128 levels at 1e-13. Nothing here is derived from any axiom, no axiom is amended, no status is set, and no hypothesis is adopted."
+upstream_dependencies: []
+runner: scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py
+---
+
+# The staggered kinetic form is the other flux sector of the emergent fermion's own gauge structure
+
+**Date:** 2026-09-02
+**Type:** bounded_theorem
+**Audit:** unset; independent audit remains a separate lane
+**Status:** bounded - bounded or caveated result note
+**Status authority:** independent audit only. This source changes no axiom, primitive, framework rule, or audit verdict.
+**Primary runner:**
+[`scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py`](../scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py)
+**Runner cache:**
+[`logs/runner-cache/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.txt`](../logs/runner-cache/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.txt)
+**Parents:** none. Every premise used below is declared in this note.
+
+The framework's landed staggered kinetic clause fixes a sign on every nearest-neighbour link and calls the resulting sign field unique as a local `Z2` gauge class. A
+separate construction supplies a fermion on the coarse sublattice `2Z^3`, one mode per coarse vertex, with a `Z2` gauge structure of its own whose gauge-invariant
+content on a face is a face stabilizer of eigenvalue `+-1`. The question this note answers is whether those two sign structures are the same object. They are:
+transport of one encoded excitation around a coarse face equals that face's stabilizer exactly, phase included, so the staggered sign field is the sector in which
+every face stabilizer is `-1`, and the plain field is the sector in which every one is `+1`.
+
+## Machine status
+
+```yaml
+actual_current_surface_status: bounded-support
+target_claim_type: bounded_theorem
+claim_type_reason: "Exact finite-cluster theorems identifying the landed staggered kinetic sign field with the all-minus face-stabilizer sector of the coarse-lattice emergent fermion, together with the exact Clifford, spin-taste, point-group and chirality structure of that kinetic form. Every statement is symbolic, integer, Z[i], F2/Z4 or exhaustive; the tagged numerical items are cross-checks of statements already established exactly."
+trace_class: frontier_discovery
+target_claim_id: null
+target_blocker_text: null
+source_of_blocker_text: frontier_question
+reachability_to_target: unknown_frontier
+artifact_role: theorem
+next_trace_action: "Run independent audit on this self-contained finite-cluster theorem, and route to its owner the science-level question this note does not decide: which flux sector a dynamical clause would select."
+conditional_surface_status: null
+audit_required_before_effective_retained: true
+bare_retained_allowed: false
+hypothetical_axiom_status: null
+admitted_observation_status: null
+```
+
+## Exact target
+
+The target is the conjunction of the seven statements below, exactly the runner's check groups `A`-`G`. Groups `A`-`C`, `E` and the structural content of `F` are
+exact -- symbolic identities in `sympy`, integer and `Z[i]` matrix arithmetic at zero tolerance, rank over a finite field, `F2` and `Z4` symplectic bit arithmetic,
+exhaustive enumeration -- and the items tagged `[numerical]` are floating-point cross-checks, at the stated tolerance, of what those groups establish.
+
+1. `T1` (`A`). The kinetic form: a `Cl(6)` operator on the eight-site coarse cell, with its exact square, trace and torus spectra.
+2. `T2` (`B`). The spin-taste split: an exact intertwiner, unique up to a phase, taking the operator to `sum sin p (sigma (x) T) + sum (1 - cos p)(I (x) B)`.
+3. `T3` (`C`). The sign field is a flux class, not a local gauge choice, and is a function of the fine coordinates mod `4`.
+4. `T4` (`D`). The proper cubic group on the eight zero modes is the genuine representation `2 A1 + 2 T1`, factorising into individually projective factors.
+5. `T5` (`E`). The chirality grading is `I (x) (+- T B_1 B_2 B_3)`, distinct from the taste pseudoscalar.
+6. `T6` (`F`). Face transport equals the face stabilizer, the all-`(-1)` sector exists exactly when at most one coarse period is odd, and in it the encoded hopping
+   is in the staggered local gauge class, with a witness and the same one-particle spectrum.
+7. `T7` (`G`). The many-body cross-check on the open `2x2x2` coarse cube.
+
+## Imports and authority
+
+Imported scientific authority: none load-bearing. The Bravyi-Kitaev superfast encoding, the Kawamoto-Smit staggering, the Dirac-Kahler spin-taste basis and the
+character table of `O` are standard methodology; every object is redeclared here and the runner recomputes every statement, the table's orthonormality included. No
+observational value, no fitted number and no framework premise enters any proof. Non-load-bearing pointers, carrying no grade and no dependency weight:
+
+- `EMERGENT_3D_FERMION_ONE_QUBIT_PER_SITE_SUPERLATTICE_ROLE_PATTERN_EXISTENCE_BOUNDED_THEOREM_NOTE_2026-09-02.md` (open PR #7834): the role pattern, the coarse
+  sublattice `2Z^3` and the superfast encoding on it. Pointer only; the encoding is redeclared below and recomputed by this runner.
+- `STAGGERED_DIRAC_REALIZATION_GATE_NOTE_2026-05-03.md`: the kinetic-form clause and the corner-structure clause quoted below.
+- `CHIRAL_CONTENT_IS_THE_EPSILON_D_CHIRALITY_IMPORT_DISTINCT_FROM_ORIENTATION_NARROW_THEOREM_NOTE_2026-06-08.md`: the chirality-grading row quoted below.
+- `STAGGERED_DIRAC_SUBSTEP4_LABELING_NO_GO_NOTE_2026-05-17.md`: the labelling no-go, cited as a pointer only.
+- `MINIMAL_AXIOMS_2026-06-29.md`: the four framework axioms quoted in "Setting". This note cites none of their grades and adopts no hypothesis.
+
+## Setting
+
+The four framework axioms are quoted, not amended. **Lattice**: "Physical sites are the points of the cubic lattice `Z^3`, with nearest-neighbor adjacency, standard
+translations, and proper cubic rotations about each site." **Qubit**: "Each site has a domain of local possibilities", whose "full one-site possibility domain has
+algebraic presentation `M_2(C)`". **Admissibility**: "There is one fixed nearest-neighbor admissibility rule, covariant under lattice translations and proper cubic
+rotations", and "For each site, the probability distribution over the possibilities is determined by, and varies with, the nearest-neighbor conditions." **Record**:
+"Records form", "a record locks exactly one admissible local possibility", "records are permanent", "Only records are readable."
+
+The landed kinetic clause under test reads, verbatim:
+
+> **Kinetic-form clause.** Within the declared kinetic class (the naive-Dirac kinetic form on nearest-neighbor `Z^3` links, made
+> compatible with the matter-statistics clause by site-local spin diagonalization), the kinetic operator is the staggered operator
+> `D = (1/2) Σ_{x,μ} η_μ(x) (χ̄_{x+μ̂} χ_x − χ̄_x χ_{x+μ̂})` with the Kawamoto-Smit phases `η_1 = 1, η_2(x) = (−1)^{x_1},
+> η_3(x) = (−1)^{x_1+x_2}`, unique as a local Z2 gauge class.
+
+and its corner-structure clause reads:
+
+> **Corner-structure clause.** The free staggered operator has the 8-element BZ-corner (taste-cube) doubler set, decomposing uniquely
+> by Hamming weight as `1 + 3 + 3 + 1`; the hw=1 triplet carries an exact irreducible `M_3(C)` algebra (translations + `C_3[111]`)
+> with no proper exact quotient.
+
+Everything below reads that sign field on the coarse lattice `2Z^3`, one fermionic mode per coarse vertex. Composition is **ordinary** throughout: the algebra of a
+region is the tensor product of its sites' algebras and no graded clause is used anywhere.
+
+## Obligation graph
+
+The proof is acyclic and each node after `P0` is checked by the correspondingly lettered runner group. `P0`, declared here, is the coarse lattice, the KS sign field
+on it, the eight-site cell, the superfast encoding, the encoded hop and the face stabilizers. `P1` (`A`) is the `Cl(6)` form, its square and trace and the torus
+spectra; `P2` (`B`) the intertwiner, its uniqueness and the spin-taste Hamiltonian; `P3` (`C`) the flux class and the mod-`4` structure of the sign; `P4` (`D`) the
+cubic lift, its closure, characters and factorisation; `P5` (`E`) the chirality grading and its image; `P6` (`F`) face transport, the sector's existence condition,
+the induced sign field and its gauge witness; `P7` (`G`) the many-body cross-check. The strongest supported scope is precisely `P0`-`P7`.
+
+## Definitions
+
+The **coarse lattice** is `2Z^3`; a coarse vertex `v` sits at the fine site `2v`, and the coarse edge from `v` along `e_ax` sits at the fine site `2v + e_ax`. The
+**KS sign** of the coarse bond `(v, v + e_a)` is `eta_1 = 1`, `eta_2(v) = (-1)^{v_1}`, `eta_3(v) = (-1)^{v_1 + v_2}`, the clause's phases read in coarse coordinates.
+The **plain sign** is `+1` on every bond.
+
+The **cell** is one `2x2x2` block of coarse vertices, eight modes, so the Bloch Hamiltonian of the KS hopping is an `8x8` matrix `H(q)`. The **encoding** is the
+Bravyi-Kitaev superfast encoding on the coarse lattice, code qubits on the coarse edges, direction order `-x < -y < -z < +x < +y < +z`.
+
+```text
+Gamma = (Y_1, Z_1 Y_2, Z_1 Z_2 Y_3),  Xi = (X_1, Z_1 X_2, Z_1 Z_2 X_3),  epsilon = Z_1 Z_2 Z_3,  T = diag(1,1,-1,-1),  B = (XX, XY, XZ)
+A_ij = X(edge (i,j)) * prod Z(edges ordered before it at i) * prod Z(edges ordered before it at j),  A_ji = -A_ij
+B_i  = product of the Z's on the edges incident to i,   S_f = the ordered product of the four A's around a coarse face f
+```
+
+`B_i = -1` marks the excitation; the **encoded hop** across `(i, j)` is `T_ji = (i/2) A_ji (B_j - B_i)`. A step is **legal** on a configuration when its source is
+occupied and its target empty. A **flux sector** is a choice of eigenvalue `+-1` for every `S_f` that is consistent with the `F2` relations among them.
+
+## Theorem 1 -- the kinetic form is a Clifford operator with an exact spectrum
+
+**Conclusion.** On the coarse lattice with the KS signs, with one `2x2x2` cell as unit cell:
+
+1. `Gamma_x, Gamma_y, Gamma_z, Xi_x, Xi_y, Xi_z` are six mutually anticommuting hermitian involutions -- a `Cl(6)` set -- and the Bloch block built from the hopping
+   rules equals `H(q) = sum_a [(1 + cos q_a) Xi_a + sin q_a Gamma_a]` identically in `q`.
+2. Hence `H(q)^2 = (6 + 2 sum_a cos q_a) I` and `tr H(q) = 0`, so `E(q) = +- sqrt(6 + 2 sum_a cos q_a)`, each eigenvalue fourfold.
+3. On the coarse torus `L = 4`, `64` modes, the spectrum is exactly `-2sqrt3` with multiplicity `4`, `-2sqrt2` with `12`, `-2` with `12`, `0` with `8`, and its
+   mirror. On `L = 6` (`216` modes) and `L = 8` (`512` modes) the real-space spectra reproduce the same Bloch prediction eigenvalue by eigenvalue, with `0` and `8`
+   zero modes.
+
+**Proof.** Item 1 is an exact matrix identity over `Z[i]` for the Clifford relations and a symbolic identity in the three momenta for the closed form, expanded term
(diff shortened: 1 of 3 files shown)


jonathonreilly:
Reading clarification (from the vacuum panel review of 2026-09-03 and an exact follow-up computation): the 'spin(2) x taste(4)' factorisation of the eight corner modes should be read as 8 = 2 (Weyl spinor components) x 2 (chiralities, the sign of T) x 2 (tastes). In the 2x2x1 magnetic cell the band function is H4^2 = (6 + 2 cos Q1 + 2 cos Q2 + 2 cos 2Q3) I with exactly two 4-fold nodes at (pi, pi, pi/2) and (pi, pi, 3pi/2), each one right- plus one left-handed Weyl; the 2x2x2 cell folds both onto the single 8-fold point at (pi, pi, pi), giving 2R + 2L: N_f = 2 four-component Dirac fields, vector-like, net chirality zero. All multiplicities, the 2 A1 + 2 T1 content and every number in this note stand; only the word 'taste' overcounts by including the two chiralities. A dedicated note (Lorentz invariance at the Dirac point and the taste census) is being landed with the exact statement.

JONbridgerapps:
### [P1] BUG / CodeRunnerReviewer — the advertised intertwiner normalization is never asserted

`scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:326-342`; affects `docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:149-150` and the machine claim scope at line 4.

The runner computes `scal = (U U^dag)[0,0]`, but `unit` checks only that `U U^dag = scal I` and `scal > 0`. The emitted B1 label nevertheless says `U U^dag = 16 I`, and B1 passes for any positive normalization (for example, multiplying the returned averaging intertwiner by 2 changes the claimed scalar by 4 while preserving the predicate). Thus the runner does not fail closed on a load-bearing expected value that the note promotes as exact.

Requested fix: assert the exact scalar `scal == 16` (preferably in exact Gaussian-integer arithmetic), add a mutation test that rescales or otherwise changes `U` so the normalization subcheck fails for that reason, regenerate the cache through `runner_cache.py`, and have this same reviewer session confirm the fixed runner/note pair.

### [P1] PROOF_OBLIGATION / ProofObligationReviewer — the torus existence theorem outruns the finite census

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:224-237,311-315`; `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:961-974,1027-1090`.

Theorem 6 states that on an arbitrary `L1 x L2 x L3` coarse torus the all-minus face sector exists iff every pairwise product `L_a L_b` is even. The supplied proof and runner only enumerate nine named tori (plus three open blocks). In F3, `want` is the claimed parity rule itself and the predicate merely compares it with `cons` on those nine fixtures; F4 independently solves the edge equations, but still only on those fixtures. No argument promotes the finite sample to all integer triples. This leaves the universal theorem's strongest leaf open; relative to that universal target the result is `CONDITIONAL`/incomplete, while the exact finite-census result is closed.

Requested fix: either narrow every surface (frontmatter, theorem, executable block, runner banner/output, and review record) to the exact nine named tori, or add a genuine all-`L_i >= 3` cohomological/combinatorial proof with its boundary assumptions and an independent checker that does not seed its expected answer from the claimed parity rule. Regenerate the cache after the chosen repair.

### [P1] NO_GO_OVERCLAIM / NoGoDisciplineReviewer — negative boundaries ship without the mandatory N1–N8 packet or N5 certificate

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:164-177,203-215,217-240,286-291,311-325`; `logs/runner-cache/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.txt:1-46`.

This bounded theorem contains several semantic negative claims: no site gauge carries plain signs to KS, the chirality image is never the taste pseudoscalar, the all-minus sector does not exist outside the stated torus parity condition, and the encoding has no odd-parity states. The candidate contains no landed `## No-Go Discipline Gate` (N1–N8) and its primary cached stdout contains none of the required substantive `per_element:`, `per_site:`, `per_mode:`, `per_block:`, and `lattice_wide:` N5 lines. Applying the required N1–N8 walk gives `FAIL`: N1 has no five normalized attack routes; N5 does not reconcile the sampled torus census with the generic rhetoric; N7 has an obvious live steelman (a general cohomology/relations proof or counterexample outside the nine fixtures); N8 merely points to a different labeling residual and performs no cross-cycle echo analysis. The remaining checklist items are absent rather than passed.

Requested fix: if these negative statements remain theorem content, land a complete claim-specific N1–N8 section/sidecar and the five-line honest N5 execution certificate, with the rhetoric narrowed to the resolutions actually established. Otherwise split/withdraw the negative surfaces and retain only the positive finite constructions. Regenerate the cache and rerun changed-evidence readiness.

### [P1] SEMANTIC_BRIDGE / PhysicsClaimReviewer + ImportSupportReviewer + AuditCompatibility — the zero-dependency packet imports two unlanded/unaudited physical identifications

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:20-26,63-75,84-98,252-265,329-337`; manifest entry `docs/audit/data/citation_graph_manifest.json` for this claim (`out_degree: 0`).

The finite algebra can be made self-contained by declaring a KS sign field and a BKSF edge-qubit encoding. The stronger repo-facing claim is not self-contained: calling that sign field “the framework's landed staggered kinetic form” quotes `STAGGERED_DIRAC_REALIZATION_GATE_NOTE_2026-05-03.md`, whose current ledger row at the frozen base is `audit_status/effective_status = unaudited`; calling the encoding “the emergent fermion's own gauge structure” and using it to discharge the parent spin-statistics residual relies on `EMERGENT_3D_FERMION_ONE_QUBIT_PER_SITE_SUPERLATTICE_ROLE_PATTERN_EXISTENCE_BOUNDED_THEOREM_NOTE_2026-09-02.md`, which is explicitly identified as open PR #7834 and is absent from both the base and this delta. This candidate does not reproduce that packet's fine-lattice role-pattern law, its formation/selection construction, or its exchange-statistics proof. Nevertheless it marks both sources as plain-text “pointer only,” declares `upstream_dependencies: []`, and acknowledges a zero-edge graph node. That hides precisely the physical carrier/lineage bridges needed by the title and corollary and prevents the restricted audit packet from seeing their non-retained state.

Requested fix: split the self-contained finite BKSF/KS algebra from the physical framework corollary. Either remove/rename all “framework's emergent fermion,” residual-discharge, “same law,” and “already on the table” claims so the note is only about declared abstract objects, or land/review the required parent content in the same candidate and add real markdown dependency edges to every load-bearing authority (accepting the resulting pending-chain boundary). Regenerate the citation graph/manifest and rerun the audit-readiness gate after the dependency surface is honest.

### [P1] IMPORTED_VALUE / CodeRunnerReviewer + PhysicsClaimReviewer — the torus KS witness silently fixes three Wilson-loop sectors

`scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:1094-1124,1207-1237`; affects `docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:228-240,252-260,303-305,320-322`.

On a periodic three-torus, the plaquette stabilizer eigenvalues do not determine a unique local-gauge class: three independent noncontractible Wilson holonomies remain. The runner itself exposes this by appending three loop generators in `induced_eta` and assigning them `vals += list(wilson)`, but the public call uses the undisclosed default `wilson=(1,1,1)`. F5 then advertises `eta'` as read from “the all-(-1) sector” with no KS input and the corollary says KS and plain differ by one face quantity “and by nothing else.” The witness on the `4x4x4` torus actually follows only after choosing the `(+,+,+)` global Wilson sector—the one matching periodic KS. This is a hidden boundary-condition/sector selector, not a consequence of the face sector; the open-block statements do not have this obstruction.

Requested fix: expose all three Wilson-loop eigenvalues as explicit inputs in the theorem, frontmatter, proof boundary, runner output, and import inventory; state the torus result only for the Wilson sector matching KS. Remove “and by nothing else” on periodic topology (or replace it with the correct face-plus-global-holonomy classification), and add mutations of each Wilson value showing that F5 rejects or reports the nonmatching gauge class for the intended reason.

### [P1] SEMANTIC_BRIDGE / PhysicsClaimReviewer — the checked Hermitian adjacency is not the quoted staggered Dirac operator on all named tori

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:84-90,114-115,126-140,252-264,297-305`; `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:120-140,238-254,1197-1204`.

The imported kinetic clause is the oriented, antisymmetric difference `D = (1/2) sum eta (bar_chi_{x+mu} chi_x - bar_chi_x chi_{x+mu})`. The runner instead constructs a real symmetric adjacency, adding the same `eta` to both matrix orientations, and calls its spectrum and Bloch block the landed kinetic form. The two can be related on an open region by first Hermitianizing to `2 i D` and then applying the site phase `g(v)=i^(v_1+v_2+v_3)`, but this is a local `U(1)`/`Z4` transformation, not the local `Z2` relabelling the note declares to be the exact identification. It is also not periodic unless every relevant period is divisible by four. An independent construction from the quoted formula gives, under periodic boundary conditions, eight zero modes for `L=6`; the candidate's symmetric matrix gives zero, with a maximum sorted-spectrum discrepancy of `sqrt(3)`. For `L=4,8` the phase happens to be periodic and the spectra agree. Thus the `L=6` result is effectively a different boundary-holonomy convention, and the claimed exact bridge to the quoted operator is absent.

Requested fix: define explicitly whether the target is the sign cocycle only, the Euclidean staggered difference `D`, or a Hermitian hopping Hamiltonian. If the latter is meant as a representative of `D`, land and check the `2 i D` plus `g(v)=i^sum(v)` bridge, classify its induced boundary holonomies for every named torus, and replace the `Z2`-only equivalence rhetoric. Recompute the `L=6` claim under the stated boundary convention (or remove it), then regenerate the runner cache.

### [P1] BUG / CodeRunnerReviewer — A4's six-decimal comparison does not enforce its advertised `1e-9` tolerance

`scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:257-282`; affects `docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:4,133-140,298`.

`spec_pairs` and `bloch_prediction` round every eigenvalue to six decimal places and compare formatted multiplicity tables. That is only a roughly `5e-7` bin test, despite A4 and the note promising eigenvalue-by-eigenvalue agreement at `1e-9`. Mutation evidence is decisive: adding `1e-7 I` to the entire `L=6` matrix leaves the runner at `PASS=23 FAIL=0`, even though every `L=6` eigenvalue is displaced by one hundred times the declared tolerance. The zero-mode count does not catch this mutation because the unmutated `L=6` matrix already has no zero modes.

Requested fix: compare sorted unrounded eigenvalue arrays directly with an explicit maximum absolute error `<= 1e-9` (while retaining multiplicity summaries only as diagnostics), print that maximum, add a diagonal-shift mutation that must fail A4, and regenerate the cache.

### [P1] SEMANTIC_BRIDGE / PhysicsClaimReviewer — the coarse sign field is redeclared, not the landed fine-lattice clause restricted to `2Z^3`

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:4,22-26,84-98,108-115,164-180,252-265,296-305,313-320`; `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:2-16,79-85`.

The landed clause defines nearest-neighbour fine-lattice phases as functions of the fine coordinate `x`: `eta_2(x)=(-1)^(x_1)` and `eta_3(x)=(-1)^(x_1+x_2)`. This candidate places a coarse vertex `v` at the fine site `x=2v`, but then defines its coarse sign using `v`, not `x`: `eta_2(v)=(-1)^(v_1)`, `eta_3(v)=(-1)^(v_1+v_2)`. Literal restriction of the landed functions to `x=2v` gives `eta_1=eta_2=eta_3=+1`; moreover, a coarse bond from `2v` to `2(v+e_a)` is not a nearest-neighbour bond of the quoted operator at all. The candidate's own mod-4 observation confirms the rescaling: its new signs depend on `2v mod 4`, whereas the quoted signs evaluated at those sites do not. Consequently the runner proves theorems about a newly declared KS-shaped sign law on an abstract rescaled cubic graph, not that the framework's landed fine-lattice kinetic form is the other flux sector of the construction on `2Z^3`.

Requested fix: either rename and scope the result throughout as a theorem about a newly declared KS-pattern sign cocycle on the coarse graph, removing every claim that it is the landed clause “read on” or restricted to `2Z^3`; or supply a real coarse-graining/embedding map from the fine nearest-neighbour operator to the coarse modes and derive its effective links, phases, normalization, and boundary data. Add a runner comparison against that derived map rather than changing coordinates inside the phase formula by declaration.

### [P2] REPO_GOVERNANCE / RepoGovernanceReviewer — the note denies the citation-manifest edit present in this delta

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:267-271,293-307`; `docs/audit/data/citation_graph_manifest.json:3,4037-4040`.

The “What does not move” section says that no citation manifest is created or edited, and the executable claim block reports zero registry entries created. This candidate increases the citation manifest's `node_count` and creates a node for this exact claim. Even if that generated graph row is intended as bookkeeping rather than a scientific premise/status mutation, the literal governance assertion is false and obscures what the PR changes.

Requested fix: state the precise distinction: no premise/status/axiom registry is changed, while the generated citation-graph manifest is updated with this source node (and, after dependency repair, its honest outgoing edges). Keep that wording consistent in the executable claim block and review record.

### [P2] LABELING / CodeRunnerReviewer + AuditCompatibility — several “exact” checks are tolerance-based floating-point witnesses

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:48-52,184-208,242-250`; `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:554-634,722-742,1271-1292`.

The note says the untagged structural statements are exact, and the runner labels D1-D3, E1, and G1 `[exact]`. D1/D2 rely on `np.allclose`; D3 rounds floating characters to six decimals; E1 uses `np.allclose` on the normalized floating intertwiner; and G1 constructs one random floating vector per syndrome, normalizes it, and accepts approximate eigenvector equations. NumPy's default relative tolerance is also left enabled in these calls. These are useful numerical witnesses, but they are not the advertised zero-tolerance integer/`Z[i]`/`F2` proof. In particular, G1 does not directly certify the stated one-dimensional eigenspace for each `B` configuration: it finds one vector in each and relies on unstated commutation/preservation and dimension arguments.

Requested fix: either relabel these leaves as numerical and state their actual `atol`/`rtol`, or implement the claimed exact predicates (unnormalized integer/rational representation matrices and characters for D, scaled Gaussian-integer identities for E, and explicit stabilizer commutation with the `B_i`, syndrome-subspace preservation, uniform subspace dimension, and exact rank/dimension counting for G1). Set `rtol=0` on all threshold checks and add perturbation mutations at each advertised boundary.

### [P1] BOUNDARY_CONDITION / PhysicsClaimReviewer — the symmetric sector criterion is not the periodicity condition of the fixed KS representative

`docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:4,110-112,217-237,303-305,329-332`; `scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:79-85,417-432,961-1075`.

The all-minus face equations on a torus have the axis-symmetric existence condition “at most one period odd.” The fixed representative declared here does not: `eta_2=(-1)^(v_1)` and `eta_3=(-1)^(v_1+v_2)` descend with periodic boundaries only when `L_1` and `L_2` are even (the third period may be odd). For example, the candidate declares `4x5x6` consistent, but direct evaluation of its own `eta_ks` across the `y` wrap gives `+1` rather than `-1` holonomy on the boundary `yz` plaquettes. C1 avoids the issue by testing only `L=4,6,8`; F3 then calls the broader sector criterion “the KS periodicity condition,” and the review record says the two conditions coincide. They coincide only after an unstated permutation that moves the sole odd period to the third coordinate, which changes the representative/axis convention.

Requested fix: distinguish the gauge-invariant all-minus-sector existence criterion from the descent condition for the explicitly ordered KS formula. For every mixed torus used in the claim, either require `L_1,L_2` even or declare and apply an axis permutation placing the odd period last; then test all wrap plaquettes and the corresponding noncontractible holonomies for that representative.

### [P1] BUG / CodeRunnerReviewer — D4 checks equal cocycle counts, not the claimed same cocycle

`scripts/emergent_fermion_pi_flux_sector_staggered_kinetic_form_check_2026_09_02.py:643-717`; affects `docs/EMERGENT_FERMION_PI_FLUX_SECTOR_IS_THE_STAGGERED_KINETIC_FORM_BOUNDED_THEOREM_NOTE_2026-09-02.md:4,191-198,301,324-325`.

D4 constructs the spin and taste cocycle dictionaries `ccs` and `cct`, but its predicate only checks `sum(ccs.values()) == sum(cct.values())`, plus nontriviality of each. Equal numbers of negative pairs do not mean two cocycles are the same entrywise. A fail-closed mutation demonstrates the gap: multiplying just `SPIN[1]` by `-1` changes its cocycle by a nonzero coboundary while preserving all absolute spin characters, non-coboundary class, and the count of 208 negative pairs; the unchanged runner still reports `PASS=23 FAIL=0` and says the factors carry “one cocycle.” The unmutated dictionaries happen to be equal, but that advertised fact is outside the acceptance predicate.

Requested fix: assert `ccs == cct` explicitly (and, if only equality of cohomology classes is intended, state and test that weaker invariant instead). Add the single-representative sign mutation so the same-cocycle assertion fails for the intended reason, then regenerate the cache.

## Structured lens and check summary

- **Candidate boundary/freshness:** reviewed only `68b6ad72005688218bd40343af5e25c0880f0da1...7c747b392dc56f4402c33aa4c80cb24c9d233e23` (the four listed files). `HEAD` and all four candidate files remained byte-identical to the committed candidate throughout this review. The installed review-loop contract matches the repository's latest copy; the required authority, primitive, audit, proof-search, and no-go governance documents were read before judgment.
- **CodeRunnerReviewer: FAIL.** The unmodified runner completes at `PASS=23 FAIL=0` in about one second, and the committed cache's SHA-256 matches the runner. Independent checks recover scalar `16`, the expected Clifford identities, pairwise face-stabilizer commutation on all sampled complexes, a rank-128 projected basis on the open cube, and the unmutated spin/taste cocycles are entrywise equal. However, three advertised results are fail-open: `U -> 2U` still passes B1, a `+1e-7 I` spectral shift still passes the stated `1e-9` A4 test, and a sign flip of one spin representative changes the cocycle while D4 still passes. Control mutations of the A2 coefficient, F3 parity predicate, a Wilson-loop value, and the G2 spectrum do fail their intended checks.
- **PhysicsClaimReviewer: FAIL.** The central physical identification does not close: the candidate redeclares the fine-coordinate KS law in coarse coordinates; checks a symmetric adjacency rather than the quoted antisymmetric staggered difference; suppresses the resulting `U(1)` phase/boundary bridge; fixes the torus Wilson sector silently; and conflates the symmetric all-minus-sector condition with periodicity of an axis-ordered KS representative. Independently, periodic `L=6` gives 0 zero modes for the candidate adjacency but 8 for `2 i D`, with maximum sorted-spectrum difference `sqrt(3)`.
- **ProofObligationReviewer: CONDITIONAL / FAIL for the authored theorem.** An independent GF(2) implementation agrees with the parity formula on all 64 tori with side lengths 3 through 6, supporting the conjectured/general result, but neither finite enumeration is a proof of the universal `L_1 x L_2 x L_3` iff statement. The exact finite-census leaf is supported; the universal leaf remains open.
- **ImportSupportReviewer: FAIL.** The abstract finite algebra is substantially self-contained, but the physical “framework's emergent fermion,” “same law,” and residual-discharge conclusions import an unaudited landed kinetic source and absent/open PR #7834 while declaring zero dependencies. The source packet does not reproduce the parent construction needed for those identifications.
- **NoGoDisciplineReviewer: FAIL.** The negative assertions trigger the gate, but no claim-specific N1-N8 packet or five-line N5 execution certificate is present. The missing general-proof route and scope mismatch keep N5/N7 open.
- **LabelingReviewer: FAIL.** Several tolerance/random-vector checks are presented as exact, and boundary-sensitive or newly declared objects are labelled as the landed kinetic form. The proposed-retained framing is therefore stronger than the demonstrated surface.
- **RepoGovernanceReviewer: FAIL.** Vocabulary lint, link resolution, compile, and diff hygiene pass, but the note falsely says no citation manifest was edited even though this delta adds its graph node.
- **AuditCompatibility: mechanically ready, scientifically blocked.** In a disposable exact-candidate clone, the full pipeline exits 0; strict lint has no errors; after ledger materialization, `check_changed_audit_evidence` checks this claim with zero failures. `py_compile`, all three `git diff --check` modes, changed-file vocabulary lint, cache pin/hash, and both relative links pass. Those are preflight receipts only. The hidden dependency edges, absent no-go packet, inaccurate exactness labels, and unresolved physics bridges make the packet unsuitable for an honest independent retained-grade audit in its current form. No audit verdict was applied.
- **Nature-grade adversarial disposition:** reject this candidate as authored. The finite sign/stabilizer algebra contains useful salvageable results, but the title, corollary, universal/boundary claims, and machine evidence materially outrun them.

## VERDICT: **FAIL**

## Focused same-session confirmation — candidate `ced96ef9273fd53941640810e07e3aad9d7ec750`

### [P1] REPO_GOVERNANCE / CandidateBoundary — the exact candidate drops a preparation-main governance section outside the declared four-file fix

`docs/repo/DEFERRED_DECISIONS.md:80-119` at preparation main `e3d9ff5e480d9d1e6bcc2b90849b18f0ce17dc4a`; candidate commit `ced96ef9273fd53941640810e07e3aad9d7ec750`, tree `8c5765b671c17599da713c15f2be213d14aeddd4`.

The candidate is not descended from the stated preparation main: `git merge-base e3d9ff5e... HEAD` is the old base `68b6ad72005688218bd40343af5e25c0880f0da1`. A direct tree comparison (`git diff e3d9ff5e... HEAD`) therefore has five changed files, not the declared four. The fifth delta deletes the complete parked-decision section “The C-add record-composition fork (+ POS/LOC admission)” from `docs/repo/DEFERRED_DECISIONS.md` (40 lines). This is a real resulting-tree regression, even though the merge-base PR diff hides it. It lies outside the reviewed salvage and prevents the exact candidate tree from being certified against the preparation base.

Requested fix: rebuild/cherry-pick the bounded salvage onto exact preparation main `e3d9ff5e480d9d1e6bcc2b90849b18f0ce17dc4a` (or otherwise restore that file byte-for-byte from the preparation base), then freeze and provide the resulting exact commit and tree for this same reviewer session. The direct base-to-candidate diff must contain only the four declared fixed files.

## Focused same-session confirmation — component `1cfc52a2b620191ca86686d3688b52dc32884a71`

The current-main rebuild resolves the preceding candidate-boundary finding: component `1cfc52a2b620191ca86686d3688b52dc32884a71` has tree `ad8477a35e794f6d743a4837524f1a307ebf629c`, parent `4c2a0fcfff84ce8cab84dce4d941eb9f8a58e5d5`, and an exact four-file component delta. The three scientific blobs are byte-identical to the preceding salvage candidate, so their focused scientific review carries forward.

### [P1] MUTATION_COVERAGE / CodeRunnerReviewer — the advertised face-incidence mutation is a tautology and a real incidence-bit corruption survives the full runner

`scripts/bravyi_kitaev_superfast_pi_flux_sign_class_finite_cubic_graph_check_2026_09_02.py:280-313,637-641,664-677`; `docs/BRAVYI_KITAEV_SUPERFAST_PI_FLUX_SIGN_CLASS_FINITE_CUBIC_GRAPH_BOUNDED_THEOREM_NOTE_2026-09-02.md:130-145`; `logs/runner-cache/bravyi_kitaev_superfast_pi_flux_sign_class_finite_cubic_graph_check_2026_09_02.txt:17`.

The purported incidence mutation never calls or alters `face_incidence`, never changes the rows sent to `solve_f2`, and never re-runs check C. It constructs a new `Pauli` with one `z` bit toggled and asks whether that structurally different object is unequal to the original face operator. That predicate is true by construction and does not test the advertised face-incidence validation. An in-memory mutation that instead toggles one actual bit in the first `face_incidence` result leaves every check—including C and I—at `PASS`, and the complete mutated runner exits `0` with `PASS=9 FAIL=0`. This directly contradicts the note/cache assertion that a corrupted face-incidence bit is rejected. In the same block, the advertised `1e-7 I` one-particle mutation is also not what is executed: lines 658-660 shift `encoded_spectrum` and compare it with itself, rather than perturbing either one-particle matrix/spectrum in check F.

Requested fix: route mutations through the same acceptance helpers used by the primary claims. For incidence, validate the geometric four-edge boundary/rank or relation invariant and inject a real incidence-mask bit flip that makes that validation fail. For the spectral falsifier, add `1e-7 I` to exactly one one-particle comparison target and call the same `rtol=0, atol=1e-9` predicate used by F. Require both mutated executions to return false, regenerate the cache, and provide the exact fixed component to this reviewer session.

### [P1] PROOF_OBLIGATION / ImportSupportReviewer — the note never defines the encoded hopping or the “legal hop operator” carrying its central spectrum/transport claim

`docs/BRAVYI_KITAEV_SUPERFAST_PI_FLUX_SIGN_CLASS_FINITE_CUBIC_GRAPH_BOUNDED_THEOREM_NOTE_2026-09-02.md:44-63,67-74,121-128`; hidden implementations at `scripts/bravyi_kitaev_superfast_pi_flux_sign_class_finite_cubic_graph_check_2026_09_02.py:181-184,604-613`.

The declared-algebra section defines `A_ij`, `B_i`, and `S_f`, but T1 then asserts a product of four “legal hop operators” and T4 asserts the spectrum of “the encoded hopping” without ever defining either operator. The runner silently chooses a single legal directed hop as `i A_(target,source)` and the many-body operator as `H_enc=(i/2) sum_(i,j) A_ij (B_i-B_j)`. The sign, orientation, factor `1/2`, and edge-counting convention are load-bearing: they fix both the transport phase and the claimed spectrum. With `upstream_dependencies: []`, those choices cannot be imported merely through the phrase “BKSF-style.” A reader cannot state, reproduce, or falsify the theorem from the note's declared objects.

Requested fix: state the exact oriented-edge convention and formula for `H_enc`, define occupancy through the `B_i` eigenvalue, derive its source-occupied/target-empty restriction `i A_(target,source)`, and define the even-parity free-Fock comparison target (including normalization). Make the runner notation match those definitions verbatim.

### [P2] LABELING / CodeRunnerReviewer — the exactness inventory omits floating-point sector-basis acceptance

`docs/BRAVYI_KITAEV_SUPERFAST_PI_FLUX_SIGN_CLASS_FINITE_CUBIC_GRAPH_BOUNDED_THEOREM_NOTE_2026-09-02.md:121-133`; `scripts/bravyi_kitaev_superfast_pi_flux_sign_class_finite_cubic_graph_check_2026_09_02.py:455-489,587-602,627-629`.

The note calls the stabilizer-projected basis “exact” and says only the two eigenvalue comparisons use floating point. In fact, `deterministic_sector_basis` builds complex NumPy vectors, accepts a seed using `np.linalg.norm(vector) > 1e-12`, normalizes numerically, and its non-`None`/shape result is part of check G, which is labelled `[exact]`. The exact face/star ranks and dimensions are independently sound, but this extra numerical acceptance is not exact `F2/Z4` arithmetic. Check H is honestly numerical; check G and the evidence inventory should not blur that boundary.

Requested fix: keep G purely on exact rank/dimension/group-consistency predicates and move the floating basis existence/construction into the numerical H evidence, or implement the basis/orbit nonvanishing in exact arithmetic. Update the note and cache labels accordingly.

## Focused same-session confirmation — final component `a1a64d36bc856c3ae330786972d3f2a02b678c52`

The exact final component resolves every focused finding above. It has tree `730c7e15291153b2572424106be018d04657a183`, parent `4c2a0fcfff84ce8cab84dce4d941eb9f8a58e5d5`, and exactly the three bounded BKSF artifacts plus the regenerated citation manifest in its component delta.

- The preparation-main regression is absent: the component is rebuilt on the reviewed train and does not alter `docs/repo/DEFERRED_DECISIONS.md`.
- `relation_inventory` now checks that every exact stabilizer relation is also a relation of the supplied geometric face-incidence masks, check C consumes that predicate, and I passes a genuinely bit-flipped incidence-mask list back through that same logic. An independent in-memory first-call mutation of the real `face_incidence` function now makes C fail and the runner exit `1` (`PASS=8 FAIL=1`).
- The one-particle falsifier now shifts `one_particle_spectrum` by `1e-7` and applies the stated `rtol=0, atol=1e-9` comparison; it no longer mutates the encoded many-body spectrum under the one-particle label.
- The note explicitly declares `h_(j<-i)=(i/2)A_ij(B_i-B_j)`, derives its legal-hop restriction `i A_ji`, and declares the positively oriented-edge sum defining `H_enc`. These definitions match the runner's operator order, sign, and normalization.
- Check G now contains only integer rank/dimension/configuration-count predicates. The projected basis, its `1e-12` nonzero threshold, numerical normalization, Hermiticity check, and `rtol=0, atol=1e-11` spectrum comparison are assigned to the explicitly numerical H surface in both note and cache.

## Structured lens and check summary

- **Candidate boundary/freshness: PASS.** Frozen component HEAD `a1a64d36bc856c3ae330786972d3f2a02b678c52`, tree `730c7e15291153b2572424106be018d04657a183`, parent `4c2a0fcfff84ce8cab84dce4d941eb9f8a58e5d5`; clean worktree at final freeze; exact four-file component delta. Review remained confined to the salvaged note, runner, cache, manifest, and necessary support sources.
- **CodeRunnerReviewer: PASS.** Direct execution gives `PASS=9 FAIL=0`; stdout matches the cache byte-for-byte and runner SHA-256 `f982e3249ef7056816fe215d5d9459eaedbd858a94842fcee2a8bb6c1ea5af1d` matches the cache. `py_compile`, component `git diff --check`, and both note links pass. All advertised mutations are now directed at the named target; a separately injected real incidence corruption fails closed.
- **Independent mathematical cross-check: PASS.** A separate canonical-Pauli-string implementation reproduces `A_ji=-A_ij`, Hermitian involutory face operators, all face/B commutators, transport phases, and relation phases on all four fixtures. It obtains `(V,E,F; face rank)` values `(8,12,6;5)`, `(27,54,36;28)`, `(64,144,108;81)`, and torus `(64,192,192;126)`, with torus cycle rank `129` completed by the three Wilson generators. Every declared staggered face has holonomy `-1`; the periodic Wilson triple is `(+,+,+)`. On the cube the independent face-plus-star rank is `12` with no minus-scalar relation, and the exact one-particle characteristic polynomial is `(lambda^2-3)^4`.
- **PhysicsClaimReviewer: PASS at the authored boundary.** The candidate is now only an abstract finite-graph theorem. It makes no physical-lattice, landed-law, emergent-fermion, continuum, spin/taste, Clifford, or role-pattern identification. The Hermitian adjacency and encoded hopping are explicitly declared rather than conflated with an imported physical Dirac operator.
- **ProofObligationReviewer: PASS.** Claims are limited to open `2x2x2`, `3x3x3`, `4x4x4`, and periodic `4x4x4` with explicit `(+,+,+)` Wilson data. Exact algebra/rank/gauge leaves and numerical spectrum leaves are separated. No finite census is promoted to a universal torus theorem.
- **ImportSupportReviewer: PASS.** The bounded objects, boundary data, sign field, legal-hop operator, encoded Hamiltonian, and spectral target are declared locally; `upstream_dependencies: []` is honest for the retained abstract claim.
- **NoGoDisciplineReviewer: not triggered.** The note contains scope exclusions and an explicit refusal of universal/physical identification, not a semantic impossibility theorem or residual-closing negative result.
- **LabelingReviewer: PASS.** `bounded_theorem`, `proposed_retained`, and `Audit: unset` match the evidence. Exact `F2/Z4` claims and floating basis/spectral corroboration are now labelled separately with their tolerances.
- **RepoGovernanceReviewer: PASS for the component.** The rebuilt delta is exactly four files, changes no governance/authority surface, and the regenerated manifest adds one zero-dependency node with the corresponding node-count increment. The old overclaimed physical note/runner/cache are absent from the resulting component delta.
- **AuditCompatibility: PASS for review readiness, without applying a verdict.** The cache is pinned and reproducible, runner timeout metadata is present, paths resolve, and the note explicitly keeps audit status unset. The coordinator's combined-train full-pipeline/strict-lint/changed-evidence rerun remains a mechanical post-content preflight; it is not treated here as scientific validation. No audit verdict was applied.
- **Nature-grade adversarial disposition:** accept the narrowly stated finite theorem packet. The original physical/universal/Clifford claims were removed rather than cosmetically weakened, and the retained formal core is reproducible, falsifiable, and bounded by the fixtures actually checked.

## VERDICT: **PASS**

PR #7883, opened 2026-09-03
feat(readout): record statistics of the half-filled sea are determinantal

jonathonreilly:
## Summary

Conditional bounded theorem note + exact runner + cache: **on the half-filled staggered sea, the record statistics of the emergent fermion form a determinantal point process whose kernel is the sea's own one-particle projector `P`.** The finished set of records at `S` has odds `det P_S` (exact over `Q(sqrt3)` and `Q(sqrt2)` on the two open clusters, and verified against the Jordan-Wigner ground state pattern by pattern); the odds at a site given any records elsewhere are the **Schur complement** of `P` (0 mismatches against explicit Lueders conditioning); PR #7858's twelve forbidden corner pairs are determinantal zeros in closed form, `det P_uv = 0`; a record at separation `d` shifts a site's odds by exactly `2 P_vu^2`, with `|P_vu|` falling as the inverse cube of the distance and vanishing exactly on separations of the wrong parity; the six neighbours enter additively; and no finite record set forces a value unless it covers the whole support of the kernel's row. Admissibility's "distribution determined by the neighbourhood conditions" is, on this vacuum, an explicit kernel formula whose reach is not capped and arrives through the lattice.

- `docs/RECORD_STATISTICS_OF_THE_HALF_FILLED_SEA_ARE_DETERMINANTAL_CONDITIONAL_BOUNDED_THEOREM_NOTE_2026-09-03.md` (324 lines)
- `scripts/record_statistics_of_the_half_filled_sea_are_determinantal_check_2026_09_03.py` (`TOTAL: PASS=28 FAIL=0`, 8 s)
- `logs/runner-cache/record_statistics_of_the_half_filled_sea_are_determinantal_check_2026_09_03.txt`

## Theorems

- **T1** the finished set's odds are determinantal (cube and block; exact quadratic-field arithmetic; value multiset; the zeros; no pair minor vanishes at half filling).
- **T2** the empty vacuum's forbidden pairs (PR #7858) are determinantal zeros, `det P_uv = P_uu P_vv - P_uv^2 = 0`, recovering the 384 cancellation zeros from `P` alone; the manifold projector has none.
- **T3** conditioning is the Schur complement; the joint law `P(A occ, B empty) = (-1)^|B| det (P - 1_B)_{A u B}`; edge records push forward through constant fibres; order independence.
[...]

PR #7876, opened 2026-09-03
[review-loop salvaged and landed] finite cube conditioning and equality kernel

jonathonreilly:
## Summary

Bounded theorem note + exact runner + cache on what a **tick** does when record formation is interleaved with the pre-record hopping dynamics, on the emergent-fermion cube's `N = 2` sector, for a fully **stipulated** tick model (geometric formation with probability `p`, Lueders conditioning, post-record Hamiltonian `H_R` = the hop terms on unrecorded sites; a Model B alternative named). The result: Lueders conditioning alone is not an update clause. With any evolution between records the finished set's odds depend on the schedule, the selection-rule zeros of the stipulated ground state are vacated at first order in the gap with exact coefficient `1/8`, and **no pre-record state is invariant under all the post-record Hamiltonians** (joint kernel of dimension 0, closed exactly). The count of records still supplies a clock.

- `docs/RECORD_TICKS_ADMIT_NO_INVARIANT_PRE_RECORD_STATE_BOUNDED_THEOREM_NOTE_2026-09-03.md` (319 lines)
- `scripts/record_ticks_admit_no_invariant_pre_record_state_check_2026_09_03.py` (`TOTAL: PASS=24 FAIL=0`, 13 s; exact except deterministic unsampled propagator evaluations tagged numerical and one fenced seeded witness)
- `logs/runner-cache/record_ticks_admit_no_invariant_pre_record_state_check_2026_09_03.txt`

## Theorems (cube, `N = 2` record sector, 896-dim)

- **T1** the sector, the exact diagonal gauge onto Jordan-Wigner, and the ground state's 384 cancellation zeros (the corner pairs sharing an x-face).
- **T2** if every record forms before any evolution, the finished set reproduces the pre-record Born diagonal exactly (PR #7858's order independence recovered as the `tau = 0` boundary).
- **T3** one record already vacates the zero set at first order: `var(H_R) = 1/4` or `3/8` after any single record, never 0; forbidden mass `m(t) = t^2/8 + O(t^4)` exactly.
- **T4** the joint kernel of `{H_R - H_R'}` over all single-site record sets, with `{H_R - H}`, has dimension 0: no nonzero pre-record state is invariant under all post-record Hamiltonians (closed exactly by union-find over 69888 unit-ratio conditions with a `Z4` phase).
[...]

PR #7895, opened 2026-09-03
feat(readout): a relaxation tick is well-posed and loses the sea statistics

jonathonreilly:
## Summary

Bounded theorem note + exact runner + cache turning the vacuum panel's candidate tick wording ("Between records, the lattice settles into its lowest-energy arrangement") into a fully stipulated model `M_R` (Lueders formation in a declared order, then replacement by the ground state of `H` restricted to the records) and computing what it does on the emergent cube against the reference any tick has to reproduce, the half-filled sea's own record statistics. **`M_R` is well-posed and memoryless (the restricted Hamiltonian is exactly PR #7876's `H_R`, so the state is a function of the records alone), but it is not the sea's registration:** every one of the 4096 patterns becomes admissible (none of the sea's 2112 zeros survives: 1856 charge zeros and the 256 cancellation zeros on the eight closed corner stars), the odds depend on the order in which the sites record (`TV(identity, reverse) = 1/3` exactly; maximal pairwise `TV = 0.727` over 32 declared orders), and the readable charge is no longer sharp (`var Q = 13/12`), though `<Q> = 0` and `<N> = 4` survive on average. Bonus exact facts: the sea's own record odds are flat `2^-k` with nothing forced for `k <= 3`; at `p = 1` (all sites at once) the distribution is the sea's Born distribution; conditioning alone costs `sqrt 3 / 3` of energy per record (`<H_R> = -(12 - k)/sqrt 3`); and `Z_e` is the unique one-site operator commuting with every corner parity, so the record basis is already fixed and the open content of a tick is the formation rule (rate and odds), not the basis.

- `docs/A_RELAXATION_TICK_IS_WELL_POSED_AND_LOSES_THE_SEAS_RECORD_STATISTICS_BOUNDED_THEOREM_NOTE_2026-09-03.md` (328 lines)
- `scripts/relaxation_tick_well_posed_loses_sea_record_statistics_check_2026_09_03.py` (`TOTAL: PASS=26 FAIL=0`, 45 s; no seeds, every order a literal; trees enumerated whole)
- `logs/runner-cache/relaxation_tick_well_posed_loses_sea_record_statistics_check_2026_09_03.txt`

## Theorems

- **T1** the restriction identity `P_S H P_S = sum_{e not in R} T_e` (every hop flips exactly one edge qubit): `M_R` is memoryless.
- **T2** the sea's flat odds and its zeros (support `1984 = 62 x 32`; `2112` zeros; the eight closed corner stars).
- **T3** under `M_R` every pattern becomes admissible; `TV(M_R, sea) = 1283/2880`; PR #7876's Model A at `tau = 0.5` is closer.
- **T4** order dependence (`1/3` exactly; `0.727` maximal over the declared set, a lower bound on the spread over all `12!` orders).
- **T5** half filling survives on average, the charge does not stay sharp; the `N`-superselected variant restores it but still loses the 256 cancellation zeros (executor correction kept: the relaxed state is not `N`-sharp at the 1456 degenerate nodes, at ticks 5, 6, 8, 9, 11).
- **T6** relaxation is not the sea held to the records (`F = 0.9626` at every one-record block; energy lowered at every block, peak drop at `k = 9`).
- **T7** the pointer-basis table.

## Boundary

- One cluster, one sector, one filling; the tick model and its tie-break stipulated and derived from nothing; an obstruction for one stipulated model, not a no-go; the panel's sentence is quoted as a candidate wording and not called wrong; nothing here says what the framework's tick is.

## Context

- Parent: PR #7876 (record ticks admit no invariant pre-record state; its `tau`/`H_R` interface); pointers to PR #7883 (the sea's determinantal record statistics), PR #7842 (the selection-rule zeros), PR #7858, PR #7879.
- Part of the owner-authorised 24-hour matter-to-TOE campaign (2026-09-02/03); the vacuum panel's tick question.

### DIFF
--- a/docs/A_RELAXATION_TICK_IS_WELL_POSED_AND_LOSES_THE_SEAS_RECORD_STATISTICS_BOUNDED_THEOREM_NOTE_2026-09-03.md
+++ b/docs/A_RELAXATION_TICK_IS_WELL_POSED_AND_LOSES_THE_SEAS_RECORD_STATISTICS_BOUNDED_THEOREM_NOTE_2026-09-03.md
@@ -1,0 +1,328 @@
+---
+claim_id: relaxation_tick_well_posed_loses_sea_record_statistics
+claim_type: bounded_theorem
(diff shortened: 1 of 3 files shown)


jonathonreilly:
**Narrowing from PR #7904, reported for this note's reader.** The emergent sea's cancellation zeros follow an exact determinant criterion on the sea's projector (`det P_T = 0` iff the empty band carries a state supported inside the corner set `T`); the closed-corner-star description is the flat-band special case (`h^2 = cI`). One cube taller (the `2x2x3` slab) the zeros are 120 corner patterns, most not stars, and on the `6^3` and `8^3` tori at PR #7883's own optimal twists no closed star is singular and no exact zero is found in the scanned range. The computations here stand on their clusters; the further reading that the cube's zeros register a known three-dimensional selection rule is withdrawn to the flat-band clusters it was computed on.

PR #7900, opened 2026-09-03
feat(readout): joint formation on a corner's record set keeps the sea's zeros

jonathonreilly:
## Summary

Bounded theorem note + runner + cache extending PR #7895 from site-wise to neighbourhood-wise record formation, testing the owner's statement (2026-09-03) that "the whole neighbourhood has to move together because it is its own shared condition", read at the level of formation since records are permanent. **The unit of record formation matters, and the unit that fits is a corner's own record set.** With the unitary tick between events (PR #7876's Model A), joint formation on the three edges at one corner keeps every one of the sea's 256 cancellation zeros (3 of 4 declared orders; 4 of 40 in the lexicographic window), and schedules of **disjoint** corner sets reproduce the sea's record statistics **exactly** (`TV ~ 1e-15`) at every `tau` tested. The reason is exact and about shape: after a corner set forms jointly, the conditioned sea is an eigenvector of what remains of the law (8/8 outcomes; 448/448 for the closed corner star), so the evolution between events cannot disturb it; after a single edge (0/2) or a face's four edges (0/16) it is not, and the face unit is worse than the corner unit on every measure. Under relaxation between events (PR #7895's `M_R`) nothing short of the whole cluster keeps more than 64 of the 256 zeros, so the panel's candidate and the corner unit do not combine. Order independence short of the whole cluster is not achieved by any unit; the mechanism for the non-disjoint corner schedules that keep the zeros is stated as open. A candidate formation sentence, offered as a candidate wording and not as axiom text: *"Records form together on a corner's record set; between formations the law runs on."*

- `docs/JOINT_FORMATION_ON_A_CORNERS_RECORD_SET_KEEPS_THE_SEAS_ZEROS_UNDER_THE_UNITARY_TICK_BOUNDED_THEOREM_NOTE_2026-09-03.md` (328 lines)
- `scripts/joint_formation_corner_record_set_keeps_sea_zeros_check_2026_09_03.py` (`TOTAL: PASS=24 FAIL=0`, 67 s; no seeds; 16 declared orders plus a 40-permutation sweep, all literals)
- `logs/runner-cache/joint_formation_corner_record_set_keeps_sea_zeros_check_2026_09_03.txt`

## Theorems

- **T1** the control: sequential Born equals the sea and is order-independent for every unit; U1 reproduces PR #7895.
- **T2** the unitary tick with corner units keeps all 256 cancellation zeros; disjoint-corner schedules reproduce the sea exactly at `tau = 0.1, 0.5, 1.234567, 2.0`.
- **T3** the eigenvector mechanism (corner 8/8, closed star 448/448; edge 0/2, face 0/16).
- **T4** relaxation keeps at most 64 zeros below the whole cluster (the first star's coset and its antipode's); `P(Q = 0)` at most `0.841`; relaxation is the identity at U2 and worse at U4.
- **T5** order dependence shrinks with unit size but never vanishes short of U5; event counts 12, 6, 4.5, 2.75, 1.

## Boundary

- One cluster, one sector, one filling; units, orders, rules, tie-break and `tau` values stipulated; the owner's statement quoted as the question, the panel's wording as a candidate; nothing derived from the axioms; no claim about what the framework's tick is. Executor refinements kept (`P(Q = 0)` at U4 is `0.667 .. 0.750`; last-digit TV differences from the real-skew propagator).

## Context

- Parents: PR #7895 (site-wise formation), PR #7876 (Model A), PR #7883 (the sea's Born statistics), PR #7842 (the zeros), PR #7891 and PR #7899 (the neighbourhood mechanism).
- Part of the owner-authorised 24-hour matter-to-TOE campaign (2026-09-02/03). A generality test on the `2x2x4` slab is running separately.

### DIFF
--- a/docs/JOINT_FORMATION_ON_A_CORNERS_RECORD_SET_KEEPS_THE_SEAS_ZEROS_UNDER_THE_UNITARY_TICK_BOUNDED_THEOREM_NOTE_2026-09-03.md
+++ b/docs/JOINT_FORMATION_ON_A_CORNERS_RECORD_SET_KEEPS_THE_SEAS_ZEROS_UNDER_THE_UNITARY_TICK_BOUNDED_THEOREM_NOTE_2026-09-03.md
@@ -1,0 +1,328 @@
+---
+claim_id: joint_formation_corner_record_set_keeps_sea_zeros_unitary_tick
+claim_type: bounded_theorem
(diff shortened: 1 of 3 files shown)


jonathonreilly:
**Narrowing from PR #7902, reported for this note's reader.** The eigenvector mechanism of Theorem 3 has an exact condition: the record layer factors out, and the conditioned sea is an eigenvector of the remaining law iff the corner's site vector lies in exactly two one-particle eigenmodes, one occupied and one empty. The cube satisfies it because pi flux makes both bands flat. One cube taller, on the `2x2x3` slab, it holds at the four degree-4 corners and fails at all eight degree-3 corners by about a tenth of the norm, and on the periodic slab at every corner. So the candidate sentence in this note's corollary item 5 is an exact property of flat-band clusters, not a principle of the lattice; everything stated here on the cube stands. PR #7902 also shows the two-corner marginal test is blind and only the full record distribution discriminates.

PR #8039, opened 2026-09-08
[physics-loop] plane formation law and induced diagonal interaction

JONbridgerapps:
The handed-off campaign left open whether its finite formation laws define a consistent law on the whole plane and which interaction that law represents. This extends the current main's corrected rectangle result to the plane and identifies its exact pair potential: axial −log K plus a chosen-diagonal +log(K²) term from the formation normalizers.

For every positive nonconstant orbit-weight triple, the two diagonal classes differ and neither has the original four-neighbor static conditional. Randomizing between these two classes retains the same conditional discrepancy; the equal mixture restores square-lattice spatial symmetry. Opposite-corner equality was already proved on main and is not claimed as new. No physical formation order, Gaussian instrument, quantum action or uniqueness among all Gibbs measures is selected.

Validation: independent proof and canonical-port review; 163 exact live checks of complete 2×3 laws/subrectangle marginals, six-value 3×3 conditionals and 96 trimming schedules. The schedules are not a full 3×4 probability census. Two actual normalizer mutations fail under optimized Python. A fresh isolated three-file -I -OO closure passes in 2.494s at 54.13MiB. Vocabulary and staged whitespace checks pass; citation topology adds one node and leaves every existing node unchanged.

Based on latest main, with no dependency on the stale handoff stack or open Gaussian block. Status is conditional-support. Original evidence, source-bound reviews and closure receipts are retained. Formal audit and the combined landing pipeline/strict lint/changed-evidence pass remain pending; no audit verdict is applied here.


JONbridgerapps:
Landed the independently reviewed source for unit 8039 on main at 7927f9e170566b351b2cf6f801cb7176f8b87226.

The plane-law interaction proof is preserved with the original all-label, mixed-boundary and nonconstant equal-weight controls. The final runner passes 281 checks.

Every original constituent path is accounted for in the [immutable disposition map](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/f0dc54ec165910386941bfb3920d1c718fb6ed15/backlog_evidence/review-drain-20260915/review-8039.json), with no deferred science; [findings and final confirmation](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/f0dc54ec165910386941bfb3920d1c718fb6ed15/backlog_evidence/review-drain-20260915/review-8039.md) give the exact scope. One integrated full pipeline, strict lint and changed-evidence pass covered all six science rows in this batch. [Combined landing receipt](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/f0dc54ec165910386941bfb3920d1c718fb6ed15/backlog_evidence/review-drain-20260915/train02-landing.json). Generated audit authority was stripped; no audit verdict was applied.

Frozen original head: 96827c8023a008b5616464fe5d66369b8fa4424f. Closing after current-main source verification and an unchanged-head check. The same-repository source branch is deleted with an exact SHA lease.

PR #8045, opened 2026-09-08
Control local native ring dynamics at its natural timescale

JONbridgerapps:
The virtual-pair expansion alone did not control dynamics long enough for a ring exchange to occur. This block proves a conditional, volume-independent local-expectation estimate at fixed rescaled ring time: a fourteen-order local normal form gives O(epsilon squared) error for supplied full-carrier dynamics, dressed ice preparation and consistent local readout.

The proof uses explicit strong supports for overlapping charge stars and separate fast and slow propagation estimates. It retains all L4 winding four-cycles. Sufficient constants are explicit and extremely conservative; no practical coupling threshold, phase, photon, physical selection or preparation mechanism is claimed.

Validation: 11,091 exact predicates; an eight-file isolated premise/runtime closure; four actual subprocess mutation failures; complete independent canonical review and graph-only confirmation. An initially surviving bare-X mutant exposed a circular control, now corrected by an independent absolute fourth-order matrix check; both histories are retained. The citation manifest adds one node and three edges without changing prior entries.

Stacked on #8044. No main merge, audit verdict or authority-ledger change. Full pipeline, strict lint and changed-evidence checks remain for an actual main-landing candidate.

PR #8046, opened 2026-09-08
Prepare native ring dynamics with a smooth Hamiltonian ramp

JONbridgerapps:
The local ring-time theorem required a dressed initial ice state. This block proves that a smooth ramp of the same supplied Hamiltonian, starting from an ordinary ice state, gives the same local ring-time expectation target with O(epsilon squared) error and no separately applied dressing operation.

The proof includes the time-dependent local frame, endpoint matching and the fast backward cone over the complete ramp-plus-hold protocol. A separate slow-ice estimate replaces the effective ramp state by the original ice density matrix. Arbitrary-profile controls retain the nonzero ff''/8 scalar and derivative-driven second generator. Final dressed readout remains supplied; bare local readout has O(epsilon) error. This is not cooling, phase identification or selection of the Hamiltonian, schedule or initial ice.

Validation:18,626 exact standard-library predicates; seven-file isolated closure; three actual semantic mutation failures; complete independent canonical review. Citation graph adds one node and four edges with prior entries unchanged. Original proofs and source-bound controls are preserved.

Stacked on #8045. No main merge or audit verdict. Full pipeline, strict lint and changed-evidence checks remain for an actual main-landing candidate.


JONbridgerapps:
Reviewed source content landed on main at bd2e2b6b0a3c75d61f111b783957229d783ac684. Independent review and full constituent dispositions: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/review-8046.md

PR #8096, opened 2026-09-13
[physics-loop] formation-order-covariance bounded_theorem bounded-support

jonathonreilly:
## Summary

- Derives, from the current four-axiom memo and the realized-state primitive, that sequential single-site formation on `Z^3` has no proper-cubic-invariant total order.
- Characterizes order-blind sequential products by the adjacent-pair exchange identity, and proves that the interior isotropic binary solutions are exactly the independent (constant) rules, which do not vary with neighbour conditions. The forcing identity is that the three-site-path matching residual equals `(pminus - pplus)^2`.
- Exhibits a positive pair-of-orders witness for a declared covariant Ising-type kernel (`e^J = 2`): path3 total variation `9/50` with all-minus masses `8/25` versus `4/17`; plaquette 24 orders collapse to 4 laws; `2x3` 720 orders give twenty-eight laws, corners-first versus row-major differs on 44 of 64 configurations with total variation `135/578`.
- Selects neither a physical order nor a physical rule. Separating clauses (order-blindness, joint covariant formation, a covariant order-law, or registered order) are recorded as owner decisions and are not adopted.

This is the first executed science block of a TOE-derivation campaign that starts from the axiom text rather than a supplied Hamiltonian, carrier, or clock. It does not close a named ledger row.

## Packet

- Note: `docs/FORMATION_ORDER_COVARIANCE_AND_ISOTROPIC_BINARY_ORDER_BLIND_CLASSIFICATION_BOUNDED_THEOREM_NOTE_2026-09-13.md`
- Runner: `scripts/formation_order_covariance_and_isotropic_binary_order_blind_classification_2026_09_13.py`
- Cache: `logs/runner-cache/formation_order_covariance_and_isotropic_binary_order_blind_classification_2026_09_13.txt`
- Loop pack: `.claude/science/physics-loops/toe-formation-order-derivation-20260913/` (`HANDOFF.md`, `TRACE_GATE.md`, `CLAIM_STATUS_CERTIFICATE.md`, `OPPORTUNITY_QUEUE.md`, `REVIEW_HISTORY.md`)

## Verification

```bash
python3 scripts/formation_order_covariance_and_isotropic_binary_order_blind_classification_2026_09_13.py
```

`TOTAL: PASS=38 FAIL=0`. Cache written through `scripts/runner_cache.py` `execute_and_write_cache` with `AUDIT_TIMEOUT_SEC = 120`. Vocab lint: 0 violations.

## Promotion value gate

[...]

PR #8093, opened 2026-09-13
docs(campaign): design TOE derivation campaign by underdetermination witnesses

jonathonreilly:
## Summary

- Adds the meta note `docs/TOE_DERIVATION_CAMPAIGN_AXIOM_SUFFICIENCY_BY_UNDERDETERMINATION_WITNESSES_NOTE_2026-09-13.md`, recording an audit of `origin/main` at `b8c9d9d819` together with the 81 open pull requests #8010 to #8090, and designing a derivation campaign from that audit.
- Audit finding: no result on `main` or in the open pull requests is a consequence of the four axioms and the three approved primitives alone. Every derivation body supplies at least one of a Hamiltonian or action, a carrier or encoding, a clock, a formation unit, a rate or an order, Born weights, a Gauss kernel, or a readout instrument before it computes. The ledger's zero retained rows measure that supplied structure.
- Method: for each load-bearing target, either derive it from the axiom text and the primitives alone, or exhibit a witness pair of models that satisfy every axiom sentence and the primitives and differ on the target, exhibited by an exact finite runner. Each witness is paired with the minimal separating clause, phrased in the axiom document's register and recorded as an owner decision, not adopted. The precedent is `docs/ADMISSIBILITY_COVARIANT_Q8_CONDITIONAL_LAW_PAIR_BOUNDED_THEOREM_NOTE_2026-08-13.md`.
- Eleven blocks sit on the seams where the text is read as underdetermined: formation order and order-blind rules; possibility covariance and the two readings of the `Cl(3,0)` sentence; handed rules under proper rotations only; menus from neighbours and the Born overlap function; continuum-alphabet readability; clock and rate; formation unit; record-process dynamics; composition and law selection; glued groups as support rules; candidate assembly and gravity.
- The note asserts no theorem, grants no status, and proposes no axiom, primitive, import, comparator, or framing.

## Test plan

- `python3 scripts/vocab_lint.py --report-only <note>`: 0 violations.
- Every backticked `docs/` citation in the note resolves to a file on `origin/main`.
- Meta note only: no runner and no cache are added.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8102, opened 2026-09-13
science(admissibility): formation-order multiset key and order census

jonathonreilly:
## Summary

Formation-order block of the axiom-sufficiency campaign (design note in #8093). One bounded theorem note, one exact runner, one pinned cache.

**Physics.** For a covariant product-class nearest-neighbor rule on the six-projector menu, the formation law of a formation order is fixed by one feature of the order: the multiset of recorded-neighbour sets of size at least two. Sites that formed with zero or one recorded neighbour contribute constant normalizers and drop out of every comparison between orders. This is a corollary of the landed identity B1 of the 2026-09-06 classification note, made explicit.

On every declared window and rule the map from multisets to laws is injective, so the census of laws is the census of multisets:

| window | orders | acyclic orientations | laws |
| --- | --- | --- | --- |
| path3, P4, star4, cycle4 | 6, 24, 24, 24 | 4, 8, 8, 14 | 2, 3, 5, 4 (landed census reproduced) |
| 2x3 rectangle | 720 | 98 | 28 |
| open 2x2x2 cube | 40,320 | 1,862 | 542 |

- The monotone-box orders (linear extensions of the product partial order: 5 on the rectangle, 48 on the cube) give one law, shared by 48 and 120 orders respectively. This is the first three-dimensional instance of the 2026-09-07 monotone-order result.
- On the rectangle the reversed order gives the same law as row-major because the point reflection fixes the multiset. On the cube the reflection moves the multiset and the reversed order gives a different law, with total variation `8207/7830108` at `(3,1,2)`. The two-dimensional coincidence is a symmetry of the window, not a property of formation orders.
- The uniform order mixture differs from the static law and from the monotone law on both windows (exact total variations on the rectangle; exact positive pointwise gaps on the cube). The mixture is a witness-generating device, not a candidate law: nothing in the axioms or the realized-state primitive supplies a measure over orders.
- The adjacent-pair exchange identity holds only at empty backgrounds for the menu rules and everywhere for the constant rule.

No order is selected as physical. No rule, coupling or reading is selected. No axiom or primitive is changed. The clause that would collapse the 542 cube laws to one is named as an owner decision, not adopted.

**Relation to concurrent work.** #8096 (sister session, same day) proves the covariance and order-blind results on the binary alphabet and computes a binary Ising-type witness. Its theorems and clauses are cited as concurrent work, not used and not claimed. Its binary numbers are re-derived here by a different implementation path (denominator form versus sequential product) with exact agreement, as an independence check for both.

**Not executed.** The `3x3` window named in the design (9! orders, 6^9 configurations) exceeds the declared budget and is listed as an extension target.

## Verification

```
python3 scripts/admissibility_formation_order_menu_order_mixture_monotone_box_and_cube_census_2026_09_13.py
TOTAL: PASS=55 FAIL=0
```

Exact integers and rationals throughout; stdout under 6000 characters; 29 declared mutations each fail in exactly their declared family (census run locally, every mutation exits 1 with expected family equal to observed family). `vocab_lint --report-only` reports no violations. Docs-only change; no CI.

## Review record

Single Fable 5.1 seat, no subagents (owner instruction), no independent checker seat. Independence rests on the mutation census, the re-derivation of the landed B5 and E5 numbers and of the concurrent note's binary numbers by a second implementation path, and the reproduction of the 2026-09-07 note's five linear extensions on the rectangle.


JONbridgerapps:
Reviewed source content landed on main at 173381a6c85174f0b8d570958ac0588867fee67f. Independent review and full constituent dispositions: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/review-8102.md

PR #8107, opened 2026-09-14
science: readability of the covariant rule in the continuum Bloch alphabet

jonathonreilly:
## Result

Readability block of the derivation campaign (design PR #8093). One bounded theorem note, one exact runner, one cache.

**Physics.** In the continuum Bloch alphabet (every letter a unit vector on the sphere, the domain the Qubit axiom supplies) the finished-window record law is fully readable: the exact Fisher information of the record law at the order-blind base point, over the six-parameter covariant rule family (edge Legendre terms `P_1..P_4`, the soldered term `S_1`, the cubic term `H`), is

- `I[P_i, P_j] = delta_ij |E| / (2i+1)`, `I[S_1, S_1] = I[P_1, S_1] = |E| / 9`, `I[P_k, S_1] = 0` for `k >= 2`, `I[H, H] = 16 n / 525`, `H` uncorrelated with every edge term,

with rank 6 on both the `2x3` window and the `2x2x2` window. Every direction of the covariant rule is read from records alone; the only correlated pair (`P_1`, `S_1`) has information angle `1/3`.

**The same computation on the finite alphabets collapses**: binary letters give rank 1 (`2x3`) and rank 2 (`2x2x2`); six-axis letters give rank 3 on both windows, with the null directions named (`H`, `P_1 - P_3`, `P_4 - (5/12) P_2`). Readability of the law is a property of the alphabet the Qubit axiom supplies, not of the geometry of the window. This resolves the earlier binary-alphabet readability no-go as an alphabet effect.

**Formation order.** For every total formation order (720 on `2x3`, 40320 on `2x2x2`) and every normalised conditional class, the record law is `1 + theta T_1 + O(theta^2)` with the same static `T_1`, so the order is unreadable at first order in the rule strength. At second order the order enters through an apex kernel: the product class carries `-(1/3) sum_w iota(w) v_q . v_q'` (apex-less, arm pairs only) and the additive class `- sum_w iota(w) (v_x . v_q)(v_x . v_q')` (apex-resolved). Exact second-order Gram distances between order classes:

| window | class | static vs uniform mixture | static vs monotone box | monotone vs reversed |
|---|---|---|---|---|
| `2x3` | product | 2/27 | 2/27 | 0 |
| `2x3` | additive | 38/243 | 2/9 | 8/27 |
| `2x2x2` | product | 16/81 | 2/9 | 0 |
| `2x2x2` | additive | 32/81 | 2/3 | 8/9 |

[...]

PR #8105, opened 2026-09-14
science: handed-rule pseudo-scalar census and parity-odd record traces

jonathonreilly:
## Summary

Second block of the TOE derivation campaign by underdetermination witnesses (design note in #8093; first block #8102). This block asks whether the Admissibility sentence, "one fixed nearest-neighbour admissibility rule, covariant under lattice translations and proper cubic rotations", permits a rule that differs from its mirror image, and whether such a rule leaves a trace in record statistics.

**Setting.** Six Bloch-axis letters on the nearest-neighbour cross of `Z^3` (46,656 arm configurations, 279,936 centre-arm pairs), three readings of the covariance clause: the site label unsoldered from the frame with internal `SO(3)`, unsoldered with internal `O(3)` (conjugation included), and soldered to the frame as in the `Cl(3,0)` presentation.

**Results (all exact, runner `TOTAL: PASS=46 FAIL=0`, about 22 s).**

- Handed (mirror-odd) covariant rule terms exist in every reading. Burnside census of cubic-invariant rule terms on the window: 90 / 43 / 4540 handed against 370 / 257 / 5138 achiral for the three readings.
- Lowest handed degree per reading: five (unsoldered `SO(3)`, a letter determinant times the even neighbour coupling), six (unsoldered `O(3)`; odd degrees are excluded by the antipodal map), and two when soldered, where the handed terms of degree two form a one-dimensional space spanned by a lattice curl, the centre letter dotted into `sum_i r_i x q_i` (a Dzyaloshinskii-Moriya form). Completeness up to degree four is verified by projecting every cubic-invariant family onto the odd part.
- Explicit admissible rules: an even base rule and four handed rules, each normalised, strictly positive and varying with the neighbour condition; the mirror image of each handed rule is exactly its sign-reversed partner.
- Parity-odd record correlators are exact: under a centre-last formation law the handed rules leave traces `32/243`, `320/1377`, `2/27`, `2/15` in their own textures, zero for the even rule, negated by the mirror rule; the static (Gibbs) law gives different magnitudes with the same signs; when the centre is recorded before the full cross the trace vanishes, so the order-mixture trace is the centre-last trace divided by seven. Even record statistics are equal for a rule and its mirror.
- Reading dependence: the letter determinant is handed under `SO(3)`, excluded under `O(3)`, and achiral when soldered; the degree-six texture is handed in every reading; the curl-type textures are handed only when soldered.
[...]

PR #8110, opened 2026-09-14
science: graded zeros against order-blind covariant record rules

jonathonreilly:
## Summary

Composition and law selection block of the TOE derivation campaign (design note in #8093; earlier blocks #8102, #8105, #8107). The question: are the graded zeros of the composition discriminator the record statistics of some covariant, order-blind nearest-neighbour rule on `Z^3`, and if not, what is the minimal separating clause?

**Result.** At any single fixed particle number, yes. Across sectors, no, and the two ways a finite-window test can treat the exterior fail on complementary sub-collections.

- **Exact sectors.** Jordan-Wigner graded hopping on the `2x3`, `2x2x2` and `3x3` windows at coupling zero; ground energies `-(2+sqrt 2)`, `-(1+2 sqrt 2)`, `-6`, `-4 sqrt 2`; simple lowest eigenvalue with gap at least `1/4` certified by inertia; zero sets of sizes 3, 2, 12, 8. The `2x3` three-particle sector is new.
- **Mechanism and its limit.** The fixed-point character mechanism of the discriminator predicts 3 of 3, 12 of 12 and the four centre-crossing lines of `3x3`, but 0 of the 2 three-particle zeros on `2x3` and none of the four non-centre lines of `3x3`. Six of the 25 zeros are genuine and not character zeros.
- **Positive kernels have no zeros.** The ungraded kernel is Perron-positive in every sector, so every kernel with strictly positive local factors, including the product and additive kernels of the readability block, has empty zero set. Grading changes the support, not just the weights.
- **Static families.** A single fixed formation order with forbidden step classes reproduces only the `2x3` three-particle zeros (20 of 180 order classes; 0 of 180 and 0 of 840 elsewhere); bond-type and star-local rules reproduce no sector; the discriminator's bond-product comparator puts positive weight on every zero.
- **Sequential support rules, fixed particle number.** Exact counts of order-blind rules whose realisable patterns are the complement of the zero set: 476 / 201 / 8750 / 421 (exterior treated as unrecorded) and 782 / 2607 / 8750 / 1116 (exterior treated as recorded-empty), with forced and never-forbidden step classes listed.
- **Cross-sector impossibility.** Exterior-unrecorded: 160 particle-number-blind rules cover the three discriminator clusters jointly, but the same-cluster pair (`2x3` at two and three particles) admits 0. Exterior-empty: 39 rules cover the same-cluster pair, but the discriminator triple admits 0. All four sectors jointly: 0 in both conventions. A rule that can see the window wall does cover the same-cluster pair (existence only).
- **Witness pair and decision points (recorded, not adopted).** Graded composition on `Z^3` versus ordinary composition with the landed superlattice role-pattern encoding. Separating clause: a composition (grading) clause or an encoding clause. Also recorded: the fixed-particle-number zero pattern is not a fingerprint of grading, its sector dependence is; the exterior convention of any finite-window record-rule test is load-bearing.

## Files

- `docs/COMPOSITION_LAW_SELECTION_GRADED_ZEROS_ORDER_BLIND_RULES_BOUNDED_THEOREM_NOTE_2026-09-13.md`
- `scripts/composition_law_selection_graded_zeros_order_blind_rules_2026_09_13.py` (`TOTAL: PASS=63 FAIL=0`, about 80 s, exact `Q(sqrt 2)` arithmetic)
- `logs/runner-cache/composition_law_selection_graded_zeros_order_blind_rules_2026_09_13.txt`

## Verification

- Runner re-run in the worktree: 63 pass, 0 fail; cache generated through `scripts/runner_cache.py`.
- Independent floating-point recomputation of every energy, gap and zero set (agreement to `1e-14`) and an explicit all-orders enumeration of four rule counts, all matching.
- Eight mutations (dropped grading sign, flipped character sign, wrong energy, wrong cube bond criterion, ungraded ground vector, collapsed exterior convention, two off-by-one count expectations), each caught.
- Docs-only plus one runner and one cache; vocabulary lint clean.


JONbridgerapps:
Reviewed source content landed on main at ebb89a16a3fb846ada2389c7960343d4194a8042. Independent review and full constituent dispositions: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/review-8110.md

PR #8114, opened 2026-09-14
science: support-rule witness, Gauss law as glued covariant NN support

jonathonreilly:
## Summary

Wave A **support-rule** block of the TOE derivation campaign (design note PR #8093; earlier blocks #8102 formation-order, #8105 handed-rule census, #8107 readability, #8110 composition and law selection). The target is whether a rotation-covariant nearest-neighbour admissibility rule's zero-probability set can be a local constraint on finished records whose free part carries a gauge structure, i.e. whether Gauss's law is a *glued* record group in the owner's hierarchy (glued / breakable / free).

**Result.** On the 2x2x2 and 3x3x3 windows with pure Z_2 or Z_3 site alphabets, every rotation-covariant linear nearest-neighbour rule is a sum rule `a r_v + b (sum of neighbours)`; the group-closed covariant rules on the cube form five nested subgroups; their glued groups (minimal codewords) are rigid clusters and the pair-derived link field has flux dimension 0 for Z_2. No Gauss-type zero set appears. With the parity roles of the landed role-pattern note supplied (vertex / link / plaquette / centre), the Gauss check matrix coincides under both exterior conventions inherited from the composition block, its zero set has dimension 19 (the graph of a divergence: 12 links + 6 plaquettes + 1 centre free), and the glued groups are exactly the 28 simple loops of Q_3, the 444 simple paths (dipoles) and 7 free units, total 479, with flux dimension 5.

**Witness pair.** Model P (pure site alphabet, sum rule: rigid clusters, flux 0) versus Model R (roles + Gauss support: a Z_2 gauge theory with matter). Both satisfy every axiom sentence and primitive; they differ on the target.

**Order-blind hard support with holes.** A forming site's support is the set of values meeting every check it completes; a link completing two checks with conflicting demands has empty support and stays unrecorded (Record: "a site need not carry a record"). For every formation order the fully recorded patterns are uniform on the Gauss set and number `2^(n-d)`; all order dependence sits in the hole mass `1 - 2^(-d)`. Exact d-histograms over all 9! plane orders (3/10, 3/5, 1/10; E[2^-d] = 5/8) and all 20! cube orders (E[d] = 12/7, E[2^-d] = 924167/2598960, hole mass about 0.644) agree with an independent inclusion-exclusion over the 108 matchings of Q_3.

**Decision points recorded, none adopted.**
[...]

PR #8122, opened 2026-09-14
science: possibility-covariance witness, soldered vs unsoldered fork

jonathonreilly:
## What this is

The **possibility-covariance** block of the TOE derivation campaign designed in
#8093, and the first block of wave B. Wave A landed as #8102, #8105, #8107,
#8110 and #8114.

The Qubit axiom gives each site the possibility domain with algebraic
presentation `M_2(C)` and says a `Cl(3,0)`-compatible real-algebra presentation
"may be used equivalently and adds no further primitive structure". The Lattice
axiom gives translations and proper cubic rotations. Admissibility says there is
one fixed nearest-neighbour rule, covariant under exactly those lattice
symmetries, whose distribution over the possibilities varies with the
nearest-neighbour conditions.

Put together, the axiom text leaves one question open: **is the internal vector
frame of the possibility domain the same frame as the lattice axes, or an
independent one?** The automorphism group of `M_2(C)` is `PSU(2) = SO(3)`, and
the `Cl(3,0)` presentation carries a grade involution extending it to `O(3)`.
Nothing in the axioms says those internal rotations are the lattice rotations.
So the covariance group on the possibility side has two readings, *unsoldered*
(internal rotations act independently of lattice rotations) and *soldered* (the
three `Cl(3,0)` generators are the three lattice axes and the two groups act
diagonally). This block computes what each reading permits and separates them.

## Method

One source-only triple. The runner works on the one-site conditional domain: a
centre Bloch vector and six neighbour slots, 21 rational coordinates, with
empty slots meaning unrecorded. For each of the ten orbits of recorded-slot
patterns under the cubic group it computes the dimension of the covariant rule
space, graded by degree in the Bloch coordinates, for the trivial, vector and
lattice-vector output types under all three readings.

Four independent routes agree:

- exact Molien-Weyl character sums over the stratum stabiliser, with the
  internal Haar average taken as a Laurent residue;
- an in-runner second route using no character sum at all: the intersection of
  the kernels of the three `so(3)` derivations, then a Reynolds projector, then
  an exact rank over the rationals, blockwise by multidegree, cross-checked on
  120 tabulated dimensions with zero mismatches;
- a Frobenius orbit count reproducing the soldered degree-one row;
- an out-of-band floating-point nullity computation over 72 values, using
  neither a character sum nor a derivation, zero mismatches.

## Results

- **Occupancy-only Bloch response.** Under both unsoldered readings the space of
  covariant rule terms that depend on which neighbours are recorded but not on
  their Bloch vectors, and return a Bloch vector, is zero-dimensional on every
  stratum. Under the soldered reading it is one-dimensional exactly on the five
  strata whose recorded-slot barycentre is nonzero, and zero where the
  barycentre vanishes. This reproduces the landed solder-boundary intertwiner
  theorem and extends it to a full stratum census.
- **Linearity.** Unsoldered, no covariant rule term is linear in the Bloch
  vectors at all, on any stratum, in either the internal `SO(3)` or the internal
  `O(3)` reading. Soldered, linear terms exist on every occupied stratum, in
  counts matching an independent Frobenius orbit count.
- **The empty neighbourhood.** With no recorded neighbours the law is a
  distribution on the Bloch ball, and its covariant content is the invariant
  ring on the sphere. Unsoldered that ring is one-dimensional through degree ten,
  so every unsoldered-invariant law has the full rotation-invariant moments.
  Soldered, a second invariant first appears at degree four. The two readings
  therefore **agree** on the antipodal fair coin, so the Born-price fair-coin
  result does not distinguish them, and first separate at a fourth-order record
  statistic: the squared-probability sum is `3/5` against `1`, and the mixed
  axial fourth moment is `1/15` against `0`.
- **Registered frame.** Explicit exhibits: the soldered occupancy-to-Bloch term,
  the unique unsoldered cubic invariant of degree three with its 48 monomials,
  five spanning quadratics, and the map sending a record's internal frame to a
  rotated one. A record can register the identification the law withholds, which
  is the record-writer orientation pattern seen from the possibility side.
- **Witness pair.** Two models satisfying every axiom sentence and every
  primitive, agreeing on the empty neighbourhood and separating `2/3` against
  `1/2` on one recorded neighbour, one soldered and one not.
- **Roles.** Translations act transitively on the parity cell, so the
  superlattice role labels used by the composition and support blocks cannot be
  law content under translation covariance. They are next-nearest-range data.
  This answers the A5 decision point directly.

## Decision points recorded, not adopted

The minimal separating clause is "the three `Cl(3,0)` generators are the three
lattice axes". It is recorded as a decision point, with the fourth-order record
statistic above as the observable that would settle it. No clause is adopted and
no axiom text changes here.

## Verification

```bash
python3 scripts/possibility_covariance_soldering_fork_invariant_rules_2026_09_14.py
```

Expected summary line: `TOTAL: PASS=94 FAIL=0` (followed by the elapsed time).
About 15 s, stdout 5901 characters, exact rational arithmetic, no dense space of
dimension 2^11 or more. Nine mutations run to completion, nine caught; a tenth
substitution is reported in the note as a non-defect with its reason.


JONbridgerapps:
Reviewed source content landed on main at 46073d34d92ae1fe47f7f9e2b126f18188a0c01a. Independent review and full constituent dispositions: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/review-8122.md

PR #8128, opened 2026-09-14
Construct positive principal-flux transfer and bound its Hamiltonian limit

jonathonreilly:
The supplied principal-flux penalty Hamiltonian previously had only a small-step transfer construction. This note constructs a strictly positive Gaussian completion on the same clock configurations, proves its gauge projection and Hamiltonian limit, and gives the extensive error bound `E [29 t² + t (6 K + 16 λ)] ε²`. It also derives the first local logarithmic correction.

An exact rational negative quadratic form on a physical three-plaquette strip shows that the unique-principal-step kernel need not be positive at finite step. The positive completion differs from it by at most `6 E (ε t)^4`; both have the stated Hamiltonian limit. This counterexample is restricted to its explicit boundary and parameters.

Validation: the canonical runner passes nine finite check families; all 18 deliberately introduced mathematical faults are detected. The runner cache is fresh, source/compile/vocabulary/diff checks pass, and the generated citation manifest adds one zero-dependency node. The focused changed-evidence check has no failures and checks no pre-existing audited claim.

This is a proposed bounded theorem referral for independent review. It does not establish a volume-uniform ground-state phase, a native action selected by the axioms, or a TOE. Independent review, formal retained status, full pipeline, strict lint, and combined integration gates remain pending. No axioms, primitives, editable prompts, or audit verdicts are changed.


JONbridgerapps:
Reviewed source content landed on main at 9c59d327433d52f5189d91c38517bbd90e5b3bf6. Independent review and full constituent dispositions: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/review-8128.md

PR #8150, opened 2026-09-15
[physics-loop] admissibility-induced-law block16 — bounded_theorem — bounded-support: the static law is not a mixture of formation laws on any window with a cycle — a one-record flip raises every order's weight; the exact criterion on cycle-free windows

JONbridgerapps:
## What this PR is

Block 16 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `fa0200b439`): its parents are block 01 and the formation-order census note of 2026-09-13, both on `main`. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-16 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (X1–X3 proved for every finite window and every non-constant rule; executed at `(3,1,2)`; conditional on the records-only reading, the six-axis menu and the product rule as supplied conditions; audit required).
- Note: `docs/ADMISSIBILITY_RULE_STATIC_LAW_IS_NOT_A_MIXTURE_OF_FORMATION_LAWS_ON_ANY_WINDOW_WITH_A_CYCLE_FLIP_MONOTONICITY_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_static_law_not_a_mixture_of_formation_laws_flip_monotonicity_2026_09_15.py` (17 checks, 10 mutations, exact rational and symbolic arithmetic, about 25 s).

## Why this block

PRs #8148 and #8149 excluded rate and unit clauses on the plaquette through its symmetries. The normalizer lemma of #8149 gives a symmetry-free mechanism that settles every window at once, and closes the "randomized formation" route to the static reading completely for value-blind randomization.

## What it says

- **X1 (the flip lemma).** For an order `σ` let `w_σ(v) = μ_σ(v)/Π_edges K(v_e)`; the static law is a constant multiple of `Π_edges K`. Flipping one record of the constant pattern from `b` to `−b` (to an orthogonal value when `p = q`) raises `w_σ` for every order, strictly iff the flipped site lies in a recorded set of size at least two — because each affected normalizer strictly decreases by `K_k(b,…,b) − K_k(−b,b,…,b) = ((p−q)/Z_1)[(p/Z_1)^{k−1} − (q/Z_1)^{k−1}]`.
[...]

PR #8147, opened 2026-09-15
[physics-loop] admissibility-induced-law block13 — bounded_theorem — bounded-support: the record two-point function of a causal Gaussian formation law is a heat kernel in level time, never the lattice Green function; massless iff value-covariant, then no stationary law

JONbridgerapps:
## What this PR is

Block 13 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `5aca149c42`): block 07's Gaussian-instance note on `main` is the only declared parent; blocks 09 and 12 (open PRs #8139, #8146) are referenced for context only. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-13 records (contract with the supervisor's lens pass, the control and the refuting pass with outputs, results, review history, certificate, trace gate, queue refresh). Supervisor-run throughout (owner directive 2026-09-15: after a conclusion, pull up and find the next high-leverage science; no subagents; proofs over computation).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (T1–T5 proved for every positive weight and noise; conditional on the records-only reading, the monotone class and the Gaussian instance as supplied conditions; audit required).
- Note: `docs/ADMISSIBILITY_RULE_CAUSAL_GAUSSIAN_FORMATION_LAW_RECORD_TWO_POINT_FUNCTION_HEAT_KERNEL_NOT_LATTICE_GREEN_FUNCTION_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_causal_gaussian_formation_law_record_two_point_function_heat_kernel_2026_09_15.py` (23 checks, 15 mutations, exact rational and symbolic arithmetic, about 3 s).

## Why this block

The derivation campaign's verdict table (#8093) leaves the record-dynamics seam "to be tested": whether a Laplacian or transverse kernel appears in the finished record statistics of a formation law with no tick supplied, and, for the gravity node, whether a record statistic can have the lattice Green function as its two-point function. Nothing on `main` or in the open PRs computes a formation law's two-point function at long distance. This block answers the seam exactly on the Gaussian instance.

## What it says

For the real Gaussian instance of the formation law (each site draws `w` times the sum of its three recorded predecessors plus independent noise `σ²`; gain `g = 3w`, equal weights by covariance) in the monotone class on the half-space of `Z³` given the level-0 records:

[...]

PR #8146, opened 2026-09-15
[physics-loop] admissibility-induced-law block12 — bounded_theorem — bounded-support: the strong-coupling side of the 3D formation law — a noisy majority eroder in level time; finite-W metastability; the ordered phase as one exact obligation

JONbridgerapps:
## What this PR is

Block 12 of the physics-loop campaign `admissibility-induced-law-20260906`, **stacked on block 11 (PR #8142)** — the base is the block-11 branch, so the diff shows block 12 only. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-12 records (contract with the supervisor's lens pass, the control and the refuting pass with outputs, results, review history, certificate, trace gate, queue refresh). Supervisor-run throughout (owner directive 2026-09-15: keep going on the strong-coupling phase; no subagents; proofs over computation).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (S0–S5 proved for every positive triple, S4 for every finite cross-section; S6 an obligation, not a claim; conditional on the named supplied readings; audit required).
- Note: `docs/ADMISSIBILITY_RULE_STRONG_COUPLING_FORMATION_LAW_LEVEL_AUTOMATON_ERODER_METASTABILITY_ORDERED_PHASE_OBLIGATION_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_strong_coupling_formation_law_level_automaton_eroder_metastability_2026_09_15.py` (20 checks, 13 mutations, exact rational and symbolic arithmetic, about 10 s).

## What it says

[...]

PR #8151, opened 2026-09-15
[physics-loop] admissibility-induced-law block17 — bounded_theorem — bounded-support: the static six-axis law at strong coupling has long-range order and several Gibbs states on Z² and Z³ for p ≥ 216 max(q,r) — reflection positivity through site planes, chessboard, contour counting (3D by the in-plane count)

JONbridgerapps:
## What this PR is

Block 17 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `2ed54cb83a`): its parent is block 03 (the static law's uniqueness region), on `main`. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-17 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** — both the two- and the three-dimensional theorems are unconditional; three standard mathematical theorems are imported at definition level; audit required.
- Note: `docs/ADMISSIBILITY_RULE_STATIC_SIX_AXIS_LAW_STRONG_COUPLING_LONG_RANGE_ORDER_AND_SEVERAL_GIBBS_STATES_REFLECTION_POSITIVITY_CHESSBOARD_PEIERLS_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_static_six_axis_law_strong_coupling_long_range_order_several_gibbs_states_2026_09_15.py` (20 checks, 13 mutations, exact rational and symbolic arithmetic, about 15 s).

## Why this block

"One law or several" for the static reading of the six-axis rule: block 03 settled the weak-coupling side (one Gibbs state where the one-site influence is small). The ordered side was open on `main`, and blocks 13–16 (PRs #8147–#8150) make the static reading the one that carries the repository's Green functions.

## What it says

- **T1.** The weight matrix has eigenvalues `Z_1`, `p − q` (three times) and `p + q − 2r` (twice); the full cube group preserves it and acts transitively on the six values.
- **T2.** On every torus of even side the static law is reflection positive for reflections through planes containing sites, **for every positive weight matrix** — the shared plane makes `E[F · F∘θ]` a sum of squares. No definiteness of the coupling is needed, so the argument covers indefinite triples such as `(5,2,4)`.
- **T3, T4.** The chessboard estimate (re-proved by the iterated quadratic-form inequality), and the disseminated bound: `μ_L(every bond of one direction disagrees)^{1/N} ≤ 6m/p`, `m = max(q, r)`.
- **T5 (two dimensions).** Planar cut–cycle duality and an explicit count of enclosing dual cycles (`(n/2)·3^{n−1}`) give, for `ε = 6m/p ≤ 1/36`, i.e. **`p ≥ 216 max(q, r)`**, `μ_L(v_0 ≠ v_x) ≤ 5/24 + 45/256 = 295/768 < 1/2` for every `x` within a quarter of a box of side at least 20: long-range order on every large torus.
- **T6.** Every limit point of the torus laws is a translation-invariant Gibbs state with that long-range order; a unique Gibbs state would be extremal, hence tail-trivial, hence short-range correlated with `μ(v_0 = v_x) → 1/6` — contradiction. **The static law on `Z²` has at least two Gibbs states for `p ≥ 216 max(q, r)`**; the symmetry orbit of a non-symmetric extremal state has between two and six members.
- **T7 (three dimensions).** Restrict to a coordinate plane through the origin: two differing records are separated by bad bonds of the plane they share, so the planar count of T5 applies inside the plane, while the chessboard estimate and the disseminated bound are the three-dimensional ones with the same value. **The static law on `Z³` has at least two Gibbs states for the same threshold `p ≥ 216 max(q, r)`**, with no surface-connectivity lemma. (Folded after opening: the first version stated this conditionally with a crude threshold.)

The thresholds are existence constants and are not claimed optimal; the band between block 03's thresholds and these is undecided. The formation law's phase (block 12's obligation, PR #8146) is untouched.

## Review surface

- Refuting pass (`CHECKER_block17_findings.md`, disjoint machinery): the sum-of-squares decomposition against the direct expectation; the two-step chessboard form through the other site plane; polyomino placements for the contour counts; exact partial sums of both series; a second adjacency implementation for the three-dimensional counts — all consistent. Three drafting defects fixed before the census (a contour count that forgot the domino's two orientations; a threshold checked in the wrong units; forbidden tokens); the conditional three-dimensional statement replaced by the in-plane count after opening.
- Census: 13 mutations, each failing in exactly its declared family. Gates: pipeline ×2, `audit_lint --strict` OK, vocabulary lint clean on the block's files, `git diff --check` clean; only the intended manifest is included from generated audit data.
- N-gate with N5 resolution lines in cached stdout; classical names only under Prior art and Imports; no floating-point literal in the runner.

## Not claimed

Optimal thresholds; the band between block 03's thresholds and these; the formation law's phase; the number of extremal states beyond the orbit bound; any coupling as physical.

## Verification

```bash
python3 scripts/admissibility_rule_static_six_axis_law_strong_coupling_long_range_order_several_gibbs_states_2026_09_15.py
python3 scripts/admissibility_rule_static_six_axis_law_strong_coupling_long_range_order_several_gibbs_states_2026_09_15.py --list-mutations
```


JONbridgerapps:
Reviewed positive source salvage landed on main at 19b73e32772a990f4d9d0792013347943bc80818. Full disposition and reason: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/drain8151-report.md

# PR8151 partial landing disposition

Final independent source review: **PASS for positive identities and explicit conditional implications**. Preserve the original branch and all22 original versions in `docs/work_history/review_loop/pr8151/original-manifest.json`.

Accepted: the pair-matrix spectrum, real-valued site-reflection sum of squares, the actual all-direction event count with its stated weight hypotheses, exact finite checks and contour arithmetic. The contour/limit consequences retain explicit uniform bad-bond and correlation hypotheses. The missing bound is an obligation, not a proved conclusion.

Rejected as written: the reflection dissemination step identifies the wrong event; its orbit reaches only parity-selected transverse rows. It does not establish the claimed unconditional phase threshold216. A nonuniform six-value distribution can have48 group images, so the asserted at-most-six orbit bound is false; equality of one-site laws does not establish equality of states. The unconditional phase claims and negative certification remain deferred with complete original proofs recoverable.

Five finding groups were repaired or narrowed. The original reviewer `/root/review_8011` independently checked reflection geometry, group-orbit and algebra controls and confirmed the final source/evidence in the same session. Sole capture20/0,29.601seconds,1,812,561,920-byte sampled peak under180seconds/3GiB; no violation or scientific retry. Final cache and combined pipeline/strict/evidence checks have separate receipts. Review applies no audit verdict.

Detailed immutable records on ai/execution include `drain8151-final-review.json`, `drain8151-unit-final.json`, original inventory, decisive countercontrols and execution receipt. Preserve this branch for the unlanded phase conclusions.

PR #8153, opened 2026-09-15
[physics-loop] admissibility-induced-law block19 — bounded_theorem — bounded-support: the unsoldered static law's ordered phase carries the lattice Green function in its transverse channel — infrared bound above, zero-field Bogoliubov bound below, long-range order for β > 3√3π/8

JONbridgerapps:
## What this PR is

Block 19 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `2ed54cb83a`): its parent is the landed possibility-covariance note of 2026-09-14. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-19 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (G1–G5 proved for every `β > 0`, G3–G5 for `β > 3√3π/8`; conditional on the pure-state sphere as the possibility domain, the unsoldered reading and the exponential overlap as supplied conditions; three standard mathematical theorems imported at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_UNSOLDERED_SPHERE_STATIC_LAW_ORDERED_PHASE_TRANSVERSE_CHANNEL_CARRIES_THE_LATTICE_GREEN_FUNCTION_INFRARED_AND_BOGOLIUBOV_BOUNDS_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_unsoldered_sphere_static_law_ordered_phase_transverse_channel_green_function_2026_09_15.py` (17 checks of the algebraic skeleton, 13 mutations, exact symbolic).

## Why this block

The derivation campaign's gravity node (#8093) needs "a covariant scalar record statistic whose two-point function is the lattice Green function". Block 13 (PR #8147) showed a causal formation law is parabolic and cannot supply it; block 17 (PR #8151) showed the soldered six-axis static law orders with a discrete symmetry and has no massless channel. The one remaining candidate inside the axioms' readings is the static reading with a *continuous* possibility symmetry. This block shows it delivers.

## What it says

For the unsoldered static law on the sphere with the exponential overlap `e^{β s·s'}` on tori of even side:

[...]

PR #8154, opened 2026-09-15
[physics-loop] admissibility-induced-law block20 — bounded_theorem — bounded-support: the unsoldered sphere static law has no long-range order on any coordinate plane or line; the Green-function channel needs the third dimension

JONbridgerapps:
## What this PR is

Block 20 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `2ed54cb83a`): its parent is the landed possibility-covariance note of 2026-09-14. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-20 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (H1 proved for `d ≤ 3` and every `β > 0`; H2–H3 on every plane and line torus with `L ≥ 2`; H4 a placement corollary conditional on blocks 17 and 19 (open PRs) for their halves; conditional on the pure-state sphere as the possibility domain, the unsoldered and static readings and the exponential overlap as supplied conditions; three standard facts imported at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_UNSOLDERED_SPHERE_STATIC_LAW_NO_LONG_RANGE_ORDER_ON_PLANES_AND_LINES_THIRD_DIMENSION_LOAD_BEARING_FOR_THE_GREEN_FUNCTION_CHANNEL_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_unsoldered_sphere_static_law_no_long_range_order_on_planes_and_lines_2026_09_15.py` (20 checks, 13 mutations, exact, under a second).

## Why this block

Block 19 (PR #8153) locates the gravity node's kernel in the transverse channel of the ordered unsoldered static law on `Z³`. The Lattice axiom names `Z³`. This block asks whether the third dimension is doing work there, and answers: yes. A plane never orders under that reading, at any coupling strength, and neither does a line.

## What it says

For the unsoldered static law on the sphere with the exponential overlap `e^{β s·s'}` on tori `(Z/2LZ)^d`:

- **H1.** Block 19's zero-field lower bound `⟨|ŝ¹(k)|²⟩ ≥ (M_N²/3)²/(βE(k) + 4/(3N))` holds in every dimension `d ≤ 3`; the dimension enters only through the count of `N` bonds per direction. Re-proved here so the note stands alone (the generator identity `L ŝ¹(k) = N^{1/2} m̂` supplies the order-parameter lower bound directly, where main's quantum no-order notes needed an unsupplied Ward lemma).
- **H2.** On the plane, `Σ_{k≠0}|k|^{−2} ≥ (N/π²)H_{L−1}` (sup-norm shells of `8j` wavevectors with `|n|² ≤ 2j²`), `E(k) ≤ |k|²`, and `4/(3N) = |k_min|²/(3π²)`; on the line the `2⌊√L⌋` smallest wavevectors have `βE(k) + 4/(3N) ≤ (βπ² + 2/3)/L`.
- **H3.** With the sum rule `Σ_k⟨|ŝ¹(k)|²⟩ = N/3`: `M_N⁴ ≤ (3π²β + 1)/H_{L−1}` on the plane and `M_N⁴ ≤ (6π²β + 4)/√L` on the line (`L ≥ 2`). **The long-range-order parameter tends to zero as the torus grows, at every `β > 0`.**
- **H4.** Hence block 19's Green-function channel exists on `Z³` above its threshold and on no coordinate plane or line, while the soldered six-axis static law orders on the plane (block 17, PR #8151): the third dimension of the Lattice axiom is load-bearing for the gravity node's kernel under the unsoldered static reading, and the menu (continuous or discrete), not the lattice, decides whether a plane orders.

In plain words: the same inequality that, in three dimensions, sandwiches the transverse record correlator between multiples of the inverse lattice Laplacian, forces the order to vanish on a plane, because the plane's sum over wavevectors of `1/|k|²` grows without bound. The kernel that the gravity lane needs is a three-dimensional feature of this reading.

## Review surface

- Refuting pass (`CHECKER_block20_findings.md`, disjoint machinery): H1 on the exactly solvable two-site instance at four couplings (with the sum rule exact there); the plane's and the line's sums by direct evaluation to `L = 64` and `L = 400` at three couplings; the sphere average in a second parametrization; Parseval on a random field. **One finding, folded:** the line's block count `2⌊√L⌋` fails at `L = 1` (where `−n` is outside the wavevector range), so the line statements carry `L ≥ 2`. The control corrected the contract's first draft before the primary (the plane's truncated last shell; the line's small term).
- Census: 13 mutations, each failing in exactly its declared family (one mutation was tightened after a 12/13 census). Gates: pipeline ×2, `audit_lint --strict` OK, vocabulary lint clean on the block's files, `git diff --check` clean; only the intended manifest is included from generated audit data.
- The bounds are the known weak ones of this route (the plane's bound bites only for `H_{L−1} > 3π²β + 1`); the rate is not claimed.

## Not claimed

The rate at which the plane's correlations decay; the Born overlap; the kernel's normalization on `Z³`; any reading, rule, coupling or dimension as physical.

## Verification

```bash
python3 scripts/admissibility_rule_unsoldered_sphere_static_law_no_long_range_order_on_planes_and_lines_2026_09_15.py
python3 scripts/admissibility_rule_unsoldered_sphere_static_law_no_long_range_order_on_planes_and_lines_2026_09_15.py --list-mutations
```


JONbridgerapps:
Reviewed positive source salvage landed on main at d0e809a52de34a6e1f086dee7b341dc1f96a07c7. Full disposition and reason: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/drain8154-report.md

# PR8154–8157 bounded source review

The four supplied sphere-model mathematical results are accepted with explicit scope corrections: finite-torus magnetization bounds in dimensions one and two; weak-coupling comparison, unique DLR law and correlation bounds; exact cubic-walk return sums and a sufficient magnetization threshold; and planar complex-rotation correlation upper bounds. The full retained proofs and mathematical imports remain explicit. The corrected component-Fourier parent from PR8153 is the actual landed input.

Material fixes remove unsupported physical-channel and universal negative conclusions. A vanishing magnetization lower bound does not exclude all Green kernels; the weak-coupling model does not classify all physical massless channels; the return-sum threshold is sufficient rather than the true critical value; and a coupling-dependent upper exponent does not exclude a faster fixed power. Finite-window/torus restrictions and the noninteger-radius shell estimate are corrected. Historical Monte Carlo is labeled as such. These deferred conclusions are not certified by relabeling them.

All 88 original path/mode/blob versions are recoverable under docs/work_history/review_loop/pr8154, 22 per constituent. The four original branches must be preserved for deferred negative and physical work after close-with-reason. No durable positive theorem is dropped. No audit verdict is applied.

Original independent reviews and focused corrections are bound in drain8154-original-review.json and drain8156-original-review.json, with 143 and 57 independent controls. Four sole final captures each passed 20 checks. Actual external process-tree limits were 1 GiB, with 180/300/300/180 second timeouts. Elapsed times were 1.05/4.35/35.51/2.11 seconds; peaks were below 122 MB. These finite controls support the indexing and arithmetic; complete proofs carry the infinite-volume statements.

Final source confirmation is drain8154-final-review-v2.json on current-main base9ff59dc9caf61444b8bf3bd3d0d13ea53ace1dac. It verifies all 102 source/cache paths and the harmless additive helper-map input changes from PR8025. No science was rerun. Operational adapter and old-base freshness failures remain preserved; only the current-base final preflight and combined gate authorize landing.

PR #8157, opened 2026-09-15
[physics-loop] admissibility-induced-law block23 — bounded_theorem — bounded-support: on a plane the unsoldered sphere static law's correlations decay at least algebraically at every coupling — complex rotations with explicit exponent

JONbridgerapps:
## What this PR is

Block 23 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `47dd808347`): its parent is the landed possibility-covariance note of 2026-09-14. It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-23 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (P1–P4 proved for every `β > 0` on every plane window containing the disc, with any exterior records, on tori, and for every infinite-volume static law on the plane should one exist; conditional on the pure-state sphere as the possibility domain, the unsoldered and static readings and the exponential overlap as supplied conditions; two standard facts imported at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_UNSOLDERED_SPHERE_STATIC_LAW_PLANE_CORRELATIONS_DECAY_ALGEBRAICALLY_AT_EVERY_COUPLING_COMPLEX_ROTATIONS_EXPLICIT_EXPONENT_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_unsoldered_sphere_static_law_plane_algebraic_decay_complex_rotations_2026_09_15.py` (20 checks, 13 mutations, exact).

## Why this block

Block 20 (PR #8154) proves the unsoldered sphere static law never orders on a plane, with only a logarithmic rate for the order parameter. What replaces block 19's Green-function channel on a plane? This block gives the rate of decay of the correlations themselves, uniform in the volume: a power law with a coupling-dependent exponent.

## What it says

Write each record as `(ρ cos φ, ρ sin φ, ζ)`.

- **P1.** For any real function `a` on the sites (zero at exterior sites), `|⟨s_0^1 s_x^1 + s_0^2 s_x^2⟩| ≤ exp(a_x − a_0 + β Σ_b (cosh(a_y − a_z) − 1))`, and `|⟨s_0·s_x⟩| ≤ (3/2)` times the same. The angles are shifted by `i a_y` (each period integral is unchanged), the observable picks up `e^{a_x − a_0}`, and each bond weight's modulus grows by at most `e^{β(cosh(a_y − a_z) − 1)}` since `cos φ cosh a ≤ cos φ + (cosh a − 1)`.
- **P2.** With `a_y = γ log((1+R)/(1+d(y)))` inside the disc of radius `R = d(x)` and zero outside — on any window containing the disc, or on the torus with the torus distance — `|a_y − a_z| ≤ γ/(1 + min(d(y), d(z)))` on every bond, and `Σ_b (cosh(a_y − a_z) − 1) ≤ 2γ² cosh γ (1 + 8H_R)` by sup-norm shells of at most `8j` sites.
- **P3.** Hence `|⟨s_0·s_x⟩| ≤ (3/2) e^{18βγ²cosh γ} (1+R)^{−(γ − 16βγ²cosh γ)}`, and with `γ = min(1, 5/(256β))` and `cosh 1 ≤ 8/5`:
  **`|⟨s_0·s_x⟩| ≤ (3/2) e^{9/16} (1 + d(x))^{−κ(β)}`, `κ(β) = 5/(512β)` for `β ≥ 5/256`, `κ(β) = 1 − 128β/5 ≥ 1/2` below** — uniformly in the window, the exterior records and the side.
- **P4.** On the torus `M_N² ≤ 6e^{9/16}(1+L)^{−min(κ,1)} + 1/(4L²)` (a power of the side where block 20 had `(log L)^{−1/2}`); every infinite-volume static law on the plane has the same decay; and since the exponent depends on `β` while a Green-function kernel's power is fixed by the dimension, **no Green-function-type kernel exists on planes at any coupling** — the rate-level counterpart of block 20's dimension placement.

In plain words: on a plane, the record field's memory of a far site fades like a power of the distance whose exponent gets smaller as the coupling grows but never reaches zero; in three dimensions above the threshold it fades like `1/r` with an exponent that has nothing to do with the coupling. That is the difference between "no kernel here" and "the kernel lives here".

## Review surface

- Refuting pass (`CHECKER_block23_findings.md`, disjoint machinery): the full shift bound against a heat-bath Monte Carlo truth on a `4×4` window at two couplings; the shift function's Lipschitz bound on a `24×24` torus with wrapping; the harmonic bound to `10⁶`; the exact bond sums on `Z²` (a factor ten below the shell bound — room recorded, not claimed); the optimization by a grid; direct torus shell sums to `L = 200`. No findings.
- Census: 13 mutations, each failing in exactly its declared family. Gates: pipeline ×2, `audit_lint --strict` OK, vocabulary lint clean on the block's files, `git diff --check` clean; only the intended manifest is included from generated audit data.
- The constants are the crude ones of this route; the true decay of the three-component model on the plane is faster and is not claimed.

## Not claimed

The true rate of decay; sharpness of `κ(β)` or the prefactor; the Born overlap; the existence of infinite-volume laws on the plane; any reading, rule, coupling or dimension as physical.

## Verification

```bash
python3 scripts/admissibility_rule_unsoldered_sphere_static_law_plane_algebraic_decay_complex_rotations_2026_09_15.py
python3 scripts/admissibility_rule_unsoldered_sphere_static_law_plane_algebraic_decay_complex_rotations_2026_09_15.py --list-mutations
```


JONbridgerapps:
Reviewed positive source salvage landed on main at d0e809a52de34a6e1f086dee7b341dc1f96a07c7. Full disposition and reason: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/drain8154-report.md

# PR8154–8157 bounded source review

The four supplied sphere-model mathematical results are accepted with explicit scope corrections: finite-torus magnetization bounds in dimensions one and two; weak-coupling comparison, unique DLR law and correlation bounds; exact cubic-walk return sums and a sufficient magnetization threshold; and planar complex-rotation correlation upper bounds. The full retained proofs and mathematical imports remain explicit. The corrected component-Fourier parent from PR8153 is the actual landed input.

Material fixes remove unsupported physical-channel and universal negative conclusions. A vanishing magnetization lower bound does not exclude all Green kernels; the weak-coupling model does not classify all physical massless channels; the return-sum threshold is sufficient rather than the true critical value; and a coupling-dependent upper exponent does not exclude a faster fixed power. Finite-window/torus restrictions and the noninteger-radius shell estimate are corrected. Historical Monte Carlo is labeled as such. These deferred conclusions are not certified by relabeling them.

All 88 original path/mode/blob versions are recoverable under docs/work_history/review_loop/pr8154, 22 per constituent. The four original branches must be preserved for deferred negative and physical work after close-with-reason. No durable positive theorem is dropped. No audit verdict is applied.

Original independent reviews and focused corrections are bound in drain8154-original-review.json and drain8156-original-review.json, with 143 and 57 independent controls. Four sole final captures each passed 20 checks. Actual external process-tree limits were 1 GiB, with 180/300/300/180 second timeouts. Elapsed times were 1.05/4.35/35.51/2.11 seconds; peaks were below 122 MB. These finite controls support the indexing and arithmetic; complete proofs carry the infinite-volume statements.

Final source confirmation is drain8154-final-review-v2.json on current-main base9ff59dc9caf61444b8bf3bd3d0d13ea53ace1dac. It verifies all 102 source/cache paths and the harmless additive helper-map input changes from PR8025. No science was rerun. Operational adapter and old-base freshness failures remain preserved; only the current-base final preflight and combined gate authorize landing.

PR #8166, opened 2026-09-16
[physics-loop] compact-response bounded_theorem conditional-support

jonathonreilly:
The lifted compact-Hamiltonian bridge previously lacked a volume-uniform cubic source estimate. This proposal derives that estimate and an all-p Hessian bound, then uses the actual time-interpolated carrier to establish Gaussian magnetic subsequences for an explicitly defined real comparison law at fixed positive coupling. A separate conditional construction confines integer-current fillings to regions with exponential component tails.

The compact-to-comparison normalized measure and source map remains open. Sparse defects and pointwise agreement on small raw curls do not supply that map. This proposal claims no compact photon phase, unique Maxwell covariance, native-law selection, TOE, axiom update, independent review or audit status.

The four-note argument is in the [source guide](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/02bfa4eb99fce8329619a476775299567f735d4b/.claude/science/physics-loops/toe-compact-response-20260916/README.md): [bridge derivatives](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/02bfa4eb99fce8329619a476775299567f735d4b/.claude/science/physics-loops/toe-compact-response-20260916/notes/BRIDGE_CUBIC_SOURCE_BOUND.md), [convex comparison](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/02bfa4eb99fce8329619a476775299567f735d4b/.claude/science/physics-loops/toe-compact-response-20260916/notes/CONVEX_EXTENSION_OF_LIFTED_KERNEL.md), [conditional integer fillings](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/02bfa4eb99fce8329619a476775299567f735d4b/.claude/science/physics-loops/toe-compact-response-20260916/notes/SPARSE_INTEGER_FILLINGS_AND_REAL_LIFTS.md), and [comparison Gaussian subsequences](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/02bfa4eb99fce8329619a476775299567f735d4b/.claude/science/physics-loops/toe-compact-response-20260916/notes/GAUSSIAN_MAGNETIC_LIMIT_FOR_THE_COMPARISON_LAW.md). The geometric application explicitly depends on proposed path bounds at #8165 head c228dbcfc0a6b71639316915f460c32dbf479723; its abstract probability premise is stated separately.

[...]

PR #8159, opened 2026-09-16
Personal TOE derivation milestone: rotor limits, native response, and open bridges

jonathonreilly:
Preserve the personal TOE derivation campaign as a review and recovery milestone. The packet contains provisional results in several supplied models, organized into six review units in the [review map](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ccca36e505a4cc5d0cd4735b377c15d69042fc41/.claude/science/physics-loops/toe-fixed-clock-field-20260915/README.md). Selective independent review is still required before scientific integration.

The strongest candidate bridges are fixed-coupling cyclic equilibrium states to integer-Gauss rotor states; periodic free-matter holonomy selection and the full coupled fixed-box slow spectrum; physical Maxwell/neutral-matter correlations when coupling tends to zero before volume grows; and boundedness of the complete mixed native sixth-order kernel, which leaves the known singleton infrared pole uncancelled. Later notes construct positive weights with a common free cone and derive exact response identities that distinguish static twist screening from a photon witness. The source-spectrum, regulator and formation/field explorations retain their explicit domains and failed attempts.

The fixed-positive-coupling charged phase, original fixed-clock full-field limit, physical law selection and TOE completion remain open. No axiom update, independent audit verdict or blanket acceptance of the exploratory packet is claimed. The user extended the active campaign; discovery continues on a separate current-main branch after this checkout is removed.

Validation includes exact cochain and CAR identities, distinct spectral/Fourier implementations, analytic bounds and targeted finite discriminators. Initial failures, missed spectral multiplicity and repairs are preserved. All169 personal-review input hashes match; all26 review-map links resolve; vocabulary checks and git diff checks pass. These authenticate source and challenge calculations, rather than supply independent scientific review. The [author review](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ccca36e505a4cc5d0cd4735b377c15d69042fc41/.claude/science/physics-loops/toe-fixed-clock-field-20260915/review/MILESTONE_PERSONAL_REVIEW.md) gives coverage and limitations, and the [manifest](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ccca36e505a4cc5d0cd4735b377c15d69042fc41/.claude/science/physics-loops/toe-fixed-clock-field-20260915/review/MILESTONE_MANIFEST.json) binds the packet.

[...]

PR #8168, opened 2026-09-16
[physics-loop] admissibility-induced-law block25 — bounded_theorem — bounded-support: the three-dimensional formation law has an ordered phase — stability of the noisy level automaton with an explicit threshold; six invariant laws; block 12's S6 closed

JONbridgerapps:
## What this PR is

Block 25 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `6dda46fc1a`): its parent is block 01's finite-window note on `main`; the level-time reading of the formation law is declared in the note. It opens for review one bounded-theorem note, its exact runner (with the whole contour construction embedded and executed exhaustively on small cones) and pinned cache, the manifest acknowledgment, and the pack's block-25 records. Supervisor-run throughout (owner directive 2026-09-16: sync main, read the latest PRs, take the highest-leverage science; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (T0–T6 proved for the noisy majority automaton for every `ε ≤ 7/10⁶`; T7 for the monotone formation law of the six-axis product rule; conditional on the records-only reading, positivity, the six-axis menu and the monotone order as supplied conditions; three standard facts imported at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_THREE_DIMENSIONAL_FORMATION_LAW_ORDERED_PHASE_STABILITY_OF_THE_NOISY_LEVEL_AUTOMATON_EXPLICIT_THRESHOLD_SIX_INVARIANT_LAWS_BOUNDED_THEOREM_NOTE_2026-09-16.md`. Runner: `scripts/admissibility_rule_three_dimensional_formation_law_ordered_phase_stability_noisy_level_automaton_2026_09_16.py` (22 checks, 13 mutations, exact).

## Why this block

Block 12 (PR #8146) read the three-dimensional formation law level by level as a three-predecessor majority automaton with one-sided noise `ε(p,q,r) ≈ max(q,r)/p`, proved the noiseless majority erodes finite islands, and reduced "does the formation reading order at strong coupling?" to one obligation (S6): that the noisy eroder, started from the all-zero level, keeps the density of ones small forever. The owner's new drafts #8159–#8166 are his live campaign on the compact U(1) photon bridge; S6 was the last open obligation of the campaign's record-dynamics row and nobody held it. This block closes it.

## What it says

For the noisy majority automaton `η_x = maj(η_{x−e₁}, η_{x−e₂}, η_{x−e₃}) ∨ ζ_x` in level time on `Z³`, with i.i.d. Bernoulli(`ε`) noise and all zeros at levels `≤ 0`:

- **T1–T4 (the explanation tree).** A one at `x` is explained by a tree of *arrows* (site to predecessor) and *forks* (site to sibling) through one-sites, rooted at `x`, in which every node has at most one arrow to a predecessor, the nodes with none are exactly the noise sites (`n` of them), and there are **exactly `n − 1` forks and at most `3(n − 1)` arrows**. The construction refines *spanned clusters* along the three charges `M_k(z) = z_k − τ(z)/3` (sum zero; each rises by exactly `1/3` along a suitable predecessor of any one), using the connected cause graph of a cluster, the excuse identity `Σ_k M_k(Excuse_k(v_k)) = Span + 1`, and a spanning lemma that redistributes a span over a minimal cluster–fork tree without changing its total. The span rises by at least one per refinement and equals the fork count at the end, and the forks form a tree on the noise nodes.
- **T5.** Such trees lift injectively to admissible subtrees of the regular tree of the twelve edge types, whose weight sum with `t` per arrow and `s` per fork obeys a three-term recursion; an exact rational super-solution bounds it below `391/100` at `(t, s) = (91/1000, 1000/107653)`. (The `66103` trees with at most four edges are counted exactly against the recursion's coefficients.)
- **T6.** Hence `P(η_x = 1) ≤ (391/100)·ε` uniformly in `x` for every `ε ≤ ε₀ := 7/10⁶`: the density of ones is at most **`3/10⁵`**.
- **T7.** With block 12's domination re-proved (`ε(p,q,r) = max` of three exact deviations, strictly decreasing in `p`), the formation law started from a constant plane keeps every later site within `3/10⁵` of that value whenever `ε(p,q,r) ≤ ε₀` — at `(p, 1, 2)` for every integer `p ≥ 285718` — and its level automaton has **at least six distinct translation-invariant invariant laws**, one per value. The formation reading orders in three dimensions.

In plain words: records forming one after another, each copying the majority of the three that came before it, can hold a memory forever once the rule's preference is strong enough; the proof counts the ways a dissenting record can be explained by noise and shows there are too few of them. The constant is the route's (simulation puts the automaton's true threshold near `0.05`), and that is stated, not hidden.

**Sharpened after opening (same day).** The first version of this PR counted trees by a depth-first word with a mark per node (`2·96^k` trees with `k` edges), which gave `ε₀ = 1/(2·96⁴) ≈ 5.9·10⁻⁹` and `p ≥ 339738628`. Reading the marks off the tree (a node is a noise site iff it has no arrow to a predecessor), using that each node has at most one such arrow, weighting arrows and forks separately under `forks = n − 1`, `arrows ≤ 3(n − 1)`, and bounding the lifted count by an exact super-solution of its generating-function recursion raised the threshold by a factor `1189`. A candidate lemma on the way — that every arrow points away from the root — was refuted by the executed trees (`538` of `4308` reach a kept pole from below through its own arrow) and dropped; the count keeps both directions. At `ε = 7·10⁻⁶` a `300×300` simulation gives a density of ones `8.3·10⁻⁶` against the bound `3·10⁻⁵`, so the bound is now within a factor four of what the automaton does there. The remaining slack of the route is the arrow budget (`3` per fork proved, `12/7` observed) and the union bound itself, not the count (the recursion's domain ends near `7.4·10⁻⁶`).

## Review surface

- The construction is executed, not only proved: on every noise configuration of the depth-2 backward cone (`1024`), on every depth-3 configuration with at most four noise sites (`2321` explained), on random cones of depth `3`–`10` and on line and triangle seeds, the runner and controls check the tree property, the marks, the spanning identity at every refinement, and `edges ≤ 4(n − 1)` (largest ratio observed `2.875`).
- One finding, caught by the control before the contract and recorded: the first draft let two forks sharing a point be adjacent in the cause graph, which skipped the cluster between them; the very first configuration then violated the bound. The tree must alternate clusters and forks.
- Refuting pass (`CHECKER_block25_findings.md`, disjoint machinery): the spanning lemma on 169 random abstract cluster–fork systems with an independent implementation; the noise map against floating-point enumeration at 200 random real weights; an independent graph implementation on 363 explained random cones; the automaton simulated on a `96×96` plane (stable to `ε = 0.05`, collapse at `0.08`). Sharpening refuter: the recursion iterated in floating point (fixed point below the certificate; divergence at `1.06 s`); the crude binomial split as a second valid bound (`44` times weaker); the Peierls partial sum over the enumerated family trees; an independent recount of the upward arrows; the simulation at `ε₀`.
- Census: 13 mutations (`certificate_wrong` replaces the first count's `series_constant_wrong`), each failing in exactly its declared family. Gates: pipeline ×2, `audit_lint --strict` OK, vocabulary lint clean on the block's files, `git diff --check` clean; only the intended manifest is included from generated audit data.

## Not claimed

The location of the true threshold; the formation law's phase between block 08's uniqueness region and this one; other menus or orders; any reading, rule or coupling as physical.

## Verification

```bash
python3 scripts/admissibility_rule_three_dimensional_formation_law_ordered_phase_stability_noisy_level_automaton_2026_09_16.py
python3 scripts/admissibility_rule_three_dimensional_formation_law_ordered_phase_stability_noisy_level_automaton_2026_09_16.py --list-mutations
```

Expected final line: `TOTAL: PASS=22 FAIL=0`.


JONbridgerapps:
Reviewed positive source salvage landed on main at 012c276a64c80cf5256d68fa7dbb06939586153b. Full disposition and reason: https://github.com/jonathonreilly/qubit-lattice-axiom-framework/blob/ai/execution/backlog_evidence/review-drain-20260915/drain8168-landing-report.md

# PR8168 — bounded partial salvage landed

The complete scoped T0–T7 formation-law proof and source-bound evidence landed on main at `012c276a64c80cf5256d68fa7dbb06939586153b` in train82. All 27 original path/mode/blob/hash identities remain recoverable, including historical failed controls and optimizer evidence. The sufficient threshold theorem retains its supplied hypotheses; finite simulations do not determine the true infinite-volume threshold. Current-main static-phase claims remain conditional.

The original reviewer confirmed all 32 source/evidence paths, then checked new-main interactions without recapturing unchanged evidence. The single stdout-only capture passed 22 checks in 2.190586 seconds, with sampled process-tree RSS 189,022,208 bytes within 900 seconds/512 MiB. It writes no primary JSON. Thirteen retained mutation definitions are not claimed as fresh successful executions.

One full pipeline, strict lint and changed-evidence pass succeeded: one checked claim, zero evidence/control failures. All 57 bound inputs matched. Generated audit outputs were preserved as validation history and stripped before landing; topology acknowledgment remains. No auditor or scientific audit verdict ran or was applied.

## Closure and recovery

This is partial salvage. Broader negative certification and unsupported threshold conclusions remain deferred or unpromoted under the detailed original dispositions. **Preserve the original branch** at `c3bbf3758eb59b1d6c997389ae35738ce986dcbc` as recovery for those conclusions; do not delete it.

Exact dispositions and provenance are in `drain8168-final-v1-review.json`, `drain8168-unit-final.json`, `drain8168-train81-addendum.json`, `train82-collected-units.json` and `train82-landing.json` under `ai/execution:backlog_evidence/review-drain-20260915/`.

PR #8156, opened 2026-09-15
[physics-loop] admissibility-induced-law block22 — bounded_theorem — bounded-support: block 19's ordering threshold sharpened to beta > 76/100 — 3G(0) pinned between 75/100 and 76/100 by exact return counts of the cubic walk

JONbridgerapps:
## What this PR is

Block 22 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `be5e5e07f6`): its parent is the landed possibility-covariance note of 2026-09-14 (through the placement only). It opens for review one bounded-theorem note, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-22 records. Supervisor-run throughout (owner directive 2026-09-15: don't stop; assess the next lane at each conclusion; no subagents).

- Claim type: **bounded_theorem**. Surface status: **bounded-support** (V1–V3 proved and executed exactly; V4 a placement corollary conditional on blocks 19 and 21 (open PRs) for their halves; two standard facts imported at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_UNSOLDERED_SPHERE_STATIC_LAW_STRONG_COUPLING_THRESHOLD_SHARPENED_RETURN_SUM_OF_THE_CUBIC_WALK_THREE_G_ZERO_BELOW_SEVENTY_SIX_HUNDREDTHS_BOUNDED_THEOREM_NOTE_2026-09-15.md`. Runner: `scripts/admissibility_rule_unsoldered_sphere_static_law_strong_coupling_threshold_sharpened_return_sum_2026_09_15.py` (20 checks, 13 mutations, exact, about twenty seconds).

## Why this block

Blocks 19 and 21 (PRs #8153, #8155) bracket the coupling window of the gravity node's kernel on `Z³`: the unsoldered sphere static law orders and carries the Green-function channel for `β > 3G(0)`, and has no channel for `β < √3/6 ≈ 0.289`. Block 19 bounded `G(0)` crudely, giving `β > 3√3π/8 ≈ 2.04`, so the window was a factor seven wide. This block pins the constant.

## What it says

For the constant `3G(0) = 3(2π)^{−3}∫ dk/E(k)` of block 19's infrared route:

[...]

PR #8507, opened 2026-09-20
ail36: if records re-form at every tick — the static law is the stationary law, the light-cone variant is a bilayer ferromagnet with long-range order, and a persistent source sits in the lattice Green function (conditional on a clause the axioms exclude)

JONbridgerapps:
## What this is

Block 36 of the admissibility induced-law campaign. A hand-off PR for the owner's review loop. Independent of the other open campaign PRs; base `main`. **Do not merge as a claim of the framework: the note is conditional on a clause the axioms memo excludes.**

## The point

Under the axioms as written every site is recorded once ("A site never carries more than one record; records are permanent."). Blocks 13, 26 and 35 showed that a continuous menu then loses its memory and that the gravity node's Green-function kernel does not appear. This block computes what the opposite clause would give: a record at every site at every tick, each drawn from the covariant rule given the previous tick's records on a symmetric stencil.

- **T1.** With the axioms' own nearest-neighbour rule, re-recording has the comparator as its stationary law: asynchronous re-recording is reversible for the static law of block 01; synchronous re-recording splits into two alternating chains whose space-time checkerboard carries the static law. Blocks 17 to 23 and 29 then describe the equilibrium of a re-recorded lattice.
- **T2.** With the site's own previous record in the past (a light-cone stencil) the chain is reversible for the product of the rule's normalizers, and two consecutive ticks carry the pair law on a doubled graph, which for even side is the ordinary bilayer: two stacked cubic lattices with a rung at every site.
- **T3.** Long-range order for the sphere menu: the magnetization bound `1 - (3/(2 beta))(G_L + H_L)` from the bilayer's reflections (bond planes and the layer swap), the domination of the twisted partition function, the infrared bound with the two bands `E` and `14 - E`, and the sum rule. Positive on large even tori for `beta > (3/2)(I_0 + I_2) < 0.5914` (the number uses block 22's bracket, open PR #8156, restated).
- **T4.** In the linearized law the mean around a persistent source is seven times the lattice Green function: a one-over-distance potential, no drift, full cubic symmetry.
- **Executed, not claimed.** No plateau at `beta = 0.5`, 0.41 at 0.6, 0.76 at `beta = 1` for sides 16 to 48; the potential around a persistent field source at 0.96 to 0.99 of the linear prediction and symmetric; an independent sampler of the equilibrium laws agrees with both chains to 0.001.

## Where the inputs came from

The derivation attempts and cross-family referee reports on the `ai/probes` branch (light-cone formation, long-range order, re-recording, the axioms map). Every step used is re-proved in the note; the six-neighbour statement (T1) and the bilayer map are the supervisor's.

## Checks

- Runner: `TOTAL: PASS=24 FAIL=0`, exact arithmetic; the infrared bound is verified exactly on a small two-valued instance.
- Mutation census: 9 of 9 fail in their own family.
- Refuting pass: an acceptance sampler of the equilibrium laws, disjoint from the chains' heat bath.

## Owner decision this block prices

Whether records re-form. The one-line contrast: as written, every place is written once and each new record reads the neighbours already written; under the clause, every place is written again at every tick and each new record reads its neighbourhood's records from the tick before.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `0b4d93c55133d72002408e095399cb5c13ce56ee` in 5efa36e7c357ae2a62ee586a5407f90982f6ded9 as part of the combined ten-PR kernel and formation batch.

- Declare supplied linear comparator; derive sphere mean derivative A(n beta)/n separately from aligned noise variance. Historical simulations do not prove limits or physical laws.
- Positive site selection and bounded sphere or finite alphabet minorization; factorized initial data for independence; correct bilayer-layer halves; project mean and exclude staggered pole. Import only scalar bound from current cubic-walk parent.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block36-re-recording-stationary-law-bilayer-long-range-order-green-function-response-20260920` at `0b4d93c55133d72002408e095399cb5c13ce56ee`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8512, opened 2026-09-20
ail38: how records form — the static law is outside the hull of every adapted formation law on a window with a cycle, and under a fixed predecessor structure every rate, order and unit clause gives one law

JONbridgerapps:
## What this is

Block 38 of the admissibility induced-law campaign. Hand-off PR for the owner's review loop; independent of the other open campaign PRs; base `main`.

## The result

The axioms say records form and leave open how: order, rate, unit. Blocks 14, 15 and 16 recorded witnesses that these choices change the law; block 16 left value-dependent orders open beyond the plaquette.

- **T1.** Let the next site be chosen by any rule that looks at everything recorded so far (value-blind orders, clocks with value-dependent rates, deterministic strategies). On any window whose graph has a cycle every such scheme gives each constant pattern strictly less weight than the static law does, by the factor `Z/D_min < 1`. The static law is outside the closed convex hull of all adapted formation laws, separated by the indicator that all records agree, with exact margin `6 p^|E| (1/Z - 1/D_min)`. On a window without a cycle `Z = D_min`, as block 14 found for trees.
- **T2.** If each site has a fixed set of predecessors, forms only after them and reads exactly them, then the order among ready sites, any clock rates and forming an antichain jointly all give the same law, the product of the rule's kernels. On level-ordered `Z^3` a level is an antichain.
- **What is left of "how".** The clause witnesses of blocks 14 and 15 are all cases where the clause changes which neighbours a forming record reads. The open question is one question: which neighbours are already written when a site is written.

Exact values at (3,1,2): plaquette `Z = 20784 < D_min = 22464`; 2x3 rectangle `6000000 < 7008768`; cube `6982520832 < 10933678080`; path and star `Z = D_min`. Control: undirected path of 3, total variation `1/72`.

## Provenance

Both statements came back from the probes' derivation rounds on `ai/probes`, refereed across model families. I re-derived both arguments and recomputed every number with my own elimination code before writing.

## Checks

- Runner `TOTAL: PASS=15 FAIL=0`, exact arithmetic; seven mutations, each failing in its own family.
- Refuting pass: brute enumeration of `Z`, a linear program for the value-blind case (infeasible), Monte Carlo of an adversarial adapted scheme (`0.0201 +- 0.0003` against the bound `0.02163` and the static `0.02338`) and of a gated scheme.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `94a63b77f61a88f787f3a03b939bc79986231d15` in 5efa36e7c357ae2a62ee586a5407f90982f6ded9 as part of the combined ten-PR kernel and formation batch.

- Self-contained counterexample, causal ancestry excludes negative coordinates; fixed read sets and explicitly factorized joint kernel; zero coordinate-sum displacement; exact table/control assertions.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block38-static-law-outside-adapted-hull-and-causal-structure-clause-independence-20260920` at `94a63b77f61a88f787f3a03b939bc79986231d15`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8530, opened 2026-09-20
ail39: records that move — pair-weight transit has the static law as equilibrium, the binding scale is a constant the rule cannot see, reflection positivity bounds it from below, and where records clump and jam

JONbridgerapps:
## What this is

Block 39 of the admissibility induced-law campaign. Hand-off PR for the owner's review loop; independent of the other open campaign PRs; base `main`. **Conditional on a supplied motion clause and a supplied binding scale. Neither is in the axioms memo, and nothing is adopted.**

## The reading

Owner, 2026-09-20: records move. A site holds at most one record at a time; an empty site can form a record by the rule; a record can transit to a neighbouring empty site when the neighbourhood allows. I read the complete axioms memo first: "A site never carries more than one record; records are permanent." forbids two records on a site at once and the destruction of a record, not a record leaving. Motion is not in the memo, which lists update laws and persistence dynamics among its open gates.

## Results (exact, finite windows)

- **T1.** A record choosing between its position and an empty neighbouring position in proportion to the pair weights it would have there is in detailed balance with the static law on the occupied set. The campaign's comparator is the equilibrium of moving records.
- **T2.** Multiply every pair weight by a scale `c`: the axioms' normalized distribution does not change; the motion does, exactly when a move changes the number of record-record bonds. A record with `k` agreeing neighbours escapes to an isolated position with probability `1/(1 + (cp)^k)`.
- **T3.** If a move is instead accepted with the rule's normalized probability at the destination, the chain is blind to the scale and reversible for no law (a four-move cycle with products `1/2592` against `1/2376`).
- **T4.** Formation at an empty site at rate `z Z_x`, content drawn from the rule, is the creation half of the same equilibrium, and the only such rate. The neutral scale `6/(p+q+4r)`: an empty neighbour weighs what a record of random content weighs on average.
- **T5.** With the empty state counted as a state of a site, the law with vacancies is reflection positive through bond planes exactly when the scale is at least the neutral scale (and `p >= q`, `p + q >= 2r`). A principle the framework already uses bounds the constant from below.

## Executed, not claimed

Cubic lattice, density 0.3, line `(p,1,2)`: records stay spread out below an onset and gather into aligned clusters above it: near `p = 10` at the neutral scale, between 5 and 6 at scale 1. The fraction of accepted moves falls to a few percent at strong preference with seven tenths of the sites empty. The simulator reproduces the literature onset for content-less records (between 2.43 and 2.7 against 2.427) and agrees with a site-by-site reference.

## Checks

- Runner `TOTAL: PASS=19 FAIL=0`, exact arithmetic; nine mutations, each failing in its own family.
- Refuting pass: exact stationary laws by linear solve, disjoint from the runner's move-by-move checks.
- Pipeline twice on the final content (both `PASS ... graph_delta=acknowledged`), changed-evidence check, audit lint strict, diff check, vocabulary lint.

## Owner decisions this block sets up

The motion clause; and the scale: a registered primitive, or fixed at the neutral value by reflection positivity or by the no-privilege principle.


JONbridgerapps:
Corrigendum pushed (31e5d0300e). The rare-formation remark after T4 holds conditionally: given the records present (their number and contents), the arrangement just before a formation is close to the static law for those records.

It does not make the growing population a sample of the static ensemble. Which contents are present, and how many, is set by the formation history: the pre-birth arrangement weighed by the insertion rates. #8545 derives the formation-event law and gives an exact six-site counterexample at the third birth.

T1 (motion's detailed balance) and T4's rate identity stand. The change is text only; the runner output is unchanged and the cache is re-pinned.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `31e5d0300e07357ef4de73c210288fe6615d6f19` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Scope the explicit nonreversible example, positive fugacity and spectral inequalities; unit-removal birth law; incorporate refreshed PR8530 conditional-population correction and distinguish event-rate sampling; separate simple cycles from overlapping 2-core contractions.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block39-records-that-move-pair-weight-transit-binding-scale-clumping-20260920` at `31e5d0300e07357ef4de73c210288fe6615d6f19`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8546, opened 2026-09-20
ail40: the binding scale pinned at the neutral value — the pair weight is the rule's likelihood ratio, records do not bind without a cycle, and all binding is agreement around loops

JONbridgerapps:
## What this is

Block 40 of the admissibility induced-law campaign. Hand-off PR for the owner's review loop; independent of the other open campaign PRs; base `main`. **The value is a proposal under block 39's moving-records reading. Nothing is adopted and no primitive is registered.**

## Context

Block 39 (PR #8530) found that once records move, the overall scale `c` of the pair weights, which the axioms' normalized distribution cannot see, becomes a binding constant, with a floor from reflection positivity at `c_0 = 6/(p+q+4r)`. The owner asked whether the constant can be derived, and accepted the recommendation to proceed with the most promising value. A length cannot give a pure number, so the value has to come from the rule.

## Results (exact, finite windows)

- **T1. Four characterizations of one value.** At `c_0`, and only there: the pair weight is the rule's one-neighbour probability divided by chance (a likelihood ratio, so an empty bond, of weight 1, carries no information); an empty site acts on each neighbour exactly as a record of uniformly random content; an empty site next to one record forms at the empty-space rate; and the law with vacancies is on the reflection-positivity floor. The formation rate becomes a likelihood ratio of the neighbourhood's coherence.
- **T2. No binding without a cycle.** At `c_0`, on any window without a cycle, the occupancies are exactly independent with density `6z/(1+6z)`, and the contents follow the rule's tree law.
- **T3. All binding is agreement around loops.** A fully occupied cycle of length `n` carries `1 + 3 l_1^n + 2 l_2^n`, from the two non-trivial eigenvalues of the rule's one-neighbour probability: `433/432` on a plaquette at (3,1,2), `1.29` at (12,1,2).
- **T4. What any other value adds.** On a window without a cycle the occupancy is the content-less lattice gas with bond activity `c/c_0`: a glue, or a repulsion, the rule does not account for.

## Executed, not claimed

With the scale pinned, the onset of alignment on the line `(p,1,2)` depends on the density alone: above `p = 16` at density 0.1, between 8 and 12 at 0.3, between 6 and 8 at 0.5, below 6 at 0.7, heading for the full-lattice value 3.6 to 3.7 of block 28.

## What the note does not say

That weights are likelihood ratios is a reading of three sentences of the axioms, not a derivation from them. Pinning one constant settles one constant. No strength of any long-range potential is computed: the framework has no definition of a source yet.

## Checks

- Runner `TOTAL: PASS=16 FAIL=0`, exact arithmetic; eight mutations, each failing in its own family.
- Refuting pass: an independent enumeration of the occupancy laws.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `361166e7eaccd7d80cd1e3d4cb94b4ad7fcdbc21` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Scope the explicit nonreversible example, positive fugacity and spectral inequalities; unit-removal birth law; incorporate refreshed PR8530 conditional-population correction and distinguish event-rate sampling; separate simple cycles from overlapping 2-core contractions.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block40-binding-scale-pinned-likelihood-ratio-no-glue-loops-20260920` at `361166e7eaccd7d80cd1e3d4cb94b4ad7fcdbc21`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8547, opened 2026-09-20
ail41: what a source is when records move (mass = record count; no action across empty sites; the tilt channel's charge is signed; the transit Laplacian is sourced by production)

JONbridgerapps:
## ail41: what a source is when records move

Hand-off PR (independent vs `main`; never merged by the author). Owner, 2026-09-20: "ok lets define what a source is and go from there".

**What it says (plain).** The gravity lane on `main` already lists what a source has to be: a local, positive, additive, covariant density entering a field equation with operator `−Δ_lat`. Applied to records that move:

1. **Mass is the record count.** The symmetry acting on contents carries every content to every other, so such a density cannot tell contents apart (T1).
2. **Nothing acts across empty sites.** Two sets of records separated by empty sites weigh exactly the product of their weights, at any binding scale; their contents are independent (T2).
3. **Through a medium the field of a mass is of short range.** Holding a record raises the mean occupancy nearby by the pair correlation; exact on the ladder (about tenfold per step), exactly zero on windows without a cycle at the neutral scale; on the cubic lattice the executed range is about one lattice step or less away from the clumping onset (T3 + controls).
4. **The long-range channel carries the wrong charge.** By the symmetry of the declared isotropic rule, the content field around a record is a number times its content vector: odd, zero on average, and zero for a record of unread content. The mass has no charge in the tilt channel. In the quadratic model two held tilts interact at one-over-distance, like tilts attracting, opposite tilts repelling; the two-tilt field is linear in the tilts but is not the sum of the one-tilt fields (T4). For the six-axis menu the campaign has found no kernel of long range at all. The lemma fails for direction-dependent rules (refuter W4), listed as a route.
5. **The one Laplacian the mass does enter is transit.** With one record per site and symmetric hops the mean occupancy obeys the lattice heat equation exactly, so steady excess production carries a Green-function halo: the gravity lane's operator with production, not presence, as the source. At the neutral scale the formation rate next to records of unrelated contents equals the void's on average, next to a single record or a flat face it is exactly the void's, and next to two or more agreeing records it is larger (`13/12, 5/4` at (3,1,2); `46/21, 2348/343` at (12,1,2)) (T5).

**Not claimed.** Any long-range attraction between masses, any force law, any identification of the production field with a gravitational potential, the infinite lattice, direction-dependent rules, any adoption. The note lists where an attraction could still come from (amplitude layer; transit field under net production; clumping onset; common tilt; direction-dependent rules).

**Evidence.**
- Runner: `TOTAL: PASS=19 FAIL=0`, exact arithmetic, self-scanned; 13 mutations, each failing in its own family only.
- Refuting pass with disjoint machinery (site-level enumeration on general graphs): seven witnesses pass, including the `2×4` periodic ladder by enumeration of all `7⁸` site states against the transfer matrix (exact equality).
- Controls: screening simulator and outputs (side 16, one seed) under the campaign pack's `specs/`.
- Audit pipeline: two passes (`graph_delta=acknowledged`), changed-evidence check and strict lint clean.

**Status.** bounded-support; conditional on block 39's moving-records reading (PR #8530) and, where used, block 40's neutral scale (PR #8546); nothing adopted or registered; unaudited.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `93efe618fa94d4607f762a698b7cfa2c8f00a000` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Full quadratic pinning cost and normalization distinction; formal response with actual source saturation; mixed-content formation counterexample; positive screened domain and zero-charge massless periodic boundary case; no arbitrary body-size or exact decay-length claim.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block41-what-a-source-is-when-records-move-20260920` at `93efe618fa94d4607f762a698b7cfa2c8f00a000`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8548, opened 2026-09-20
ail42: records interacting through the odds of unformed sites (contents interact, masses do not; field equation massless on 5p = 7q + 4r; records are boundary values, not additive sources)

JONbridgerapps:
## ail42: records interacting through the odds of unformed sites

Hand-off PR (independent vs `main`; never merged by the author). Owner, 2026-09-20: "why couldnt records interact via the neighborhood probabilities?", then "well keep going until we have something solid or need to probe a different direction".

**Short answer.** They can. Block 41's "no action across empty sites" (PR #8547) is a theorem about a law in which an empty site has no state. The axioms' sentence on the probability distribution speaks of every site and does not say a condition must be a record. This note gives the unformed site its odds, lets them be a condition for its neighbours, and works out both forms of that reading. It is a supplied reading; nothing is adopted.

1. **Summed form** (an unformed site is one unknown content): two records across a gap of `n` bonds weigh `1 + 3λ₁ⁿ + 2λ₂ⁿ` (equal contents), `1 − 3λ₁ⁿ + 2λ₂ⁿ` (opposite), `1 − λ₂ⁿ` (orthogonal) against chance; the average over contents is exactly `1`, and every arrangement of `k` records weighs `z^k Z`. Contents interact across empty sites; masses do not, at any distance (T1).
2. **Self-consistent form** (each site's odds from its neighbours' odds): exact derivative `K₁ − 1/6` per neighbour, spectrum `λ_i(6 − E(k))`, field equation `(−Δ + m²)v = source/λ₁` with `m² = (1 − 6λ₁)/λ₁`; **no mass term exactly on `5p = 7q + 4r`, where the declared triple `(3,1,2)` lies**; beyond it the uniform field is unstable (T2).
3. **Proved regime**: an elementary contraction of the ratio metric (`R(ωx, ωy) ≤ R/((1−κ) + κR)`, proved at scope and sharp to `999/1000`) gives existence, uniqueness and geometric decay of a record's influence for `6(1−κ) < 1` (T3).
4. **What crosses is content**: the map is covariant under the rotation of all contents; the six-outcome odds have no content-blind component; a content-blind weight sees the leans at second order, `1 + 3λ₁² Σ m_y·m_{y'}` (T4).
5. **"No record" as a seventh possibility** gives mass a first-order channel of strength `ρ(1−ρ)(g−1)/(1+ρ(g−1))`, `g = c/c₀`: **exactly zero at the neutral scale**, screened with glue, massless only on its own surface (T5).
6. **Records are boundary values, not additive charges**: the linear field of a body of agreeing records is the reaching probability of a killed walk; interior records carry the charge `1 − 6λ₁` (zero on the massless surface); the body's strength is its capacity, which is subadditive. Executed near the surface: cubes of `1, 8, 27, 64` records act like `1, 2.2, 3.2, 4.0` single records (T6).

**Bottom line.** The record layer does have a field across empty space with the lattice Laplacian as its operator. Three things separate it from a field of the kind the gravity lane asks for: its first-order charge is content, not mass; it has a mass term except on tuned surfaces; and its sources do not add.

**Evidence.** Runner `TOTAL: PASS=20 FAIL=0`, exact arithmetic, 14 mutations each failing in its own family only. Refuting pass with disjoint machinery (enumeration, symbolic differentiation, hill-climbing): 8 witnesses pass; two of them upgrade the runner's point checks to identities in symbolic `(p, q, r)` and `(ρ, g, p, q, r)`. Floating-point control for the nonlinear fields (not claimed): four-digit agreement with the screened Green function at range `0.77`; `r·v(r)` flat on the massless surface with iteration counts growing like the square of the side. Two pipeline passes (`graph_delta=acknowledged`), changed-evidence check and strict lint clean.

**Not claimed.** Anything about the nonlinear field on or beyond the massless surfaces, the infinite lattice, a force law, a gravitational reading, any adoption. The two forms are different objects; neither is claimed to approximate the other.

**Status.** bounded-support; conditional on the supplied reading (and, for T5, on blocks 39/40); unaudited.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `cf6aae77d7ffdda9463f71469ede2744c2e6b7b2` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Full quadratic pinning cost and normalization distinction; formal response with actual source saturation; mixed-content formation counterexample; positive screened domain and zero-charge massless periodic boundary case; no arbitrary body-size or exact decay-length claim.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block42-interaction-through-the-odds-of-unformed-sites-20260920` at `cf6aae77d7ffdda9463f71469ede2744c2e6b7b2`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8549, opened 2026-09-20
ail43: the fields of the record layer arrive by diffusion (settling over a distance costs the square of that distance in ticks; a carrier of a long-range field needs a speed)

JONbridgerapps:
## ail43: the fields of the record layer arrive by diffusion

Hand-off PR (independent vs `main`; never merged by the author). Owner, 2026-09-20: "keep going until we have something solid or need to probe a different direction".

**What it says.** Blocks 41 and 42 (PRs #8547, #8548) found two fields in the record layer with the lattice Laplacian as operator: the halo of wandering records around a place that produces them, and the odds of unformed sites around a record. Block 42 listed three differences from a field of the kind the gravity lane asks for (content charge, mass term off tuned surfaces, non-additive sources). This note adds a fourth that applies to both and would survive repairing the other three: **both are set up by a random walk, so a field over a distance `R` costs of the order of `R²` ticks.**

- **T1** on a torus of side `L` the slowest mode keeps at least half its deviation for `n ≤ 1/(2hE_min)`, hence for at least `L²/(8π²h)` ticks (`(1−x)ⁿ ≥ 1−nx`, `E_min ≤ (2π/L)²`).
- **T2** on the infinite lattice the `s`-tick kernel has mean-square reach exactly `6hs`; the share beyond `R` is at most `6hs/R²`; the response to a switched-on source is the sum of the kernels.
- **T3** by exact differentiation of block 42's odds map, one linearized iteration is `6λ₁` times a step of the simple walk (lean `1/6` at (3,1,2), `3/23` at (5,2,4)); block 42's control shows the same law in the full nonlinear map (iterations `3484, 6693, 10875` at sides `15, 21, 27`).
- **T4** a mass term buys speed only by giving up range: a range `1/m` still costs of the order of `(range)²/h` ticks.
- **Corollary (units)**, with the lattice unit `a` left free as the registered scale reference requires: `a ≥ R²/(2cT)`. Comparators only: `0.09 mm` for the Earth–Sun distance and the age of the universe.

**Context read on `main` for this block.** The gravity lane's weak-field packet is a supplied quadratic action (its cited note says in its own words that the field equation "is supplied to the loop … it does not derive the operator from the framework axioms"); the record-pair source note names three missing pieces (a Record-local source, a local conservation/attachment law with cadence and zero mode, an absolute unit). The carrier of a long-range field between moving bodies has to have a speed; neither record-layer field does.

**Evidence.** Runner `TOTAL: PASS=13 FAIL=0`, exact arithmetic, 8 mutations each failing in its own family only; refuting pass by whole-path enumeration and explicit torus relaxation (4 witnesses); two pipeline passes (`graph_delta=acknowledged`), lints clean.

**Not claimed.** Anything about carriers that are not reversible walks (the amplitude layer; records that keep their direction of travel), any value of the lattice unit, any gravitational statement, any adoption.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `a6af3f2d571cc23aed9674ab44f999fc55de94a5` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Separate discrete model and continuous generator, low-mode lower bound with alternating-mode exception; corrected mass denominator; body-inclusive reflection balance; conditional local-equilibrium sound; degenerate sphere redraw classes.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block43-record-layer-fields-arrive-by-diffusion-20260920` at `a6af3f2d571cc23aed9674ab44f999fc55de94a5`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8553, opened 2026-09-21
ail45: the wind law (records are conserved, so a capturing body's inflow is inverse-square; wind confirmed parameter-free; F = K Q1 Q2/r^2 symmetric; the inviscid sign-change prediction refuted and re-scoped)

JONbridgerapps:
> **Correction, 2026-09-21 (author; see the comments below and #8563, #8558).** (1) T5's reason for the free-streaming value (unbiased sampling) is wrong; the value is right for a wind in local equilibrium. (2) T3's spherical symmetry is an assumption and fails for the executed flows: the wind of a capturing body depends on the lattice direction (`+13 %` along the axes, `−13 %` along the body diagonals at this note's parameters). The force law here was executed **along a lattice axis only**; off the axes `K/K₀ = 0.735 ± 0.046` along a body diagonal against `0.907 ± 0.042` along an axis. The shell-averaged wind profile is unaffected.

## ail45: the wind law

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550); nothing adopted.

**Why capturing bodies attract inverse-square (block 44, executed).** Records are never destroyed, so whatever a body captures has to cross every surface around it.

**Exact.**
- **T1, T2** number and momentum are conserved locally, with explicit observables: the current `j(x→y) = [record at x pointing to y, y empty] − [the reverse]`, and a momentum flux (a moving record carries `e_d`; an exchange passes `e_d` forward and `e_{d'}` back; scattering passes the mean change across its bond). Checked at every site of all 12636 two-record configurations of the `3³` torus. Hence a body capturing `Q` per tick draws `Q` through every surface around it, and the force on a body is the momentum flux through any enclosing surface.
- **T4** the second-order momentum flux of product states: `√3 Π_ij = (ρ/3)δ_ij + A g_i g_j + B g² δ_ij`, `A = (3/5 − ρ)/ρ`, `B = −1/(5ρ)` (sphere menu; from the uniform moments `1/3, 1/5, 1/15`). One record per site lowers `A` by one and makes it negative above `ρ = 3/5`. The six-axis menu has **no** `g_i g_j` term: it is not isotropic at second order.
[...]

PR #8554, opened 2026-09-21
ail46: the potential of the wind solves the lattice field equation with the capture rates as sources; the inertial gas has radiation's equation of state (a correspondence of form with the weak-field packet, no gravitational claim)

JONbridgerapps:
> **Correction, 2026-09-21 (author; see the comment below and #8563).** The zero circulation of the stationary wind is an assumption of this note and fails for the executed flows: hopping along lattice axes gives the gas a viscous term of cubic symmetry, the potential inflow does not solve the creeping equations, and the executed wind depends on the lattice direction. T1 (uniqueness of a circulation-free current) and T3 (the conserved tensor) stand; the identification of the executed wind with the Green-function gradient does not. The superposition check in the comments is an average over directions and does not test isotropy.

## ail46: the potential of the wind, and the conserved tensor of the inertial gas

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550) and block 45's conservation laws (PR #8553); nothing adopted; no gravitational claim.

**Exact.**
- **T1** on a torus the bond fields with zero divergence and zero circulation are exactly the three constant fields (exact rank: 189 of 192 on the `4³` torus, 78 of 81 on the `3³` torus). So a mean current whose divergence is fixed by the capture rates and whose circulation vanishes is unique up to constants, and it is the lattice gradient of `Φ = −G * (Q − mean Q)`: **the potential of the wind solves the lattice field equation `−Δ_lat Φ = −(Q − mean Q)` with the capture rates as its sources.** On the infinite lattice, the same with `G₀` for fields that vanish at infinity.
- **T2** the wind that builds up from rest under the linearized conservation equations is a lattice gradient: zero circulation.
- **T3** the conserved tensor (density, momentum density, mass current, momentum flux; conserved locally and exactly by block 45): in equilibrium the pressure is exactly the density times the speed over three **at every density**, exclusion or not — the equation of state of radiation — and the mass current is `(1 − ρ)` times the momentum density, so the tensor is symmetric only in the dilute limit.

**Correspondence of form (not a claim).** The weak-field packet on `main` posits an action with `φ = G₀ P₀ ρ`, a local positive additive source and a bilinear test response. Within the inertial clause the same three statements hold for the wind, each for a reason: the Laplacian from the permanence of records, the source the capture rate (`0.24 ×` the record count for bodies transparent to the gas — executed in the campaign controls), the bilinear response the momentum of captured records (block 45, executed), `P₀` the uniform part of the density.

**What it costs** (carried from blocks 44–45): encounters must re-draw contents (executed in the campaign: without that there is no sound); matter must capture records, which makes it grow and drags it when it moves; and the capture rate is the record count only for transparent bodies.

**Evidence.** Runner `TOTAL: PASS=11 FAIL=0`, exact arithmetic, 7 mutations each failing in its own family only; refuting pass with disjoint machinery (rational rank by elimination over fractions; three sinks on the `6³` torus; pressure by bond-event enumeration): 3 witnesses pass; two pipeline passes, lints clean.


JONbridgerapps:
**Executed after the note was frozen (supervisor, campaign control; not claimed):** T1 says that a circulation-free mean current whose divergence is fixed by the capture rates is the lattice gradient of `Φ = −G * Q`, so the wind of several capturing bodies is the superposition of their potential flows. For one body that is automatic by symmetry; for two it is not. Two capturing balls of radius 3 at separation 16 (sphere menu, side 64, density 0.289, `γ = 1`, 1500 samples of the mean momentum field, `Q = 7.80, 7.69`): the best scalar fit of the measured field to `√3/(4π(1−ρ)) Σ_a Q_a (x_a − x)/|x − x_a|³` is `0.983, 0.991, 0.998` over the sites whose nearest body is at distance 6–9, 9–13, 13–18 (4099, 10540, 23018 sites), and the rms residual per component, `0.0083, 0.0082, 0.0082`, equals the sampling noise `0.0080`; on the midplane the component along the joining axis is pure noise (`0.0064`). No free parameter. The control script (`superposition.py`) is in the campaign scratch and is queued for `ai/probes`.

JONbridgerapps:
**Correction (author's own finding, block 51, https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/8563): the zero-circulation assumption fails for the executed flows.**

This note derives zero circulation only for the linearized evolution from rest and **assumes** it for the stationary wind ("block 45's executed wind profile is consistent with it to a few per cent" — but that profile is a shell average and cannot test it). Block 51 measures the wind by direction and finds it direction-dependent with scattering (axes / body diagonals `1.30 ± 0.02` at density `0.3`, `γ = 1`), and shows why: the streaming of records along lattice axes gives the gas a viscous term of cubic symmetry, `ν_lat(∇²g_i + ∂_i²g_i)`; acting on a potential flow it is not a gradient (curl `−105xy(x² − y²)/r⁹` for `1/r`), so the potential inflow does not solve the creeping equations, and the solution that does has circulation and is `f(direction)/r²` at every distance.

T1 (a circulation-free current with given divergence is the gradient of the Green-function potential) and T3 (the conserved tensor) stand. What falls is the identification of the executed stationary wind with that gradient. The superposition check posted above fitted one scalar coefficient per distance band and is an average over directions; it does not test isotropy.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `b9398e0fec91a0e238f4770aba99daf988746928` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Retain -g_i g_j, distinguish scaled stress, require spherical closure and first-harmonic balanced capture geometry. Finite bond Hodge identities only, rest tensor symmetry and units scoped; transparent-body force retained as conditional coefficient algebra.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block46-the-potential-of-the-wind-and-the-conserved-tensor-20260921` at `b9398e0fec91a0e238f4770aba99daf988746928`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8558, opened 2026-09-21
ail48: free capturing bodies are carried by the wind — capture law, anisotropic collisionless shadow, dilution law of motion and its window

JONbridgerapps:
> **Correction, 2026-09-21 (author; see the comment below and #8563).** One sentence is withdrawn: "The inverse-square law with one coefficient needs collisions", together with the comparison with "`9/4` in every direction for the collisional law". With scattering the wind of a capturing body is still direction-dependent (axes to body diagonals `1.30–1.35`); `9/4` is the value of block 45's spherically symmetric closure, not of the gas. The same control confirms this note's T3 at finite distance (`0.180, 0.143, 0.116` measured against `0.176, 0.147, 0.122` exact). T1, T2, T4, T5 are unaffected.

## ail48: free capturing bodies are carried by the wind

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550) and a supplied body clause; nothing adopted; no gravitational claim.

**What it settles.** Blocks 45 and 47 (PRs #8553, #8556) held the bodies fixed; the decision record (PR #8555) carried only an estimate of what happens when they are let go. This block gives the exact capture law, says where the single force coefficient is valid, and derives the law of motion of a free capturing body with the window in which a fall is accelerated.

**Exact.**
- **T1 capture law.** A capturing site takes up a record of content `s` in proportion to `|s|₁` (1 along an axis, `√3` along a body diagonal): captured records are a biased sample of the gas (exact witness).
- **T2 first-harmonic winds.** `⟨|s|₁⟩ = 3/2`, `⟨|s|₁ s_i s_j⟩ = δ_ij/2`, so in a wind with content law `(1 + 3u·s)/(4π)` every captured record brings `u` on average. **This corrects block 45's T5**, whose reason for the free-streaming value (unbiased sampling) was wrong; the value `K₀` is right for a gas in local equilibrium, so it is the collisional value. Bodies need equally many exposed faces along the three axes (a plate is a witness against).
[...]

PR #8556, opened 2026-09-21
ail47: transparent bodies — capture rate is the record count; force is the product of the record counts over distance squared

JONbridgerapps:
> **Scope correction, 2026-09-21 (author; see the comments below and #8558, #8563).** `G = K₀q₁²` is the coefficient of block 45's spherically symmetric, local-equilibrium closure. The pairs here were executed **along a lattice axis only**; the wind of a capturing body, and the executed force between capturing bodies, depend on the lattice direction (`K/K₀ = 0.907 ± 0.042` along an axis, `0.735 ± 0.046` along a body diagonal, for solid bodies at this note's density and scattering rate).

## ail47: transparent bodies — the force is the product of the record counts

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550) and block 45's wind law (PR #8553); nothing adopted; no gravitational claim.

**The gap it closes.** Block 41 (PR #8547) forced the record count as the only local additive invariant source. Block 45 executed the wind law `F = K Q₁Q₂/r²`, whose charge is the capture rate, and found that for dense bodies the capture rate follows the surface, not the record count. The two meet in the transparent limit.

**Exact (product-state expectations).**
- **T1** a capturing site takes up `q₁ = (√3/2)ρ` records per tick (sphere menu; `ρ` for the six axes).
- **T2** a body captures through its exposed faces, `faces × ρ/(4√3)`: separate sites add; compact bodies share faces (78, 174, 294, 486 faces for balls of radius 2–5).
- **T3** with block 45's free-streaming coefficient, two transparent bodies attract with `G N₁N₂/r²`, `G = K₀q₁² = 3√3 ρ/(16π(1 − ρ))`.

**Executed (not claimed; the prediction was computed first).**
- capture rate per site `0.236, 0.241, 0.234` for 3, 9, 15 sites scattered in a ball of radius 6 (kinetic `0.251`), falling to `0.135` at 180 sites as the ball becomes opaque; per exposed face `0.041–0.045` for solid balls (kinetic `0.042`);
[...]

PR #8559, opened 2026-09-21
ail49: the attraction is tied to growth — a body in balance draws no wind; pull follows net uptake; growth rate fixed by the force coefficient

JONbridgerapps:
## ail49: the attraction is tied to growth

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550) and a supplied emission clause; nothing adopted; no gravitational claim.

**The question.** In blocks 45–48 (PRs #8553, #8554, #8556, #8558) bodies attract because they capture records, and a body that captures grows. Block 48 left one route open: a body that gives back what it captures. This block works it.

**Exact.**
- **T1 no net uptake, no wind.** Records are conserved, so the flux of the mean current out of any region is minus the net uptake inside it. A body whose emissions equal its captures sends no flux through any surface around it: no inverse-square wind.
- **T2 independent records.** Emission through the faces with the cosine law is the time reverse of capture: emitted records are the same directed walks as the ones removed, and at the capture rate the surplus cancels block 48's multinomial deficit at every site — the gas around the body is uniform and no site feels a force, at any distance, in any direction. Uniform emission instead leaves a residue with no net number and the force coefficient `|r̂|₁(|r̂|₁ − 3/2)`: a push near the axes, a pull near the body diagonals, angular mean `1 + 4/π − 9/4 ≈ 0.023`.

**Closure (blocks 45, 47).**
- **T3 pull and push separate.** The wind a body makes follows its **net** uptake; the momentum a body takes up follows its **gross** capture (cosine-law emission carries `(2/3)n` per face, zero over a body). Pushes: `β₁` and `β₂` times `K₀Q₁Q₂/r²` with `β` the kept fractions. Action and reaction between two bodies differ; the gas carries the difference; two bodies in balance do not attract at this order.
- **T4 the exchange rate.** `q₁ = (8π/3)(1 − ρ)G`: a body's relative growth rate is `8π(1 − ρ)/3` times the coefficient of the pull it exerts, for any kept fraction. A mass steady over `T` ticks pulls with a coefficient below `3/(8π(1 − ρ)T)`.

**Executed (not claimed; prediction written while the control ran).** Two solid balls of radius 3 at separation 16, side 96, `ρ = 0.3`, `γ = 1`, two seeds per configuration; pushes over the reference `K₀Q₁Q₂/r²`:
- both capture: `0.92 ± 0.05`;
- one in balance (captures 7.24, emits 7.24 per tick), one capturing: the body in balance is pushed with `0.75 ± 0.06` and pulls the other with `0.10 ± 0.07`;
- both in balance: `0.05 ± 0.07`.
The push on the body in balance sits below the capturing pair's by `0.17 ± 0.08`. A second control (pinned body in a uniform wind) locates the cause: a dilute body of about 15 sites in balance takes up `0.85, 0.83, 0.80` (`± 0.03`) of `u` per captured record at `γ = 0, 1, 4`, against `0.97`, `0.89` when it only captures, while **one site** in balance takes up `0.97 ± 0.03` against `1.00 ± 0.02`: the shortfall is the body capturing its own emissions again, which bring back what they took and not the wind's content. (The supervisor's first guess — emitted records diluting the wind through collisions — was refuted by the `γ = 0` run; the gates were stopped and the note corrected before opening.)

**What it means for the lane.** Within these clauses a body attracts in proportion to the rate at which it grows, and a body that has stopped growing has stopped attracting. With blocks 41–48: the record layer supplies the source (the record count) and, on this evidence, not the carrier.

**Evidence.** Runner `TOTAL: PASS=14 FAIL=0`, exact arithmetic, 8 mutations each failing in its own family only; refuting pass with disjoint machinery (symbolic integration; an exact site-by-site transport solve; path enumeration; quadrature): 5 witnesses pass; two unproved sentences about the next-order force were withdrawn before the gates; two pipeline passes, lints clean.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `589280e3798ef2e39433fad5248c37cc89aed712` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Propagate cubic-viscosity correction, fixed-ray vector remainder and Hessian operator bound, separate growing-active-site idealization, explicit uniform-emission routing, monopole-only no-go and constant-parameter growth bound.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block49-the-attraction-is-tied-to-growth-20260921` at `589280e3798ef2e39433fad5248c37cc89aed712`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

JONbridgerapps:
**Explanation of the executed balanced-body push (2026-09-23, supervisor), from a refereed probes unit.**

`J:derive:recapture-share-of-a-body-in-balance:a1` (worker `w-macbookpro90c72-jbb54`, Claude Opus 5.5) worked out why a balanced body's push per capture falls below the capture-only value. Its referee `w-macbookpro90c72-j8897`, a Grok model, confirmed the result at `γ = 0` and small density.

**The result.**
- A balanced body recaptures a share `f` of its own emissions.
- For a reflection-symmetric ensemble, the exposed normals of a polycube sum to zero, so emissions carry no mean momentum. Hence `push = push_co (1 − f)`.
- An emission never returns to its own site or to a lattice neighbour of it, so a single site has `f = 0`.
- `f = 0.1337` for 15 sites uniform in a radius-4 ball, `0.1441` for the executed bodies weighted by capture, and `0.1086` for a solid radius-3 ball.
- Over the executed 40-tick window the cumulative share is `0.1278`, which predicts `0.97(1 − 0.1278) = 0.846` against the executed `0.846 ± 0.032`.

**Not exact.** The worker's shadow-delay model, and its parts (b) and (c), remain estimates. The identity `push = push_co(1 − f)` holds for the reflection-symmetric average, not for one asymmetric body.

PR #8555, opened 2026-09-21
pack: decision record for the moving-records lane (blocks 39-52): what is established, the forks, the cost of the capture picture, a correction (the wind is not isotropic), and the revised recommendation (source yes, carrier no)

JONbridgerapps:
## Decision record: moving records, sources and carriers (blocks 39–46, with an addendum for blocks 47–49)

Pack-only hand-off PR (no note, no runner, no audit data; never merged by the author). Written at the end of the owner's 12-hour campaign on this lane (2026-09-20T23:37Z → 2026-09-21).

`.claude/science/physics-loops/admissibility-induced-law-20260906/DECISION_RECORD_moving_records_sources_carriers_20260921.md` is one page for the owner:

- **What is established** across blocks 39–46 (open PRs #8530, #8546, #8547, #8548, #8549, #8550, #8553, #8554), exact unless marked executed, including two campaign controls run after block 45: for bodies transparent to the gas the capture rate is the record count and **the force is the product of the record counts over the square of the distance with the coefficient predicted** (`F/(N₁N₂) = (1.44 ± 0.10)·10⁻⁴` against `1.45·10⁻⁴` at separation 16); and without the scattering that re-draws contents there is no sound.
- **What the gravity lane on `main` actually has**: a supplied field equation (its own cited note says so), three named missing pieces, and no committed dynamics in either layer.
- **A correspondence of form** between the wind law and the weak-field packet's three statements — not a gravitational claim.
- **The forks that are the owner's**: glue or the neutral scale; whether an unformed site's odds are a condition; inertia; whether an encounter may re-draw a record's content (waves need it; the Record axiom says a record "locks exactly one admissible local possibility"); capture and transparency; the rule's weights are not yet in the inertial clause.
- **The cost of the capture picture in one inequality** (supervisor's estimate): growth and drag share one rate, and attraction wins only when the gas inside an orbit weighs much less than the central body.
- **Recommendation** (not a decision): settle the re-drawing fork first; it gates both the odds reading and inertia. — *superseded by the addendum:*
[...]

PR #8562, opened 2026-09-21
ail50: inertia joined with the rule's pair weights by a clock — a global clock is exact, a local clock is exact for pairs and fails at three records

JONbridgerapps:
## ail50: inertia joined with the rule's pair weights by a clock

Hand-off PR (independent vs `main`; never merged by the author). Part of the owner's 12-hour campaign on this lane. Trace class: frontier discovery, within block 44's supplied inertial clause (PR #8550) and block 39's supplied law with vacancies (PR #8530); nothing adopted; no gravitational claim.

**The question** (decision record PR #8555, fork 6). Block 44's inertial gas has a structureless equilibrium and never uses the rule's weights `(p, q, r)`; blocks 39–40 have the rule's law as equilibrium and no inertia. Can records travel along their contents, with number and momentum conserved event by event, **and** have the rule's law with vacancies as their stationary law?

**Exact (six-axis menu, periodic windows).**
- **T1 global clock.** Block 44's events form a bijection on configurations: every record has exactly one event and exactly one predecessor. So if every event of `C` runs at the rate `1/π(C)`, `π` is stationary — any weights, any density — and the configurations are visited in the same order as before; only the waiting times change. Not local: a record's clock reads every pair weight of the configuration.
- **T2 local clock.** With the event of the record at `x` at the rate `1/π_x(C)` (a record lingers where its content agrees with its neighbours'), `out(C) − in(C) = Σ W₀[A_b(u) − A_x(u)]` over the records that have another record directly behind them. Zero for two records and whenever no such pair has a third record next to it; **not zero at three records**: at `(3, 1, 2)` it fails at 3168 of the 70200 three-record configurations with a record at the origin (28512 of all 631800), by at most `3 = (cr − 1) + (cp − 1)`.
- **T3.** For two records the stationary law is the rule's pair law; at the neutral scale `c₀ = 6/(p + q + 4r)` the mean pair weight is one, and `c₀r = 1` exactly when `p + q = 2r`, which holds at `(3, 1, 2)`. Without re-drawing the dynamics is not irreducible (collinear pairs never leave their line): stationarity, not uniqueness.
- **T4.** Re-drawing a bond's two contents on their momentum class with probabilities proportional to `π`: detailed balance, momentum conserved, local.

**Open, and named:** whether some local clock is exact at all densities. The balance conditions are linear in the rates, so on the three-record sector it is a finite linear feasibility problem.

**Evidence.** Runner `TOTAL: PASS=14 FAIL=0`, exact arithmetic, 8 mutations each failing in its own family only. Refuting pass with disjoint machinery (forward flux accounting over the full sectors, 12636 and 631800 configurations; symbolic weights; a continuous-time simulation) returned three findings, all folded: the orbit count (9×, not 27×), the closed classes of collinear pairs (the simulation first gave `2 : 2/3 : 2`, then `2.97 : 1.01 : 2.00` per reachable configuration), and a floating-point zero in a symbolic check. Two pipeline passes, lints clean.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `e21017edf0e5a6738584c29532e6d383e87887a8` in 6bf62ae06ccbea85255e791d7e57b2b800ef4710 as part of the combined fourteen-PR moving-record and capture batch.

- Weighted redraw cannot cancel streaming defect, pin counts; consistent sink sign, elliptic continuum scope and projected generic damping; homogeneous-reservoir aggregate capture, alpha=c price and higher-order lattice terms.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block50-inertia-joined-with-the-rules-weights-by-a-clock-20260921` at `e21017edf0e5a6738584c29532e6d383e87887a8`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8568, opened 2026-09-21
ail53: no master clock — a neighbour-determined, scale-covariant tick rate obeys the lattice averaging equation; records enter as additive sources (first block of the source-link direction)

JONbridgerapps:
## ail53: no master clock — the lattice field equation, its zero mode and additive record sources from one clause

Hand-off PR (independent vs `main`; never merged by the author). First block of the **source-link** direction, on the owner's instruction "ok lets work the source link into the amplitude layer". Trace class: frontier discovery, within a supplied clause; nothing adopted; no gravitational claim; **no parked decision touched** (statistical bridge, C-add fork, M₄(ℂ) enlargement stay parked).

**Foundations read first.** `docs/MINIMAL_AXIOMS_2026-06-29.md` and `docs/repo/DEFERRED_DECISIONS.md` were read in full before framing. The axioms contain no amplitude dynamics and no time metric (rates, update laws, Born weights and source/action identification are open gates). On `main` the weak-field packet's three pieces — operator, source, response `S = L(1 − φ)` — are supplied, and its own cited note says the field equation "is supplied to the loop".

**The idea.** The packet's response law already says the potential *is* a shift of a local rate. The axioms define no clock. So ask what law a local tick rate can obey when only **ratios** of rates mean anything.

**The clause (supplied, not adopted).** Every site has a positive tick rate; the rate at a site is determined by the rates at its six nearest neighbours by one rule covariant under translations and proper cubic rotations (the form the Admissibility axiom has for the odds); multiplying every rate by `t` multiplies the answer by `t`.

**Exact.**
- **T1.** A linear, covariant, shift-symmetric nearest-neighbour law for `u = log w` is `c₀(u_x − average of the six neighbours) = s_x`: the lattice Laplacian. No mass term — a mass term would single out a rate, i.e. a master clock.
- **T2.** *Every* rule of the class, linear or not, has that law as its weak-field form (all six first derivatives equal `1/6` at the uniform point, by the degree-one identity and the transitivity of the rotations). The second order is **not** forced (`(p − 1)d²/6` for the power mean of order `p`). The geometric mean is exactly linear in `u`. The product of the six rates — the form the campaign's rule has for unnormalized weights — has degree six, is not scale covariant, and its symbol changes sign.
- **T3.** On a closed lattice only the zero-sum part of the source enters and `u` is fixed up to a constant no ratio can see: the packet's zero-mode projection.
- **T4.** A record cannot pin an absolute rate (that needs a master clock). A covariant clause gives it a ratio `κ` to its neighbours' mean — the additive source `log κ`. The field of **any** arrangement is exactly the sum of single fields; the pair term is symmetric; with `κ < 1` clocks run slow near records. **One number is left: `κ`.**
- **T5.** A test record whose hops are timed by the clock of the site it sits on has the stationary law `1/w = exp(−u)`: it gathers where clocks run slow. Steady masses, exact additivity, the same in every direction at long range (refuter: slope of the field against `1/r`, `−0.1467` along an axis and `−0.1413` along a body diagonal against `−0.1432`; block 51's wind differed by 30 %) — but a **drift, not an acceleration**.

**Correspondence (of form).** The packet's operator, its zero-mode projection and the additivity of its source follow from the clause, with `φ = −u` and block 41's record count as the source. What remains supplied: `κ`, the non-linear completion, inertia, and any delay of the field. Acceleration needs a record whose *phase*, not only its waiting time, is timed by the local clock — the amplitude layer; that is the next block.

**Evidence.** Runner `TOTAL: PASS=16 FAIL=0`, exact arithmetic (exact dual numbers for the derivatives), 10 mutations each failing in its own family only; refuting pass with disjoint machinery (symbolic solving; a sparse solve on a `41³` box; a transform solve; a simulation): 5 witnesses pass; two findings folded (a wall artefact in the first far-field test; a mis-scaled check); one supervisor's argument (records as pinned values → capacity) withdrawn in the lens pass because a pinned absolute rate is not scale covariant. Two pipeline passes, lints clean.


JONbridgerapps:
Corrigendum pushed (c4e6a23f6c): T5's remark that the record's mean velocity follows the gradient is wrong for T5's own timing. Hops timed by the clock of the site the record sits on are unbiased, so its expected displacement is zero away from contact; what follows the gradient is the flux of its probability, `(m_x w_x − m_y w_y)/6` across a bond. The record lingers where clocks run slow rather than being carried there. Found while building block 95 (#8860), which proves this (its T3) and extends T5 to records that are both sources and movers. Text-only; runner output unchanged (cache re-pinned for the note's new fingerprint).

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `c4e6a23f6c24c7f5de6d94bedbd749458452c604` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

- Separate the supplied log-linear completion; nonzero-total torus obstruction and finite residence law; exact phase-force counterexample, fixed-rate similarity and independent conditional ray scope.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block53-no-master-clock-forces-the-lattice-laplace-equation-20260921` at `c4e6a23f6c24c7f5de6d94bedbd749458452c604`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8570, opened 2026-09-21
ail54: a phase timed by the local clock — every packet falls towards slow clocks; force = energy × gradient exactly; the walk with the qubit as coin ties weight to inertia (source-link direction, block 2)

JONbridgerapps:
## Block 54 — second block of the source-link direction (hand-off; independent vs `main`)

Owner's instruction: *"ok lets work the source link into the amplitude layer"*. Block 53 (#8568) ended on a gap: a record whose **waiting time** is timed by the local clock drifts towards slow clocks, but a drift has no inertia. This block takes the step it named: a record whose **phase** is timed by the local clock.

**Objects and clause (supplied, not adopted).** One moving record carried as a two-component amplitude (the site's qubit as its content); a nearest-neighbour generator; a positive rate field (any — block 53's clause supplies one); and: *the amplitude at a site advances in that site's own time.* The axioms memo contains no amplitude dynamics and no time metric; these are candidate fillings of open gates.

### Exact (runner `PASS=24 FAIL=0`, exact arithmetic, 12 mutations each failing in its own family)

- **T1 — the generator.** Hermitian, translation-invariant, rotation-covariant nearest-neighbour generators form a **three-parameter family**: a constant, plain hopping, and one term `β Σ σ_j sin k_j` that couples content to motion — its velocity operator *is* the content (block 44's "content = direction of travel" is its expectation value). With inversion **added** as a symmetry (the Admissibility axiom names proper rotations only — flagged): a content that reverses, as a direction does, leaves that term alone (the walk with the qubit as coin); a content inversion does not touch removes it. One walker has **no rest energy**: nothing in `M₂(ℂ)` anticommutes with all three content matrices. No larger site algebra is used; parked entry 4 untouched.
- **T2 — timing the phase.** The clause gives `wH`; conserved quantity `Σ|ψ_x|²/w_x` (the same `1/w` weight as block 53's drifting record); similar to the hermitian `√w H √w`. Changing every rate by one factor only changes the unit of time.
- **T3 — force = energy × gradient, exactly.** In a uniform gradient `H_wⁿ T_a = λ_aⁿ T_a H_wⁿ` (λ_a = ratio of the clocks): a translated packet evolves faster by exactly the clock ratio, and in a state of energy `E` the wave vector along `a` falls at the constant rate `(λ_a − 1)E`. Every generator, every state, no ray approximation. Stated for a slab: with a uniform gradient everywhere the rates are unbounded and rays escape at `t = 1/g`.
[...]

PR #8571, opened 2026-09-21
ail55: action and reaction — the rate field's source is the amplitudes' energy density; matched pulls and a kept ledger fix the clock law's second order (source-link direction, block 3)

JONbridgerapps:
## Block 55 — third block of the source-link direction (hand-off; independent vs `main`)

Owner's instruction: *"ok lets work the source link into the amplitude layer"*. #8568 gave the rate field its law, #8570 gave the amplitude its response (pull = energy × gradient, exactly). The weak-field packet's third supplied piece is the **source**, read off the amplitude as `ρ = |ψ|²`. This block asks what the source has to be.

**Premise (supplied, not adopted):** the rate field obeys a static law and the books balance — the amplitudes' energy plus an energy of the field does not change in time. The axioms contain no conserved energy; this is declared as the premise.

### Exact (runner `PASS=18 FAIL=0`, exact arithmetic, 10 mutations each failing in its own family)

- **T1 — the ledger identity.** Change the rate at one site and the amplitude's energy changes by the energy it has there: `d⟨H_w⟩/du_x = e_x = Re χ_x†(H_wχ)_x`, and `Σ e_x = ⟨H_w⟩`.
- **T2 — what the source must be.** The ledger is kept when the law's source is the **energy density** with its mean removed. With any other source `s` it moves at the exact rate `Σ (e_x − s_x) du_x/dt`. A body at rest: `e = m w |χ|²` — the packet's `|ψ|²`, times the rest energy and the local rate. A moving body: its energy, not its rest energy.
- **T3 — every pull matched.** #8570 made the pull on a body proportional to its energy. Two bodies pull each other equally and oppositely at every separation **iff** each sources in proportion to its energy. If every body sourced alike (a count), a heavy and a light body left alone would push themselves along.
- **T4 — no master clock, again.** The ledger must keep its form under a change of the unit of rate ⇒ the field's own energy has weight one (`√(w_x w_y) f(u_x − u_y)` on bonds). Then (i) the unit of rate cannot be a variable — otherwise the total would have to vanish and like energies would repel — so the source enters with its mean removed: the packet's projection, a second time; (ii) the coupling `γ` is a pure number; (iii) **the second order of #8568's clock law, which that block left free, is fixed**: through second order every such law is the averaging law for `√w` (power mean of order ½); `f` enters at fourth order.

In the continuum T4 is Einstein's second static theory of 1912 (`Δ√c = (k/2)√c σ`), which he reached by the same action-and-reaction argument; named under Premises and Prior art. What is added: exact lattice statements, the exact failure rate of any other source, the link to #8568's free second order, and the identification of the packet's `|ψ|²` as the rest-body form.

### Executed, not claimed — two walkers on a ring sourcing the field that times them

| source | both at rest: wave-vector changes fail to cancel by | ledger moved | one body moving: fail by | ledger moved (T2 predicts) |
|---|---|---|---|---|
| energy density | `1.6×10⁻⁵` | `−1×10⁻¹¹` | `2.6×10⁻⁵` | `−1.4×10⁻⁹` (0) |
| probability density | **`0.50`** | `+1.37×10⁻⁵` | `0.14` | `−3.94×10⁻³` (`−3.94×10⁻³`) |
| rest form `m w |χ|²` | `9.8×10⁻⁴` | `−2.94×10⁻⁶` | **`0.41`** | `+4.57×10⁻⁴` (`+4.57×10⁻⁴`) |

### Where the three blocks leave the weak-field packet (correspondence of form; no gravitational claim)

| supplied on `main` | within three supplied clauses |
|---|---|
| operator `−Δ_lat`, zero mode projected | no master clock (#8568); again from the weight-one ledger (here) |
| test response `S = L(1 − φ)`, `F = +m∇φ` | phase timed by the local clock (#8570): `φ = −u`, `m` = energy |
| source `ρ = |ψ|²` | the books balance (here): energy density; `m w |χ|²` at rest |
| coupling | one pure number `γ`; #8568's `κ = exp(−(γ/6)E/w̄)` if records are pulled as amplitudes are |

### Not settled, and said so
Whether an amplitude that has formed **no record** sources anything — a question for the record reading; the note gives the form the result takes if only records source (T3 + the corollary). The ledger itself is a premise. Mean field, static law (no delay). `γ`, `f` beyond second order, lengths: not derived. No statistical statement is used or compared; parked entries untouched; nothing adopted.

Refuting pass with disjoint machinery (symbolic ledger identity; symbolic bump test for general `f`; ten bodies on a `16³` torus; finite-difference weight-one identity; dense-exponential two-walker run): 5/0. Three control artefacts folded (see the Review record). Two pipeline passes; lints clean. Author checks only.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `525c6500c56b178a241a2e17cbf18a46485e57f2` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

- Canonical dynamics and sufficient field equations explicitly supplied; total-ledger multiplier corrected; gamma/6 and bump expansion corrected; grounded positive diagonal model, fixed-diagonal pair derivative, necessary alternative-bond mass bound only.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block55-action-and-reaction-the-rate-fields-source-is-the-amplitudes-energy-density-20260921` at `525c6500c56b178a241a2e17cbf18a46485e57f2`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8573, opened 2026-09-21
ail56: the strong field, exactly — bodies at rest make the clock law linear in √(rate); the ledger is a surface term, bounded by a capacity (source-link direction, block 4)

JONbridgerapps:
## Block 56 — fourth block of the source-link direction (hand-off; independent vs `main`)

#8568 found that the rate field's sources add exactly — at weak field. #8571 fixed the clock law's second order and left one function beyond it. This block takes the simplest member (field energy `(2/γ) Σ_bonds (φ_x − φ_y)²`, `φ = √w`) and asks what happens when the field is **not** weak. It is exactly solvable.

**Everything here carries one further choice that #8568–#8571 leave open: the bond energy.** The refuting pass was pointed at exactly that, and found a boundary (below). T5 names the principle the choice amounts to.

### Exact (runner `PASS=17 FAIL=0`, exact rational solves on a `7×7×7` box, 12 mutations each failing in its own family)

- **T1 — the law is linear.** For bodies at rest (`e = m w`) the clock law `(12/γ) φ_x (φ_x − avg φ) = −m_x φ_x²` divided by `φ_x` is the *linear* problem `((1 − A) + (γ/12)M) φ = 0`, `φ = 1` on the walls. One solution; `0 < φ ≤ 1`; more energy anywhere slows every clock. Linearity does not even need rest: for any fixed amplitude `e_x = φ_x (Kφ)_x` with `K` independent of the rates. Rest supplies `K = diag(m) ≥ 0`, and with it the bounds. *(An earlier sentence of my draft said the opposite; corrected before opening.)*
- **T2 — the ledger is a surface term, and it is `Σ m φ`.** For *every* weight-one bond energy the ledger of a static arrangement equals a sum over wall sites: the outside sees the ledger. For the simplest one it is exactly `Σ m_i φ_i` — between the bare energies `Σ m_i` and the energies in the field `Σ m_i φ_i²`.
- **T3 — saturation and a capacity bound.** One body: `φ_0 = 1/(1 + x)`, `x = (γ/12) g_0 m`. Its ledger stays below `12/(γ g_0)` (`7.9/γ` on `Z³`) however much is put there; its energy in the field peaks at `x = 1`; its clock stops only in the limit. Any region: ledger `< (12/γ) × capacity`, reached as the clocks on it stop.
- **T4 — pairs.** Closed form; the defect of a pair's ledger is bilinear in the "charges" `m φ`, leading term `(γ/6) m_1 m_2 c`; `∂(ledger)/∂c < 0` exactly: the ledger falls as the bodies approach **at any field strength**.
- **T5 — what would make the choice (a supplied principle).** #8570's clause makes an amplitude's energy *bilinear in the root rates* (`Σ φ_x φ_y h_xy`, `h` rate-independent). If the local clock times the field's own energy the same way, then nearest-neighbour range + covariance + "uniform rates solve the empty-space law" leave `c Σ_bonds (φ_x − φ_y)²` and nothing else; the whole ledger is one quadratic form in `φ`.

### Boundary found by the refuting pass
A second admissible bond energy, `(1/γ) Σ (w_x − w_y)²/(w_x + w_y)` — weight one, even, same second order — has a non-linear static law, agrees at weak field to `10⁻⁵`, and **stops the clock at a body at the finite bare energy `18/γ`**, with no static solution beyond. So *whether and where clocks stop is a property of the bond energy*, not of the clauses. What is general: the surface term, and everything through second order. The second member violates T5's principle (it fails the parallelogram identity).

### Executed, not claimed
Stopped balls of radius 2, 4, 6 in boxes up to `81³`: ledgers `0.85, 0.90, 0.93` of `8πR/γ` after extrapolation in box size — proportional to the **radius**, `1.30, 0.35, 0.15` per site against `7.9` for one site alone. Outside a stopped ball `1 − φ = a/r − b` with `a` within 1 % of `γ ×` ledger`/(8π)`. A pair at `x = 2.5` each: defect `0.08` of its weak-field value, within 10–1 % of `(γ/6)(mφ)² g` at separations 2–12.

### Boundaries
Bodies at rest with a rest energy are supplied (one walker has none, #8570). Nothing moves: no delay, no record formation. Walls are held at the ambient rate (they fix the unit of rate; the closed lattice's multiplier would break linearity). `γ` not derived. No gravitational claim; nothing adopted; no parked decision touched.

Refuting pass with disjoint machinery (direct non-linear solve with no division by `φ`; finite-difference stationarity and wall sum; symbolic pair; forty random arrangements; a second bond energy; the bilinear family symbolically + a random complex amplitude under the walk): 6/0. Two pipeline passes; lints clean. Author checks only.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `d6c6e2572de656c33fdb78a0c3b14b28d18d9c5c` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

- Canonical dynamics and sufficient field equations explicitly supplied; total-ledger multiplier corrected; gamma/6 and bump expansion corrected; grounded positive diagonal model, fixed-diagonal pair derivative, necessary alternative-bond mass bound only.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block56-the-strong-field-exactly-linear-in-root-rate-ledger-and-capacity-20260921` at `d6c6e2572de656c33fdb78a0c3b14b28d18d9c5c`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8572, opened 2026-09-21
ail decision record: the source link in the amplitude layer — state of knowledge after blocks 53–56 (pack-only)

JONbridgerapps:
One page for the owner, pack-only (a single file under the campaign pack; no note, runner or audit surface changes).

**What it holds**
- The three supplied clauses, one sentence each: **no master clock** (#8568), **the phase is timed by the local clock** (#8570), **the books balance** (#8571).
- Ten established rows with evidence addresses, exact unless marked executed — including the two things that qualify the headline: one walker has **no rest energy** in `M₂(ℂ)`, and packets **drift sideways** by `gT/(2k)` (a wave correction; "everything falls alike" is a statement about rays).
- **Addendum for block 56 (#8573), rows 11–16:** with the simplest admissible field energy the strong field is exactly solvable — the clock law is *linear* in `√w`; the ledger is `Σ m√w` and a surface term; a site's ledger saturates at `7.9/γ`; a region's is bounded by its capacity (∝ radius); pairs attract at any field strength. **Boundary:** another admissible field energy stops a clock at the finite bare energy `18/γ`. A supplied principle that would make the choice: the local clock times the field's own energy as it times an amplitude.
- The weak-field packet piece by piece: operator + projection, test response, inertia, source, coupling — what each becomes within the clauses. One pure number `γ` is left.
- **Nine forks that are the owner's**: clocks per site; a moving record as an amplitude; the ledger; whether an amplitude that has formed no record sources anything; rest energy; lengths (a field of clock rates gives half of the comparator's bending of light); delay; `γ`; and whether the field's own energy is timed as an amplitude's is.
- A layman paragraph and the integration order (#8568 → #8570 → #8571 → #8573).

Nothing is adopted; no parked decision is touched or re-raised; author checks only. `audit_lint --strict` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8578, opened 2026-09-21
ail57: a delay for the rate field — referred to its neighbours its motion does not propagate; referred to the distant clocks it does, at a speed set by the local rate (source-link direction, block 5)

JONbridgerapps:
## Block 57 — a delay for the rate field (hand-off; independent vs `main`; fifth block of the source-link direction, first of the 12-hour campaign)

The clock law of #8568 / #8571 / #8573 is **static**: move a body and the rates everywhere change at the same instant (decision record #8572, fork 7). This block asks what motion of its own the rate field can have when there is no master clock.

**Premise (the reading of "no master clock" for motion):** the parameter `t` in the equations is only a label — `t → f(t)`, `w → w/f'` changes nothing observable. #8570's amplitude law and every ledger term of weight one are *already* unchanged by this (that is what weight one means). It bites only now.

### Exact (runner `PASS=14 FAIL=0`, exact arithmetic, 9 mutations each failing in its own family)

- **T1 — what a kinetic term may see.** A change of parameter shifts every `du_x/dt` by the same amount ⇒ a kinetic term survives iff it sees only **differences** between sites (`M·1 = 0`). The obvious term `Σ (du/dt)²/w` does not survive: it needs a `t` that is more than a label.
- **T2 — referred to neighbours, nothing travels.** A nearest-neighbour covariant form with `M·1 = 0` is a multiple of the lattice operator of `Σ_bonds (v_x − v_y)²` — **the same operator as the field's weak-field energy**. Every mode swings at one frequency `w̄/√(γκ)`; the field of a source switched on is its *static profile at every instant, everywhere at once*. Longer finite range: still no branch through zero frequency. Non-linearly: the accelerations need the inverse of a weighted lattice operator, which couples every site to every other.
- **T3 — referred to the distant clocks, a front.** Referring every rate to one clock of the system (held walls, or the lattice mean) survives the change of parameter and gives a wave equation: a front at **`c` sites per local tick**. Price: it is *not a nearest-neighbour law*, and `c` is a second pure number.
- **T4 — a master parameter** gives the same waves plus a uniform mode in which every clock slows together, `ψ₀² = ψ₀(0)² + a t²` (`ψ = w^{−1/2}`) — a motion no comparison of clocks could show.
- **T5 — a field that is not a rate** (unchanged by a change of parameter) carries its on-site kinetic term through any change of parameter and has a front with nearest-neighbour terms only. #8570's amplitude is such a field; a field of lengths would be another.

**Consequence for the owner's forks:** delay (fork 7) and lengths (fork 6) are one fork. With no master clock and nearest-neighbour laws the rate field cannot carry a delay; a delay needs a reference to distant clocks, or a field that is not a rate.

### Executed, not claimed
Two walkers 300 sites apart on a line with held walls, the full non-linear law referred to the wall clocks, field switched on from the ambient rate:

| | front expected | `k_B` before | after | ledger (with kinetic part) |
|---|---|---|---|---|
| `c = 1`, ambient rate 1 | 300 | `−2×10⁻⁹ … −2.5×10⁻⁴` (t = 100 … 250) | `−0.0018 … −0.0201` (t = 300 … 450) | `−8×10⁻⁸` |
| `c = 2` | 150 | same pattern | | `−8×10⁻⁸` |
| ambient rate ×4 | 75 | same pattern | | `−3×10⁻⁷` |

From the static field, slow motion follows #8573's static law to 0.7–1.2 % at 0.03 of the limiting speed (2–4 % at 0.1). In a `61³` box half-rise times are `r/(c w̄) + 0.25…0.5` along axes and body diagonals at two ambient rates. Under the neighbours-only law `u/u_static` is the same number at distances 1, 6 and 21 at every time.

### How the first version of this theorem failed
My first statement was "with no master clock the rate field has no waves". Trying to break it, I found the way round: a kinetic term referred to the lattice mean or to held walls has `M·1 = 0` and gives ordinary waves. The no-go is for finite range; the note is a trichotomy (neighbours / a clock of the system / a master parameter) plus what else can carry a delay.

### Boundaries
The premise is a reading (stronger than #8568's sentence; the weaker reading gets T4). Which reference the framework would state is the owner's. `c` not derived. Wave statements are weak-field; the non-linear coupled motion is executed on a line. The note does not show that the static law's instant response can carry a usable signal, nor that it cannot. Lengths are named, not treated. No gravitational claim; nothing adopted; no parked decision touched.

Refuting pass with disjoint machinery (symbolic generic `f(t)`; generalized eigenproblems on a `6³` torus; a non-linear elliptic solve; propagation at two ambient rates; the uniform mode on a ring): 5/0. Two pipeline passes; lints clean. Author checks only.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `621c0c56378574c17622f21ebd69e9294639909b` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block57-a-delay-for-the-rate-field-nearest-neighbour-kinetic-terms-do-not-propagate-20260921` at `621c0c56378574c17622f21ebd69e9294639909b`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8579, opened 2026-09-21
ail58: what a formation event does to the ledger and to the far field — the monopole never jumps, the dipole does (source-link direction, block 6)

JONbridgerapps:
## Block 58 — what a formation event does (hand-off; independent vs `main`; sixth block of the source-link direction, second of the 12-hour campaign)

Decision record #8572, **fork 4**: does an amplitude that has formed no record source the rate field (#8571's reading), or do only records (#8568's)? This block does **not** decide it and assumes **no rule for where records form**. It asks a narrower exact question: *when a spread amplitude at rest is replaced by one record at a site, what happens to the ledger and to what the outside sees?* In #8573's setting both sides are exactly solvable (a spread amplitude at rest is a set of bodies at rest with bare energies `m|χ_x|²`).

### Exact (runner `PASS=14 FAIL=0`, exact rational solves on a `7³` box, 9 mutations each failing in its own family)

- **T1 — the ledger through the event.** Kept iff the record's bare energy is `m' = E/(1 − (γ/12) g_y E)` — *more* than the amplitude's (a point slows its own clock more than a spread does) — and only if `E < 12/(γ g_y)`: an amplitude that shows more than one site can show **cannot become one record with the books balanced**.
- **T2 — what the walls see.** For every discrete-harmonic `h`: `Σ_wall bonds (1 − φ_inside) h_wall = 6 Σ q_x h_x`, `q_x = (γ/12) m_x φ_x`. With `h = 1`: the monopole. Keep the ledger and **the monopole never jumps**, wherever the record forms. With `h` a coordinate: the dipole, exactly. (Refuting pass: quadrupole and octupole too.)
- **T3 — the dipole jumps** by `Q(y − X̄)`, `X̄` the centre weighted by `|χ_x|² φ_x`; no choice of the record's energy removes it. For **any** odds `p_y` the mean jump is `Q(Σ p_y y − X̄)`: zero for odds by the charge density `|χ|²φ`; **not** zero for odds by `|χ|²` or by the energy density `|χ|²φ²` (they miss by equal amounts on opposite sides — the amplitude's own field makes the rate differ across it). The note proposes no rule; it states one exact constraint any rule would face.
- **T4 — if only records source:** nothing is seen before the event and `m'φ'_y` after — **a monopole from nothing at every formation event.** The two readings of fork 4 differ at leading order in what the outside sees.

### Executed, not claimed
`41³` box, weak (`γ = 0.05`) and strong (`γ = 1`) coupling, records at three sites: wall flux changed by `2×10⁻⁷` / `4×10⁻⁶` (the accuracy of the probe for `g_y`); first moments of the flux jump by `Q(y − X̄)` to five digits; centres at `γ = 1`: charge `−1.8225`, probability `−1.8449`, energy `−1.7999`. Under #8578's wall-referred law the change reaches `r = 8, 16, 24` at `t ≈ 6, 13.5, 21`.

### Next to the parked statistical postulate — how this block behaves
No odds are assumed, derived or proposed; every statement about odds is an identity in arbitrary odds; registry entry 1 is named as *not used*; the fences are verbatim; no wake condition is claimed.

### A supervisor expectation that was wrong
#8571 makes the *energy* density the source, so I expected odds by energy density to hold the centre. They do not: the centre the walls see is weighted by the **charge** density `|χ|²φ`.

### Boundaries
#8573's setting throughout (simplest bond energy; supplied bodies at rest with a rest energy; held walls; static law). Motion, several records and records that move are not worked. No gravitational claim; nothing adopted.

Refuting pass with disjoint machinery (root-finding for the record's energy; higher harmonic weights; thirty random amplitudes for the three densities; concavity of the ledger): 4/0. Two pipeline passes; lints clean. Author checks only.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `e526d22c1841c4d53170ad945fdb46b968832bff` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block58-what-a-formation-event-does-to-the-ledger-and-to-the-far-field-20260921` at `e526d22c1841c4d53170ad945fdb46b968832bff`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8581, opened 2026-09-21
ail59: bond rates and lengths — a body at rest sources no length, and the bending of rays carries one more free number (source-link direction, block 7)

JONbridgerapps:
## Block 59 — bond rates and lengths (hand-off; independent vs `main`; seventh block of the source-link direction, third of the 12-hour campaign)

#8578 ended: a delay needs a reference to distant clocks *or a field that is not a rate* — lengths, say. Decision record #8572, fork 6, carries one blunt fact: a field of clock rates alone bends a ray past a body by **half** of the comparator's amount. This block asks what lengths can be in the framework's vocabulary and whether they supply the other half.

**Widening (supplied, not adopted):** every bond has a crossing rate `c_b` of its own; a bond's **length** is the pure number `ℓ_b = √(w_x w_y)/c_b`. #8570 is `ℓ = 1`.

### Exact (runner `PASS=15 FAIL=0`, exact arithmetic, 10 mutations each failing in its own family)

- **T1 — what one walker sees.** A walker with no rest energy (and #8570's has none) never meets a site rate: its generator contains bond rates only. For it, "clock rate" and "length" are not two things.
- **T2 — each rate is sourced by the energy it times.** Bond rate ← hop energy on that bond; site rate ← on-site energy; they add to the whole. **A body at rest has exactly zero hop energy on every bond**: it slows site clocks and sources no bond rate and no length. A walker loads the bonds it crosses, and no others.
- **T3 — a law for bond rates: three numbers, one scalar.** #8568's uniqueness argument, run for bonds, leaves three free numbers. But only the mode with all directions equal is a long-range field; differences between directions are the scalar's second derivatives times `(α − δ + β/2)/(12β)` — a tide, two powers of distance below the field.
- **T4 — bending against falling.** A slow body falls by `∇ log(site rate)`, a ray crossing the gradient bends by `∇ log(bond rate)`: ratio `d log c/d log a` — 1 for locked lengths, 2 for `ℓ = w̄/w`, 0 if bond rates ignore site rates.
- **T5 — what could make it 2.** Not a local *formula* for the length (a length is a pure number, a rate is not ⇒ `ℓ = G(w_x, w_y)` sees only `w_x/w_y`: the gradient, never the depth). A local *law* can: a cross term in the field's energy gives `ℓ = (w̄/w)^β`, `β` free; the ambient rate enters through held walls, as in #8578. **Ratio = `1 + β`: the clauses of #8568–#8578 give 1; the comparator's value is 2 — one more free number.**

### Executed, not claimed
A packet of the walk crossing a uniform gradient in a `64³` box bends by `−1.6164` with locked lengths and `−3.1783` with `ℓ = w̄/w` (ratio 1.966; ray clouds to 0.3 %). A slow body on a line falls alike under both clauses (ratio 1.02).

### Two first-draft claims I broke before the gates
1. "Anisotropic bond-rate modes are short-ranged" — the refuting pass's transform solve showed a **power law**: they are slaved to the scalar's second derivatives (coefficient now derived and checked at three off-axis points).
2. "No local law ties a length to the depth of a rate" — it stood in the file name. True for an algebraic clause; **false for a field equation** (the cross term). Claim id, file names, T5 and the headline were rewritten.

### In plain terms
What #8568–#8573 built is a theory of clock rates. It makes slow bodies fall correctly by construction and bends rays by half the comparator's amount. Lengths that stretch where clocks slow would supply the other half, and the vocabulary has room for them — but nothing found so far forces them to stretch by the right amount, or at all.

### Boundaries
Bond rates are a supplied widening (#8570's clause by itself forces the product form). Comparing a ray with a slow body needs an on-site rest energy, which one walker does not have; if rest energy is internal hop energy the ratio is 1. T3 and T5(b) are linear weak-field statements with held walls. No law of motion for lengths. The comparator is quoted as a comparator: no gravitational claim; nothing adopted; no parked decision touched.

Refuting pass with disjoint machinery (symbolic rays and symbols; transform solve on `48³`; random dense checks; symbolic weights; a coupled quadratic minimization): 6/0. Two pipeline passes; lints clean. Author checks only.


JONbridgerapps:
Corrigendum pushed (7ff9b1540a): the runner's printed scope line still carried the first-draft wording 'anisotropic bond-rate modes are short-ranged', which the refuting pass had broken (they are slaved to the scalar's second derivatives: a power law). Docstring, checks and note were already right. Cache re-pinned; two pipeline passes; lints clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `7ff9b1540a3f53ff0072f6102d01968c127d151d` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block59-bond-rates-and-lengths-a-body-at-rest-stretches-no-lengths-20260921` at `7ff9b1540a3f53ff0072f6102d01968c127d151d`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8593, opened 2026-09-21
ail63: what the walk conserves — momentum flows on bonds; relabellings reach second neighbours; the nearest-neighbour frame misses by cos k̄ (source-link direction, block 11)

JONbridgerapps:
## Block 63 — what the walk conserves (source-link direction, block 11; campaign block 7)

Hand-off PR against `main` (independent of the open stack; blocks 54 and 62 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 62 (#8592): the member for the full symmetric stretching answers a static source only if the stress is divergence-free; the walk's site-placed frame response, executed there, is not — by a second-order term that also depends on placement at the same order. What does the walk conserve *exactly*, what generates that current, and how does block 62's response differ from it?

**Answer (exact identities of block 54's walk, identity frame; every state).**
- **T1.** `π_j = Re ψ†S_jψ` has a continuity equation with a bond current `J_a^j(x → x + e_a) = ½Re[ψ†(x + e_a)σ_a(S_jψ)(x) + (S_jψ)†(x + e_a)σ_aψ(x)]` — a three-site object.
- **T2.** For `G_ξ = ½Σ{ξ_j, S_j}`: `i[H, G_ξ] = Σσ_a½{C_a[d_aξ_j], S_j}` — a deformation on the *bonds* along `a` (stretch for `j = a`, tilt for `j ≠ a`) that hops along `a` and differences along `j`: second-neighbour reach. Its response is `J`; on stationary states `J` is divergence-free exactly.
- **T3.** `J = ½[Θ(x) + Θ(x + e_a)] − ½Re(d_aψ)†σ_a(d_aS_jψ)`. For a plane-wave pair `J = e^{iq_a/2} cos k̄_a Θ`: **block 62's defect is the factor `cos k̄_a`**; the law's symbol `sin k_a − sin k'_a = 2 sin(q_a/2) cos k̄_a` is not a function of `q` alone.
- **T4.** `H` reaches nearest neighbours; `i[H, G_ξ]` reaches second neighbours of both kinds. No deformation confined to nearest-neighbour hops has `J` as its response.
- **T5.** `d(ψ†σ_cψ)/dt = 2Σε_{cad}Θ_d^a − [b_c(x) − b_c(x − e_c)]`: the antisymmetric part of the site response is the torque on the coin; a pure divergence on stationary states.

[...]

PR #8590, opened 2026-09-21
ail60: a ledger linear in the rates — every clock a multiplier, the ledger a wall term, and the curvature member doubles the bending (source-link direction, block 8)

JONbridgerapps:
## Block 60 — a ledger linear in the rates (source-link direction, block 8; campaign block 4)

Hand-off PR against `main` (independent of the open stack; blocks 53–59 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 59 (#8581) left the ratio of a ray's bending to a slow body's fall at `1 + β`, `β` free, with the clauses so far giving half the comparator's bending. Block 55 (#8571) T4(b) could not let the unit of rate be a variable, because a positive field energy plus positive content never sums to zero. What does a weight-zero field (lengths) change?

**Answer (exact, within supplied clauses and a declared family).**
- **T1.** For any ledger of weight one, the derivatives in all the log-rates sum to the ledger. Every rate varied ⇒ ledger zero; walls held ⇒ the ledger *is* the walls' term (block 56's surface term was an instance).
- **T2.** Lengths allow a field energy *linear in the rates*, `F = Σ w_x G_x(lengths)`. Then every rate is a multiplier: a one-site relation between energy per tick and the lengths; the rates have no stiffness of their own; the lengths' equation is linear in the rates; a closed lattice has no static configuration with positive content.
- **T3.** On `Kℓ^p(aΔλ + bq)`: `β = a/(2(ap − b))`; bilinear members `1/p`; the member `p = 1` is `−K ×` rate `×` volume `×` scalar curvature of the stretched lattice (exact symbolic identity) and has **`β = 1`: bending twice the fall.** Weak field: `Δλ = −e/(4Kw̄)`, `Δu = (e + τ)/(4Kw̄)`; `γ = 1/(4K)`.
- **T4.** That member's strong field for bodies at rest, exactly: `χ = 1 + ΣQ_i g_i`, `Q_iχ_i = m_i/(8K)` (unique positive solution; minimum of a strictly convex function); `N = wχ` solves a linear equation; no clock stops at any strength; block 53's rule holds at every strength with the lengths as weights, `κ` in closed form; the ledger is `Σ m_i/χ_i` — below the bare energies, above block 56's, unbounded; far away the lengths carry `2Q`, the rates `P + Q` (equal at first order only).
[...]

PR #8595, opened 2026-09-21
ail64: bond strains and plaquette curls — blindness to the coin's axes forces the curvature member (β = 1), and needs a blind content (source-link direction, block 12)

JONbridgerapps:
## Block 64 — bond strains and plaquette curls (source-link direction, block 12; campaign block 8)

Hand-off PR against `main` (independent of the open stack; blocks 54, 60, 62, 63 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 63 (#8593): the walk keeps its books on the bonds. Block 60 (#8590): the curvature member gives `β = 1` but is *declared*. Put the field where the walk keeps its books, and ask what is left to declare.

**The variables (supplied).** A strain `B_a^j` on every bond — stretch for `j = a`, tilt for `j ≠ a` — coupled as a relabelling's deformation is; relabellings shift it by `d_aξ_j`; the mismatch around a plaquette is its *curl*.

**Answer.**
- **T1 (lattice, exact).** Curls are blind to relabellings; `∂⟨H[B]⟩/∂B` is block 63's conserved bond current; the field equations of *any* function of the curls are divergence-free identically. The static equations are consistent at first order for every stationary state: the first item of block 62's bill is met exactly.
- **T2 (continuum, exact symbolic).** Among densities with at most two derivatives built from the curls — six numbers, the inversion-odd one included, since the Lattice axiom has proper rotations only — blindness **point by point** to a rotation of the coin axes that varies from place to place leaves the volume and **one** combination, `(c_1, c_2, c_3, c_4) ∝ (1, 2, −4, −8)`. Point by point is what a ledger linear in the rates requires (every rate multiplies its own site's density). Blindness up to a divergence leaves `c_4`, hence `β`, free.
- **T3.** That combination is minus volume × scalar curvature of `eᵀe`: block 60's member, no longer declared.
- **T4.** On the family `β = c_4/(4c_1 + 2c_2 + 4c_3)`: free without blindness (and the antisymmetric strain is then a field), **one** with it.
- **T5 (the cost; lattice, exact).** A blind field energy gives a rotation of a site's three forward bonds no field equation, while the walk's response to it, `J_a^j(x) − J_j^a(x)`, is non-zero at all 64 sites of a `4³` torus for an exactly stationary state. **A blind field energy needs a blind content**, which block 54's walk is not.

I stopped the first run of the gates to add T5: I found it while scoping the next block, and the headline should not ship without its cost.

**Evidence.** Runner `scripts/admissibility_rule_bond_strains_and_plaquette_curls_a_field_energy_per_local_tick_that_does_not_see_the_coins_axes_2026_09_21.py`: `TOTAL: PASS=17 FAIL=0`; 11 mutations, each failing in its own family only. Refuting pass W1–W4 with disjoint machinery: dense `H[B]`; a relabelling moves the spectrum at second order only (`1.333 s²`); **exact frames at rational points, all orders in the strain**: the blindness conditions' null space is the volume and `(−1/8, −1/4, 1/2, 1)`; `D* + √g R = 0` exactly for a finite strain. Findings folded: a sign (inverse frame taken for the co-frame gave `β = −1`); an unverified sentence about the weaker demand, then computed and added to the runner; a miscount. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.

**Next.** A walk that does not see a varying rotation of its coin axes.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `e5688665736b60adf84e9eb84276e8c4169c2a17` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block64-bond-strains-and-plaquette-curls-blindness-to-the-coins-axes-forces-the-curvature-member-20260921` at `e5688665736b60adf84e9eb84276e8c4169c2a17`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8596, opened 2026-09-21
ail65: the blind walk — a scalar hop weighted by the twist of the coin makes a varying rotation of the coin axes a symmetry (source-link direction, block 13)

JONbridgerapps:
## Block 65 — the blind walk (source-link direction, block 13; campaign block 9)

Hand-off PR against `main` (independent of the open stack; blocks 54, 62, 63, 64 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched (the one-site algebra stays `M_2(C)`).

**Question.** Block 64 (#8595) T5: a field energy that does not see the coin's axes needs a content that does not see them either, and block 54's walk does. What exactly does a rotation of the coin axes that varies from site to site do to the walk, and what must be added for the walk not to see it?

**Answer (exact, first order).**
- **T1.** `−(i/2)[θ·σ, H] = ½Σ_j{(θ × e_j)·σ, S_j} + ½Σ_aC_a[d_aθ_a]`. Turning the coin differently at each site is a rotation of the frame at every site **plus** a coin-independent symmetric hop on every bond, weighted by the **twist**: the change along the bond of the rotation about the bond's own axis. No twist, no hop (uniform rotations; rotations about axis 1 that vary only across axis 1).
- **T2.** Couple a site rotation field `ϑ` with both terms: a coin rotation is then the shift `ϑ → ϑ + θ`. The walk is blind; the added term is hermitian and nearest-neighbour. Without the hop this fails at first order.
- **T3.** The response to `ϑ_c(x)` is `½ d(ψ†σ_cψ)/dt` — block 63's torque identity read as a balance law — zero at every site for stationary states; without the hop it is non-zero at every sampled site of an exactly stationary state.
- **T4.** At long wavelength the hop is `½ div ϑ = −⅛ ε·T`: the inversion-odd scalar that block 64's blindness *removed* from the field's energy is what the content *needs*. It is allowed because the Lattice axiom has proper rotations only.

I expected an on-site potential from the continuum; the lattice gives a hop weighted by the twist. The naive generalisation to a varying frame failed in scratch and is not claimed. **Not removed:** block 64 T5's torque, which belongs to a rotation of a site's *bonds* — a different lattice deformation from a rotation of the *coin* (they agree at long wavelength only).

**Evidence.** Runner `scripts/admissibility_rule_the_blind_walk_a_scalar_hop_weighted_by_the_twist_of_the_coin_along_the_bond_2026_09_21.py`: `TOTAL: PASS=13 FAIL=0`; 9 mutations, each failing in its own family only. Refuting pass W1–W4 with dense matrices and *finite* rotations: `|U H U† − H[sθ]|` is `1.86 s²` with the hop and `1.10 s` without; spectrum shift `2.03 s²` against `0.6 s`; a numerically found stationary state: `3×10⁻¹⁷` against up to `5.6×10⁻³`. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.

**Next.** One formulation carrying the symmetric six on the bonds and the rotation three on the sites; the fall as a consequence of the ledger's consistency.


JONbridgerapps:
**Corrigendum pushed.** The Result's item 3 and the plain-terms paragraph overstated what the blind walk removes.

- The torque that disappears for stationary states is the coin's own torque, the response to the coin's twist `ϑ` (T3).
- Block 64 T5's torque is a different quantity. It is the walker's response to a rotation of its bonds, and it is not removed. This note's Boundaries already said so.
- On the lattice, a rotation of a site's bonds and a rotation of its coin differ by an exact symbol. That is harvest issue #8659, a worker result not yet refereed by another model family.
- So a field energy blind to both rotations still has no consistent static equations with the walker.

The change is text-only. No theorem, check or number changed, and the runner output is unchanged. The pipeline and lints are clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `fb37a985ba166358495f7878d06a41776f60744d` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block65-the-blind-walk-a-scalar-hop-weighted-by-the-twist-of-the-coin-20260921` at `fb37a985ba166358495f7878d06a41776f60744d`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8597, opened 2026-09-21
ail66: the fall is owed by the ledger — a field energy per local tick requires the content's stress gradient to equal its weight (source-link direction, block 14)

JONbridgerapps:
## Block 66 — the fall is owed by the ledger (source-link direction, block 14; campaign block 10)

Hand-off PR against `main` (independent of the open stack; blocks 54, 55, 60, 63, 64 are restated, not required). Nothing new is supplied; nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** The decision record (#8572) lists clause B (the phase keeps local time ⇒ force = energy × gradient, block 54) and clause C with block 55's matched pulls (the source is the energy density) as separate supplied items. Blocks 60–64 built a field energy counted per local tick, from the strains of a frame, blind to relabellings. Given that field energy, is the law of fall still a separate clause?

**Answer.**
- **T1 (continuum, exact symbolic).** For `F = ∫e^u D(e)`, `D` *any* member of block 64's family: `E_j^a∂_be^j_a − ∂_a(E_j^ae^j_b) + U∂_bu = 0` identically — invariance under relabellings with the rates carried along. Checked through second order in nine strain functions and the log rate (third order of the ledger: where the fall first appears) for three members; it fails without the rates' term. It does not need the curvature member.
- **T2.** With the rates' constraint `U = −𝔢` and the strains' equations, *any* content coupled through such a ledger must have a stress gradient equal to `𝔢 ∂u` — its weight. **What sources the clocks is what is pulled; a content with no stress gradient cannot be static.** Block 54's force law and block 55's matched pulls are one consistency condition.
- **T3 (lattice, exact, every state).** `i[φ, S_j] = −C_j[d_jφ]`: a relabelling makes a *weighted hop* of a rate, not a rate. `d⟨G_ξ⟩/dt = Σ(d_aξ_j)J_a^j[φψ] − Σξ_jf_j` with a closed-form force density; hydrostatic balance at every site on stationary states.
- **T4.** `f = 𝔢 du` at leading order in the lattice spacing (a factor `cos k` for a plane wave, because the lattice momentum is `sin k`).

**What is not said.** Clause B is *not derived*: the identity forbids inconsistent content, it does not produce clocked content. The status of the force law changes from supplied to required. T1 has no exact lattice form here; the two sides agree at leading order in the wave vector.

**Evidence.** Runner `scripts/admissibility_rule_the_fall_is_owed_by_the_ledger_a_field_energy_per_local_tick_requires_the_contents_stress_gradient_to_equal_its_weight_2026_09_21.py`: `TOTAL: PASS=12 FAIL=0` (~50 s); 8 mutations, each failing in its own family only. Refuting pass W1–W4: a falling packet — `d⟨S⟩/dt` equals minus the total force to five digits, force/weight `0.89–0.93 = cos k`; stationary states on a `6×5×4` torus — `|div J + f| < 10⁻¹⁶` with `|f|` up to `5×10⁻³`. One refuter test (on a ring) turned out to be empty — `0 = 0`, because such stationary states carry no momentum current — and was replaced. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Corrigendum pushed.** The plain-terms conclusion, "things fall because otherwise the books could not be balanced", holds in the continuum only.

- On the lattice, block 64's curl-built members, with the rates as site multipliers, are exactly unchanged by a relabelling of the strains alone, whatever the rates.
- So their identity has no rate term, and their static equations admit no content that feels a force. On the lattice these ledgers forbid the fall rather than owe it.
- This is harvest issue #8644, a worker result not yet refereed by another model family.
- The continuum statement needs a lattice member whose relabelling carries the rates. No such member has been constructed; the problem is queued on ai/probes.

The change is text-only, in the plain terms, the Boundaries and the Review record. No theorem, check or number changed, and the runner output is unchanged. The pipeline and lints are clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `c4afb022c22e1f4e677675fab6326b0f839e55e3` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block66-the-fall-is-owed-by-the-ledger-20260921` at `c4afb022c22e1f4e677675fab6326b0f839e55e3`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8598, opened 2026-09-21
ail67: a formation event in the curvature member — the ledger can always be kept; the clocks' far field then jumps at second order (source-link direction, block 15)

JONbridgerapps:
## Block 67 — a formation event in the curvature member (source-link direction, block 15; campaign block 11)

Hand-off PR against `main` (independent of the open stack; blocks 58 and 60 are restated, not required). Nothing is adopted; no gravitational claim. **The parked statistical postulate is neither used nor approached:** no odds are assumed, derived or proposed; every statement about odds is an identity in arbitrary odds.

**Question.** Fork 4 of the decision record (#8572) — does an amplitude that has formed no record source anything? — got an exact discriminator in block 58 (#8579), under block 56's clock law: one far-field coefficient, and a threshold. Block 60's curvature member (#8590) has *two* far-field coefficients and an unbounded ledger. Same question, there, exactly (bodies pinned at rest; no rule of formation assumed).

**Answer.**
- **T1.** The ledger of a walled region is `8KΣQ`. One record of charge `ΣQ` — bare energy `8KQ'(1 + Q'g_yy)` — keeps it, for **every** amplitude: no threshold. At the centre it must carry more bare energy than the amplitude had, by `8KΣΣQ_xQ_{x'}(g_yy − g_{xx'})`. Keep the bare energy instead and the ledger falls.
- **T2.** What the walls see of the *lengths* are exactly the moments of the charges. With the ledger kept the monopole does not jump; the dipole jumps by `Q'(y − X̄)`. Over *any* odds the mean jump vanishes for odds `∝ p/χ` — block 58's `p√w` at weak field — and not for `p` or `p/χ²`.
- **T3.** What the walls see of the *clocks*, `ΣQ_xw_x`, is **not** kept: it jumps at second order in the charges, `−(m' − m)/(4K)` (`1.152 → 0.871` here). The record that would keep it is another record (bare energy `22.8` against `11.6`, executed). **No record keeps both far-field coefficients**, and nothing in these fields is delayed (block 60 T5).

**Scope.** The clocks' jump is a property of bodies *pinned* at rest (`P < Q` at strong field, block 60); whether bound bodies with internal motion remove it is on `ai/probes`. Fork 4 is not decided: the note takes the side on which an unformed amplitude sources, and says what the other side gives (first-order jumps in both coefficients).

**Evidence.** Runner `scripts/admissibility_rule_a_formation_event_in_the_curvature_member_the_ledger_can_always_be_kept_and_then_the_clocks_far_field_jumps_2026_09_21.py`: `TOTAL: PASS=13 FAIL=0` (rational charges chosen first, so every number is exact; both stationarity conditions at all 125 sites); 8 mutations, each failing in its own family only. Refuting pass: brute-force stationary points of the full non-linear ledger in 250 variables, no unit-source potential in the search — every number reproduced. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `3bea45f6d8dddddd0ce44e94e083bb597b86967a` in c3f8c47a58bfba48c1d9e030d6d79ecda00a294e as part of the combined fifteen-PR rate, frame and current batch.

- Distinguish flux of 1-N from leading clock coefficient P+Q; explicit example versus general expansion with possible zero coefficient; monotone bounded record flux requires S<1/(2g_yy); no dynamical formation claim.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block67-a-formation-event-in-the-curvature-member-20260921` at `3bea45f6d8dddddd0ce44e94e083bb597b86967a`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8602, opened 2026-09-21
physics-loop: admissibility induced-law block 70 - the species-exchange maps: what each varying field becomes

JONbridgerapps:
## Block 70 — the species-exchange maps (source-link direction, block 18; campaign block 14)

Hand-off PR against `main` (independent of the open stack; blocks 54, 62–65, 68, 69 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Blocks 68 (#8599) and 69 (#8601) compared the walk's eight species through `H²` in *uniform* fields. What maps lie behind those comparisons, do they hold for fields that vary, and what do they say about the count of species?

**Answer (exact operator identities, every state, every field).**
- **T1.** `V_n = R_nU_n` — the site signs `(−1)^{n·x}` with a half turn of the coin — gives `V_nHV_n = det(D_n)·H`, in any rate field. `det D_n` is the species' sense (handedness): four of each. Site densities are unchanged; the energy density is multiplied by `det D_n`.
- **T2.** What each *varying* field becomes: rates — unchanged; frame — `ρEρ` (same lengths, mirrored angles); twist of the blind walk — `ρϑ` (the blind walk is blind for every species); reach-two strain — `BD`, a change on one side only (stretch ↔ squeeze); **reach-three strain — unchanged**.
- **T3.** With rates and reach-three strains, in any configuration: the sixteen branches (species × sign of the energy) are **two classes of eight exact copies** — same fields, same site densities (even maps; `ΘV_n` for odd ones). The inversion `x → −x` carries one class into the other in the mirror-image fields; the Lattice axiom names proper rotations only. The checkerboard `U_(111)` reverses the energy of everything that hops an odd number of steps — rates, frames, twists, reach-three strains — so their spectra are symmetric about zero. The reach-two term hops an even number: it alone breaks this. Exact witness: `tr H³ = −735/8192` on a `4×4×4` torus.

**Executed (dense matrices, random fields).** All forty identities hold as matrix identities with residual zero; every level has multiplicity two in every case; spectrum asymmetry `≤ 9e-15` for rates with reach three, frame, twist — and `1.2e-2` for reach two.

**What it means.** The species question of block 68 ("kept or removed?") becomes a question of multiplicity under the couplings that respect the maps; reach two acquires a second cost besides showing the species different lengths.

[...]

PR #8603, opened 2026-09-21
physics-loop: admissibility induced-law block 71 - amplitudes of negative energy: exact twins as test bodies, exact opposites as sources

JONbridgerapps:
## Block 71 — amplitudes of negative energy (source-link direction, block 19; campaign block 15)

Hand-off PR against `main` (independent of the open stack; blocks 53–56, 60, 70 are restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 70 (#8602): the walk's sixteen branches are exact copies as *test bodies*, and half of them have negative energy. Blocks 53–60 stated every theorem about *sources* for positive energies. What do the same clauses say when the energy is negative, and could a rule of the supplied kind exclude such amplitudes?

**Answer (exact).**
- **T1 — twins source oppositely.** `e_x[A_nψ] = −e_x[ψ]` at every site, any rate field, same site density. Block 55's ledger makes the energy density the source *with its sign* (its identity and its law do not use the sign; its T4(b) does).
- **T2 — pairs.** Pulls stay matched for either sign. A ray starting from rest accelerates at `−w∇w` on either branch, so everything falls towards slow clocks: positive pairs attract, negative pairs **repel**, a mixed pair **chases** (the positive one recedes, the negative one follows), ledger and momentum kept.
- **T3 — strong field (curvature member).** One body of either sign: `w_0 = (1 + g_0m/(2K))^{−1/2}`, `χ_0 = (1 + r)/2`. A negative body at rest has a static field with positive rates **iff `m > −2K/g_0`**; its clock rate grows without bound at the bound, where the unit-source potential is a zero mode of `−Δ + Q/χ`. The quadratic's second root solves both field equations with a *negative* rate. Many bodies: positive static rates iff that operator is positive definite (both directions proved). Block 56's member: same zero mode at `−2/(γg_0)`.
- **T4.** The symbol of the positive-energy projector is discontinuous at each of the walk's eight zeros: no rule of finite (or summable) reach commuting with translations keeps exactly the positive energies.

**Executed.** Two walkers on a ring sourcing the field that times them (block 55's control with a negative branch): displacements `+12.5, −6.0` (positive pair), `−16.7, −9.9` (mixed), `−21.3, +11.3` (negative pair); wave-vector changes cancel to `1.5e-6`, ledger to `2e-10`. Box `9×9×9`: `w_0` equals the closed form at six sizes up to `0.999` of the bound (`31.6`).

**What it hands the owner.** A fork about the *content*, not the field: which amplitudes are present. The supplied clauses are silent, and no short-range rule can sort them.

**Evidence.** Runner `scripts/admissibility_rule_amplitudes_of_negative_energy_fall_like_the_others_and_source_the_opposite_field_2026_09_21.py`: `TOTAL: PASS=12 FAIL=0`; 7 mutations, each failing in its own family only. Recorded: the gates were stopped once to correct a theorem sentence (T3(a) claimed positive lengths for both roots without restricting to negative bodies); the negative pair's repulsion was seen in the control first and then added as an exact check; the second root was examined and excluded by a check, not ignored. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Note after the owner's landing of block 54 (c3f8c47a58, 2026-09-23, supervisor).**

As landed on main, block 54 withdraws the exact force on every packet, and its T4 is a conditional ray model.
- A ray at rest, or moving across the gradient, accelerates as `−w²∇u`, towards slow clocks.
- A ray moving along the gradient has coefficient `−w² + 2v²`, so above `|v| = w/√2` it accelerates away from slow clocks.

T2's "everything falls towards slow clocks, whatever its own sign" is therefore the ray statement for rays at rest or moving across the gradient. The twins fall exactly as the originals do in either case, so the sign-symmetry content of T2, and the source side (T3 onward), are unaffected.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `e7ef26762e497345834dacc1c70f978fbd7242fc` in 7445cc7a50e2c7631d70dc8d9a065d0653ffa6ef as part of the combined thirteen-PR species, momentum and projected-state batch.

- State weak translation-invariant pair assumptions, existing-length condition for rate positivity and exact projector domain; retain exact clocked commutators and component signs; restore stationary hypotheses and omitted coframe stress terms; no universal field-solution conclusion.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block71-amplitudes-of-negative-energy-as-sources-20260921` at `e7ef26762e497345834dacc1c70f978fbd7242fc`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8611, opened 2026-09-22
physics-loop: admissibility induced-law block 76 - the filled sea's energy as the ledger's field term: chessboard of clocks invisible; the sea induces a clock stiffness, not the curvature member

JONbridgerapps:
## Block 76 — the filled sea's energy as the ledger's field term (source-link direction, block 24; first block after the fork probe)

Hand-off PR against `main` (independent of the open stack; blocks 54, 55, 63, 69, 70, 71 restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched. Context: `FORK_PROBE_source_link_20260922.md` on the decision-record branch (#8572) — a four-lens panel on the owner's forks; this block is its first-ranked probe.

**Question.** Block 71's fork: which amplitudes are present. The reading that matches established physics fills every negative-energy state (holes are positive sources; no chase). That reading also offers a completion: the sea's own energy is a functional of clocks and lengths — book it as the field's term and the field's energy is *induced*, not supplied; γ becomes a number; per-tick counting and blindness might be automatic. Does it?

**Exact.**
- **T1 — a chessboard of clocks is invisible.** The hop across a bond carries `√(w_x w_y)`; on a chessboard of fast and slow clocks every bond product is the ambient rate: `φHφ = H` at any amplitude. The walk sees the rates only through bond products.
- **T2.** Under the reach-two coupling the sea's energy is even in the strain (block 70's checkerboard map + tracelessness): no first-order response to strain.
- **T3.** With anticommuting composition the two-record energy density is the sum of the one-record densities (exact, 144 components); a hole in an occupied negative state sources positively.
- **T4.** With the measured signs, the reading's linearised static law is repulsive if the volume term is kept, attractive with `γ = 1/κ` if it is removed, and by T1 singular at the chessboard mode either way.

**Executed (tori 6³–12³).** `c₀ = −1.193` per site. Rates' polarisation gradient part `+0.0236–0.0240` per lattice `|q|²` across six modes in three directions at L = 12, converging from L = 6: **`κ = 0.095`** — the induced energy is **not counted per local tick**; it is block 56's member (half bending), with `γ_ind = 10.5`. Strain stiffnesses under reach three: `+0.0035, +0.0021, +0.0004` — ten to sixty times smaller. The chessboard mode costs `≤ 1e-13`; the corner-type modes' gradient parts equal `|c₀|/2 × |q|²_lat/12` exactly, as T1 predicts.

**Verdict.** The free massless sea does not induce the curvature member on this lattice. Fork (i) stays a genuine supplied clause. Fork (v)'s known-physics reading rests on an exchange sign the axioms do not supply (flagged in T3).

**Evidence.** Runner `scripts/admissibility_rule_the_filled_seas_energy_as_the_ledgers_field_term_chessboard_clocks_invisible_2026_09_22.py`: `TOTAL: PASS=13 FAIL=0`; 8 mutations, each failing in its own family only. Recorded: the exact zero was seen in the control first and proved after; the census caught curly apostrophes in the runner's fence and two names outside the allowed sections. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `2fc37b9f66dd04517137a7addadfeeb7f416d114` in 7445cc7a50e2c7631d70dc8d9a065d0653ffa6ef as part of the combined thirteen-PR species, momentum and projected-state batch.

- Require differentiability for derivative statements, separate declared comparator from checkerboard-null exact Hessian and its source compatibility; correct full periodic negative eigenstate. All-axis exchange survives at zero offset; corner energy magnitudes are not automatically rest masses; staggered gap uses absolute mass.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block76-the-filled-seas-energy-as-the-ledgers-field-term-20260922` at `2fc37b9f66dd04517137a7addadfeeb7f416d114`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8613, opened 2026-09-22
physics-loop: admissibility induced-law block 78 - one record per site is an interaction, not a free sea

JONbridgerapps:
## Block 78 — one record per site is an interaction, not a free sea (source-link direction, block 26; third after the fork probe)

Hand-off PR against `main` (independent of the open stack; blocks 54, 55, 71, 76 restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched. Context: `FORK_PROBE_source_link_20260922.md` (#8572), §4 item 1.

**Question.** Block 76 (#8611) probed a *free* sea of records with anticommuting composition. The axioms say one record per site at a time — a hard-core condition, with no sign under exchange. What are two of the axioms' records under that exclusion?

**Answer (exact, rings of 4–7 sites, the reduced clocked walk).**
- **T1.** The two-record generator restricted to configurations with the records on different sites has `tr H²/dim = 2/3, 3/4, 4/5, 5/6`, against `6/7, 8/9, 10/11, 12/13` for the free antisymmetric pair and `10/9, 12/11, 14/13, 16/15` for the free symmetric pair. **The exclusion is an interaction, not a free pair of either sign.**
- **T2.** Under exclusion, antisymmetric, symmetric and sign-free generators have the same `tr H²/dim` on every ring, the same `tr H⁴/dim` on 5, 6, 7 and `tr H⁶/dim` on 5, 7; the sign shows only through the parity of even rings (fourth order on 4, sixth on 6). **The exchange sign barely matters; the exclusion does the work.**
- **T3.** For two orthogonal one-record states in a varying rate field, the antisymmetric pair's energy is exactly the sum of the one-record energies under the free generator (block 76 T3) and **not** under the hard-core generator: the ledger's source of two records is not the sum of the one-record sources.

**Executed.** Full spectra on rings 4–8 (hard-core antisymmetric and symmetric coincide to `1e-15` on odd rings, differ on even ones) and on a `4×4` torus with the two-dimensional walk (free 496/528 states, hard-core 480); three records on a ring of 6: 220 → 160 states.

**What it means.** Block 76's free sea becomes a comparator. If the framework's records fill a sea, it is an interacting many-record problem, and block 76's "a hole sources as a positive body" (which used additivity) has to be re-derived for it.

**Recorded.** The first additivity test compared expectations of `H²`, a two-body operator on the pair — wrong test, caught by the runner; rebuilt with the generator itself. T2 sharpened after the control's full spectra showed the signs differing where the fourth traces agree.

**Evidence.** Runner `scripts/admissibility_rule_one_record_per_site_is_an_interaction_not_a_free_sea_two_records_under_exclusion_2026_09_22.py`: `TOTAL: PASS=10 FAIL=0`; 6 mutations, each failing in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `88a8e6752e01205879181758e70b2a47f2bad275` in 7445cc7a50e2c7631d70dc8d9a065d0653ffa6ef as part of the combined thirteen-PR species, momentum and projected-state batch.

- Supply composition and compression explicitly; preserve finite trace and counterexample scope; distinguish whole-generator from kinetic-only clocking, literal five-profile nullities from three-dimensional wall claims. Fixed-state density identity retained; antisymmetric opposite-coin scalar-momentum witness replaces ordered same-coin witness; projector noncommutation does not prove compressed-observable nonconservation.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block78-one-record-per-site-is-an-interaction-not-a-free-sea-20260922` at `88a8e6752e01205879181758e70b2a47f2bad275`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8615, opened 2026-09-22
physics-loop: admissibility induced-law block 80 - many records under exclusion: the source is the projected density; the jammed sea stiffens the clocks the opposite way

JONbridgerapps:
## Block 80 — many records under exclusion (source-link direction, block 28; fifth after the fork probe)

Hand-off PR against `main` (independent of the open stack; blocks 54, 55, 63, 76, 78 restated, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 78 (#8613): the axioms' records compose under exclusion, not freely. Which of the ledger's identities survive, and what does the interacting crowd do to the clocks — against block 76's free sea (#8611)?

**Exact (two records, ring with varying rates).**
- **T1.** The ledger's source is the *projected* pair's own energy density: the exact `u_x`-derivative of the quadratic form equals `e_x^{(2)}` at every site (not the unprojected pair's), and the densities sum to the energy (weight one). Block 55's law is about the state the records are in; exclusion changes the state, not the law.
- **T2.** The exclusion commutes with site fields (a chessboard of clocks stays invisible, block 76 T1) and with the simultaneous translation (crystal momentum conserved).
- **T3.** It does not commute with the one-step momentum: block 63's conserved momentum is not conserved for records under exclusion — the crowding pushes.

**Executed (sparse many-record ground states).** Under exclusion the free sea's filling is a record on nearly every site — a jam. On 2D tori its clock response has the **opposite sign** to the free sea's (`−0.027, −0.043` per `|q|²` against `+0.025, +0.015`) and a volume term three to four times smaller; at **half filling** the mobile crowd has the free sea's sign and size (`+0.014` vs `+0.015`). Strains through the reach-three coupling: free `+0.0002`, crowd `+0.0004`, jam `−0.0017` per `|q|²` — **the interaction supplies no lengths' stiffness either.** The 2D free sea's clock stiffness (`+0.017–0.023`) was checked against the 3D one (`+0.024`) before any sign was read; 1D was discarded (no gradient part at the longest mode).

**What it means.** The bookkeeping survives the axioms' composition rule; the sea reading does not carry over as a sea. The induction route (fork probe §3) is closed for free and for interacting records alike.

**Recorded.** The site-resolved real part was first computed wrongly (right for the total, wrong per site) and caught by the runner; the `∂/∂u` factor was caught by weight one; the note first lacked a `## Theorem T3` so the injection mutation could not bite (T2(c) became T3); the `5×3` torus was dropped from the control for time; the pipeline rejected a mid-pass edit of the control ("classifier inputs changed") and was re-run from scratch.

**Evidence.** Runner `scripts/admissibility_rule_many_records_under_exclusion_source_projected_density_interacting_sea_2026_09_22.py`: `TOTAL: PASS=9 FAIL=0`; 7 mutations, each failing in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `c7e551d185c9874915f2605740fc04a302914d11` in 7445cc7a50e2c7631d70dc8d9a065d0653ffa6ef as part of the combined thirteen-PR species, momentum and projected-state batch.

- Supply composition and compression explicitly; preserve finite trace and counterexample scope; distinguish whole-generator from kinetic-only clocking, literal five-profile nullities from three-dimensional wall claims. Fixed-state density identity retained; antisymmetric opposite-coin scalar-momentum witness replaces ordered same-coin witness; projector noncommutation does not prove compressed-observable nonconservation.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block80-many-records-under-exclusion-the-ledger-and-the-interacting-sea-20260922` at `c7e551d185c9874915f2605740fc04a302914d11`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8628, opened 2026-09-22
physics-loop: admissibility induced-law block 82 - what can gap the walk inside the qubit: no invariant term at any reach; chessboard or axis stripes on sites; alternating lengths give one rest energy; exclusion cages

JONbridgerapps:
## Block 82 — what can gap the walk inside the qubit (source-link direction, block 30; first after the rest-energy panel)

Hand-off PR against `main` (independent of the open stack; blocks 54, 77–81 restated or placed, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 81 (#8626) left rest energy without an in-framework source. A four-lens panel on the supervisor's dossier named the routes; the synthesis (supervisor's) was to settle first what a rest energy *can* be inside `M₂(ℂ)`: the complete map of backgrounds that separate the walk's two branches at its eight species points, and what the Record axiom's exclusion does to the one background the lane had in hand.

**Exact.**
- **T1 — no covariant translation-invariant term of any reach.** The rotations permute the eight corners in orbits `1, 1, 3, 3`; at each corner the stabiliser (order 24 or 8) acts irreducibly on the coin (commutant exactly the scalars). So every covariant translation-invariant generator, whatever its reach, is scalar at each species point: the branches touch there. Blocks 54 T1(d) and 77 are the reach-0 and reach-1 cases.
- **T2 — on sites: the chessboard scalar and the axis stripes, nothing else.** The backgrounds `m₀(x) + m(x)·σ` anticommuting with the walk form exactly a 4-dimensional space: `c₀ε` and `c_j(−1)^{x_j}σ_j` (a content read by the coin that leans along one axis and alternates along it). The three stripes anticommute pairwise: one rest energy `√(Σc_j²)` for all eight species; mixed with `ε` the gap is `||c₀| − |c||`. Block 81 T4's "content order gives no gap" was for the uniform and chessboard contents; the stripe is the content order that does gap — and the static law never weighs it heaviest (`qp²` per site against `p³`, `q³`).
[...]

PR #8632, opened 2026-09-22
physics-loop: admissibility induced-law block 83 - alternating lengths are a relabelling: invisible to the walk and free for the ledger; a rest energy needs the bond's own amplitude

JONbridgerapps:
## Block 83 — alternating lengths are a relabelling (source-link direction, block 31; second after the rest-energy panel)

Hand-off PR against `main` (independent of the open stack; blocks 54, 62–64, 82 restated or placed, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 82 (#8628) found that an alternation of the hop amplitudes gives every species the rest energy `√3β|δ|` and named the lane's lengths as the one field-based candidate. In the lane's own couplings — block 62's site frame, blocks 63–64's relabelling strain with its family of field energies — what *is* an alternation of the lengths, and can anything in the lane select or feel it?

**Exact.**
- **T1 — invisible and free.** Through the frame, `½{e_jσ_j, S_j}` with `e_j = 1 + δ(−1)^{x_j}` is *exactly* the free walk (`(−1)^{x_j}` anticommutes with `S_j`). Through block 64's strain coupling the added operator is *exactly zero*. The alternation is the relabelling `B_a^a = d_aξ_a`, `ξ_a = −(δ/2)(−1)^{x_a}`: all curls vanish, the volume sum is `N`, so every field energy of block 64's family — curvature member included — charges nothing for it.
- **T2 — no frame or strain pattern gives a first-order rest energy.** Every term `½{X, S_j}` (any site or bond function `X`, any coin matrix) has exactly zero matrix elements inside the sixteen-state corner subspace, because `S_j` annihilates every corner plane wave on either side. The coupling that does gap reads the bond's *own* amplitude, `Σσ_a(1/2i)(t_a(x)T_a − T_a†t_a(x))`; its corner block squares to exactly `3δ²`. The lane's couplings see a bond through an average over its ends or a difference across it; for an alternation both vanish.
- **T3 — the sea never rises under an alternation, but never sees one.** For the bond-amplitude coupling the sea's energy is concave (Ky Fan) and even in `δ` (translation by one step carries `+δ` to `−δ`, exactly), so largest at `δ = 0`; on the ring of four the energies are `±|δ|, ±1` (twice each), `E_sea = −2 − 2|δ|`. In the lane's own couplings the alternation is not in the generator at all.

**Executed (control).** On `6³` the frame- and strain-coupled alternations differ from the free walk by exactly `0` in every matrix element; the bond-amplitude one gaps at `√3δ`. A smooth random strain through block 64's coupling has corner matrix elements `~10⁻¹⁸` and moves the sixteen zero modes by `4·10⁻⁵` (species points shifting off the grid): no gap. The sea on `8³`: `−1.190 → −1.732` under the bond-amplitude alternation, constant to every digit under the frame-coupled one.

**What it means.** The rest-energy panel's leading route (decision record row 48) closes on the lane's own terms: an alternation of the lengths is a coordinate choice — invisible to the walk, free for the ledger — and cannot be selected or felt. Block 82's alternating-length rest energy needs a coupling that reads a bond's own amplitude, which is not a relabelling's deformation and is not in the lane. Whether the axioms admit such a coupling is the owner's question. Remaining in-framework route: the two-record exclusion problem on `4³`.

**Evidence.** Runner `scripts/admissibility_rule_alternating_lengths_are_a_relabelling_invisible_and_free_rest_energy_needs_the_bonds_own_amplitude_2026_09_22.py`: `TOTAL: PASS=14 FAIL=0`; 8 mutations, each failing in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Corrigendum (same day, before review; being pushed now — text only, no theorem, check or number changes).**

The note's first version said that a coupling reading a bond's own amplitude "is not in the lane". That is wrong: block 59 (#8581) gives every bond its own crossing rate `c_b` and its generator `H = Σ_b c_b h_b` times the hop across each bond by that rate — exactly the bond-amplitude coupling — with each bond rate sourced by the hop energy it times (block 59 T2). So blocks 82 T3(a) and 83 T2(b) apply to block 59's bond rates directly: an alternation of the bond rates gives all eight species the rest energy `√3β|δ|`.

What the note establishes stands: the lengths of blocks 61–64 (site lengths; frame; relabelling strain) cannot do it — an alternation there is a relabelling, invisible and free. The next block asks whether block 59's bond rates alternate spontaneously (the sea's exact gain against the bond law's alternation stiffness).

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `3944a9633f3ef9e209645753d93109f834089d10` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

- Retain finite graph weights and scoped symbols; supply reflection-positive parameter conditions; restore large-field gaps, two-site zero operator and scalar corner-zero exceptions; first-order corner projection does not exclude higher orders; low-range cancellation does not cover reach-three strain.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block83-alternating-lengths-are-a-relabelling-invisible-and-free-20260922` at `3944a9633f3ef9e209645753d93109f834089d10`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8641, opened 2026-09-22
docs(campaign): clock-and-rate races, invisibility, and hole mass

jonathonreilly:
Wave C (clock-and-rate) of the TOE derivation campaign by underdetermination witnesses (design note 2026-09-13). One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**Clocks always reshape the order law and need not reach the records.** Independent exponential clocks with a covariant rate law induce the sequential race as an exact order law. The neighbour-count clock moves order masses across [1/3360, 1/252] against the uniform 1/720 — yet for the chain rule of a positive pair measure, an order-blind conditional rule, the finished-record law equals the measure exactly under the count, squared-count, inverse-count and value clocks. The mechanism is exact: race factors telescope to one for every fixed value pattern, so clocks spend no probability on values. A covariant clock leaves no record trace when the conditional rule is order-blind and the support has no holes.

**Order-sensitive rules see the clock.** The local nearest-neighbour rule fails the formation-order block's exchange identity, and the three clocks produce three distinct exact finished laws; the edge-agreement expectation separates the clocks pairwise and from the static measure. The exhibited order-blind rule conditions on every formed site while the local rule is clock-sensitive — the locality against order-blindness tension, consistent with the sister classification that local order-blind binary rules are constant.

**The hole mass reads the clock even though the rule is order-blind.** On the support-rule note's parity-role plane, the hard-support completion mass E[2^-d] is 5/8 under the uniform clock (that note's value, recomputed by two independent dynamic programmes and a from-scratch 9! enumeration) and about 0.706, 0.797, 0.580 under the count, squared-count and inverse-count clocks — four distinct exact rationals, ordered by clock eagerness: cluster-eager clocks complete checks before links can close them. Conditional on completion, the record law stays uniform on the Gauss set for every value-blind clock, so the clock is visible in how often full records exist and invisible in what they say.

**Decision points recorded, nothing adopted:** a uniform-clock clause; an eagerness clause; a value-clock clause; order-blindness of the physical rule (which erases the clock from records except through hole masses); time as record count. The realized-state primitive supplies no comparison measure and none is imported.

## Verification

`python3 scripts/clock_and_rate_covariant_race_order_laws_record_visibility_and_hole_mass_2026_09_22.py` prints `TOTAL: PASS=22 FAIL=0` (exact Fractions, 1.4 s, stdout under the 6000-character ceiling) and exits nonzero on any failure. Canonical cache included. Mutation census: 8 of 8 defect mutants caught; two further mutants diagnosed as genuine non-defects (supplied-example parameter freedom).


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `10f826fc3793f35b5da7884531d048566494e871` in 5efa36e7c357ae2a62ee586a5407f90982f6ded9 as part of the combined ten-PR kernel and formation batch.

- Real-sphere analytic domain with rational controls; orbit/infinite-order proofs and exact census. Last-site parity rule plus general per-order uniformity proof. Verify all512 dual characters and LR formed-neighbour conditional; scope exclusions to supplied conditionals.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Unsupported stronger conclusions were narrowed; original commits remain recoverable through this PR. No auxiliary calculation was silently discarded.

PR #8643, opened 2026-09-22
docs(campaign): formation-unit joints, block-chain equality, hole dial

jonathonreilly:
Wave C (formation-unit) of the TOE derivation campaign by underdetermination witnesses (design note 2026-09-13). One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**A joint law may equal a sequential law — realised exactly.** The cube partitions into two covariant site-plus-all-neighbour units (the in-window restriction of the proposed site-plus-six-neighbours unit; the antipodal map swaps them, the diagonal C3 preserves each). Under the block-chain joint law — the forming unit draws its pattern from the pair measure conditioned on all formed sites, unformed sites marginalised — the finished law equals the measure exactly in both unit orders, on the cube and the 2x3 window, just as every single-site order does. For order-blind chain content the formation unit is invisible.

**Product-fill joints are exactly visible for local rules.** When the unit's sites draw independently given the formed exterior alone, the first unit's three internal edges agree at exactly 1/2 — the correlation a sequential process builds through conditioning is erased — and the edge-agreement expectation separates sequential (about 7.877), product-fill (6.9) and grouped (about 8.007) processes from each other and from the static measure (about 8.533), all as exact rationals.

**On the constrained window the unit is a dial the hole mass reads, and completed records cannot see it.** For the support-rule note's hard-support plane, with the joint law stated as uniform on unit patterns consistent with every check the unit completes: check-aligned units and the whole-window glued unit complete with probability 1; single sites with 5/8 (the support-rule and clock-and-rate value, recomputed); the misaligned partition that leaves the shared link to form last and alone with exactly 1/2 (one variable, two completed checks, over-determined half the time — single-site formation is the misaligned extreme of this dial). In every case the completed-record law is the uniform law on the Gauss set. Together with the clock-and-rate block: for order-blind content, both the clock and the formation unit are read exclusively by hole masses — process choices change how often full records exist, never what they say.

**Decision points recorded, nothing adopted:** single-site unit; covariant-set unit with product fill; covariant-set unit with block-chain fill (chain content cannot see it); support-aligned glued units. A formation-model decision would pick a point on the dial and does not by itself amend the Admissibility sentence.

## Verification

`python3 scripts/formation_unit_sequential_vs_joint_covariant_sets_block_chain_equality_and_glued_unit_hole_dissolution_2026_09_22.py` prints `TOTAL: PASS=19 FAIL=0` (exact Fractions, 0.4 s, stdout under the 6000-character ceiling) and exits nonzero on any failure. Canonical cache included. Mutation census: 7 of 7 defect mutants caught; two further mutants diagnosed as genuine non-defects (supplied-example parameter freedom).


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `7a91226fcb050781657c717c3bf84257c23e0b39` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

- Restrict comparison; assert exact agreement expectations.

PR #8646, opened 2026-09-22
docs(campaign): record correlators, NN-sentence exclusion, rigidity

jonathonreilly:
Wave D (record-dynamics reframing) of the TOE derivation campaign by underdetermination witnesses (design note 2026-09-13). One source-only triple: bounded-theorem note, exact runner, canonical cache. No tick, clock, Hamiltonian or evolution law enters any computation.

**Correction before landing (2026-09-22, pushed to this PR):** the first version called nearest-neighbour locality an unrecorded separating clause and presented the nearest-neighbour and all-pairs laws as a witness pair "agreeing on every stated sentence". Admissibility already reads "one fixed nearest-neighbor admissibility rule", with the distribution "determined by, and varies with, the nearest-neighbor conditions". The all-pairs law violates that sentence under both readings, so the decision point is withdrawn. The runner now checks both readings exactly, and the files are renamed to match.

## The physics

**A tick-free record correlator, and it is not a Laplacian kernel.** The nearest-neighbour pair measure is an admissible static law: every single-site conditional given all other sites depends on the neighbours alone, checked on all 256 configurations. Its connected record correlator is exactly 3^-(distance) on every pair of the 8-site path. The Dirichlet Green function of the path Laplacian is exactly (min+1)(8-max)/9, which is linear. A scaling matched at distance one fails at distance two, so the mismatch is in the algebraic decay type, not a fit residual.

**The nearest-neighbour sentence excludes long-range pair structure.** The supplied all-pairs measure is covariant and order-blind, and its record correlator is distance-blind at exactly 835/867. It is not admissible, though. Under the static reading its site-0 conditional moves when only the far site flips (128/129 against 32/33). Under the formation reading its chain rule conditions on non-neighbours. A control shows the static check also rejects a single next-nearest bond. One sharper point is left open for the next block: under the formation reading, even the nearest-neighbour measure's chain rule conditions on a formed non-neighbour (5/9 against 4/9 at distance two). So which order-blind laws are admissible formation rules is a real question.

**Glued supports are a separate correlation type.** Under the uniform Gauss-set law of the parity-role plane, every one-point and two-point parity is exactly 0. Every check parity and every dual-code sum is exactly 1 at full window span, and non-dual observables vanish. This is exact long-range multi-point order from support alone, and no two-point probe sees it.

The no-law-level-state proposal remains a hypothesis, no photon or field is identified, and the transverse vector kernel is a named boundary.

## Verification

`python3 scripts/record_dynamics_correlator_kernels_nearest_neighbour_sentence_and_glued_support_rigidity_2026_09_22.py` prints `TOTAL: PASS=18 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in about 0.2 s, and prints 2834 characters of stdout. The canonical cache is included. Mutation census: all 9 defect mutants are caught, two of them only after added controls (an empty-set parity guard and a next-nearest-bond control). One further mutant is a masked non-defect.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `30fc512cb7c76d877f023fa8c94cad9851b7fa4c` in 5efa36e7c357ae2a62ee586a5407f90982f6ded9 as part of the combined ten-PR kernel and formation batch.

- Real-sphere analytic domain with rational controls; orbit/infinite-order proofs and exact census. Last-site parity rule plus general per-order uniformity proof. Verify all512 dual characters and LR formed-neighbour conditional; scope exclusions to supplied conditionals.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Unsupported stronger conclusions were narrowed; original commits remain recoverable through this PR. No auxiliary calculation was silently discarded.

PR #8657, opened 2026-09-22
physics-loop: admissibility induced-law block 85 - the crowd under exclusion also gains from an alternation of the bond rates: its ground energy never rises, and the jam is blind

JONbridgerapps:
## Block 85 — the crowd under exclusion also gains from an alternation of the bond rates (source-link direction, block 33; fourth after the rest-energy panel)

Hand-off PR against `main` (independent of the open stack; blocks 59, 76, 78, 80, 84 restated or placed, not required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Question.** Block 84's balance (bond rates alternate below a threshold stiffness) used the free sea as the source of the gain, a comparator since the axioms' records compose under exclusion (block 78). Does the crowd of records under exclusion gain from the alternation too?

**Exact.**
- **T1 — the crowd's ground energy never rises.** At every filling, translation by one site carries the compressed generator at `+δ` exactly to the one at `−δ` (checked exactly on the ring of four for one, two and three records), so the ground energy is even in `δ`; as a minimum of affine functions it is concave; hence largest at `δ = 0`. Two records on the ring of four: `−√2` at `δ = 0`, `−√218/10` at `δ = 3/10` (exact algebraic numbers).
- **T2 — the sea's own filling is a jam, which is blind.** Under an alternation no one-record energy vanishes (`E² ≥ 2δ²` in 2D, `3δ²` in 3D), so the free sea fills a record on every site; that jam's compressed generator is exactly zero. The crowd's balance is a statement about fillings below the jam.
- **T3 — a single record gains nothing** on the ring of four (its lowest energy is exactly `−1` at every `δ` tried).

**Executed (block 80's sparse machinery, 2D tori).** The crowd's ground energy per site falls under the alternation at every filling below the jam, with the sea's sign and a smaller size: quadratic coefficients `−0.19, −0.09` (`3×3`, 4 and 6 of 9 records), `−0.30, −0.12` (`4×3`, 6 and 8 of 12), `−0.03` (`4×4`, 4 of 16) against the free sea's `−0.35, −0.35, −0.24`; on `3×3` the coefficient grows with filling to five records and is zero at the jam.

**What it means.** Block 84's direction survives the axioms' composition: a crowd of actual records, none sharing a site, prefers the long-short alternation as the sea does, less strongly; the crowd's threshold stiffness lies below the sea's `χ/12` by a filling-dependent factor (a tenth to nine tenths on these tori). A lattice packed solid prefers nothing. Not computed: a three-dimensional crowd coefficient; the size of the rest energy at the crowd's threshold.

**Recorded.** A first runner ordered algebraic eigenvalues with floating-point conversions; the source scan caught them and exact comparisons replaced them.

**Evidence.** Runner `scripts/admissibility_rule_the_crowd_under_exclusion_also_gains_from_an_alternation_of_the_bond_rates_ground_energy_never_rises_jam_blind_2026_09_22.py`: `TOTAL: PASS=13 FAIL=0`; 6 mutations, each failing in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `ccc2fb1c7138c67734d3839e94f25e97cc6a9307` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

- Distinguish continuum susceptibility from finite-grid cusps; prove the restricted linear-rate sufficiency separately; defer false endpoint-minimum inference and crowd-threshold bounds; zero Hessian equality and unexamined directions remain open; compare explicitly supplied arithmetic/log constraints; retain exact spectral and unbounded-objective identities.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block85-the-crowd-under-exclusion-also-gains-from-the-alternation-20260922` at `ccc2fb1c7138c67734d3839e94f25e97cc6a9307`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8663, opened 2026-09-22
docs(campaign): order-blind NN formation, unsoldered constancy

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. This block asks what the Admissibility text itself says about the clock decision's candidate "order-blind physical rule". It uses the formation reading (a nearest-neighbour rule applied sequentially along a formation order). One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**Order-blind nearest-neighbour formation carries no correlation beyond nearest neighbours, on every alphabet.** Suppose a nearest-neighbour rule never fails and is order-blind on a window. Then every set of pairwise non-adjacent sites is independent with the one-site law: form those sites first, and none of them has a formed neighbour. On the 3-site path, the order-blind rules below give ends that are exactly independent (every pair mass 1/48², resp. 1/24²), while adjacent sites correlate at exactly 1/72.

**Order-blind strictly positive rules have one exact form, forced already on the 7-site cross.** Exchange between the centre and one leaf of the site-plus-six-neighbours window gives r(a|N) = r0(a) ∏ φ_d(a,b). The factors are bond-symmetric, φ_d(a,b) = φ_-d(b,a), and normalised on every neighbour condition. Conversely, every such family is order-blind on every window.

**Under the unsoldered reading the text forces order-sensitive formation.** The factor is the same for every direction, so two neighbours holding one value give E[φ²] = E[φ]² = 1. Hence φ = 1 and the rule is constant. Admissibility requires the distribution to *vary* with the nearest-neighbor conditions. So no strictly positive admissible rule is order-blind under this reading: two formation orders of the 7-site cross give different finished-record laws. This holds on every alphabet, including the Bloch sphere, and extends the landed binary classification. The clock decision's candidate "order-blind physical rule" is excluded here: formation order is written into the records.

**Under the soldered reading, order-blind rules that vary exist on the generic orbit and the sphere, and not on the high-symmetry orbits.**
- On the 6-, 8- and 12-point cube orbits, exact eigenline algebra shows every order-blind positive rule is constant. On the 12-point orbit, the only common zeros of the pair conditions are complex.
- A single 24-point orbit carries an explicit varying order-blind rule, built from the sign of the transverse coordinates on the caps a·d = 3.
- So do the 48-point chiral pair (chirality on the same caps) and the Haar sphere (a centred profile on disjoint polar caps {a·d > 3/4}, with nearest-neighbour correlation exactly 1/73728).

Each rule is checked exactly: covariance, bond symmetry, normalisation on all 4096 neighbour-condition classes, and one finished law across all six orders of the path. All of them are correlated at nearest neighbours only.

**What moves in the decision structure.** Locality is Admissibility text, not a decision; the record-dynamics PR #8646 was corrected in place for this before landing. The clock decision loses its order-blind candidate under the unsoldered reading with positive rules. In every reading, a record correlation at distance two or more from an admissible rule that never fails certifies order-sensitive formation or a support with holes. The static reading of the same sentence admits geometric correlations at every distance. The two readings are recorded side by side with this exact separation, and nothing is adopted.

## Verification

`python3 scripts/order_blind_nn_formation_independence_unsoldered_constancy_soldered_escape_2026_09_22.py` prints `TOTAL: PASS=27 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in about 16 s, and prints 4291 characters of stdout. The canonical cache is included. Mutation census: all 11 mutants are caught, one after an added eigenspace-dimension assertion. Two controls give the exact non-existence procedure teeth: it rejects forms that share a real zero, and it confirms the 24-point solution.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `7754a0010f19d1ab417749fb7fa838298cb2bfb1` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

PR #8664, opened 2026-09-23
docs(campaign): order-blind formation with zeros and holes

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. This block follows #8663. That PR proved that under the unsoldered reading a *strictly positive* order-blind nearest-neighbour rule is constant. This block covers the two cases it left open, zero conditionals and rules that can fail, for every alphabet. One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**Zeros don't help.** A never-failing unsoldered order-blind rule is constant on every condition that occurs. Exchange makes the adjacent-pair law symmetric. The 3-site bent path with two equal leaves then forces E[ρ(b|·)²] = E[ρ(b|·)]², so the variance vanishes and ρ(b|a) = r0(b). No positivity is used. With Admissibility's variation sentence, every never-failing admissible unsoldered rule is order-sensitive. Example: copy-when-unanimous has zeros, never fails, and varies. Its variance gap is exactly 1/4, and its two leaves differ with probability 1/2 leaves-first but never centre-first.

**Rules that can fail: strict order-blindness is exactly pair form.** Strict order-blindness means every order gives the same finished sub-law, holes included. For an unsoldered rule this holds exactly when r(a|N) = r0(a) ∏ K(a,b) with K symmetric and sub-normalised. By the first result, a varying rule of this kind must have holes. Its hole mass and its completed-record law are both independent of the formation order, so the clock cannot be read from any of its records.

**Rejection and forcing separate on the clock.** Checked on the 7-site cross over all 5040 orders:
- Binary rejection exclusion (draw a coin, fail on a clash with a formed neighbour) gives one sub-law, with completion 1/64 for every order.
- Forcing exclusion (take the opposite of unanimous neighbours, fail if they disagree) has the same completed law but order-dependent completion: 6 values, from 1 (centre first) to 1/32 (leaves first).

So the hole mass that read the clock in the clock-and-rate block belongs to forcing rules. Rejection rules' hole masses are clock-blind.

**The static pair law returns as a completed-record law.** Take the soft pair rule K(a,b) = (3/4)(1 + ab/3). It is strictly order-blind (all 5040 orders of the cross) and it varies. Conditional on completion, its record law on the 8-site path is *exactly* the static nearest-neighbour pair measure 2^(agreeing bonds), with correlations 3^-d at every distance. Its completion is (3/4)^7 for every order. So the static reading's laws are completed-record laws of order-blind formation with holes. The two readings agree on what completed records say and differ only by a hole mass. Binary rejection gives exact staggered order at every distance (completion 1/128 on the path). Six-axis rejection gives distance-2 agreement 1/5 and distance-3 agreement 4/25.

**The unsoldered admissible formation laws split exactly in two.** Never-failing rules are order-sensitive: the formation order enters the records. Order-blind rules carry holes, have pair form, and have a clock-blind hole mass. Their completed records can hold long-range correlation, paid for by a hole mass that grows with the window.

## Verification

`python3 scripts/order_blind_formation_zeros_and_holes_unsoldered_pair_form_2026_09_22.py` prints `TOTAL: PASS=14 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in about 10 s, and prints 2477 characters of stdout. The canonical cache is included. Mutation census: all 8 defect mutants are caught. One further mutant is a masked non-defect: every mean is 0 by flip symmetry, which an explicit check now pins.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `8c400e65f929e6b33c9ae268a70a6efbc969ae35` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

PR #8666, opened 2026-09-23
docs(campaign): holes as unrecorded sites make the order visible

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. This block follows #8663 and #8664, which left one order-blind escape under the unsoldered reading: rules with holes, *with failed histories dropped*. It asks what the axiom text says a hole is. One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**The text reads a hole as an unrecorded site.** Three sentences settle it. Record says "When present, a record locks exactly one admissible local possibility … A site with no record cannot be read." Qualification says "A state is a configuration of records." And a law gives an answer where its supplied condition holds. So a site with no admissible possibility carries no record, its neighbours cannot read it, and formation continues. Nothing discards the history. The drop convention behind the completion masses is a conditioning the text does not supply.

**Under that semantics, unsoldered order-blind rules are constant.** Take a rule normalised on its domain. On an adjacent pair, the later site may fall outside the domain but the earlier never does. So order-blindness puts every occurring condition inside the domain; the star extends this to every condition. The rule then never fails, and #8664's variance theorem makes it constant. Treating deficits as formation failures that leave sites unrecorded gives the same answer: order-blindness forces a constant hole rate, and then constant values.

**So every admissible unsoldered rule makes the formation order visible, with or without holes.** The clock decision's candidate "order-blind physical rule" is excluded outright. What remains is *which* order law holds. Hole locations carry the order. On the 7-site cross, #8664's clock-blind hole rules give 7 distinct finished-state laws under the text's semantics, against 1 under the drop convention.

**Combined with the previous blocks.** Under the text's semantics, order-blind admissible formation exists only under the soldered reading, on generic alphabets, with nearest-neighbour reach. So *every admissible formation law whose records correlate beyond nearest neighbours makes the formation order visible*, in both readings of covariance. The completion masses of the clock-and-rate and formation-unit blocks are the probabilities of having no unrecorded site under this semantics, so they stand.

## Verification

`python3 scripts/holes_as_unrecorded_sites_unsoldered_order_blind_rules_constant_2026_09_22.py` prints `TOTAL: PASS=10 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in 0.15 s, and prints 1810 characters of stdout. The canonical cache is included. Mutation census: 8 of 8 caught, one after pinning the soft rule's values.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `3fabd0e9fe96bfa40d9fd8fee1f0d5684d6756cb` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

PR #8667, opened 2026-09-23
docs(campaign): ice support under nearest-neighbour formation

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. The landed spin-half cubic-ice notes derive Coulomb correlations as *uniform* averages over ice configurations, from the Rokhsar–Kivelson ground state of a supplied Hamiltonian. This block asks whether nearest-neighbour *formation* produces that support and measure at all, read with the text's hole semantics (#8666: a site with no admissible possibility stays unrecorded). One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**No nearest-neighbour rule keeps the ice rule in every order.** Links are not neighbours of each other. So when a vertex's six links form first, they are independent under any rule, in either reading, and P(exactly 3 of 6 occupied) = 20p³(1−p)³ ≤ 5/16.

**Without soldering, a vertex cannot coordinate its links at all.** When the vertex forms first, a direction-blind rule gives all six links the same conditional, so they are independent and identically distributed given the vertex's record, and the bound 5/16 holds again. The maximum over 729 two-record rules is exactly 5/16.

**With soldering, vertex records can enforce the rule locally, but the order sets a defect price.** A vertex can record one of its 20 ice configurations and have its links copy. The vertex is then unrecorded with probability 0, 0, 0, 0, 1/8, 3/8, 11/16 when k = 0..6 of its links formed first.
- On two vertices sharing a link: in vertex-first order the shared link is unrecorded with probability 1/2. In link-first order both vertices are recorded only with probability 25/256. Conditional on no unrecorded site, both orders give exactly the uniform law on the window's 200 ice configurations.
- On a 24-link plaquette a *sweep* order avoids unrecorded sites entirely. But it weights the 10016 ice configurations 1/8000 or 1/12000 rather than uniformly. A record statistic separates the two laws: 62/125 under the sweep against 124/313 uniform.

**For the photon lane.** The uniform ice measure behind the landed Coulomb correlations is a completed-record law of nearest-neighbour formation only when histories with unrecorded sites are discarded, which the text does not do. The established formation routes to the Coulomb phase use soldering, plus an order law or joint formation units; otherwise defects depend on the order. Without soldering, the link-first and vertex-first orders are blocked (≤5/16). Other unsoldered orders are not classified. Those are exactly the recorded decision groups, now with exact prices.

## Verification

`python3 scripts/ice_support_formation_nn_rules_vertex_bound_soldered_vertex_records_and_conditioning_2026_09_22.py` prints `TOTAL: PASS=9 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in about 0.1 s, and prints 1945 characters of stdout. The canonical cache is included. Mutation census: 8 of 8 caught. Three were caught only after the checks were rewritten to compare sets rather than counts that coincide by symmetry.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `00ab89679274df1f33a95929ae1c976a8f4b88ad` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

- Restrict to named protocols; correct plaquette distinct-link count from 24 to 20.

PR #8648, opened 2026-09-22
docs(campaign): assembly graph with minimal decision sets

jonathonreilly:
Wave E (assembly) of the TOE derivation campaign by underdetermination witnesses. This PR is **re-assembled in place**, now as the thirteenth edition. Inputs:
- the next-steps continuation (#8663, #8664, #8666, #8667, #8669, #8670, #8671);
- the record-dynamics correction (#8646);
- registered frames (#8676);
- the static ice frame source (#8679);
- uniform ice by formation (#8686), directed ice (#8687) and relational frames (#8691);
- Coulomb measure selection (#8698) and formed flux transport (#8701);
- two planar squares and one cube (#8715), relational spiral rigidity (#8717), two adjacent cubes (#8719), joint cell units (#8720), static spiral local rigidity (#8724), linear sweep transport (#8726), the landed torus (#8727), and seven relational letters (#8729);
- multi-qubit units (#8731), planar windows of two to five squares (#8735), static global rigidity of relational letters (#8743), the global octant it needs (#8744), layer units on infinite prisms (#8740), the flux stiffness of their layer transfer (#8746), relational letters in alternating pairs (#8750) and in four-angle cycles (#8752), the sweep contrast for the cycle letters (#8756), and the rotation-covariant rule's results (#8854, #8856, #8857).

It is one source-only triple: the graph note, its verification runner and the canonical cache.

**What changed from the first edition.** The first edition put gravity behind four decisions: clock, locality, roles and soldering.
- Locality is Admissibility text ("one fixed nearest-neighbor admissibility rule"), so its decision group is withdrawn.
- The text excludes the clock group's order-blind candidate under the unsoldered reading. The group becomes "which order law".
- Roles can be supplied *or* registered by formation (#8669). The photon route can run through the static reading *or* through formation (#8667).
- **Third edition:** soldering gains an alternative in both formation routes (#8676). Coordinate letters with an order law put the lattice frame in the record. The ice support and the role letters are then read from the labels, and soldered rules are emulated.
- **Fourth edition:** the static route needs a frame source too (#8679). Read statically, the landed uniform ice measure breaks the nearest-neighbour sentence in occupation form. Vertex records that name directions (soldered records or coordinate letters) restore it.
- **Fifth edition:** the formed photon route reaches the uniform ice measure only through a coordinator (#8686). Without one, formed ice is directed and causal (#8687). Letters may be fixed, or relational spirals under possibility covariance (#8691). The minimal sets are unchanged.
[...]

PR #8669, opened 2026-09-23
docs(campaign): star constraints and soldered role registration

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. The assembly graph (#8648) left gravity behind the roles decision: the superlattice roles are either supplied next-nearest data, a window extension, or a sector clause. The landed roles note (2026-09-04) pins the parity skeleton *statically* with nearest-neighbour tables over role letters. This block asks what *formation* needs in order to register the skeleton as record content. One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**Formation enforces pair constraints, not star constraints.** The later site of each adjacent pair checks it, so pair constraints hold in every order. A star constraint jointly constrains a site's six neighbours. When the centre forms first and each neighbour sees only the centre, no site ever sees the whole star. Without soldering, those neighbours are identically distributed given the centre's record, which gives these bounds:
- the skeleton's link profile (vertex letters on the axis, plaquette letters across it) holds with probability at most **16/243**;
- Gauss parity with an odd vertex formed first holds with probability at most **1/2**;
- the ice count holds with probability at most **5/16**.

Gauss parity is absorbed when the vertex forms *last*. A soldered vertex can dictate one of its 32 consistent link sets.

**Scalar role letters form wrong skeletons without leaving a hole.** With unsoldered letters V, L, P, C, pair constraints hold, yet completed states leave the skeleton with no unrecorded site. That is up to 7/8 of completed mass on sampled cube orders, and V–L–P along one axis on a path.

**Soldered, axis-labelled role letters register the skeleton; the order law sets the price.** Here each site's letter fixes every neighbour's letter. Results on the 2×2×2 cube, over all 40320 orders:
- Completion is exactly (1/8)^(k−1) for k independent nucleations (histogram 8640 / 24480 / 5760 / 1440).
- Conditional on completion, the state is one of the 8 skeleton phases, uniformly. The phase is registered by the first nucleus.
- Connected growth always completes.
- The uniform order law completes with probability 599/2048.

**Decision structure.** Registering the roles by formation needs soldering, a single-nucleus order law, and an eight-letter alphabet. Locality is axiom text (#8646), and the clock is reduced to the order law (#8663, #8664, #8666). On the registered-roles route, the first assembly's account of gravity (clock, locality, roles, soldering) therefore becomes **soldering, the order law and the alphabet**. The second-edition assembly (#8648) computes gravity's minimal decision sets exactly. These bounds hold in the broadcast order (the centre forms first and each neighbour sees only it). Unsoldered orders that might register a frame are not classified here. *(Corrected in place: an earlier version of this PR said the four decisions "collapse to two"; that version omitted the alphabet group.)*

## Verification

`python3 scripts/star_constraints_under_nn_formation_broadcast_bounds_and_soldered_role_registration_2026_09_22.py` prints `TOTAL: PASS=11 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in 0.3 s, and prints 2100 characters of stdout. The canonical cache is included. Mutation census: 8 of 8 caught, one after pinning the exact 7/8 share.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `616ba568037e6676bbfd29288a13bbe73170e283` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

- Add continuous derivative, parity and p(1-p) proofs.

PR #8670, opened 2026-09-23
docs(campaign): order laws for extended soldered supports

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. #8669 reduced the decisions behind downstream gravity to the order law and soldering. It also found that a soldered star-constrained support (the parity-role skeleton) completes with probability exactly (1/8)^(k−1) for k independent nucleations. This block asks which order laws can form such supports on large windows. One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**Independent clocks nucleate everywhere.** Under independent identical clocks (the uniform race), a site nucleates exactly when it forms before all its neighbours, with probability 1/(1 + degree). Over all orders, E[k] is exactly 2 on the 5-site path, 38/15 on the 3×3 plane and 2 on the cube, which is 1/7 per site on Z^3. Sites with disjoint neighbourhoods nucleate independently. That bounds the skeleton's completion by 8(7/8)^m − 7(6/7)^m → 0, below 1/100 at m = 60. **Independent local clocks cannot form extended soldered supports.**

**Defect-free formation needs formation without a first event.** The Lattice axiom says no site is privileged, so an order law on Z^3 is translation-invariant. Its nucleations then form a translation-invariant random set, which is either empty or has positive density. Completion on growing windows therefore needs an order law in which *every* formation is preceded by a neighbour's, and *none* is first.

**Such laws exist and are covariant.** Lexicographic sweeps over the 48 signed frames have one nucleation per cube, and their uniform mixture is rotation-invariant although no single sweep is. Under every sweep the skeleton completes with certainty, with the same completed law for all 48 frames. On Z^3 a sweep has no nucleation at all. Each realisation carries a frame of lattice directions that the completed records do not show.

**For the order-law decision:** the route to extended soldered supports (roles, ice) is a front-like order law with no first formation. Which one holds, and whether every such covariant law fixes a direction in each realisation, remain open.

## Verification

`python3 scripts/order_laws_nucleation_density_and_sweep_laws_2026_09_22.py` prints `TOTAL: PASS=10 FAIL=0` and exits nonzero on any failure. It uses exact Fractions, runs in about 0.9 s, and prints 1564 characters of stdout. The canonical cache is included. Mutation census: 7 of 7 defect mutants caught; one relabelling non-defect (a reversed frame is another frame).


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `0945593a2c8137d75f53a073c03a7defe3ad0f29` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

- Withdraw broad exclusion; preserve iid bound; no infinite process existence claim from a non-well-founded order.

PR #8671, opened 2026-09-23
docs(campaign): the soldering menu, four cubic actions on qubits

jonathonreilly:
Next-steps campaign after the TOE derivation campaign by underdetermination witnesses. In the second-edition assembly (#8648), every established formation route to downstream gravity uses the soldering decision. The possibility-covariance block treated soldering as a two-way fork. This block asks how many readings the axiom text actually admits, and what each lets formation build. One source-only triple: bounded-theorem note, exact runner, canonical cache.

## The physics

**The covariance sentence admits exactly four actions on possibilities.** Admissibility requires covariance under proper cubic rotations but doesn't say how a rotation moves a qubit's possibilities. That action is a homomorphism from O (≅ S4) into the qubit's automorphisms (SO(3) on the Bloch ball). Up to conjugacy there are exactly four:
- **trivial**: the unsoldered reading;
- **a sign twist**: diag(1, s, s), with s the sign of the axis permutation;
- **axis soldering**: s(g)|g|, which permutes the three axes;
- **full soldering**: g itself.

Proof: the S4 character table, plus the fact that a real 3-dim representation's determinant is sign^(#A2 + #E + #T2). Exactly four of the eight splittings have determinant 1; T2 and A1+E fail on a quarter-turn.

**Each action blinds a site to some bond directions.** Kernel elements fix every possibility but move bond directions. The kernels have orders 24, 12, 4, 1 and leave 1, 1, 3, 6 orbits of directions. So a site can send one instruction per orbit to neighbours formed after it. Exact bounds (trivial / sign twist / axis / full):

| constraint | trivial | sign twist | axis | full |
|---|---|---|---|---|
| ice count (3 of 6) | 5/16 | 5/16 | 1/2 | 1 |
| odd Gauss parity | 1/2 | 1/2 | 1/2 | 1 |
| parity-role link profile | ≤16/243 | ≤16/243 | certain | certain |

**What the menu decides, in the broadcast order** (the centre forms first and each neighbour sees only it):
- Registering the parity-role skeleton needs at least *axis* soldering.
- Forming the ice support (the photon route) by vertex broadcast needs *full* soldering.
- Gauss parity by broadcast also needs full soldering; a checker formed last absorbs it under any action.
- The sign twist behaves like the unsoldered reading throughout.

Orders in which neighbours also see other formed sites and relative geometry are not classified here. In those orders an unsoldered rule might register a frame of its own.

## Verification

`python3 scripts/soldering_menu_four_cubic_actions_on_qubit_possibilities_and_formation_bounds_2026_09_22.py` prints `TOTAL: PASS=10 FAIL=0` and exits nonzero on any failure. It uses exact Fractions and integer matrices, runs in 0.5 s, and prints 1852 characters of stdout. The canonical cache is included. Mutation census: 8 of 8 caught, two after strengthening.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `a9acb94bf0479b80b59aa35295477b66a9edf8cd` in cb2e2e6132b9dac501a24662496b74cd2dee12bb as part of the combined eight-PR formation/covariance batch.

- Add continuous derivative, parity and p(1-p) proofs.

PR #8650, opened 2026-09-22
Physics: generated rotor fields and prepared transverse waves

jonathonreilly:
This extends #8635 from a spin-half ring model to a compact rotor with electric and magnetic terms generated by the same hard-core record hopping. A separate compact-state residual proof then gives prepared transverse harmonic waves on a fixed finite box, followed by an explicitly ordered continuum limit on a finite list of modes.

The construction supplies integer-spin link memory, a selected Gauss background, local dressed preparation, quantum hopping and a declining formation rate. Microscopic scales grow with spin. The wave result additionally selects zero global electric flux, a small-angle packet and a fixed wave/time normalization. It does not establish a thermodynamic Coulomb phase, natural vacuum selection, positive limiting record production, native implementation, full QFT or TOE closure.

Both units have separately sealed independent derivations and bounded author-source comparisons, with no substantive correction requested. Exact finite coefficients, weighted cutoff controls, compact-packet propagation and preserved countercontrols are included. The publication package verifies 83 byte-identical artifacts, 89 selected local seal bindings and the 129 inherited artifacts; all 17 added Python sources compile. Scientific numerical outputs were authenticated, not all re-executed during packaging.

Start with .claude/science/mobile-record-formation-20260920/campaign12h_third/ROTOR_WAVE_PUBLICATION_README.md. Third-party full texts are omitted with URL/hash/read receipts retained. Frozen source-time “review pending” labels are superseded only by the included bounded comparisons, not by formal retained status. This is a draft stacked on the local-field PR; newer charged-matter work is outside its scope.

PR #8672, opened 2026-09-23
Physics: repeated record formation and generated field dynamics

jonathonreilly:
Repeated record formation needs a dynamical description that retains both the field interactions and the matter motion created by births. This change supplies two controlled limits for a specified quantum lattice model, with explicit boundaries between them.

- Balanced formation: a finite local Lindblad generator with O(epsilon) full-density error, a fixed-volume rotor limit, and exact successive-refill probabilities. The same four-leaf center refills twice with eventual probability 11/15 for coherent birth channels and 2/3 for resolved channels; exact dark states account for the rest.
- Field interactions with finite formation: a different scale retains rapid second-order matter motion, finite fourth-order field interactions and finite birth jumps. A full-cluster spectral rotation and dissipative corrector prove a spin-independent fixed-volume O(epsilon) approximation. On the initially vacant-B cubic sector, a joint large-spin limit gives generated electric and magnetic field dynamics until the first formation, with an exponential first-event clock and controlled time/mark probabilities.

- Post-event fast motion: a self-contained six-site calculation gives a uniform-spin microscopic approximation on compact fast times and an exact vacancy probability from a physical infinite-path reduction. A separate source-informed review reconstructs the actual formation output, field winding, complete finite-spin sectors and probability law. No later birth is possible in this one-vacancy control.

- First-event information: each retained first mark acts isometrically on the initial rotor-field information. An exact degree-three cyclic control checks all 18 channel/recovery maps. Independent review required explicit zero-rate handling and removal of an overly broad matter-erasure loss statement; both repairs are preserved and acknowledged. The recovery is logical, not a supplied physical reversal of record formation.

Independent validation: the balanced/locality notes have a separate-context blind reconstruction and source comparison. The higher-scale note has a completed source-informed review. Neither required a correction. Exact complete-sector coefficients, embedding identities, locality and dark weights were independently checked. Matrix-unit controls test the large off-block source and its cancellation. Selective microscopic endpoints agree within 7e-14 and 4.4e-15; full trajectory grids were not independently rerun. All 186 selected artifacts authenticate, 33 Python sources compile, and inherited publication verifiers pass.

Scope: quantum degrees of freedom, background, Hamiltonian, preparation and reservoir are supplied. The full target retains rapid matter motion after births; only the pre-first-event sector has the field-only limit. A persistent photon regime with repeated formation, finite-energy fuel, volume-uniform microscopic error and native derivation remain open. This is not formal retained/audit status. A historical different-scaling three-site companion is included for shared-seal provenance but was authenticated only. Preserved whitespace exceptions belong to sealed historical source/diff bytes and are documented explicitly.

Stacked on #8650. Start with `.claude/science/mobile-record-formation-20260920/campaign12h_third/REPEATED_FORMATION_PUBLICATION_README.md`, which links the theorem notes and their actual independent reports.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `fe6dc2c5ef061fa1e0051063d49178f23b872c13` in 9d8596c758f116c572bcc5ee46cfe58ec07701da as part of the combined four-PR mobile record foundations batch.

- Publish complete13 selected arguments with explicit model assumptions, accurate scopes and canonical runners. Preserve all523 original path dispositions and recovery branches;490ancillary paths remain deferred, not silently discarded.
- Run the reviewed controls in fresh temporary directories in declared order and include generated scientific JSON in authenticated stdout. Everyhelper is bound as an actual runtime input; all13executions passed.
- Place21support scripts at discoverable paths; declare static TYPE_CHECKING imports, execution timeouts and concrete inputs. Confirm packet and graph helper sets agree for all13notes and refresh evidence after the fix.
- Keep only the used geometry/divergence functions; preserve excluded cooling source on the recovery branch. Add exact cube circulation-basis Gram and logical image-recovery checks, including the zero-rate qualification. No physical record reversal or general quantum postulate is inferred.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `codex/mobile-record-repeated-formation-20260923` at `fe6dc2c5ef061fa1e0051063d49178f23b872c13`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8676, opened 2026-09-23
docs(campaign): registered frames carry the lattice frame unsoldered

jonathonreilly:
## Summary

A rule under the unsoldered reading never sees a bond direction. This block
shows that its letters can still make the finished record carry the lattice
frame, and then carry what the ice and role routes needed soldering for.

- **Coordinate letters.** Letters (c, t) with c in Z_4^K and a small tag.
  A record carries a frame when c(x) = c0 + sigma(x) mod 4, with sigma sending
  the three lattice directions to generators of different axes.
- **Sweeps force the frame.** In a sweep every site forms after its three
  back-neighbours (open PR 8670). On Z^3 the complete stationary records of the
  three-neighbour rule are exactly the frames aligned with the sweep. The proof
  uses a plaquette lemma and a frame lemma. A complete search on the torus
  Z_4^3 finds exactly 6 such records with c(0) = 0.
- **What a frame carries.** A link read as occupied when its odd label
  coordinate is 1 gives the ice support at every vertex. The label parity
  gives the role letters. An emulator reproduces the soldered ice
  vertex-record rule of open PR 8667 exactly in each frame.
- **Price at a first formation.**
  - In corner growth the corner's three forward neighbours are i.i.d. under
    any unsoldered rule, so the frame costs at most (m-1)(m-2)/m^2. For
    coordinate labels that is (K-1)(K-2)/K^2, and the corner rule attains it:
    2/9 at K = 3 and 3/8 at K = 4.
  - From a centre the frame costs at most 1/972.
  - A designed order reaches the frame with probability 1.
- **What moves.** The soldering decision gains an alternative: the alphabet
  (coordinate letters, 320 letters at K = 3) with the order law. Tonight's
  soldering-necessity results (PRs 8667, 8669, 8671) still stand in the
  broadcast order.

## Files

- `docs/REGISTERED_FRAMES_COORDINATE_LABELS_SWEEP_RIGIDITY_CORNER_PRICE_AND_SOLDERED_EMULATION_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/registered_frames_coordinate_labels_sweep_rigidity_corner_price_and_soldered_emulation_2026_09_22.py`
- `logs/runner-cache/registered_frames_coordinate_labels_sweep_rigidity_corner_price_and_soldered_emulation_2026_09_22.txt`

## Verification

```bash
python3 scripts/registered_frames_coordinate_labels_sweep_rigidity_corner_price_and_soldered_emulation_2026_09_22.py
```

- Result: `TOTAL: PASS=16 FAIL=0`, stdout 2435 characters, about 1.3 s,
  exact Fractions.
- Mutation census: 14 of 14 defect mutants caught.
- Covariance: every arising configuration was checked under the 24 rotations.

No rule, reading, alphabet or order law is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `b68cf70d9d330f563d0f22a624b3cf96200f3ef3` in e5b39ea4e232e1181acfa49b25f7b22a39912489 as part of the combined four-PR frame/ice-model batch.

- State K=3 in the scope, theorem and runner; retain K=4 corner counting separately; restrict designed-order claim to tested boxes.
- Supply simplex averaging argument and declare shifted role phase for the emulated star.

PR #8679, opened 2026-09-23
docs(campaign): static ice measure needs a frame source

jonathonreilly:
## Summary

The assembly's static route to the ice support takes the landed uniform ice measure as the record law. Whether that law meets the nearest-neighbour sentence under the static reading was a declared open edge. This block settles it. All results are exact, on the fine torus Z_4^3, which is the landed note's L = 2 torus: 9600 ice configurations, 880 of them at zero flux, both counts reproduced.

- **Occupation form fails.** A link's conditional given all other sites is fixed by next-nearest links: the other five links at either endpoint. If vertex, plaquette and cube records carry nothing about the links, every link takes both values under identical neighbour records. This holds in all states and in the zero-flux sector alone.
- **Soldered vertex records restore it.** Let each vertex record the set of occupied bond directions. Then every site's value is a function of its six neighbours' values, the link marginal stays uniform, and the support is invariant under the 24 rotations acting on positions and records.
- **Records that cannot tell a vertex's links apart fail.** Under the unsoldered reading this happens on one vertex star. It also happens on the 2x2x2 torus when plaquette and cube records are constant, whatever the vertex alphabet: the two links along an axis join the same vertex pair, and every ice state has an axis whose two links differ.
- **Coordinate letters (#8676) restore it without soldering.** An exhaustive plaquette lemma shows the static frame rule is rigid, so the complete static records are the 48 signed frames. Six neighbour labels fix a site's label. Every site's value is a function of its neighbours' values up to rotation, checked on 1228800 site-state pairs. Ice, link-vertex agreement and roles are read from the labels.
- **What moves.** The static route needs a frame source too: soldering with supplied roles, or coordinate letters with roles read from the labels. With the formation routes (#8676), every recorded route to the ice support needs soldering or the coordinate-letter alphabet. The assembly (#8648) recomputes gravity's minimal sets.

## Files

- `docs/STATIC_ICE_MEASURE_NEAREST_NEIGHBOUR_ADMISSIBILITY_NEEDS_A_FRAME_SOURCE_SOLDERED_VERTEX_RECORDS_OR_COORDINATE_LETTERS_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/static_ice_measure_nearest_neighbour_admissibility_needs_a_frame_source_2026_09_22.py`
- `logs/runner-cache/static_ice_measure_nearest_neighbour_admissibility_needs_a_frame_source_2026_09_22.txt`

## Verification

```bash
python3 scripts/static_ice_measure_nearest_neighbour_admissibility_needs_a_frame_source_2026_09_22.py
```

`TOTAL: PASS=13 FAIL=0`, 2319 characters of stdout, about 57 s, exact integer arithmetic. Mutation census: 11 of 11 defect mutants caught (ten by a FAIL line, one by a nonzero exit). No reading, record scheme or alphabet is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `16abf2925d9fb897feb27fe3477353492e483524` in e5b39ea4e232e1181acfa49b25f7b22a39912489 as part of the combined four-PR frame/ice-model batch.

- Assert 9600; disclose rotation subsampling and source covariance proof; require joint independence for auxiliary-record extension; explain why full link conditionals are point masses.

PR #8686, opened 2026-09-23
docs(campaign): uniform ice by formation needs a coordinator

jonathonreilly:
## Summary

The landed Coulomb correlations are uniform averages over ice configurations. #8667 reached the uniform ice measure by formation only by conditioning, and #8666 found that conditioning is not in the text. This block asks whether any nearest-neighbour formation law reaches the uniform measure with no unrecorded site. Everything is exact.

- **The square's loop blocks it for local records.** The window is one square: 4 vertices and 20 links, with 10016 ice configurations (tr T^4, T = [[4,6],[6,4]]). Records are local: a vertex's record is equivalent to its six links, and a link's record is its occupation. Then no order and no rule reproduces the uniform measure without an unrecorded site. The proof:
  - Take the last-formed of the square's 8 sites. Everything formed before it spans a forest, so the law formed so far is a directed Markov law on a forest.
  - Between that site's two neighbours the forest has one path. If the path has a collider, the ends are independent; if it has none, they are independent given any inner site.
  - Under the uniform measure the ends are dependent both marginally and given the middle site, for all 8 choices of last site, so neither case fits.
  - #8667's sweep is one instance: it completes, but with masses 1/8000 or 1/12000.
- **The plaquette site coordinates the square exactly.** The plaquette site is the nearest neighbour of the square's four links. It forms first and records their pattern, drawn from the measure's marginal, a law invariant under the square's 8 symmetries. The links copy it and the vertices fill in uniformly. The result is exactly 1/10016 per configuration, with no unrecorded site. The coordinator's record names its links, so it needs a frame source (#8676, #8679).
- **The coordination does not simply extend.** On two squares sharing a link (501632 configurations), each square's other links depend on the other square's given the shared link. So a second coordinator that sees only the shared link cannot supply the law.
- **Plans also work, at a price.** Records carrying a whole configuration reproduce the uniform law, but that content is non-local.
- **What moves.** The ice support is formed with a frame source. The uniform measure behind the Coulomb correlations needs more: a face coordinator (verified on one square), non-local records, or conditioning, which the text does not supply. Multi-plaquette coordination is an open edge of the formed photon route.

## Files

- `docs/UNIFORM_ICE_BY_FORMATION_THE_SQUARE_LOOP_OBSTRUCTION_THE_PLAQUETTE_COORDINATOR_AND_PLANS_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/uniform_ice_by_formation_square_loop_obstruction_plaquette_coordinator_and_plans_2026_09_22.py`
- `logs/runner-cache/uniform_ice_by_formation_square_loop_obstruction_plaquette_coordinator_and_plans_2026_09_22.txt`

## Verification

```bash
python3 scripts/uniform_ice_by_formation_square_loop_obstruction_plaquette_coordinator_and_plans_2026_09_22.py
```

- Result: `TOTAL: PASS=9 FAIL=0`, 1530 characters of stdout, about 8 s, exact Fractions.
- Mutation census: 12 of 12 defect mutants caught.
- Two relabellings were diagnosed as non-defects. Both are symmetries of the law.

No rule, record scheme or order law is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `b650036ea7c055066a8f2034fb6c1c6e982111b1` in e5b39ea4e232e1181acfa49b25f7b22a39912489 as part of the combined four-PR frame/ice-model batch.

- Run actual plan propagation through the connected formation order; remove unsupported partial-record exclusion.

PR #8687, opened 2026-09-23
docs(campaign): directed ice from sweep formation

jonathonreilly:
## Summary

Formation reaches the ice support along a sweep, with no unrecorded site (#8667, #8670) and with a frame source (#8676, #8679). With local records it cannot reach the uniform ice measure (#8686). This block asks what the sweep-formed ice looks like at long distances. The answer is exact: it is directed and causal.

- **The rule is linear in the mean.** In arrow form the ice rule is 3 in and 3 out at every vertex. A vertex that picks its forward arrows uniformly among the consistent ones gives each forward arrow conditional mean B/3, where B is the back sum. The conditional covariance is (3/2)(1 - B^2/9) times the sum-zero projector.
- **Flux fluctuations walk forward.** The mean response to one back-link is exactly the multinomial random walk with steps e_1, e_2, e_3. Each layer carries mass 1, the transverse variance is 2n/9, and the peak on the diagonal decays like 1/m. The symbol D(k) = 1 - (1/3) sum e^(-i k_j) vanishes only at k = 0 and expands as (i/3) sum k_j + (1/6) |k|^2 + ...: first order along the diagonal, second order across it. That is directed diffusion.
- **The formed measure is causal.** On a 4x4x4 box with independent uniform inflow, second moments close exactly. Every back sum has variance 3, and same-layer arrows are uncorrelated. The covariance of any two arrows is R/3 inside a forward cone and 0 outside it; this is checked for all 192 x 192 forward pairs.
- **The uniform measure has no direction.** On the landed 2x2x2 torus (9600 states) its arrow covariance is invariant under all 48 cubic symmetries.
- **What this means for the photon lane.** The sweep keeps the Gauss law exactly, but it forms a directed phase, not the landed isotropic Coulomb phase. The uniform measure has to be restored by a coordinator (#8686) or by conditioning.

## Files

- `docs/DIRECTED_ICE_SWEEP_FORMATION_MAKES_ICE_CORRELATIONS_CAUSAL_FLUX_PROPAGATES_AS_A_DIRECTED_RANDOM_WALK_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/directed_ice_sweep_formation_flux_propagates_as_a_directed_random_walk_2026_09_22.py`
- `logs/runner-cache/directed_ice_sweep_formation_flux_propagates_as_a_directed_random_walk_2026_09_22.txt`

## Verification

```bash
python3 scripts/directed_ice_sweep_formation_flux_propagates_as_a_directed_random_walk_2026_09_22.py
```

- Result: `TOTAL: PASS=12 FAIL=0`, 2194 characters of stdout, under a second, exact Fractions.
- Mutation census: 12 of 12 defect mutants caught.
- One further mutant was diagnosed as a non-defect: its constant multiplies VB - 3, and every back sum has variance 3, so that factor is always 0.

No rule, order law or physical identification is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `d07822fa6381db1ee466e0e1b0ff1d156f53a357` in e5b39ea4e232e1181acfa49b25f7b22a39912489 as part of the combined four-PR frame/ice-model batch.

- Give Stirling/ratio proof, per-coordinate 2n/9 and total 2n/3, reciprocal periodicity, and distinct-arrow covariance scope.

PR #8691, opened 2026-09-23
docs(campaign): relational frames in one qubit via spiral records

jonathonreilly:
## Summary

Coordinate letters (#8676) are one of the two frame sources behind every route to gravity (#8648, fourth edition). The Qubit axiom says "No possibility is privileged". Read as possibility covariance, a rule must commute with every rotation of the Bloch sphere, so fixed letters are unavailable. This block asks whether the frame can still live in one qubit's own possibilities. All arithmetic is exact and rational.

- **Relational letters exist.** Spiral records b(x) = R_n(t . x) b0 are built from three Pythagorean rotations, with (cos, sin) = (3/5, 4/5), (5/13, 12/13) and (8/17, 15/17). The six signed angles are distinct, and the positive ones obey the plaquette lemma.
- **Spirals are stationary in a sweep.** The spiral rule reads a great circle and orientation from three back-neighbour values and returns the unique point that each value reaches by one of the rotations. It is symmetric in its inputs, so it is unsoldered, and it commutes with rotations of the sphere. It reproduces a spiral and a rotated spiral at every site of a 4x4x4 box.
- **The record carries the frame.** Each core site reads every neighbour's bond direction from the rotation between their values. Ice, roles and emulation then follow as in #8676 and #8679. Under the static reading each core site is the unique covariant completion of its back neighbours, and also of its forward neighbours.
- **First formations miss it.** A lone child's law is symmetric about its parent's value, so its only atoms are the parent's poles, and spiral values never sit there. Two stationary spirals that agree at a parent differ at its child. So after the first lone child, spiral registration at a first formation fails almost surely.
- **What moves.** Under possibility covariance the alphabet routes survive inside one qubit, exactly where the assembly puts them: in the sweep and under the static reading. The first-formation prices of #8676 need fixed letters.
- **Not established.** Rigidity (whether every stationary sweep record is a spiral) is left open.

## Files

- `docs/RELATIONAL_FRAMES_IN_ONE_QUBIT_SPIRAL_RECORDS_UNDER_POSSIBILITY_COVARIANCE_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/relational_frames_in_one_qubit_spiral_records_under_possibility_covariance_2026_09_22.py`
- `logs/runner-cache/relational_frames_in_one_qubit_spiral_records_under_possibility_covariance_2026_09_22.txt`

## Verification

```bash
python3 scripts/relational_frames_in_one_qubit_spiral_records_under_possibility_covariance_2026_09_22.py
```

- Result: `TOTAL: PASS=7 FAIL=0`, 1402 characters of stdout, under a second.
- Mutation census: 9 of 9 defect mutants caught.
- One further mutant removes a redundant guard and was diagnosed as a non-defect.

No reading, rule, alphabet or order law is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `33a1714a8efbcc0bf136b58523248f5d3660b293` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

- Declare full real sphere, fixed finite candidate targets after two formed values, local conditional laws and face-connected sequential growth from one seed; preserve the valid symmetry bound.

Issue #8631, opened 2026-09-22
probe HIT: J:derive:lightcone-long-range-order:a1


JONbridgerapps:
Task `J:derive:lightcone-long-range-order:a1` (derivation, round 2, attempt 1 of 5), worker `w-jonathonsmac4f50-j4031` (claude-opus-5-5).

**Files on `ai/probes`:**
- `probes/work/derive/lightcone-long-range-order/w-jonathonsmac4f50-j4031/ATTEMPT.md`
- `probes/work/derive/lightcone-long-range-order/w-jonathonsmac4f50-j4031/check.py`
- log `logs/probes/J:derive:lightcone-long-range-order:a1/w-jonathonsmac4f50-j4031__8c34881f__20260922T203536Z.json` (`TOTAL: PASS=6 FAIL=0`, `CHECK PASS`)

**HIT line:**
```
HIT: the reflection that swaps the layers of the doubled graph is reflection positive with the slab halves (Gamma = bilayer torus under (x, a) -> (x, a XOR parity(x)), exact) and NOT with the layer halves (crossing matrix I + A_nn, <d, M d> = -64 on a cube); with the slab halves, long-range order of pi for beta > (3/2)(I0 + I2) = 0.590494 (re-derivation of a5's threshold, same model family)
```

**What is new here.**
- The layer swap `(x, a) ↦ (x, 1 − a)` with the two **layers** as halves is not reflection positive for any `β > 0`. Every edge crosses it, and the crossing matrix is `M = I + A_nn` (the 7-stencil), whose eigenvalue at `(π,π,π)` is −5.
- A staggered difference on a 2×2×2 cube gives `⟨d, Md⟩ = −64` exactly, so a 2×2 kernel minor is negative.
- So the slab halves `{(x,a): a = parity(x)}` are necessary for the reflection used by #8463 (a5) and by `w-macbookpro90c72-j451b`. Referees of those attempts should check that they use these halves.

**What is re-derived (the route of #8463 and j451b, which I read after my own plan failed).**
- `(x, a) ↦ (x, a ⊕ parity(x))` is an exact isomorphism of Γ onto the bilayer torus (edge by edge, L = 4 and 6).
- Its bond-plane and slab reflections have only mirror-pair crossing edges and cover every edge. So Gaussian domination holds for all fields.
- The spectrum `{E, 14 − E}` equals `{E, E + 2}`.
[...]

Issue #8463, opened 2026-09-20
probe HIT: J:derive:lightcone-long-range-order:a5


JONbridgerapps:
Science triage of a derivation HIT.

- log: `logs/probes/J:derive:lightcone-long-range-order:a5/w-jonathonsmac4f50-ja59c__23d218ff__20260919T160933Z.json`
- work: `probes/work/derive/lightcone-long-range-order/w-jonathonsmac4f50-ja59c/` (`ATTEMPT.md` + `check.py`, ai/probes 23d218ff)
- worker: w-jonathonsmac4f50-ja59c (claude-opus-5)
- review: Claimed for referee: long-range order of the light-cone formation law's stationary law pi (sphere menu) for beta > beta_0 = (3/2)(I_0 + I_2) in [0.5879, 0.5931]; the round-1 gap (staggered twist) is closed by the layer swap with halves A|B, whose crossing edges are exactly the vertical edges, so Gaussian domination holds for all fields and both infrared bands close the sum rule; 6 checks pass (G1 exhaustive graph facts on L=4,6 incl. the bilayer isomorphism, P1 Legendre positivity, N1 numerical Ising-analogue domination incl. twists, S1 spectrum, T1 rigorous I_0/I_2 brackets from 2000 exact return counts, T2 finite-volume bounds: at beta=1 <|m_0|^2> >= 0.217 at L=24 vs executed 0.578).

HIT lines:
```
SUMMARY: PROVED (ATTEMPT.md S1-S9, finite facts CHECKED here) long-range order of pi (sphere menu) for beta > beta_0 = (3/2)(I_0 + I_2) in [0.5879, 0.5931]: the layer swap with halves A|B (the two parity classes) is a reflection of Gamma whose crossing edges are exactly the vertical edges, which closes the round-1 gap (the staggered twist Z(t 1_A) <= Z(0) follows from one Cauchy-Schwarz step); with the bond-plane x layer-swap reflections every edge crosses a reflection, so Gaussian domination holds for all fields, the infrared bound holds on both bands {E, 14 - E}, and <|m_0|^2>_pi >= 1 - (3/(2 beta))(G_L + H_L) on every even torus
[...]

PR #8170, opened 2026-09-16
[physics-loop] admissibility-induced-law block26 — bounded_theorem — bounded-support: the unsoldered formation law in level time — gain-one spin waves with the local-limit constant, finite planes forget, and the simulated algebraic decay of the initial plane's memory (the reading decides memory for the sphere menu)

JONbridgerapps:
## What this PR is

Block 26 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `6dda46fc1a`): its parent is block 01's finite-window note on `main`; the sphere menu, the level order and the level automaton are declared in the note. It opens for review one bounded-theorem note with a frontier-discovery artifact, its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-26 records (control and refuter specs with their outputs). Supervisor-run throughout (owner directive 2026-09-16: keep the campaign running for twelve more hours; no subagents).

- Claim type: **bounded_theorem** (T1–T5); trace class **frontier_discovery** for the executed decay. Surface status: **bounded-support** (T1–T5 proved for every `β > 0`; the nonlinear infinite-plane decay simulated and labelled a conjecture; conditional on the records-only reading, positivity, the sphere menu and the level order as supplied conditions; three standard facts named at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_UNSOLDERED_FORMATION_LAW_IN_LEVEL_TIME_GAIN_ONE_SPIN_WAVES_LOCAL_LIMIT_CONSTANT_FINITE_PLANES_FORGET_AND_THE_SIMULATED_ALGEBRAIC_DECAY_OF_THE_INITIAL_PLANE_MEMORY_BOUNDED_THEOREM_NOTE_2026-09-16.md`. Runner: `scripts/admissibility_rule_unsoldered_formation_law_level_time_gain_one_spin_waves_local_limit_constant_finite_planes_forget_2026_09_16.py` (20 checks, 14 mutations, exact).

## Why this block

The campaign knows the six-axis menu orders under both readings (blocks 17, 25) and the sphere menu orders under the static reading (block 19: `β > 76/100`, block 22). The formation reading of the sphere menu was unworked; block 13 linearized it — gain one, variance `1/(3β)`, a logarithmically growing field — and left "whether an aligned region exists at all" open. This block proves what can be proved about the nonlinear kernel and the linear theory exactly, then runs the nonlinear law and compares it with the six-axis law at the same weight.

## What it says

[...]

PR #8692, opened 2026-09-23
physics-loop: admissibility induced-law block 90 - light-cone formation keeps memory in 3+1: its stationary law is one layer of a reflection-positive bilayer, ordered above beta = 0.5905

JONbridgerapps:
## Block 90 — light-cone formation keeps memory in 3+1: its stationary law is one layer of a reflection-positive bilayer, ordered above β = 0.5905 (formation lane)

Hand-off PR against `main` (independent of the open stack; block 19 #8153 supplies the machinery, blocks 26/28 #8170/#8172 are placed, none is required). Nothing is adopted; no gravitational claim; no parked decision is touched.

**Why this block.** Under the owner's reading ("records form"), blocks 26 and 28 found that the sphere formation law in level time loses its common direction on two-dimensional planes. The probes' scans found that it keeps one on three-dimensional planes, and the probes' derivation units proposed a proof route for the light-cone clause. This block re-derives that route, checks every finite step exactly, and compares the proved coupling with an independently executed onset.

**Exact.**
- **T1 — light-cone formation is reversible; level order is not.** With the past `x + {0, ±e_j}` the level-to-level chain satisfies detailed balance with `π(s) = Π_x Z(h_x(s))`. The pair law of two successive levels is the sphere ferromagnet on the doubled graph. The formation law's own long-time law is one layer of it. With the level-ordered past `{0, −e_1, −e_2, −e_3}` a three-step cycle `a → b → −a → a` has forward over backward probability ratio `e^{8β}`, so that chain is reversible with respect to no law.
- **T2 — a bilayer.** `(x, a) ↦ (x, a ⊕ parity(x))` maps the doubled graph onto the bilayer torus edge by edge (tori 4 and 6). The spectrum `E(k)`, `E(k) + 2` is checked eigenvector by eigenvector on `4³`.
- **T3 — reflection positive with slab halves, not layer halves.** Across the 13 reflections of `4³` no vertex is fixed, the crossing edges are mirror pairs, and every edge is covered. Block 19's argument then gives Gaussian domination. With the layers as halves, a `2×2` kernel minor is negative for every `β`: `⟨d, Md⟩ = −64` on the cube and `−20N` on the torus.
- **T4 — memory.** `⟨|m₀|²⟩ ≥ 1 − (3/(2β))(G_L + H_L)`. The exact thresholds are `β_4 = 18239/35840` and `β_6 = 27735979/51891840`. **Long-range order holds for `β > (3/2)(I₀ + I₂) = 0.5905`**, with `I₀ = W/6` in closed form and `I₂` executed.

**Executed.** An independent simulator puts the onset in `(0.55, 0.58)`. The probes' scans put it in `(0.55, 0.60)`. The proved bound is 2 to 8 per cent above it. Above the onset the plateau tends to a positive limit, with `m∞ ≈ 0.633, 0.759, 0.853` at `β = 0.75, 1, 1.5` over planes 16 to 96. The level-ordered clause also keeps a plateau (onset `(1.1, 1.3)`), but that is not proved.

**What this does not say.** It does not show whether the light-cone law forgets on two-dimensional planes. It does not say which formation clause the axioms intend; that is the owner's. The theorem is long-range order of the stationary law, not a memory of the first direction.

**Provenance.** The route is the probes' (issues #8463 and #8631, and the attempt of w-macbookpro90c72-j451b). All of it comes from one model family and has not been refereed by another. This block is by the same family (Claude Opus 5.5). The exact checks are independent code. An other-family referee is still owed.

**Evidence.** Runner `scripts/admissibility_rule_light_cone_formation_keeps_memory_in_3plus1_stationary_law_one_layer_of_a_reflection_positive_bilayer_2026_09_23.py`: `TOTAL: PASS=15 FAIL=0`. Each of the 9 mutations fails in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Corrigendum pushed: status and prior art.** I missed block 36 (#8507, 2026-09-20) when writing this block.

- **Status.** The clause here puts a record at every site of `Z³` at every tick. Block 36 found that clause excluded by the memo as written: "A site never carries more than one record; records are permanent." Every result in this note is conditional on it and would need an axiom change.
- **Prior art.** Block 36 already proved the light-cone chain's reversibility and the bilayer. It also proved the sum-rule bound with long-range order above `(3/2)(I₀ + I₂)`, with `β₀ < 0.5914` rigorous from block 22's bracket, and the linear response `7/E(k)`.
- **New here:** the cycle criterion for the level-ordered past, the eigenvector check, the layer-halves minor, the exact finite thresholds and the independent simulator.

The change is text-only; the runner output is unchanged and the pipeline and lints are clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `c3412ce59542e2edfaef2ae45d19cca9d390c194` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block90-light-cone-formation-keeps-memory-in-3plus1-20260923` at `c3412ce59542e2edfaef2ae45d19cca9d390c194`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8173, opened 2026-09-16
[physics-loop] admissibility-induced-law block29 — bounded_theorem — bounded-support: the transverse kernel's normalization in the ordered sphere static law — measured between the bounds, with the transverse sum rule and the spin-wave reference

JONbridgerapps:
## What this PR is

Block 29 of the physics-loop campaign `admissibility-induced-law-20260906`, an **independent PR against `main`** (base `6dda46fc1a`): its parent is block 01's finite-window note on `main`; the sphere static law and its transverse structure factor are declared in the note. It opens for review one bounded-theorem note with a frontier-discovery artifact (the measured normalization), its exact runner and pinned cache, the manifest acknowledgment, and the pack's block-29 records (control and refuter specs with outputs). Supervisor-run throughout (owner directive 2026-09-16: keep the campaign running for twelve more hours; no subagents).

- Claim type: **bounded_theorem** (T1–T3); trace class **frontier_discovery** for the measurement. Surface status: **bounded-support** (T1–T3 proved as stated; the normalization measured on finite lattices; conditional on the records-only reading, positivity, the sphere menu and the static reading as supplied conditions; two standard facts named at definition level; audit required).
- Note: `docs/ADMISSIBILITY_RULE_TRANSVERSE_KERNEL_NORMALIZATION_IN_THE_ORDERED_SPHERE_STATIC_LAW_MEASURED_BETWEEN_THE_BOUNDS_WITH_THE_TRANSVERSE_SUM_RULE_AND_THE_SPIN_WAVE_REFERENCE_BOUNDED_THEOREM_NOTE_2026-09-16.md`. Runner: `scripts/admissibility_rule_transverse_kernel_normalization_ordered_sphere_static_law_measured_between_bounds_sum_rule_spin_wave_reference_2026_09_16.py` (14 checks, 7 mutations, exact).

## Why this block

Blocks 19–23 (PRs #8153–#8157) placed the gravity node's Green-function kernel in the transverse channel of the ordered sphere static law, between `(M²/3)²/(βE(k))` and `1/(βE(k))` — a window of a factor nine or more. The queue's item "the kernel's normalization on `Z³`" is hard as a theorem and cheap to measure, with two exact anchors: the spin-wave value and the `k = 0` sum rule.

## What it says

[...]

PR #8696, opened 2026-09-23
physics-loop: admissibility induced-law block 91 - a held source under light-cone formation is answered by the inverse lattice Laplacian within a factor between m² and one

JONbridgerapps:
## Block 91 — a held source under light-cone formation is answered by the inverse lattice Laplacian within a factor between m² and one (formation lane)

Hand-off PR against `main`. It is independent of the open stack: block 90 (#8692) supplies the bilayer and Gaussian domination, block 19 (#8153) the infrared-bound machinery, and blocks 29/35 (#8173/#8180) the gravity node's question. None is required. Nothing is adopted; no gravitational claim; no parked decision is touched.

**Why this block.** The gravity node asks which kernel carries a source's influence through the records. Under light-cone formation the probes held a source at one site. They measured the lattice's `1/r` potential around it, at 0.91 to 0.99 of their linear prediction. Block 90 turned the light-cone law into an equilibrium bilayer. This block proves what the formation law's own answer to a held source must satisfy.

**Exact.**
- **T1 — a held source keeps formation reversible.** Add a uniform field `ε` and a source `h` at one site to the rule's argument. The chain still satisfies detailed balance. Its long-time law is one level of block 90's doubled-graph ferromagnet, with the source on both levels at that site (rings, symbolic weights; the marginal).
- **T2 — response is a covariance.** The stationary response is `β` times the covariance with the source site's two records. In block 90's bilayer labels this is `R̂(k) = 2β N⁻¹⟨|Ŝ_+(k)|²⟩`, the slabs' mean. A level's structure factor splits into the symmetric branch at `k` and the antisymmetric branch at `k + π`. Exact on the doubled ring of four.
- **T3 — the window.** For every even `L`, `β`, `ε ≥ 0` and `k ≠ 0`:

  **`⟨m⟩²/(⟨P_b⟩E(k) + ε⟨m⟩) ≤ R̂_⊥(k) ≤ 1/E(k)`**

  The upper side is block 90's infrared bound and holds for every menu. The lower side is a rotation-derivation inequality for the sphere menu. `⟨P_b⟩ ≤ 1` is the correlation of a record with a displaced predecessor. The same window holds for potential differences `R(0) − R(x)`, because the weights `1 − cos k·x` are nonnegative. Exact instances cover Gaussian domination and the upper bound on the Ising ladder, with `β` enclosed by rationals. The derivation algebra, the sphere integrals and the stiffness sums are checked symbolically.
- **T4.** The gain-one linear model gives `A(7β)/E(k)`. The probes' linear prediction is exactly the upper bound.

**Executed.** A two-copy simulator with a uniform field and a small source measures the response mode by mode on `16³`. At every shell it sits `1.04`, `1.01` and `1.004` times the floor at `β = 1, 2, 3`. So the answer is very nearly `⟨m⟩²/(⟨P_b⟩E(k) + ε⟨m⟩)`: the inverse lattice Laplacian at every wave vector, with its strength set by the records' alignment and bond stiffness.

**What the control taught.** Under the probes' protocol the floor first looked violated by 0.3 to 1.5 per cent. Two causes were found:
- Without a field, the record direction wanders, and a fixed-axis measurement mixes in the longitudinal response.
- A source of 8 per cent of the local field answers nonlinearly.

With `ε = 0.02` and `h = 0.1`, checked against `h = 0.05`, every shell lies in the window. The theorem did not change.

**Prior art.** For the static sphere law this two-sided window is block 19's transverse channel (#8153), run at zero field against the long-range-order parameter. Block 20 (#8154, closed and archived) made its lower side dimension-free. This block's new content is the formation law's response as a bilayer covariance, which makes the window apply to records that form. It also adds the sharper `⟨P_b⟩` floor and the measurement that the floor is nearly attained.

**Provenance.** The question comes from the probes' source-potential scans: one model family, unrefereed. The bounds are the supervisor's derivation (Claude Opus 5.5, the same family). An other-family referee is owed.

**Evidence.** Runner `scripts/admissibility_rule_pinned_source_under_light_cone_formation_inverse_lattice_laplacian_within_m_squared_and_one_2026_09_23.py`: `TOTAL: PASS=17 FAIL=0`. Each of the 8 mutations fails in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Corrigendum pushed: status and prior art.**

- **Status.** The clause here puts a record at every site at every tick. Block 36 (#8507) found that clause excluded by the memo as written, so every result is conditional on it.
- **Prior art.** Block 36 already gave the linearized response to a persistent source, `7/E(k)`, and executed the potential at 0.96–0.99 of it, symmetric.
- **New here:** the nonlinear response as an exact covariance, the two-sided window, and the mode-by-mode measurement.

For the static law, which is the equilibrium of pair-weight motion under the owner's moving-records reading, the window is block 19's. That part does not depend on the excluded clause.

The change is text-only; the runner output is unchanged and the pipeline and lints are clean.

JONbridgerapps:
**The level-ordered past's kernel, exactly (2026-09-23, supervisor), from refereed probes units.** Like this note, it is conditional on block 36's clause of re-recording at every level, which the memo excludes (see this PR's corrigendum).

`J:derive:the-level-ordered-response-kernel`, two attempts:
- a1 by `w-jonathonsmac4f50-j40ae` (Claude Opus 5.5), confirmed by `w-macbookpro90c72-j4d1b` (a Grok model);
- a2 by `jf747` (a Grok model), confirmed by `w-macbookpro9927a-ja03f` (Claude Opus 5.5).

**Results.**
- **The kernel.** Under the level-ordered past (predecessors `x, x−e₁, x−e₂, x−e₃`), the gain-one linear kernel is exactly `G(a,b,c) = (4/3) L!/(a!b!c!) 3^{−L}` on the forward octant and `0` elsewhere. Every level carries `4/3`.
- **Its shape.** It is a one-way wake along `(1,1,1)`: `2/(πr)` on the axis, which is `8/7` of this note's light-cone `7/(4πr)`, and exponentially small off the axis.
- **In the drift-free directions** it is `8/E(k)` against `7/E(k)`.
- **Against this note's bound.** With the rule's gain, the linear model's response `2A(4β)/E(k)` exceeds T3's light-cone window `1/E(k)` exactly when `β > κ*/4 ≈ 0.45`, where `κ* ∈ (1.79, 1.81)`. That is consistent with the one executed level-ordered ratio above the bound.
- **What is not proved.** The nonlinear level-ordered response is open.

So under the level-ordered past a held source does not produce an inverse-Laplacian kernel, while under the light-cone past it does (this note).

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `036826ac2e7b7ff8c3b2529c25db4839e7e4748d` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block91-a-held-source-under-light-cone-formation-20260923` at `036826ac2e7b7ff8c3b2529c25db4839e7e4748d`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8698, opened 2026-09-23
docs(campaign): positive static rules select the Coulomb measure

jonathonreilly:
## Summary

The static route to the landed equal-time Coulomb correlations rests on the landed ice note's supplied quantum Hamiltonian, which is an open edge in #8648. This block asks what, under the static reading, selects the uniform ice measure at all. Everything is computed exactly on the landed L = 2 torus: 9600 ice states, 880 of them at zero flux.

- **A hard ice rule selects nothing.** With the soldered vertex records of #8679, each site's value in every ice state is a function of its neighbours' values. So any measure on the ice states satisfies the same hard conditionals. Three such measures share them and still differ:
  - the uniform law, with flux variance 76/25;
  - the zero-flux uniform law, with flux variance 0;
  - a single state.
- **A positive rule has one law.** Let vertex records range over the 20 ice patterns, and weight each state by eps per vertex-link disagreement. This rule is nearest-neighbour: a link's odds depend only on its two vertex records, checked on 1440 perturbed configurations. It is positive, so by Brook's lemma it determines its joint law.
- **That law selects the uniform measure.** The link law is proportional to prod_v A(p_v), with A = 1 + 9 eps^2 + 9 eps^4 + eps^6 on ice patterns.
  - Given no charged vertex, the law is exactly uniform on the 9600 ice states for every eps > 0.
  - P(no charged vertex) is 0.0957, 0.9508 and 0.99949 at eps = 1/10, 1/100 and 1/1000. The shortfall is about 513 eps^2, because charges come in pairs.
  - A transfer programme over 2^24 link states gives these numbers exactly, and its ice part equals 9600 A^8.
- **What moves.** The equal-time Coulomb law needs a frame source and a positive nearest-neighbour rule in the ice limit, not a supplied Hamiltonian. The Hamiltonian is still needed for dynamics. Hard rules and formation (#8687) do not select the uniform measure.

## Files

- `docs/COULOMB_MEASURE_SELECTION_HARD_STATIC_ICE_RULES_ADMIT_EVERY_ICE_MEASURE_POSITIVE_RULES_SELECT_THE_UNIFORM_ONE_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/coulomb_measure_selection_hard_static_rules_admit_every_ice_measure_positive_rules_select_the_uniform_one_2026_09_22.py`
- `logs/runner-cache/coulomb_measure_selection_hard_static_rules_admit_every_ice_measure_positive_rules_select_the_uniform_one_2026_09_22.txt`

## Verification

```bash
python3 scripts/coulomb_measure_selection_hard_static_rules_admit_every_ice_measure_positive_rules_select_the_uniform_one_2026_09_22.py
```

`TOTAL: PASS=6 FAIL=0`, 1245 characters of stdout, about 1.5 s, exact Fractions.

The mutation census catches 9 of 9 defect mutants. It also produced two diagnosed non-defects:
- a redundant neighbour, which the ice count already determines;
- a sub-support, which has the same conditionals.

No rule, reading, limit or physical identification is adopted.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `8408efbe4e4d71aba749b2d790701dd6af4cd0f9` in 61fa847f032d25c90bca2c1273b7edb5713fe160 as part of the combined ten-PR ice formation and transfer batch.

- State support-wise compatibility; prove positive finite-joint uniqueness and limiting concentration, label ratio empirical, assert second moment76/25 and count2880.

PR #8703, opened 2026-09-23
physics-loop: admissibility induced-law block 92 - the zero-field bound on the formation bilayer: a held source's kernel without a field in 3+1, and no memory on planes

JONbridgerapps:
## Block 92 — the zero-field bound on the formation bilayer: a held source's kernel without a field in 3+1, and no memory on planes (formation lane)

Hand-off PR against `main`. It is independent of the open stack: blocks 90 (#8692) and 91 (#8696) supply the bilayer, the long-range-order bound and the response identity, and block 20 (#8154, closed and archived) supplies the argument. None is required. Nothing is adopted; no gravitational claim; no parked decision is touched.

**Why this block.** Block 91's lower side for a held source's response needs three things: a uniform field, a passage to infinite volume, and a named step from long-range order to a spontaneous direction. Block 90 left open whether light-cone formation forgets on planes. Block 20's zero-field inequality, built for the static law, works against the long-range-order parameter itself. Carried to the formation bilayer, it settles both.

**Exact.**
- **T1 — the zero-field inequality on the bilayer.** At zero field, in two or three dimensions: `u(k) ≥ (M²/3)²/(βE(k) + 2/(3V))`. The rungs join equal phases and carry no weight, so block 20's argument goes through with `V = 2N` vertices. The last step is an exact identity, simpler than block 20's case split.
- **T2 — in 3+1, a held source's window at zero field.** Block 91's response is `R̂ = βu`, and block 90 gives `M² ≥ 1 − β_L/β`. So in every finite volume:

  **`((1 − β_L/β)/3)²/(E(k) + 2/(3βV)) ≤ R̂(k) ≤ 1/E(k)`**

  No field, no limit and no import are needed. The limit floor is `(β − β₀)²/(9β²)`: at most `1/9` of the inverse Laplacian, while block 91 measures about `0.95`. The bound is valid but weak.
- **T3 — on planes, no memory.** `M⁴ ≤ (6π²β + 1/2)/H_{L/2−1}`, and a level's `⟨|m_0|²⟩ ≤ M² + 3/(20βN)`. So light-cone formation keeps no common direction on planes, at any `β`. This answers block 90's open question.

**Executed.** On planes the plateau falls with the side, from 16 to 128, at every `β` tried: `0.39 → 0.07` at `β = 1`, `0.81 → 0.73` at `β = 2`, `0.91 → 0.88` at `β = 4`. The plane bound itself is qualitative: it bites only when `H_{L/2−1} > 6π²β + 1/2`, at `L/2 ~ 10²⁵` for `β = 1`. The note says so.

**So the dimension decides.** Under the light-cone clause, forming records keep a common direction in three dimensions of space above a proved coupling, and never in two. Where they keep it, a held source is answered by a positive fraction of the inverse lattice Laplacian at every wave vector, and that holds at zero field.

**Provenance.** The argument is the supervisor's earlier block 20, carried to the bilayer. This block is by Claude Opus 5.5, the same family as the probes, and an other-family referee is owed.

**Evidence.** Runner `scripts/admissibility_rule_zero_field_bound_on_the_formation_bilayer_held_source_kernel_in_3plus1_no_memory_on_planes_2026_09_23.py`: `TOTAL: PASS=17 FAIL=0`. Each of the 7 mutations fails in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Corrigendum pushed: status.** Through blocks 90 and 91, this block inherits a clause that block 36 (#8507) found excluded by the memo as written: a record at every site at every tick. For the memo as written, planes were already seen to forget (blocks 26 and 35).

The zero-field inequality itself, block 20's argument carried to the bilayer, does not depend on the clause. Only its reading as a statement about formation does. The change is text-only; the runner output is unchanged and the pipeline and lints are clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `3e26092f3c43921e0096c930dad5a8ad1362d4b6` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block92-zero-field-bound-on-the-formation-bilayer-20260923` at `3e26092f3c43921e0096c930dad5a8ad1362d4b6`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8710, opened 2026-09-23
physics-loop: admissibility induced-law block 93 - covariance forces a reversible formation law: three covariant pasts, and the six-neighbour past settles into two copies of the static law

JONbridgerapps:
## Block 93 — covariance forces a reversible formation law: three covariant pasts, and the six-neighbour past settles into two copies of the static law (formation lane)

Hand-off PR against `main`. It is independent of the open stack: block 19 (#8153) supplies the rule and the static law, and block 90 (#8692) the doubled graph and bilayer. Neither is required. Nothing is adopted; no gravitational claim; no parked decision is touched.

**Why this block.** Blocks 90–92 took the light-cone past as a supplied clause. The decision record's row 61(a) leaves to the owner which past the axioms mean by a forming record's "nearest-neighbor conditions". This block asks what the axioms' own covariance allows. Covariance here means the Lattice axiom's 24 proper rotations, together with the Admissibility axiom's one covariant nearest-neighbour rule.

**Exact (the 3+1 reading: levels of records on Z³, each record forming from nearest-neighbour records of the level below by block 19's pairwise rule).**
- **T1 — three covariant pasts.** The rotations split `{0, ±e_j}` into two orbits. So a covariant past is the record below, its six neighbours, or all seven, with one weight per orbit (`w₀`, `w₁`). All are symmetric. The level-ordered past `{0, −e_j}` is fixed by only 3 of the 24 rotations.
- **T2 — all reversible.** Under the pairwise rule, every symmetric past satisfies detailed balance with `π = Π Z(h_x)`, at every pair of weights. Covariance alone makes forming records reversible.
- **T3 — what each does.**
  - *The record below alone:* no common direction ever forms.
  - *The six neighbours alone:* every step flips parity, so the event lattice splits into two classes that never interact. Each class settles into the **static law**, so the campaign's comparator is realized exactly inside a covariant formation law. A level's even and odd sites are independent.
  - *All seven:* block 90's bilayer, with rung `w₀` and slabs `w₁`. Thresholds on `4³` are `18239/35840`, `2857397/10250240` and `32773/71680` at `(w₀, w₁) = (1, 1), (1, 2), (2, 1)`.

**Executed.** In infinite volume the thresholds are `0.5905`, `0.3174` and `0.5398`, and `0.758` for the six neighbours alone. A level started with its even and odd halves orthogonal stays orthogonal for 400 levels under the six-neighbour past. Under all seven the halves align within ten levels.

**For the owner.** Row 61(a) becomes: which reading (3+1, or levels as diagonals of `Z³` where the level-ordered past lives), and which weights `(w₀, w₁)`. In the 3+1 reading the axioms leave nothing else.

**Evidence.** Runner `scripts/admissibility_rule_covariance_forces_a_reversible_formation_law_three_covariant_pasts_six_neighbour_past_two_static_laws_2026_09_23.py`: `TOTAL: PASS=14 FAIL=0`. Each of the 8 mutations fails in its own family only. Two pipeline passes (`rows=4880`, `graph_delta=acknowledged`); lints clean. Author checks only (Claude Opus 5.5); no independent review has taken place.


JONbridgerapps:
**Corrigendum pushed: status and prior art.**

- **Status.** This note's "3+1 reading" puts a record at every site of `Z³` at every tick. That is the clause block 36 (#8507) found excluded by the memo as written. So "what the axioms allow in the 3+1 reading" should read "what covariance allows within that excluded clause".
- **Prior art.** Block 36 already proved that synchronous re-recording from the six neighbours splits into two chains whose space-time checkerboard carries the static law. It also proved that asynchronous re-recording is reversible for the static law.
- **New here:** the classification of the covariant pasts, reversibility at every pair of weights, the weighted thresholds and the record-below past.

The owner's moving-records reading is a third reading, and it is not affected by the exclusion. The change is text-only; the runner output is unchanged and the pipeline and lints are clean.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `fa522a87d1307efc24eb0e03ccfd060d0281a70d` in fbe9f0f0154402d8afa07a8d12c920df2d9d0848 as part of the combined thirteen-PR bond and bilayer batch.

All original paths have explicit dispositions. The three canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block93-covariance-forces-a-reversible-formation-law-20260923` at `fa522a87d1307efc24eb0e03ccfd060d0281a70d`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8701, opened 2026-09-23
docs(campaign): formed flux as a persistent walk

jonathonreilly:
## Summary

Sweep-formed ice moves a flux fluctuation as a diffusive walk (#8687), and photon dynamics currently rests on the supplied Hamiltonian (#8698). This block asks what transport formation itself can give. The answer: with memory, formation moves flux like a damped wave. All results are exact.

- **The rule.** R_p keeps each arrow straight with probability p and is otherwise uniform among ice-consistent forward arrows. It:
  - keeps the ice rule exactly;
  - gives forward mean p b_i + (1 - p) B/3;
  - is covariant under the six octant-preserving axis permutations;
  - needs a frame source to pair axes (soldering or coordinate letters).
- **Memory.** A unit of flux keeps its axis with probability (1 + 2p)/3. The direction kernel K = p I + (1 - p) J/3 has eigenvalues 1, p, p, so memory relaxes by p per layer.
- **Persistent walk.** The mean flux kernel carries flux 1 per layer.
  - At p = 1 it runs straight along its axis.
  - At p = 0 it is exactly the multinomial walk of #8687, checked at 2925 sites.
- **Ballistic, then diffusive.** The spread equals the Markov occupation variance (checked exactly for layers 1-15). It grows at rate (2/9)(1 + p)/(1 - p), which is 2/3 at p = 1/2. So flux moves ballistically along an axis for about 1/(1 - p) layers and diffuses beyond.
- **Causal covariance.** On a 4x4x4 box with independent inflow (p = 1/2 and 3/4), back sums have variance 3, same-layer arrows are uncorrelated, and the covariance with an inflow arrow is the persistent-walk kernel. So the causal structure of #8687 holds across the family.
- **What moves.** Formation can carry flux like a telegraph (damped) wave, along the octant's axes, over a finite memory. Undamped isotropic waves are not produced, so photon dynamics stays with the supplied Hamiltonian.

## Files

- `docs/FORMED_FLUX_AS_A_PERSISTENT_WALK_STRAIGHT_CONTINUATION_GIVES_BALLISTIC_THEN_DIFFUSIVE_TRANSPORT_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/formed_flux_persistent_walk_straight_continuation_ballistic_then_diffusive_2026_09_22.py`
- `logs/runner-cache/formed_flux_persistent_walk_straight_continuation_ballistic_then_diffusive_2026_09_22.txt`

## Verification

```bash
[...]

PR #8715, opened 2026-09-23
docs(campaign): two planar ice squares admit no local formation order

jonathonreilly:
## Summary

This PR asks whether the uniform ice measure, the equal-time law behind the landed Coulomb photon bridge, can be produced by local formation. The answer depends on the window:

- **One square:** yes, with its plaquette site.
- **Two squares in the plane:** no.
- **One cube:** yes, with its cube site.

It answers the assembly's open edge "coordinated formation of the uniform ice measure beyond one square" (open PR #8648, open PR #8686) on those windows.

- **An exact criterion.** Records are local: links record occupations, and vertices, plaquettes and cubes record their own edges. A nearest-neighbour formation law then reproduces a record law exactly, with no unrecorded site, iff some order has this property: each site's conditional, given everything formed before it, depends only on its formed neighbours. A search over formed-site subsets decides the question.
- **One square** reproduces open PR #8686. Without the plaquette site no order works (at most 3 of 8 sites). With it, an order works.
- **Two squares (2 x 1 rectangle, 501632 ice configurations).**
  - Without plaquette sites no order works (at most 4 of 13 sites).
  - With both plaquette coordinators, still none works, under all five record schemes tested. Vertex records are a pattern, a count or nothing; plaquette records are a pattern, a count or nothing.
  - At most 7 of 15 sites form exactly with full local records, and 11 with plaquette records alone.
- **One cube (weight 6316544).**
  - With its six plaquette sites alone there is no order (at most 5 of 26 sites).
  - With its cube site, the cube-first order reproduces the law exactly: cube, then plaquettes, links, vertices.
  - Each top cell needs its own coordinator. Two squares in the plane have none above them.
- **Photon lane.**
  - In the plane, the plaquette coordinator does not extend to two squares.
  - In three dimensions every cube has a cube site. Whether two adjacent cubes can be formed exactly is not searched.
  - Where local coordination fails, formation needs non-local records (open PR #8686) or conditioning, which is not in the text.
  - Formation otherwise gives directed transport (open PRs #8687, #8701). The static reading with a positive rule selects the Coulomb law (open PR #8698).

## Files

- `docs/UNIFORM_ICE_BY_FORMATION_TWO_PLANAR_SQUARES_ADMIT_NO_LOCAL_ORDER_AND_ONE_CUBE_NEEDS_ITS_CUBE_SITE_BOUNDED_THEOREM_NOTE_2026-09-22.md`
- `scripts/uniform_ice_two_squares_no_local_formation_order_complete_order_search_2026_09_22.py`
- `logs/runner-cache/uniform_ice_two_squares_no_local_formation_order_complete_order_search_2026_09_22.txt`

## Verification

```bash
python3 scripts/uniform_ice_two_squares_no_local_formation_order_complete_order_search_2026_09_22.py
```

- `TOTAL: PASS=7 FAIL=0`, about 15 s, 1205 stdout characters, exact Fractions.
- Mutation census: 12 of 12 caught.
- The planar weights match the transfer-matrix count and open PR #8686.
- The cube weight is recomputed as a generating-function coefficient.
- A budget cap on reachable sets fails its check if it binds. It never binds here: the searches reach at most 2560 sets (planar) and 173 (cube).

## Scope

This is a bounded theorem on declared windows and record schemes. No rule, record scheme or order law is adopted. No new axiom, primitive, comparator or framing is imported. Audit is required before any retained use.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `47619bb8890c5c7a465c85fbba46c19c22bbd505` in 61fa847f032d25c90bca2c1273b7edb5713fe160 as part of the combined ten-PR ice formation and transfer batch.

PR #8719, opened 2026-09-23
docs(campaign): two adjacent ice cubes admit no local formation order

jonathonreilly:
## Summary

Open PR #8715 showed three things about exact local formation of the uniform ice measure, the equal-time law behind the landed Coulomb correlations:
- one square is formed exactly with its plaquette site;
- two squares in the plane are not formed exactly by any order;
- one cube is formed exactly with its cube site.

This PR settles the three-dimensional question it left open. Two adjacent cubes are not formed exactly by any order, even with both cube sites.

- **Closure lemma.** Suppose a site's record is a function of its formed neighbours' records. Moving it forward to the first moment it becomes one keeps every later step admissible (tower property). So a search over formed sets closed under such moves decides whether an order exists. On six windows it reproduces the plain search of #8715: same verdicts, same largest formed sets.
- **Two cubes sharing a face** (45 sites; weight 3985518592, recomputed as a generating-function coefficient). With both cube sites, no order reproduces the uniform measure. That holds for full local records and for three coarser schemes: vertex records carry nothing, plaquette records carry nothing, or both.
- **Where it stops.** With full local records the largest formed set is one cube's sites: its cube site, 6 plaquettes, 12 links and 4 outer vertices. The other cube site depends on the formed cube beyond the shared plaquette. The reason is the four shared vertices' weights C(2, 2 − e1 − e2), which form the rank-2 matrix [[1, 2], [2, 1]] over the two cubes' perpendicular edges.
- **Photon lane.** Exact local formation of the uniform measure stops at one top cell, in the plane (#8715) and in three dimensions (here). The equal-time Coulomb law comes from the static reading with a positive rule (#8698). By formation it needs plan records (#8686) or conditioning, which is not in the text. With local records, formation gives directed, causal ice (#8687, #8701).

## Files

- `docs/UNIFORM_ICE_BY_FORMATION_TWO_ADJACENT_CUBES_ADMIT_NO_LOCAL_ORDER_EVEN_WITH_BOTH_CUBE_SITES_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/uniform_ice_two_adjacent_cubes_no_local_formation_order_closed_set_search_2026_09_23.py`
- `logs/runner-cache/uniform_ice_two_adjacent_cubes_no_local_formation_order_closed_set_search_2026_09_23.txt`

## Verification

```bash
python3 scripts/uniform_ice_two_adjacent_cubes_no_local_formation_order_closed_set_search_2026_09_23.py
```

`TOTAL: PASS=5 FAIL=0`: about 220 s, 1435 stdout characters, peak about 0.8 GB.

- **Exactness.** Weights are exact integers and conditionals are compared by integer cross-multiplication. The two-cube sections run only if the validation passes.
- **Mutation census.** 11 of 11 defect mutants are caught. A closure mutant that tests the whole formed set is diagnosed equivalent on these windows.

## Scope

This is a bounded theorem on declared windows and record schemes. Larger windows are not searched. No rule, record scheme or order law is adopted. No new axiom, primitive, comparator or framing is imported. Audit is required before any retained use.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `d57f2984c89b7668093207e45f0a49310bbb7e65` in 61fa847f032d25c90bca2c1273b7edb5713fe160 as part of the combined ten-PR ice formation and transfer batch.

PR #8720, opened 2026-09-23
docs(campaign): joint cell units form uniform ice along rows, not blocks

jonathonreilly:
## Summary

With single sites, exact local formation of the uniform ice measure stops at one top cell (#8715, #8719). The candidate assembly (#8648) keeps a gravity route through joint formation units (#8643), {soldering, joint units, roles}, and lists joint-unit formation as an open edge. This PR tests the natural finite units: closed cells formed jointly.

- **Unit criterion.** Exactness with no unrecorded site holds iff some unit order has each unit's joint conditional, given the formed set, depend only on the formed sites adjacent to its new sites. The proof is the chain rule over units.
- **Rows work.** One square and two squares are exact in every order. A row of three is exact in exactly the 4 of 6 orders that do not form both ends first.
- **The 2 × 2 block does not.** No order of the four cell units is exact, and every one fails at the second unit. The reason: the two unformed cells close a loop around the centre, joining the new unit to formed vertex records it does not touch.
- **Chains of units work.** Row units on a 2 × 3 window form a chain, and they are exact exactly as cells in a row are. With only two units any split is exact, because the second unit is the whole remainder. A first draft checked row units on the 2 × 2 block, which was vacuous for that reason; it was replaced by the 2 × 3 window. On the lattice, such units are unbounded and pick a direction, as a sweep's layers do.
- **The same in three dimensions.** Cube units on two cubes sharing a face are exact in both orders (enumeration over 614912 grid patterns, with non-vacuity guards). The 2 × 2 × 1 slab of four cubes has 33 grid links and cannot be enumerated. For each of its 12 ordered pairs of first units, the runner finds two formed patterns that agree on every site adjacent to the second unit but give it different exact laws. Both laws are recomputed by an independent vectorised enumeration. So all 24 orders fail at the second unit.
- **Chains need an open axis.** Wrap three rows into a loop (a planar cylinder, 21888 grid patterns) and no order of the row units is exact, although the open 2 × 3 window has four. Units spanning an axis must be formed as a chain along an open axis, as a sweep's layers are.
- **Photon lane.** In the plane and in three dimensions, finite cell units extend the one-cell coordinator along a row but not around a block. Exact joint formation of the uniform measure then needs one of:
  - units unbounded along an axis;
  - plan letters (#8686);
  - conditioning, which is not in the text.

  The assembly's {soldering, joint units, roles} set now rests on units unbounded along an axis.

## Files

- `docs/UNIFORM_ICE_BY_JOINT_CELL_UNITS_ROWS_YES_THE_TWO_BY_TWO_BLOCK_NO_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/uniform_ice_joint_cell_units_rows_but_not_the_two_by_two_block_2026_09_23.py`
- `logs/runner-cache/uniform_ice_joint_cell_units_rows_but_not_the_two_by_two_block_2026_09_23.txt`

## Verification

```bash
python3 scripts/uniform_ice_joint_cell_units_rows_but_not_the_two_by_two_block_2026_09_23.py
```

`TOTAL: PASS=7 FAIL=0`: about 30 s, 1899 stdout characters, exact Fractions and integers. Mutation census: 15 of 15 defect mutants caught, plus one diagnosed equivalent (a cube site's record is a function of its own unit's links).

## Scope

Declared planar and cubic windows with cell units; larger cube windows are not searched. No rule, unit or order law is adopted. No new axiom, primitive, comparator or framing is imported. Audit is required before any retained use.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `05a5a16fb4367b84b3562ec37142b69353c61b34` in 61fa847f032d25c90bca2c1273b7edb5713fe160 as part of the combined ten-PR ice formation and transfer batch.

PR #8724, opened 2026-09-23
docs(campaign): relational spiral frames are locally rigid statically

jonathonreilly:
## Summary

Open PR #8691 left open whether every stationary record near a relational spiral is a spiral. Open PR #8717 answered at linear order: static rigidity, flexibility under the sweep. This PR lifts the static answer to the full nonlinear relations on 4- and 5-boxes.

- **Nullity exactly 5.** The unknowns are values on the used sites, and back and forward axes at the core sites, with unit and orthogonality constraints and every relation. Linearised at the spiral, the system has nullity 5, spanned by five exact kernel vectors:
  - the phase;
  - the two tilts;
  - a turn of the back axis at core corner (1, 1, 1);
  - a turn of the forward axis at the opposite core corner.

  A modular rank bound together with the exact vectors fixes the rational nullity.
- **Exact nonlinear families realise all five.** These are exact solutions of every static relation:
  - rotations of the whole record by (3/5, 4/5) about x, y or z;
  - a turn of either corner axis by (5/13, 12/13), with that corner's three outer neighbours rotated along.

  Their tangent vectors at the spiral are the kernel vectors. The corner tangents come from a first-order Rodrigues expansion that is independent of the linear rows.
- **Local rigidity.** By the implicit function theorem, the solution set near the spiral is a 5-dimensional manifold, and the family fills it. So every static solution close to the spiral is a rotated spiral, up to the two corner axes.
- **For relational frames.** The static reading holds a relational frame rigidly. The sweep does not (#8717). This is where the assembly's {alphabet, reading} route carries relational letters.

## Files

- `docs/RELATIONAL_SPIRAL_FRAMES_ARE_LOCALLY_RIGID_UNDER_THE_STATIC_READING_NONLINEAR_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_spiral_frames_locally_rigid_under_the_static_reading_nonlinear_2026_09_23.py`
- `logs/runner-cache/relational_spiral_frames_locally_rigid_under_the_static_reading_nonlinear_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_spiral_frames_locally_rigid_under_the_static_reading_nonlinear_2026_09_23.py
```

`TOTAL: PASS=3 FAIL=0`: about 1 s, 1073 stdout characters, exact Fractions. Mutation census: 9 of 9 caught.

## Scope

[...]

PR #8717, opened 2026-09-23
docs(campaign): relational spiral frames are rigid statically, not in a sweep

jonathonreilly:
## Summary

Open PR #8691 showed that under possibility covariance one qubit per site can still carry the lattice frame: spiral records are stationary under a covariant, unsoldered sweep rule. It left rigidity open. This PR answers it at linear order.

- **The linearisation.** In plane, only a constant phase solves. Out of plane there are two unknowns: a, the value's vertical deviation with sign reversed, and β, the axis tilt. They satisfy a(x − e_i) = c_i a(x) − s_i β(x). Every w gives the exact solution a = Π_j (c_j − w s_j)^(−x_j), β = w a.
- **Flexible on boxes in a sweep.** On an L-box, the sweep's linear solutions are exactly the phase and these modes, with nullity 3L − 3 for L = 3, 4, 5. The certificate is:
  - exact rational solutions of the full linearisation of #8691's relations;
  - independence and a rank bound modulo a prime.
- **Rigid on the lattice.** A mode is bounded on Z³ only if |c_j − w s_j| = 1 for all j. Each such circle is |w|² − 2 cot θ_j Re w = 1, so the three circles meet only at w = ±i. Those two modes are the global tilts. So every bounded linear deformation of a spiral on Z³ is a global rotation.
- **The static reading is rigid on boxes.** Here each core site completes from its back-neighbours and from its forward neighbours. A mode then needs (s_i + c_i w)/(c_i − s_i w) to agree for all i, which happens only at w = ±i. On 4-, 5- and 6-boxes the solutions are the two global tilts plus a free axis at each of the two core corners; nothing else moves a core site.
- **The sweep amplifies.** Any two back-neighbours fix (a, β), so forward propagation is unique. Consistent boundary errors propagate as their modes, growing per step by:
  - 5/3, 13/5, 17/8 at w = 0;
  - 5, 13, 34 at w = 1/2.

  On a 6-box the sweep rebuilds the w = 1/2 mode exactly. Changing one boundary value by 1/1000 stops the rule at the first site that reads it.
- **For relational frames.** The angle reading survives every linear deformation. The lattice spiral is rigid, and the static reading holds the frame rigidly even on finite windows. But a sweep does not heal boundary errors. So a near-spiral frame registered by formation needs exact boundary data, which complements #8691's first-formation result. The static route ({alphabet, reading}) is where the frame is held rigidly.

## Files

[...]

PR #8729, opened 2026-09-23
docs(campaign): seven relational letters carry the lattice frame

jonathonreilly:
## Summary

Open PR #8691 built relational letters under possibility covariance: spiral records whose neighbours differ by rotations through three Pythagorean angles. Those spirals take infinitely many values, and the PR left open **whether finitely many possibilities suffice**. This PR answers yes: **seven suffice, and no spiral does with fewer**.

- **When the rule's output is unique.** For a spiral, a second fit (another angle assignment or orientation) appears exactly when one of these holds:
  - two angles agree modulo 180°;
  - a 3-cycle shifts every angle by the same amount;
  - one angle is the mean of the other two modulo 180°.
- **Twelve values.** The angles (30°, 60°, 150°) avoid all three. Their spiral lives in Q(√3) and takes the twelve multiples of 30° on one great circle. The following are checked exactly:
  - on the 4-box every site with three back-neighbours is the rule's unique output;
  - the rule commutes with a quarter turn of the sphere;
  - each core site reads its six bond directions from the six distinct signed angles.
- **The conditions matter.** With (30°, 90°, 150°), 30 + 150 = 2 × 90, and the rule has a second output.
- **Seven, and no fewer.** Six distinct nonzero signed angles among the multiples of 360°/m need m ≥ 7. Up to m = 12, valid triples exist exactly for m = 7, 9, 10, 11 and 12.
  - For spiral inputs every value lies on one great circle and the rule's normal is the axis, so each fit is a congruence of angle multiples mod m.
  - This planar reduction matches the exact Q(√3) rule at all 27 sites of the twelve-value box. A negative control checks that the comparison can fail.
  - With it, (1, 2, 4) × 360°/7 gives a seven-value spiral that is the rule's unique output at all 64 sites of a 5-box, with the frame reading.
- **For relational frames.** Under possibility covariance one qubit can carry the lattice frame with seven relational letters. The points are defined only relative to one value and one axis. First formations still miss it (#8691). Static frames are locally rigid (#8724), and so are frames on the whole lattice (#8717).

## Files

- `docs/RELATIONAL_SPIRAL_LETTERS_WITH_FINITELY_MANY_VALUES_SEVEN_SUFFICE_AND_ARE_THE_FEWEST_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_spiral_letters_with_finitely_many_values_seven_suffice_2026_09_23.py`
- `logs/runner-cache/relational_spiral_letters_with_finitely_many_values_seven_suffice_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_spiral_letters_with_finitely_many_values_seven_suffice_2026_09_23.py
```

- `TOTAL: PASS=6 FAIL=0`, under a second, 1613 stdout characters, exact arithmetic in Q(√3) and modular arithmetic.
- Mutation census: 11 of 11 caught. Three first misses were repaired: a rotation-properness check, rejection of a triple containing an angle and its negative, and a negative control for the planar comparison.

## Scope

Spiral records, the rule of #8691, and the declared box. No reading, rule, alphabet or order law is adopted. No new axiom, primitive, comparator or framing is imported. Audit is required before any retained use.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `c447d85ccf15bb1bc0508d804f533fc91f7d5995` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

- Exclude singular w, retain Fourier-distribution proof with bounded exponential-polynomial argument, restrict rank/IFT statements to tested sizes and angles, distinguish sufficiency from degenerate uniqueness necessity.

PR #8731, opened 2026-09-23
docs(campaign): multi-qubit units leave relational first formations a fair coin

jonathonreilly:
## Summary

Open PR 8691 showed that under possibility covariance a lone child, which sees only its parent, misses its spiral value almost surely. It asked whether formation units of several qubits offer more room. This PR answers for 2 x 2 x 2 units whose law, like a lone child's, depends only on their formed neighbours: not enough.

- **A single face cannot tell which way the frame runs.** A unit attaching through one face sees four values (A0, A0 + α, A0 + β, A0 + α + β) in the planar form of a spiral.
  - Reversing the great circle, which is a rotation of the sphere, negates the view and must negate the step.
  - The half turn about the growth axis must keep the step.
  - The two views differ only by a rotation of the circle. This is checked for all 24 proper orientations, both circle orientations and all three axes.
- **So a covariant rule is at best a fair coin.** At every single-face view it gives the steps t and -t equal probability. No deterministic covariant rule continues, and each single-face attachment is right with probability at most 1/2.
- **Growth needs many of them.** Any unit that extends the formed span along an axis attaches through one face. Registering n units per side therefore has probability at most 2^(-3(n - 1)). The covariant fair coin reaches this bound exactly: exactly 1 of 64 draws registers a 3 x 3 x 3 grid of units, in each of the 48 first-unit patterns.
- **Each premise matters, shown by controls:**

  | Control | Covariance | Registers |
  |---|---|---|
  | copy an in-face sign | keeps the reversal, breaks the half turn | 12 of 48 |
  | treat the face as the back side | keeps the half turn, breaks the reversal | 6 of 48 |
  | also read two layers back (outside local formation) | keeps both | 48 of 48 |

What this means for relational frames: under possibility covariance and local formation, first formations fail for single sites (PR 8691) and, on large windows, for 2 x 2 x 2 units. The frames are still carried by sweeps without a first formation (PR 8717), or by the static reading, where they are rigid (PRs 8717 and 8724) and seven letters suffice (PR 8729).

## Files

- `docs/RELATIONAL_LETTERS_FORMED_BY_MULTI_QUBIT_UNITS_EVERY_SINGLE_FACE_ATTACHMENT_IS_AT_BEST_A_FAIR_COIN_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_letters_multi_qubit_units_single_face_attachment_is_a_fair_coin_2026_09_23.py`
- `logs/runner-cache/relational_letters_multi_qubit_units_single_face_attachment_is_a_fair_coin_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_letters_multi_qubit_units_single_face_attachment_is_a_fair_coin_2026_09_23.py
```

- Result: `TOTAL: PASS=7 FAIL=0` in about a second, with 2314 characters of stdout.
- Mutation census: 13 of 13 caught, plus one equivalent mutant. The equivalent one copies the other in-face sign; both copies are right exactly when the pattern's three axis signs agree.

## Review

One Opus 5.5 seat, with no subagents; the same seat wrote the runner and the note. The note records the independence sources: half-turned views are read off the rotated pattern, and the attachment count is checked both against the closed form and against the calls a growth run makes.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `2bdf1a26bae8b368e76d4e48f47d77c1e8952613` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

- Declare full real sphere, fixed finite candidate targets after two formed values, local conditional laws and face-connected sequential growth from one seed; preserve the valid symmetry bound.

PR #8743, opened 2026-09-23
docs(campaign): static relational frames are globally rigid for Sidon angles

jonathonreilly:
## Summary

PR #8691 left open whether every stationary record of its relational frames is a spiral. PRs #8717 and #8724 showed that spirals are linearly rigid, and locally rigid in the full nonlinear relations, under the static reading. This PR proves **global** rigidity under the static reading and records what it rests on.

- **Every static record is a spiral.** This holds for the seven letters (1, 2, 4) × 360/7 and for 30°, 60°, 150°, on every core of side at least 2. It holds up to a global rotation of the sphere.
- **The proof is additive.** Completions read a great circle from their inputs. Neighbouring completions share two distinct, non-antipodal values, so all core values lie on one circle, where each completion is additive. The rest takes four steps:
  1. **Squares.** If the angles form a Sidon set (the six sums θ_i + θ_j, i ≤ j, are distinct), every elementary square is straight or folded.
  2. **Signs.** The sign of the differences depends only on the level x_1 + x_2 + x_3.
  3. **No folds.** Induction on the level rules out every fold.
  4. **Constancy.** Straight squares make each difference constant along its axis, so the record is a spiral.
- **Sidon is exactly the rule's own conditions.** Over all 494 triples of distinct angle multiples modulo m = 5 to 12, Sidon is equivalent to conditions (a) and (c) of PR #8729.
- **The search agrees.** Complete search on cores of side 2, 3 and 4 finds exactly the 12 spirals with one core value fixed. Each is the unique completion at every core site. A full search on the side-4 box, face values included, cross-checks the core-only search.
- **Contrasts.**
  - Sets that break (c) have 48, 384 and 3072 core records.
  - With back neighbours only (the sweep reading), the Sidon sets admit records that are not spirals: 336 and 24 on the side-2 core.
- **On a torus the frame must wrap consistently.** The seven letters have static records on the side-L torus exactly when 7 divides L: the search finds none for L = 2 to 6 and the 12 spirals for L = 7. The 30°, 60°, 150° letters need 12 to divide L. Neither set has a static record on the tori that carry the landed ice measure.
- **The record carries the frame, and nothing positional.** Every core site of a static record sees the same differences, so no covariant local readout carries the role pattern. Even with the global rotation fixed, the seven letters give x and x + (1, 1, 1) the same value. This narrows a sentence of #8691 (see the companion commit there): the emulation of soldered rules goes through, but role letters and label readouts need absolute labels.
- **The rigidity uses the reading's octant.** #8691's static reading takes back neighbours x − e_i and forward neighbours x + e_i: a fixed octant, which supplies the orientation of each axis. If each core site may choose its own octant, the seven letters have 15024 records on the side-2 core, only 48 of them spirals.

What this means: given the reading's octant, a static relational record fixes the rest of the lattice frame everywhere on the core. So, given the octant, the frame source of gravity's {alphabet, reading} set (#8648) is exact for relational letters. Roles still come from fixed letters or are supplied. The per-site-octant reading, which is covariant and local, leaves the letters flexible. Other covariant local readings are not searched.

## Files

- `docs/RELATIONAL_LETTERS_UNDER_THE_STATIC_READING_ARE_GLOBALLY_RIGID_EXACTLY_FOR_SIDON_ANGLE_SETS_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_letters_static_reading_globally_rigid_exactly_for_sidon_angle_sets_2026_09_23.py`
- `logs/runner-cache/relational_letters_static_reading_globally_rigid_exactly_for_sidon_angle_sets_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_letters_static_reading_globally_rigid_exactly_for_sidon_angle_sets_2026_09_23.py
```

- Result: `TOTAL: PASS=9 FAIL=0` in about 47 s, with 3049 characters of stdout.
- Mutation census: 18 of 18 caught. Every search is capped by records, steps and a shared comparison budget, so no mutant runs past the ceiling.

## Review

One Opus 5.5 seat, with no subagents; the same seat wrote the runner and the note. The note records the independence sources:
- the full-search cross-check;
- the identification over all small triples;
- the plaquette lemma by enumeration;
- uniqueness with the completion rule of #8729;
- direct torus searches.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `34007c67010b932ac3ad12ef368af798d5a55359` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

- State Sidon sufficiency, additive-support contrasts and failure of a class-wide guarantee; retain all finite counts without classifying every nonglobal field.

PR #8744, opened 2026-09-23
docs(campaign): static relational rigidity needs a global octant

jonathonreilly:
## Summary

#8743 proved that under #8691's static reading every static record of the seven relational letters is a spiral. It also found that the rigidity uses the reading's fixed octant: with one octant per site, the side-2 core has 15024 records. This PR asks how much orientation the reading has to supply.

| How the octant is supplied | Records on the side-2 core | Spirals |
|---|---|---|
| global octant (every line along an axis alike), per octant | 12 | 12 |
| global octant, union over the 4 opposite pairs | 48 | 48 |
| one orientation per lattice line (all 4096 fields of 12 lines) | 11136 | 48 |
| one octant per site (#8743) | 15024 | 48 |

- **A global octant is enough.** Each of the 8 octants gives exactly the 12 spirals aligned with it.
- **Orienting lines is not enough.** 1300 of the 4096 line fields admit records, and one field that is not a global octant admits 108.
- **The readings are nested.** Global-octant records are line-field records, and line-field records are per-site records.

What this means: under the static reading, the relational letters fix the frame only once the reading supplies a global octant, an orientation for each axis. That is supplied structure, like a time direction. It sits inside gravity's unsoldered set under possibility covariance, {alphabet, reading, roles} (#8648, tenth edition).

No reading, rule, alphabet, unit or order law is adopted.

## Files

- `docs/RELATIONAL_LETTERS_STATIC_RIGIDITY_NEEDS_A_GLOBAL_OCTANT_SITE_AND_LINE_ORIENTATIONS_LEAVE_THE_LETTERS_FLEXIBLE_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_letters_static_rigidity_needs_a_global_octant_line_orientations_are_flexible_2026_09_23.py`
- `logs/runner-cache/relational_letters_static_rigidity_needs_a_global_octant_line_orientations_are_flexible_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_letters_static_rigidity_needs_a_global_octant_line_orientations_are_flexible_2026_09_23.py
```

- Result: `TOTAL: PASS=3 FAIL=0` in about 176 s, with 829 characters of stdout.
- The searches carry record, step and work caps.
- The cache was written with `execute_and_write_cache`, and `vocab_lint` is clean.
- Mutation census: 7 of 7 caught.

## Review

One Opus 5.5 seat, with no subagents; the same seat wrote the runner and the note.

[...]

PR #8750, opened 2026-09-23
docs(campaign): relational pair letters record the role pattern

jonathonreilly:
## Summary

#8743 showed that relational spiral letters under the static reading are rigid for Sidon angles, but that a rigid record gives every site the same view. So it carries the lattice frame and no role pattern, and under possibility covariance the assembly (#8648) has to supply roles. This PR asks whether a relational record can carry the role pattern.

**Letters in alternating pairs.** Take six angles in three pairs, and one fixed nearest-neighbour rule, the pair rule:
- the six differences at a site are signed angles with one common sense;
- the two neighbours on one lattice line use the two angles of one pair, one each;
- the three lines use the three pairs.

The rule depends only on differences, and it uses the reading's global octant (#8744) to tell back from forward. The declared letters are m = 16 with pairs (1, 7), (2, 10) and (3, 5), in steps of 22.5°.

- **Rigid.** Two conditions on the angles make every core square straight:
  - (D): the within-pair differences are distinct up to sign across pairs;
  - (E): no within-pair difference equals a difference between the other two pairs.

  Then every static record is an alternating spiral. With one value fixed there are exactly 96: 6 frames × 2 senses × 8 phases. A complete search on cores of side 2, 3 and 4 finds exactly these.
- **Roles are read.** From its forward differences each site reads the frame and its parity vector x mod 2, up to one global phase: the role pattern. The readout is covariant under rotations of the circle, and all 8 phases occur. On the side-8 torus the vertex/link/plaquette/cube classes have 64/192/192/64 sites.
- **Nearest-neighbour form.** Condition (U) holds: no two allowed stars share their six neighbours. So every site is a function of its neighbours.
- **Tori.** Records wrap only when 8 | L. The search finds none for L = 2 to 7, and the 96 spirals for L = 8.
- **Not on the landed ice torus (side 4).** Wrapping there forces every pair to sum to 180°. Then moving a site by 180° swaps its sense and types without changing its neighbours, and (U) fails:
  - all 240 decodable pair sets up to m = 24 that wrap on the side-4 torus fail (U);
  - for m = 20, pairs (1, 9), (2, 8), (4, 6), the records are still exactly the 96 spirals, but every site admits its antipode.
- **Contrasts on the side-4 letters.**
  - Without complementary types there are 2592 records, and the readout misses the role pattern on all but the 96 spirals.
  - With one octant per site there are 9216 records, of which 96 are spirals.

What this means: a static relational record can carry the role pattern. Away from the side-4 torus, the unsoldered gravity set under possibility covariance, {alphabet, reading, roles}, can drop roles and become {alphabet, reading}: the same set fixed letters need. On the landed ice torus neither single-angle nor pair letters work, so with these letters the static route there still needs fixed letters or soldering. #8752 gives four-angle cycle letters that do work on that torus. The static ice records of #8679 are checked on the side-4 torus, not on the side-8 torus.

## Files

- `docs/RELATIONAL_LETTERS_IN_ALTERNATING_PAIRS_RECORD_THE_ROLE_PATTERN_UNDER_THE_STATIC_READING_BUT_NOT_ON_THE_SIDE_FOUR_TORUS_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_letters_in_alternating_pairs_record_the_role_pattern_under_the_static_reading_2026_09_23.py`
- `logs/runner-cache/relational_letters_in_alternating_pairs_record_the_role_pattern_under_the_static_reading_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_letters_in_alternating_pairs_record_the_role_pattern_under_the_static_reading_2026_09_23.py
```

- Result: `TOTAL: PASS=18 FAIL=0` in about 47 s, with a peak of 54 MB and 2998 characters of stdout.
- Mutation census: 19 of 19 caught.

## Review


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `d07529c6b548484ad01c46e8a0b3599cd5462bde` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

- Define rules locally, make extension a supplied choice, distinguish all-star uniqueness from sampled site checks, and report the capped open-box count as at least 100000. The newer oriented classification corrects earlier flexibility rhetoric.

PR #8752, opened 2026-09-23
docs(campaign): cycle letters record the role pattern on the ice torus

jonathonreilly:
## Summary

#8750 found that relational letters in alternating pairs record the role pattern under the static reading, but not on the landed ice torus of side 4. On that torus the pairs must sum to 180°, and a site is fixed by its neighbours only up to its antipode. This PR gives relational letters that do work there.

**Four-angle cycles.** Each lattice line cycles through four angles whose sum is 0 mod m, under one fixed nearest-neighbour rule, the cycle rule:
- the six differences at a site are signed angles with one common sense;
- on each line, the back and forward differences are consecutive angles of one cycle;
- the three lines use the three cycles.

The declared letters are m = 211 with cycles (19, 31, 132, 29), (108, 43, 194, 77) and (88, 39, 126, 169).

- **Rigid.** If every solution of the square relation is straight (a finite condition: 96 straight, 0 bent), every static record is a cycle spiral. With one value fixed there are 768 of them: 6 frames × 2 senses × 4³ phases. Complete searches on cores of side 2, 3 and 4 find exactly these.
- **On the landed ice torus.** Records exist exactly when 4 divides L. The search finds none for L = 2, 3, 5, 6, 7, and the 768 spirals for L = 4 and 8. The side-4 torus is the landed ice torus.
- **Roles read there.** Each site reads the frame and its phase x mod 4 on every line, up to one global phase. That gives its parity vector (the role pattern) up to one global phase; all 64 cycle phases and all 8 parity phases occur. The readout commutes with rotations of the circle.
- **Each site is a function of its neighbours.** No two allowed stars share their six neighbours, and the rule admits exactly one value per site. The positive control is a decodable cycle set mod 37 with 24 shared stars. m is odd, so no residue is 180° and the antipodal ambiguity of #8750 cannot arise.
- **Contrasts.** A cycle set mod 37 with bent squares has 1344 core records, 576 of them not spirals. With one octant per site the side-4 torus has 24576 records, 768 of them the spirals with every axis forward. #8854 shows that all 24576 are the same spirals with axes reversed or with a raw sense per axis: under that rotation-covariant rule the letters record their own octant, so no octant need be supplied. (Corrected 2026-09-23; an earlier version read this contrast as the rigidity using the reading's global octant.)

[...]

PR #8756, opened 2026-09-23
docs(campaign): cycle letters under the sweep record no role pattern

jonathonreilly:
## Summary

#8752 found that relational letters in four-angle cycles are rigid under the static reading, which reads a site's back and forward neighbours, and that they record the role pattern on the landed ice torus. This PR asks what the sweep reading of #8743 leaves of that. The sweep reads only the three back neighbours.

- **Sweep rule.** A site's three back differences must be signed angles with one common sense, from the three different cycles. A control checks that the rule accepts one sense with three cycles and rejects mixed senses and a repeated cycle.
- **Witnesses of flexibility.** On the core of side 3, a search finds 40 sweep records within 1139801 search nodes. The first, completed on the faces, satisfies the sweep rule at all 27 core sites and is not one of the 768 cycle spirals. The spiral set is checked to contain an explicit control spiral.
- **No static extension.** All 40 records break the static cycle rule at (2, 2, 2), whose six neighbours all lie in the core, so none extends to a static record whatever the face values. The static rule accepts the control spiral at that site.
- **No role readout.** The first witness's back-step phases are not the position mod 4 plus one global phase: they take 3, 3 and 4 values on the three axes. The same readout gives one phase per axis on the control spiral.

What this means: the forward neighbours carry the rigidity. Roles from relational letters need the static reading, so the assembly's relational route to roles rests on the reading decision, as it does for single-angle letters (#8743).

## Files

- `docs/RELATIONAL_CYCLE_LETTERS_UNDER_THE_SWEEP_READING_RECORD_NO_ROLE_PATTERN_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/relational_cycle_letters_under_the_sweep_reading_record_no_role_pattern_2026_09_23.py`
- `logs/runner-cache/relational_cycle_letters_under_the_sweep_reading_record_no_role_pattern_2026_09_23.txt`

## Verification

```bash
python3 scripts/relational_cycle_letters_under_the_sweep_reading_record_no_role_pattern_2026_09_23.py
```

- Result: `TOTAL: PASS=7 FAIL=0` in about 48 s, with 1132 characters of stdout.
- Mutation census: 7 of 7 caught. Three earlier mutants were caught only after controls were added for the static rule, the spiral set and the readout.

## Review


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `528d4ee96c75035b4c748e8de690f2e8ba6eef03` in b7d460034a13bf660b1d9bf8aa3fc4fb0322f744 as part of the combined thirteen-PR relational-record batch.

PR #8840, opened 2026-09-23
physics-loop: admissibility induced-law block 94 - three readings of records form against known physics: the spacetime reading is two-dimensional, skewed and irreversible

JONbridgerapps:
## Block 94 — three readings of "records form" against known physics: the spacetime reading is two-dimensional, skewed and irreversible

Hand-off PR against `main`. It answers the owner's question of 2026-09-23: "we can probe both readings no? see which is closest to our known physics?" Known physics is a comparator only. Nothing is adopted, and no parked decision is touched.

**Three readings, not two.** The "3+1 with separate time steps" reading used by blocks 90–93 is block 36's clause: a record at every site at every tick. Block 36 (#8507) found that clause excluded by the memo as written. Corrigenda are pushed on #8692, #8696, #8703 and #8710: I had missed block 36 as prior art. The owner's moving-records reading, in use since 2026-09-20, is the third reading, and it is also three-dimensional.

**Exact.**
- **T1:** reading (i), with `Z³` as spacetime, has a level chain that is reversible with respect to no law (cycle ratio `e^{8β}`). Its propagation is skewed (`T₁₁₁ = 2/27`), and its causal cone is a triangle.
- **T2:** in reading (i), a resting source's field is reflection-odd along the step directions and logarithmic. Two resting worldlines act on each other unequally, by 0.173.
- **T3:** in reading (iii), motion by the pair weights is reversible, and formation is the only arrow.
- **T4:** the dimension of space follows from how causal cones grow: 2 in reading (i), 3 in readings (ii) and (iii).
- **T5:** there are no fronts under detailed balance, so reversible record layers carry no waves.

**Scorecard.**
- **Reading (i) is last.** Its world has two space dimensions, its law has a built-in arrow, and its resting field is lopsided and logarithmic.
- **Reading (ii) proves the most, but contradicts the text.** A tick's record would either stay, making two records on one site, or cease, which breaks permanence. Its "1/r" is a transverse channel charged by a direction, and it has no waves.
- **Reading (iii) is closest among the readings the text allows.** It has three dimensions, reversible motion and irreversible formation. But its record layer makes no reciprocal field and no waves; the gravitational matches come from the amplitude layer's supplied clauses.
- **Reading (iii)'s reversibility and order depend on a formation-rate clause.** If every empty site keeps forming records, the lattice jams and motion stops.
- **A fourth reading was named by two lenses independently.** Records form once per site and never move, and what moves is unrecorded possibility, the amplitude layer.

**Panel.** Four lenses covered relativity, statistical mechanics, gravitation and foundations. They ran on Claude Opus 5.5, after the Fable lenses hit a usage limit. I checked their load-bearing claims and wrote the synthesis.

**Evidence.** Runner `scripts/admissibility_rule_three_readings_of_records_form_against_known_physics_spacetime_reading_two_dimensional_skewed_irreversible_2026_09_23.py`: `TOTAL: PASS=13 FAIL=0`. Each of the 8 mutations fails in its own family only. Two pipeline passes (`rows=4988`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `33ca79b8a3590973c1ad00306f88dd765893bbdf` in 8cc5f114f0d703100d6a7a764a1c689c7a56d65f as part of the combined fifteen-PR clock and locality batch.

- Retain specified kernels, even-lag stationary bound and finite continuous-time spectral theorem; restore beta-zero, aliased stencils, negative-real persistent multipliers and endpoint exceptions; no universal wave or source-limit conclusion.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block94-three-readings-against-known-physics-20260923` at `33ca79b8a3590973c1ad00306f88dd765893bbdf`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8865, opened 2026-09-23
Resolve formation capacity and seven-site terminal counts

jonathonreilly:
Moving an old record off a site can reopen formation, but each actual birth still adds two records to the supplied fixed-capacity model. This PR derives the finite count bound and the homogeneous infinite-volume mean balance, exhibits a physical state whose zero initial birth rate becomes positive through motion, and gives exact terminal counts on a seven-site path. From the stated empty-B preparation, one third of histories end after one birth and two thirds after two, for either specified formation instrument.

The branch builds on [PR #8852](https://github.com/jonathonreilly/qubit-lattice-axiom-framework/pull/8852)'s compensated local-volume target. It includes frozen author and independent PRE evidence, a distinct exact post-PRE comparison, the new terminal-count proof, full source bindings, and a scope review. The independent checker completed its computation but hit a usage limit before writing a final comparison narrative; the root review states exactly what was checked. The terminal-count proof has no checker-authored final review. No formal audit or retained claim is asserted.

Validation: the publication verifier checked 66 selected artifacts; both portable exact seven-site scripts ran and reproduced the saved scientific results; the comparison matched all 72 nonzero reactivation channels and the recorded periodic Gauss data; whitespace check passed. The infinite result is a mean bound in the supplied compensated target. It does not establish pointwise infinite-volume extinction, microscopic-volume exchange, an autonomous birth energy source, or TOE closure.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `79ab61e5aee8945bf96924815c7e0f9dd4d256a5` in 1c9a243e9b393520183bb06600dec3da9182371a as part of the combined thirteen-PR compensated formation and energy batch.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `codex/mobile-record-capacity-terminal-counts-20260923` at `79ab61e5aee8945bf96924815c7e0f9dd4d256a5`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8860, opened 2026-09-23
physics-loop: admissibility induced-law block 95 - records that move on their own clocks and slow the clocks around them attract with a 1/r potential, equal both ways

JONbridgerapps:
## Block 95 — records that move on their own clocks and slow the clocks around them attract with a 1/r potential, equal both ways

Hand-off PR against `main`. It answers the owner's request of 2026-09-23: "I think we are in the third column of your chart, and we need to figure out the not from records alone bits (make sure to check latest PRs and repo though)". The third column is the moving-records reading. This block takes its reciprocal `1/r` cell; the waves cell is next. Nothing is adopted, and no parked decision is touched.

**Survey first.** I checked the latest PRs and main for a reciprocal pull:
- the capture pull (block 49) is tied to growth and is not reciprocal;
- the record layer's own fields (blocks 41–43) are diffusive and charged by content;
- the ice-rule constructions (main's cubic-ice note; #8698) are sourced by signed charges;
- #8600 and #8650 carry waves, not a pull.

The one field with a universal, additive source and no mass term is block 53's clock field.

**Supplied, all named in the note.**
- Block 53's clock clause: a record slows its site's clock by the ratio `κ`, and only ratios of rates mean anything.
- A timing clause: a hop `x → y` runs at `w_x^a w_y^(1−a)`.
- The field follows the records at once.

**Exact.**
- **T1:** in a held field the stationary law is `W(C) Π w^(1−2a)`. Timed by the site it leaves, a record lingers where clocks are slow; for one record this is block 53's T5. Timed by the bond, nothing happens. Timed by the site it enters, the effect reverses.
- **T2:** when the field follows the records, the whole gas is in detailed balance with `W(C) exp(6 log κ (1 − 2a) Σ_pairs G(r − s))`, for any number of records, with each pair counted once. The held-field product would count each pair twice.
- **T3:** timed by the site it leaves, a record's hops are unbiased. The field changes only how long it waits: zero mean displacement away from contact, and two records always tick together.
- **T4:** `G` is even and decreasing (exact on `4³` and `6³`), with far field `1/(4πr)` (executed). The pair term is attractive exactly when `(2a − 1) log κ < 0`, and it is exactly reciprocal.

**Correction.** T3 corrects block 53's remark that the record's mean velocity follows the gradient. A text-only corrigendum is pushed on #8568.

**Executed control.**
[...]

PR #8866, opened 2026-09-23
physics-loop: admissibility induced-law block 96 - waves among moving records need a coupling that time reversal flips; clocked record motion never oscillates

JONbridgerapps:
## Block 96 — waves among moving records need a coupling that time reversal flips; clocked record motion never oscillates

Hand-off PR against `main`. It takes the second cell of the owner's request of 2026-09-23 ("we need to figure out the not from records alone bits"): waves in the moving-records column. Block 95 (#8860) took the reciprocal `1/r` pull. Nothing is adopted, and no parked decision is touched.

**Survey.** The repository has four routes to waves:
- block 44: records carrying a direction, with collisions that redraw directions at fixed momentum (sound);
- the mobile-record lane's #8600: record pairs with a chiral exchange rate `γ` (curl colour waves, static at `γ = 0`);
- #8650: link rotors with quantum hopping (transverse waves);
- block 54: possibility's qubit moving reversibly (the walk).

**Exact.**
- **T1:** record motion in detailed balance never oscillates. This covers block 39's motion and block 95's clocked motion at every timing, with the field held or slaved, and every function of the configuration: slaved possibility, the slaved clock field. The spectrum is real, so there are no fronts and no orbits. Checked at all 360 moves on a ring for three timings. A circulation gives complex mode rates.
- **T2:** a record's own direction memory, without exchange, gives only short waves that die. On a line the modes are non-real iff `|sin k| > (1 − p)/p`, and the density branch is diffusive. In 3D the spectrum is closed under conjugation. As `p → 1` there are free streams along the axes, not an isotropic wave.
- **T3:** in each route, long waves come from two conserved things coupled at first order in `k` by a term time reversal flips: density and momentum; `X` and `Y` through `γ`; the two parts of the amplitude. Made symmetric, the coupling gives growth and decay, not waves. Content-blind records diffuse under every covariant rule, because no vector is invariant under the 24 rotations.

**Reading.** The piece that does not come from records alone is a coupling that time reversal flips. Inertia needs the same thing, which is why block 95's pull clumps records instead of making them orbit. In the owner's picture it can enter through a chiral exchange, through links, through possibility keeping its own qubit state, or through momentum passed between records (which needs contents to change).

**Evidence.**
- Runner `scripts/admissibility_rule_waves_among_moving_records_need_a_coupling_that_time_reversal_flips_2026_09_23.py`: `TOTAL: PASS=13 FAIL=0`. Each of the 8 mutations fails in its own family only.
- Control: clocked chains of 120 and 2016 states have real spectra to `6·10⁻¹⁵` (with a circulation, `|Im|` 0.67–2.88). The 3D direction-memory record was scanned over the zone. A planned sentence ("the density branch is real at every `k`") was withdrawn after the control.
- Two pipeline passes (`rows=4988`, `graph_delta=acknowledged`); lints clean.
- Author checks only; no independent review has taken place.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `111143f765d15adac61fa33554fbf8749136754b` in 8cc5f114f0d703100d6a7a764a1c689c7a56d65f as part of the combined fifteen-PR clock and locality batch.

- Retain specified kernels, even-lag stationary bound and finite continuous-time spectral theorem; restore beta-zero, aliased stencils, negative-real persistent multipliers and endpoint exceptions; no universal wave or source-limit conclusion.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block96-waves-among-moving-records-need-a-coupling-time-reversal-flips-20260923` at `111143f765d15adac61fa33554fbf8749136754b`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8870, opened 2026-09-23
Quantify first-birth energy spread in compensated path target

jonathonreilly:
On the seven-site compensated rotor path, the actual first birth raises the target's mean energy by `32δ/3` in the stated Hamiltonian convention. Its conditional transition-energy variance is `56δ²/9` for both resolved and coherent instruments. Adding `μN` changes the mean by `2μ` per birth without changing the variance or marked histories, so this finite-path spread remains an energy-accounting obligation even when the mean is offset.

The note uses all actual first channel vectors in the 52-state Gauss sector. A separate root rotor-path calculation matches every vector to the independently built PRE matrix from #8865 and checks the moments with exact integer and rational arithmetic. The portable runner was rerun in this worktree; its scientific result matched the committed result. The publication verifier authenticated seven artifacts and three source dependencies; the branch has a clean diff and no whitespace errors.

This is a provisional review proposal for a supplied target and one finite first event. It supplies no autonomous reservoir or measured heat cost, makes no all-time heating claim, and has no separate independent post-result review or formal audit verdict.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `4f5f210ecc33a8eba817fcfb153c9f43f81b96a6` in 1c9a243e9b393520183bb06600dec3da9182371a as part of the combined thirteen-PR compensated formation and energy batch.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `codex/mobile-record-energy-spread-20260923` at `4f5f210ecc33a8eba817fcfb153c9f43f81b96a6`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8872, opened 2026-09-23
physics-loop: admissibility induced-law block 97 - a record that cannot read the field from its own motion hops on its own clock, and with kept books records attract

JONbridgerapps:
## Block 97 — a record that cannot read the field from its own motion hops on its own clock, and with kept books records attract

Hand-off PR against `main`. Block 95 (#8860) found that records hopping in block 53's clock field attract or repel according to `(2a − 1) log κ`, with no clause fixing either factor. This block fixes both. Nothing is adopted, and no parked decision is touched.

**Exact.**
- **T1:** every rate `w_x^a w_y^(1−a)` has degree one (no master clock). Counted in ticks of the site it occupies, a record hops `(w_y/w_x)^(1−a)` times per tick and picks a direction with odds proportional to `w_y^(1−a)`. Both are free of the field only for `a = 1`, the hop timed by the site the record occupies. Timed by the bond, a record could read the field from its own hopping (`√(w_y/w_x)` hops per own tick). Timed by the site it enters, it heads for fast clocks.
- **T2:** for amplitudes, the three timings are similar generators (`WH`, `√W H √W`, `HW`): one dynamics. For records they are three stationary laws (`1/w`, `1`, `w`). So block 54's local-time clause does not decide for records; read as the arrival of a record, it would give `a = 0`.
- **T3:** with `a = 1` and block 55's kept books (`log κ = −(γ/6) E/w̄`), block 95's pair term is `−(γE/w̄) G(r)`, which is attractive. On `4³`, `U(1,0,0) − U(2,0,0) = −(γE/w̄)(228/7680)`.

**Principle supplied:** a record cannot read the field from its own motion. It is a statement about locality, not about the sign, and it has a unique solution.

**Correction made while writing.** The first draft took block 54's clause as the principle. T2 showed that clause does not decide for records, so the principle is now stated on its own, and the title and claim id were changed before the gates.

**Evidence.** Runner `scripts/admissibility_rule_a_record_that_cannot_read_the_field_from_its_own_motion_hops_on_its_own_clock_2026_09_23.py`: `TOTAL: PASS=11 FAIL=0`. Each of the 6 mutations fails in its own family only. Two pipeline passes (`rows=4988`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
**Note after the owner's landing of block 55 (c3f8c47a58, 2026-09-23, supervisor).**

As landed on main, block 55 changes two things this note relies on:
- It keeps its reciprocal point-force algebra as a separately supplied model with its own coupling, "not implicitly the gamma normalization" of its field law.
- It states that no physical source or exact packet force is established.

So T3's premise `log κ = −(γ/6) E/w̄` is the PR text's corollary. Within the landed block 55 it is an identification of the point-force model's source with the linearized field law, not a consequence.

T1, T2 and the locality principle that selects `a = 1` do not depend on it. T3's conclusion (with `a = 1`, records attract with `U = −(γE/w̄)G`) stands conditional on that identification, or on any clause that makes a record's source proportional to its energy with a negative coefficient. Block 104 (#8931) states the same identification explicitly.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `f6c81d4a73f6b0547340749e1d1655b2f92a7ef8` in 8cc5f114f0d703100d6a7a764a1c689c7a56d65f as part of the combined fifteen-PR clock and locality batch.

- Neutralize torus sources, require constant diagonal for recomputed pair law, restrict uniform directions to W=1 with rate w/12, distinguish transverse ray projection and exact normalized birth probabilities, preserve scalar onsite-energy counterexample and separate coupling hypotheses.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block97-a-record-that-cannot-read-the-field-hops-on-its-own-clock-20260923` at `f6c81d4a73f6b0547340749e1d1655b2f92a7ef8`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8873, opened 2026-09-23
Bound finite-time electric tails after record formation

jonathonreilly:
Actual record motion and repeated formation shift electric links in the compensated target. This PR proves that a fixed link with an initially finite exponential field moment keeps that moment finite at every finite time, uniformly in graph volume at fixed degree. For initial zero field, a direct absolute-value bound gives a prefactor-one Chernoff tail. The locally normal infinite target inherits the estimate through the bounded-observable volume limit in #8852.

The proof uses the actual resolved or coherent formation channels, a hard-core single-link shift lemma, weighted Lindblad estimates and monotone cutoffs. A blind independent reconstruction proved a coarser bound before reading the root theorem; its exact Gauss counterexample shows why the pair count must reach distance two. The post-PRE comparison checked the root coefficient and the separate absolute-value corollary. The publication verifier authenticated 25 artifacts and four pinned sources. Root and independent portable controls reran successfully, with matching scientific results; the 27-file branch has no whitespace errors.

This is a conditional theorem for the supplied compensated rotor model. It does not select that Hamiltonian or formation instrument, provide a microscopic spin-uniform or all-time field bound, derive an energy reservoir, or claim a photon/particle/TOE result. No formal audit verdict is applied.


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `a7e3dae5c667e9ba0c9561a9bcf87b0401e21a5a` in 1c9a243e9b393520183bb06600dec3da9182371a as part of the combined thirteen-PR compensated formation and energy batch.

- Assert all-mark spin polynomials, exact H4 moments and drift, covariance and full selected image count. Add clipped/unclipped absolute weights to the finite weighted-generator check. Independent tensor and complex-matrix controls agree.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `codex/mobile-record-exponential-field-tail-20260923` at `a7e3dae5c667e9ba0c9561a9bcf87b0401e21a5a`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

PR #8871, opened 2026-09-23
docs(ice): the unit link field fixes one stiffness for flux and correlations

jonathonreilly:
## Summary

#8869 found that the zero-flux layer chain of uniform ice carries the branch of a massless nearest-neighbour lattice field. That dispersion does not depend on the field's stiffness. This PR asks what sets the stiffness, on the cross-sections 2 × 2 to 2 × 10 and 4 × 4, with the row transfer of #8859.

- **What a Gaussian field predicts.** A divergence-free field with weight exp(−(K/2) Σ E²) on the links of the prism A × Z has flux cost c = K/2 per layer. Its equal-layer vertical correlation is S_zz(q) = √(Q/(Q+4))/K, with Q = Σ 2(1 − cos q_i).
- **The unit link field fixes K.** Every link is occupied or empty, so E = (−1)^(x+y)(2v − 1) = ±1 and ⟨E_z²⟩ = 1. The sum rule then fixes K_A = (1/A) Σ_q √(Q/(Q+4)), with no free constant.
- **The flux cost is K_A/2 on every shape.** The costs reproduce #8859 and lie within 1% of K_A/2 on every cross-section:

  | cross-section | c(2) | K_A/2 | deviation |
  |---|---|---|---|
  | 2 × 2 | 0.28082 | 0.27884 | +0.71% |
  | 2 × 4 | 0.30726 | 0.30841 | −0.37% |
  | 2 × 6 | 0.31210 | 0.31336 | −0.40% |
  | 2 × 8 | 0.31374 | 0.31501 | −0.40% |
  | 2 × 10 | 0.31451 | 0.31576 | −0.40% |
  | 4 × 4 | 0.32704 | 0.32709 | −0.02% |

- **The same constant sets the correlations.** Take K = 2c(2) from the flux cost. The zero-flux vacuum then satisfies K S_zz(q) = √(Q/(Q+4)) within 0.64% at every nonzero wavenumber on the square, and within 5% on the strips 2 × 4 to 2 × 10. The strips' largest deviation is at q = (π, 0), across the width of 2. On 2 × 2 the agreement is within 10%. The flux cost and the correlations are separate observables: the |S| = 2 top level and the zero-flux vacuum vector.
[...]

PR #8869, opened 2026-09-23
docs(ice): the layer transfer branch has a massless lattice field's dispersion

jonathonreilly:
## Summary

#8864 found that the zero-flux layer chain of uniform ice is gapless on the computed cross-sections: its gap falls like 1/b and follows the smallest transverse wavenumber. This PR resolves the chain's spectrum by wavenumber, on the strip 2 × 8 and the square 4 × 4, and asks what the whole lowest branch looks like. It uses the row transfer of #8859 and the layer transfer T of #8740.

- **Two symmetries.** T commutes with the transverse translations, so its zero-flux spectrum splits by wavenumber. T also commutes with the complement v → 1 − v, because the rule "three of six links occupied" is self-complementary. The branch rate in each wavenumber sector is D(k) = ln(λ_0/|λ_top(k)|).
- **The branch matches a massless lattice field, with nothing fitted.** The comparison is the nearest-neighbour field on Z^3: the graph Laplacian, a framework object, whose transfer rate satisfies cosh D = 1 + Σ(1 − cos q_i). It has no mass, velocity or stiffness to fit.

  | cross-section | q | D (ice) | D (field) |
  |---|---|---|---|
  | 2 × 8 | π/4 | 0.7429 | 0.7478 |
  | 2 × 8 | π/2 | 1.3165 | 1.3170 |
  | 2 × 8 | 3π/4 | 1.6581 | 1.6530 |
  | 4 × 4 | (π/2, 0) | 1.2963 | 1.3170 |
  | 4 × 4 | (π/2, π/2) | 1.7419 | 1.7627 |
  | 4 × 4 | (π/2, π) | 2.0469 | 2.0634 |

  On the strip every point is within 1%, and on the square within 2%. The square's diagonal-to-axis ratio, 1.3438, agrees with the field's 1.3385 within 1%.
- **The branch is made of single field quanta.** The vacuum is even under the complement and has a positive level. All 10 branch states computed are odd, as one quantum of the staggered field (−1)^(x+y)(2v − 1) is. They also have negative levels, the staggered sign along the transfer direction.
- **Cross-checks.** The wavenumber-0 top equals the top of the full zero-flux sector. The lowest branch points are the gaps of #8864: 8 D(π/4) = 5.9434 on 2 × 8 and D(π/2, 0) = 1.29627 on 4 × 4. Opposite wavenumbers, and (π/2, 0) with (0, π/2), are degenerate.

[...]

PR #8859, opened 2026-09-23
docs(campaign): ice flux stiffness on a square cross-section

jonathonreilly:
## Summary

#8746 found that the flux sectors of the layer transfer matrix (#8740) carry a Gaussian stiffness on cross-sections of width 2, and left square cross-sections open. Its degree-count method does not reach a 4 × 4 cross-section. This PR applies the transfer matrix one row at a time instead. A row, with its horizontal links summed out, maps the row's w bits and the links in from the previous row to the row's v bits and the links out to the next row. The frontier is 2a link bits, so the 2^16 × 2^16 matrix is never formed.

- **Checks against #8746.** trace T² = 9600 on 2 × 2, the transfer is symmetric, and the zero-flux and |S| = 2 top eigenvalues match #8746 on 2 × 2, 2 × 4 and 2 × 6.
- **The square 4 × 4 (area 16).** The top eigenvalue falls with the flux over |S| = 0, 2, 4, 6, 8. The coefficient c(S) = (A/S²) ln(λ_0/λ_S) is:

  | \|S\| | c(S) |
  |---|---|
  | 2 | 0.3270 |
  | 4 | 0.3300 |
  | 6 | 0.3352 |
  | 8 | 0.3430 |

  It stays within 3% of its |S| = 2 value up to |S| = 6: a Gaussian flux stiffness on a square cross-section.
- **Shape at equal area.** The strip 2 × 8 has c = 0.3137 at |S| = 2, about 4% below the square.
- **Strips of width 2.** c = 0.2808, 0.3073, 0.3121, 0.3137, 0.3145 for 2 × 2 to 2 × 10, rising and levelling.

What this means: layer units see one Gaussian flux stiffness on both shapes, about 0.31–0.33 per layer at these sizes. That is the cost of a uniform field in a Coulomb phase. The square sits a few percent above the strip at equal area; the large-cross-section value is not decided here. No constant is compared with an outside value.

No reading, rule, alphabet, unit or order law is adopted.

## Files

- `docs/UNIFORM_ICE_FLUX_STIFFNESS_ON_A_SQUARE_CROSS_SECTION_BY_ROW_TRANSFER_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/uniform_ice_flux_stiffness_on_a_square_cross_section_by_row_transfer_2026_09_23.py`
- `logs/runner-cache/uniform_ice_flux_stiffness_on_a_square_cross_section_by_row_transfer_2026_09_23.txt`

## Verification

```bash
[...]

PR #8740, opened 2026-09-23
docs(campaign): layer units form uniform ice on infinite prisms at zero flux

jonathonreilly:
## Summary

The assembly (#8648) lists as an open edge exact formation of the uniform ice measure "by layer units on the infinite lattice". #8720 showed that finite cell units fail around blocks and that chains of units work on finite windows. This PR runs the chain along an infinite axis.

- **Layers carry the measure through a transfer matrix.** A layer unit is the slab between heights z and z + 1: its in-plane links and the vertical links above it. The ice rule couples a unit only to the vertical links below it, so T[v, w] counts in-plane completions, and the n × n × H torus has trace T^H ice states. For n = 2, trace T² = 9600 (the landed 2 × 2 × 2 count), and trace T⁴ = 23063296 agrees with a transfer along another axis.
- **The flux through the cross-section is conserved.** The staggered vertical flux S(v) = Σ (−1)^(x+y) (2v − 1) flips sign at every layer, so |S| is conserved:

  | sector | layer states | eigenvalues |
  |---|---|---|
  | \|S\| = 0 | 6 | simple top eigenvalue 32 + 2√209 ≈ 60.91 |
  | \|S\| = 2 | 8 | +46 and −46 |
  | \|S\| = 4 | 2 | +18 and −18 |

  The zero-flux block is positive, with characteristic polynomial (x − 6)(x + 10)(x + 14)²(x² − 64x + 188), fixed exactly by the traces of its first six powers, and the polynomial annihilates the block.
- **Long prisms select zero flux.** The share of ice states that carry flux falls as 2(46/60.91)^H for even H and vanishes for odd H.
- **Layer units form the zero-flux measure exactly.** In the zero-flux sector, the law of a layer given the layer below and the far end of a long torus tends to P(w | v) = T[v, w] φ(w) / (λ φ(v)). It does not depend on the far end (deviation 8.7 × 10⁻¹⁰ at H = 16). So layers formed as a chain reproduce the limit measure, and each layer's law depends only on the vertical links below it.
- **Odd cross-sections alternate.** For n = 3 the top eigenvalues come as ±λ, and odd tori have no ice states.
- **A second cross-section agrees.** On 2 × 4 (256 layer states), the flux is again conserved (sectors 0 to 8). The zero-flux block (70 states) has an entrywise positive square and a top eigenvalue 2401.33, above every flux sector's. The layer kernel again forgets the far end.

[...]

PR #8875, opened 2026-09-23
docs(ice): test defects interact through the lattice Green's function

jonathonreilly:
## Summary

#8869 and #8871 found that the flux of uniform ice, read from layer units, behaves as one massless Gaussian lattice field whose stiffness is fixed by the unit field on every link. A field of this kind should have sources. This PR puts two test sources into the layer measure and measures how they interact, on the square 4 × 4 and the strip 2 × 8.

- **Test defects are charges.** Uniform ice occupies three of the six links at every vertex, so the staggered field E = (−1)^(x+y+z)(2v − 1) is divergence-free. A test defect is one vertex with four or two occupied links, and it carries charge ±2. Two opposite defects are a probe only: the rule stays three of six everywhere else.
- **Free energy by layer transfer.** The pair's free energy is V = −ln(Z_pair/Z_0) in the infinite prism A × Z. T is the layer transfer of the landed layer-unit note (#8740), applied by the row transfer of #8859. When the defects are z layers apart, the transfer between them runs through the |S| = 2 sector.
- **The prediction has no free constant.** A Gaussian divergence-free field gives V(r) − V(r′) = K Q²(G_A(r′) − G_A(r)). Here G_A is the Green's function of the Z^3 graph Laplacian on the prism, Q = 2, and K = 2c(2) comes from the flux cost as in #8871.
- **Symmetry.** The 4 × 4 cross-section is the four-cube as a graph, and V depends only on graph distance there: 0.46195, 0.53561, 0.56358 and 0.57797 at distances 1 to 4. Swapping the defect types leaves V unchanged.
- **The measured interaction,** as ratios of measured to predicted differences, with nothing fitted:

  | shape | pairs | ratios |
  |---|---|---|
  | 4 × 4 | in the layer, distances 2, 3, 4 | 0.9614, 0.9738, 0.9781 |
  | 4 × 4 | in the layer, from distance 2 | 1.0080, 1.0085 |
  | 4 × 4 | along the prism, z = 1 to 9 | 0.967 to 0.995 |
  | 2 × 8 | along the length | 0.981 to 0.993 |
  | 2 × 8 | along the prism, z = 1 to 9 | 0.973 to 0.995 |
  | 2 × 8 | across the width of 2 | 0.9252, 1.0563 |

  The pair attracts.
- **The string.** On a finite cross-section, defects far apart along the prism are joined by confined field lines, and V grows linearly. At z = 8 the step V(z+1) − V(z) agrees with the Gaussian's within 0.1%, and with the flux cost ln(λ_0/λ_2) within 1%. The tension, K Q²/(2A), falls as the cross-section grows.

What this means: the static field of uniform ice has sources, and they interact as the field's own lattice Green's function says, with the stiffness already fixed by the flux cost. On the square the pair free energy follows it within 4% at every separation, and within 1% beyond distance 2. This is the static Coulomb interaction of the Coulomb phase, read from layer units, with charge 2 and no free coupling. On large cross-sections the Gaussian reading gives the potential −K Q²/(4πr) = −2/(3πr); that is a reading, not a computed limit. No constant is compared with an outside value.

## Files

- `docs/UNIFORM_ICE_TEST_DEFECTS_INTERACT_THROUGH_THE_LATTICE_GREEN_FUNCTION_WITH_THE_FLUX_COST_STIFFNESS_BOUNDED_THEOREM_NOTE_2026-09-23.md`
- `scripts/uniform_ice_test_defects_interact_through_the_lattice_green_function_with_the_flux_cost_stiffness_2026_09_23.py`
- `logs/runner-cache/uniform_ice_test_defects_interact_through_the_lattice_green_function_with_the_flux_cost_stiffness_2026_09_23.txt`

## Verification

```bash
python3 scripts/uniform_ice_test_defects_interact_through_the_lattice_green_function_with_the_flux_cost_stiffness_2026_09_23.py
```

- Result: `TOTAL: PASS=7 FAIL=0` in about 24 s, with a peak of about 500 MB and 1209 characters of stdout.
- Mutation census: 12 of 12 caught.

## Review


JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `970731556520cf4ac3bb4bc5a2562b60dbd30931` in f131dcf77de58e95a1373e9f923fbfb32cddb136 as part of the combined fourteen-PR ice transfer and sampling batch.

- Retain explicit conditional structural mathematics and finite computational observations. State domains, supplied models, approximations and unresolved limits; no particle interpretation from complement parity, no universal Gaussian law from covariance agreement.
- Add seeded iterative starts and explicit eigenpair residual tests without claiming rigorous enclosures; use integer count powers; state hidden-variable contraction nonlocality, sampled-column TV scope, correct lattice anisotropic comparator and finite-sector normalization. Preserve listed finite ratios.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Unsupported stronger conclusions were narrowed; original commits remain recoverable through this PR. No auxiliary calculation was silently discarded.

Validation: fourteen fresh runners, independent mathematical controls and fourteen targeted caught defect mutations; shared pipeline, strict lint and changed-evidence gate passed (14 checked, zero failures/control failures). Final blobs verified on main.

Issue #8862, opened 2026-09-23
probe HIT: J:derive:bending-over-fall-at-strong-field:a2


JONbridgerapps:
**Task:** `J:derive:bending-over-fall-at-strong-field:a2`. Worker `w-jonathonsmac4f50-j3a05`, model `claude-opus-5-5`.
**Log:** `logs/probes/J:derive:bending-over-fall-at-strong-field:a2/w-jonathonsmac4f50-j3a05__7e8ce5e3__20260923T224936Z.json` on `ai/probes`.
**Deliverable:** `probes/work/derive/bending-over-fall-at-strong-field/w-jonathonsmac4f50-j3a05/{ATTEMPT.md,check.py}`. 10 checks pass: exact sympy, plus executed ray integrals and packets.

**Provenance.** This attempt and blocks 54 to 77 come from the same model family (Claude Opus). The HIT needs a referee from another family.

### Exact, at every point and every strength

**The two laws.**
- **A ray:** `H = (w/l)|k|`, from block 60's crossing rate `√(w_x w_y)/(χ_x χ_y)`. Its transverse acceleration is `c²∇⊥ ln(l/w)`.
- **A slow body:** block 77's rest energy gives `E = w √(m² + k²/l²)`. It falls at `−(w/l)² ∇ ln w`.

**Their ratio in block 60's exact one-body field** (`χ = 1 + Qg`, `N = 1 − Pg`, `P = Qw₀`):

`bending/fall = (2 + 3Qg₀ − Qg)/(1 + Qg₀) = 1 + 2/(1 + w₀/w(x))`, for any shape of the Green's function.
- It is linear in `g`.
- It is exactly 2 at the body and block 60's `1 + (1 + 2Qg₀)/(1 + Qg₀)` far away.
- It lies in **[2, 3) at every point**: block 60's bound holds locally, not only far away.
- The ratio integrated along a straight line is a weighted mean of it, so it is also in [2, 3).

### Why this is a HIT (the task's condition: a ratio above 3 at any strength)

- The **exact nonlinear ray deflection** (Fermat, exact index), divided by the straight-line fall, **exceeds 3 at strong field**: 3.75 at `b = 3b_c` for `Qg₀ = 7.58`.
- It **diverges at the capture radius** `r* = a(1 + w₀ + √(w₀² + w₀ + 1))`, with `a = Q/(4π)` (about 5 sites here).
[...]

PR #8878, opened 2026-09-24
physics-loop: admissibility induced-law block 98 - one field for records and waves: a wave passing a record turns by twice the record's pair energy at that distance

JONbridgerapps:
## Block 98 — one field for records and waves: a wave passing a record turns by twice the record's pair energy at that distance

Hand-off PR against `main`. Block 94's scorecard left the moving-records column's "one geometry for matter and fields" at "supplied". Block 95 (#8860) and block 97 (#8872) showed that records clump, and attract, in block 53's clock field. Block 54 showed that a walk whose phase runs on the local clock falls towards slow clocks. This block ties the two together in counted terms. Nothing is adopted, and no parked decision is touched.

**Exact.**
- **T1:** on every torus, the kernel of one record summed along a lattice line is the plane kernel. Checked exactly on `4³/4²` and `6³/6²` at every transverse offset.
- **T2:** for the far field `1/(4πr)`, the transverse gradient integrated along a straight line at impact parameter `b` is exactly `−2 · 1/(4πb)`.
- **T3:** at long waves the walk's ray tensor is the identity and its speed is one. To first order, a wave passing a record turns towards it by `δ(b) = 2|U(b)|`, with `U` block 95's pair energy. Since block 95's pair excess is `g = e^(−U)`, this is `δ(b) = 2 log g(b)`.

**In plain terms.** Measure how much more often records are found at some distance from one another, take the log and double it: that is the angle through which a wave passing at that distance turns towards a record.

**Executed.**
- The lattice kick divided by the pair energy is `2.10, 2.04, 2.008, 2.003, 2.0002` at `b = 2, 4, 8, 12, 16` (128-torus, backgrounds removed).
- Walk packets passing one record agree with clouds of rays carrying their own spreads: 1.053, 0.994, 1.083.
- The control's first design used wide packets passing close to the record, and the packets' own spreads washed out the turn. It was replaced before any comparison entered the note.

**Limits.** First order in the field, straight passage, far field, long waves. In the physics comparator light turns by twice what a clock-only field gives; the lane's lengths (blocks 59–61) are where the rest would have to come from. This is named, not worked.

**Evidence.** Runner `scripts/admissibility_rule_one_field_for_records_and_waves_wave_turns_by_twice_the_pair_energy_2026_09_23.py`: `TOTAL: PASS=11 FAIL=0`. Each of the 6 mutations fails in its own family only. Two pipeline passes (`rows=4988`, `graph_delta=acknowledged`); lints clean. Author checks only; no independent review has taken place.


JONbridgerapps:
Corrigendum pushed (6672f80380). N1 route 4 said the lengths' share of the turn was not worked in the lane. In fact:
- Block 59 (#8581) had shown that with bond rates and lengths, bending over fall is `1 + β`.
- Block 60 (#8590) had shown that its curvature member has `β = 1`.

This note's relation `δ = 2|U(b)| = 2 log g(b)` is therefore for block 54's product-form walk, with lengths one. Under block 60's member the turn doubles, to `4|U(b)| = 4 log g(b)` at first order, provided the records' pair law is unchanged by the lengths (not worked).

I found this in the probe harvest through worker result #8862, which extends block 60's ratio exactly to all field strengths; it is unrefereed and from the same model family. The change is text only; the runner output is unchanged and the cache is re-pinned.

JONbridgerapps:
Landed the reviewed, narrowed source from frozen head `6672f80380772bc01bbeda5a1cd7cf6db2d456c0` in 8cc5f114f0d703100d6a7a764a1c689c7a56d65f as part of the combined fifteen-PR clock and locality batch.

- Neutralize torus sources, require constant diagonal for recomputed pair law, restrict uniform directions to W=1 with rate w/12, distinguish transverse ray projection and exact normalized birth probabilities, preserve scalar onsite-energy counterexample and separate coupling hypotheses.

All original paths have explicit dispositions. The canonical science paths are landed with the corrections above and fresh evidence. Historical auxiliary scripts, experiment outputs, findings and campaign state are deferred and preserved on branch `physics-loop/admissibility-induced-law-block98-one-field-for-records-and-waves-20260923` at `6672f80380772bc01bbeda5a1cd7cf6db2d456c0`; they are not certified by this review. The generated citation manifest is replaced by the shared current-main manifest. No auxiliary calculation was silently discarded.

